PR #6295, opened 2023-05-17
[interop client] let the interop client send additional arbitrary metadata, controlled by a flag

apolcyn:
This is a Go analogue of an old PR in C++: https://github.com/grpc/grpc/pull/18156 

This is needed for RLS integration tests, where we sometimes need to set custom headers for routing purposes.

cc @easwars 

RELEASE NOTES: none

### DIFF
--- a/interop/client/client.go
+++ b/interop/client/client.go
@@ -128,7 +128,7 @@ func parseAdditionalMetadataFlag() (metadata.MD, error) {
 		return nil, nil
 	}
 	r := *additionalMetadata
-	addMd := metadata.New()
+	addMd := metadata.New(map[string]string{})
 	for len(r) > 0 {
 		i := strings.Index(r, ":")
 		if i < 0 {
--- a/interop/client/client.go
+++ b/interop/client/client.go
@@ -24,12 +24,15 @@
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"flag"
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"golang.org/x/oauth2"
@@ -41,6 +44,7 @@ import (
 	"google.golang.org/grpc/credentials/oauth"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/interop"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/testdata"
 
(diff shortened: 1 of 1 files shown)


interop/client/client.go
```diff
@@ -115,6 +120,37 @@ const (
...
+		r = r[i+1:]
+		i = strings.Index(r, ";")
+		if i < 0 {
+			addMd.Set(key, r)
+			break
+		}
+		addMd.Set(key, r[:i])
+		if i == len(r)-1 {
+			break
+		}
```
dfawley, line 148: I think you can remove this and rely on the loop bounds instead?

interop/client/client.go
```diff
@@ -214,6 +250,22 @@ func main() {
 	if len(*serviceConfigJSON) > 0 {
 		opts = append(opts, grpc.WithDisableServiceConfig(), grpc.WithDefaultServiceConfig(*serviceConfigJSON))
 	}
+	addMd, err := parseAdditionalMetadataFlag()
+	if err != nil {
+		logger.Fatalf("Error parsing --additional_metadata flag: %v", err)
+	}
+	if addMd != nil {
+		unaryAddMd := func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+			ctx = metadata.NewOutgoingContext(ctx, addMd)
```
dfawley, line 259: You should probably be using `metadata.AppendToOutgoingContext()` instead in case a test case ever needs to set its own metadata.  Unfortunately that takes a key/value pair slice instead of the metadata map so will need a slight reworking of `parseAddlMDFlag`.

interop/client/client.go
```diff
@@ -214,6 +250,22 @@ func main() {
...
+			return invoker(ctx, method, req, reply, cc, opts...)
+		}
+		streamingAddMd := func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+			ctx = metadata.NewOutgoingContext(ctx, addMd)
+			return streamer(ctx, desc, cc, method, opts...)
+		}
+		opts = append(opts, grpc.WithUnaryInterceptor(unaryAddMd))
+		opts = append(opts, grpc.WithStreamInterceptor(streamingAddMd))
```
dfawley, line 267: Nit: combine these lines?

interop/client/client.go
```diff
@@ -214,6 +250,22 @@ func main() {
 	if len(*serviceConfigJSON) > 0 {
 		opts = append(opts, grpc.WithDisableServiceConfig(), grpc.WithDefaultServiceConfig(*serviceConfigJSON))
 	}
+	addMd, err := parseAdditionalMetadataFlag()
+	if err != nil {
+		logger.Fatalf("Error parsing --additional_metadata flag: %v", err)
```
dfawley, line 255: Optional: IMO just put the `logger.Fatalf` inside the function and have it not return an error.  But it's fine if you don't want to change it.

apolcyn: done

PR #6297, opened 2023-05-18
authz: End2end test for audit loggers

erm-g:
End to end tests covering AuditLogging feature

### DIFF
--- a/authz/grpc_audit_end2end_test.go
+++ b/authz/grpc_audit_end2end_test.go
@@ -1,0 +1,101 @@
+package authz
+
+import (
+	"context"
+	"net"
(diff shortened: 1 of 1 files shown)


dfawley:
It seems like this is not ready.  Does it really need to be a pull request?  AFAICT there's no need for a draft pull request, as you can do everything it can do in other ways: - to run the tests, push to your fork and check actions. - to show others the code, point them at your branch/fork. - to allow others to pre-review, make a PR on your own fork.

I'll close this so that it doesn't appear in our open PR list, but let me know if there's a use case that requires draft PRs.

PR #6298, opened 2023-05-18
xds: support built-in Stdout audit logger type

gtcooke94:
This PR adds the functionality to parse and build the known StdoutLogger that we include as an implemented AuditLogger.

I added a `Name` const to identify this logger, matching the LoadBalancer design. I believe this is needed. I also exported stdout.Logger so that we could check the type in the test, though I'm thinking there might be a better way to do this. I'm open to suggestions on a better way to test.

RELEASE NOTES:
- xds: add the functionality to parse and build the known StdoutLogger through the XDS path


internal/xds/rbac/converter_test.go
```diff
@@ -110,5 +112,52 @@ func (s) TestBuildLoggerErrors(t *testing.T) {
...
 	}
+}
 
+func (s) TestBuildLoggerKnownTypes(t *testing.T) {
+	tests := []struct {
+		name         string
+		loggerConfig *v3rbacpb.RBAC_AuditLoggingOptions_AuditLoggerConfig
+	}{
+		{
+			name: "stdout logger",
```
rockspore, line 123: So if users send typed structs with the name being `stdout_logger`, we will end up creating the same logger. Can we test those as well?

rockspore: We will accept that and create stdout loggers for those configs. I don't think there is other ways here as authz policy translates the config into typed structs regardless.

internal/xds/rbac/converter.go
```diff
@@ -72,10 +75,24 @@ func getCustomConfig(config *anypb.Any) (json.RawMessage, string, error) {
...
+		if err := config.UnmarshalTo(stdoutLoggerConfig); err != nil {
+			return nil, "", fmt.Errorf("failed to unmarshal resource: %v", err)
+		}
+		return convertStdoutConfig(stdoutLoggerConfig)
 	}
 	return nil, "", fmt.Errorf("custom config not implemented for type [%v]", config.GetTypeUrl())
 }
 
+func convertStdoutConfig(config *v3auditloggersstreampb.StdoutAuditLog) (json.RawMessage, string, error) {
+	json, err := json.Marshal(config)
```
rockspore, line 89: Instead of calling this, you could simply return `{}`.

gtcooke94: Discussed offline, we don't want to simply ignore the input config and return `{}`, though it would be fine with the current impl

internal/xds/rbac/converter.go
```diff
@@ -24,13 +24,17 @@ import (
...
 	"google.golang.org/grpc/authz/audit"
+	"google.golang.org/grpc/authz/audit/stdout"
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
 const udpaTypedStuctType = "type.googleapis.com/udpa.type.v1.TypedStruct"
 const xdsTypedStuctType = "type.googleapis.com/xds.type.v3.TypedStruct"
+const stdoutType = "type.googleapis.com/envoy.extensions.rbac.audit_loggers.stream.v3.StdoutAuditLog"
```
arvindbr8, line 37:
```suggestion
const (
	udpaTypedStructType = "type.googleapis.com/udpa.type.v1.TypedStruct"
	xdsTypedStructType = "type.googleapis.com/xds.type.v3.TypedStruct"
	stdoutType = "type.googleapis.com/envoy.extensions.rbac.audit_loggers.stream.v3.StdoutAuditLog"
)
```
pls. While you are at there, might be helpful to fix typos

internal/xds/rbac/converter.go
```diff
@@ -24,13 +24,17 @@ import (
 	v1xdsudpatypepb "github.com/cncf/xds/go/udpa/type/v1"
 	v3xdsxdstypepb "github.com/cncf/xds/go/xds/type/v3"
 	v3rbacpb "github.com/envoyproxy/go-control-plane/envoy/config/rbac/v3"
+	v3auditloggersstreampb "github.com/envoyproxy/go-control-plane/envoy/extensions/rbac/audit_loggers/stream/v3"
```
arvindbr8, line 27: we had a change recently which attempts at standardizing pb named imports, so that we can rename them appropriately at import time. #6238 

@dfawley: should we make a note of this and make changes to the google3 import script?

internal/xds/rbac/converter.go
```diff
@@ -72,10 +76,24 @@ func getCustomConfig(config *anypb.Any) (json.RawMessage, string, error) {
...
 	}
 	return nil, "", fmt.Errorf("custom config not implemented for type [%v]", config.GetTypeUrl())
 }
 
+func convertStdoutConfig(config *v3auditloggersstreampb.StdoutAuditLog) (json.RawMessage, string, error) {
+	json, err := protojson.Marshal(config)
+	if err != nil {
+		return nil, "", err
+	}
+	return json, stdout.Name, nil
```
arvindbr8, line 94: Seems like there is an check for non nil error at the callsite of `getCustomConfig` and the returned json.RawMessgage and string is anyway unused, can we skip the non-nil error check from this method and rely on the callsite check? 

what i mean is:

```suggestion
	json, err := protojson.Marshal(config)
	return json, stdout.Name, err
```

arvindbr8 (changes requested):
Easwar is out of office for now. I've taken a pass at it.

internal/xds/rbac/converter.go
gtcooke94: I believe I have the typed structs named correctly

Do you have a suggested name change for `v3auditloggersstreampb`? It's a new import, but it roughly takes the same approach of the last 2 tokens before `v3` `ignored/stuff/audit_loggers/stream/v3` -> `v3auditloggersstreampb` similar to `ignored/stuff/xds/type/v3` -> `v3xdstypepb`

internal/xds/rbac/converter.go
gtcooke94: Done

internal/xds/rbac/converter.go
gtcooke94: I'm not immediately seeing any typos, where are they?

internal/xds/rbac/converter.go
arvindbr8: Sure this SGTM. I was just pinging Doug to see what is the action item for the import script. Resolving this.

arvindbr8 (approved):

authz/audit/stdout/stdout_logger.go
```diff
@@ -31,6 +31,9 @@ import (
 
 var grpcLogger = grpclog.Component("authz-audit")
 
+// Name is the string to identify this logger type in the registry
+const Name = "stdout_logger"
```
dfawley, line 35: Are all loggers going to be registered with a `_logger` suffix?  Isn't this redundant?  Is this standardized across languages or no?

internal/xds/rbac/converter_test.go
```diff
@@ -110,5 +112,60 @@ func (s) TestBuildLoggerErrors(t *testing.T) {
...
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			logger, err := buildLogger(test.loggerConfig)
+			if err != nil {
+				t.Fatalf("expected success. got error: %v", err)
+			} else {
+				switch test.name {
+				case "stdout logger":
```
dfawley, line 150: I don't think switching behavior on the testcase name is proper.  The name is there for descriptive purposes only.  Everything else should be covered by a parameter in the testcase struct.

In this case, if you want to test that the proper logger was built, you should be able to find the logger's type by looking it up in the registry, too, instead of exporting it.

Is this not something that should be tested for the other test case, too?

authz/audit/stdout/stdout_logger.go
rockspore: While I don't have strong preference on whether to keep this suffix, we are using the same name in [C++](https://github.com/grpc/grpc/blob/0d3ecd3e63cf3520d0d65b1f797cf54c1e97ac03/src/core/lib/security/authorization/stdout_logger.cc#L39) and it will be the same across languages such that the same authz policy can work everywhere.

gtcooke94: This was previously hardcoded [here](https://github.com/grpc/grpc-go/blob/a6e1acfc4420debbf347b0730e69d25b0a9c69ac/authz/audit/stdout/stdout_logger.go#L78)

I was just pulling it out to be an exported const so it could be looked up.

I'd be okay to change the name to `stdout`, what do you think @rockspore

internal/xds/rbac/converter_test.go
gtcooke94: Yes it should be tested for the other test case, only keeping it out was a holdover from when I had it combined with the other test in this file.
Will clean it up

gtcooke94: I'm not 100% clear what you mean by getting the correct logger type from the registry.

But that's also something we can kick down the road because there's only one logger type right now, so I can leave it hard-coded in as it is right now if that is okay with you

internal/xds/rbac/converter.go
arvindbr8: `udpaTypedStuctType` -> `udpaTypedStructType`

gtcooke94: Ah thank you! Fixed

internal/xds/rbac/converter_test.go
```diff
@@ -110,5 +112,56 @@ func (s) TestBuildLoggerErrors(t *testing.T) {
...
+				},
+				IsOptional: false,
+			},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			logger, err := buildLogger(test.loggerConfig)
+			if err != nil {
+				t.Fatalf("expected success. got error: %v", err)
```
dfawley, line 147: Since this is a fatal, you don't need the `else` below.

authz/audit/stdout/stdout_logger.go
dfawley: Oh wait..should this just not even be in the registry at all?  It seems like it's a built-in type that we support and not something that is supposed to be supported via the registry for generic loggers?

internal/xds/rbac/converter.go
```diff
@@ -60,22 +66,33 @@ func buildLogger(loggerConfig *v3rbacpb.RBAC_AuditLoggingOptions_AuditLoggerConf
 
 func getCustomConfig(config *anypb.Any) (json.RawMessage, string, error) {
 	switch config.GetTypeUrl() {
```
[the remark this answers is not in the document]
gtcooke94: I like that a lot more, changing

internal/xds/rbac/converter_test.go
gtcooke94: PTAL

internal/xds/rbac/converter.go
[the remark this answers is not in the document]
gtcooke94: PTAL

internal/xds/rbac/converter_test.go
```diff
@@ -110,5 +112,56 @@ func (s) TestBuildLoggerErrors(t *testing.T) {
...
+			} else {
+				_, ok := logger.(*stdout.Logger)
+				if !ok {
```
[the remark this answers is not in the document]
gtcooke94: Changed the structure of the test so this is no longer there at all

authz/audit/stdout/stdout_logger.go
gtcooke94: I believe the intention with the built in types is that we place them in the registry for use - https://github.com/grpc/grpc-go/blob/a6e1acfc4420debbf347b0730e69d25b0a9c69ac/authz/audit/stdout/stdout_logger.go#L34-L38
@rockspore please check me here

rockspore: > I'd be okay to change the name to `stdout`, what do you think @rockspore

If there is no strong preference, I'm leaning towards keeping it as is so that we don't have to change what's already in C++ and the latest gRFC PR.

> Oh wait..should this just not even be in the registry at all?

We use the same registry for built-in loggers. We just pre-register them by importing the pkg here. The registry API probably needs to be prevent the built-in types from being overwritten.

dfawley: > We use the same registry for built-in loggers. We just pre-register them by importing the pkg here. The registry API probably needs to prevent the built-in types from being overwritten.

This doesn't sound right to me.

I think the built-in types should be hard-coded and not be present in the registry.  IIUC the registry is for the custom types that users can implement.  AFAICT there's no need to put the built-in types into the registry.  It buys us nothing and seems to have some downsides.

gtcooke94: I think this is an important but tangential-to-this-pr point to iron out.

@dfawley how would you feel about keeping this PR going as if the current behavior is the right behavior.
Then, we can resolve this separately, and if we change how the stdout logger is constructed we will update that in a separate PR.
As the audit logging is currently implemented, this is the correct way to do it.

gtcooke94: Right, we tried to stay similar to the LB policy for consistency

> But they're separate because we intend for them to be used through other pathways than xDS, right?

Yes, it can come through [here](https://github.com/grpc/grpc-go/blob/59134c303c31ba2ce65da8962c050918098f2a25/authz/grpc_authz_server_interceptors.go#L51) through `NewStatic`

internal/xds/rbac/converter_test.go
```diff
@@ -169,3 +167,18 @@ func createStdoutPb(t *testing.T) *anypb.Any {
 	}
 	return customConfig
 }
+
+// Builds a config with a nonsensical type in the anypb.Any.
+func createUnsupportedPb(t *testing.T) *anypb.Any {
```
dfawley, line 172: I think you might be able to re-use https://github.com/grpc/grpc-go/blob/59134c303c31ba2ce65da8962c050918098f2a25/internal/testutils/marshal_any.go#L30 instead to simplify a little?

dfawley (approved):

internal/xds/rbac/converter_test.go
gtcooke94: Changed

internal/xds/rbac/converter_test.go
```diff
@@ -175,10 +176,5 @@ func createUnsupportedPb(t *testing.T) *anypb.Any {
 	// proto that is not accepted in our custom config parsing. This was chosen
 	// because it is already imported.
 	pb := &v3rbacpb.RBAC_AuditLoggingOptions{}
-	customConfig, err := anypb.New(pb)
-	if err != nil {
-		t.Fatalf("createStdoutPb failed during anypb.New: %v", err)
-	}
-	return customConfig
-
+	return testutils.MarshalAny(pb)
```
dfawley, line 179: I think you missed the point.  Remove `createUnsupportedPb` and replace it with `testutils.MarshalAny(&v3rbacpb.RBAC_AuditLoggingOptions{})`.  But this is fine too if you want, just remove the `t` parameter to this function since it's effectively unused.

PR #6304, opened 2023-05-22
authz: End2End test for AuditLogger

erm-g:
End to end tests covering

1. AuditLogging invocation testing
2. Audit on Allow, Audit on Deny, Audit All, No Auditing
3. SPIFFE Ids from new cert

RELEASE NOTES: none


authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,304 @@
...
+	testgrpc.UnimplementedTestServiceServer
+}
+
+func (s *testServer) UnaryCall(ctx context.Context, req *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+	return &testpb.SimpleResponse{}, nil
+}
+
+type statAuditLogger struct {
+	Stat         map[bool]int
+	stdoutLogger audit.Logger
```
rockspore, line 34: If we are not doing anything with it, why are we including the stdout logger?

rockspore (changes requested):
Left some initial comments. Please fix the tests.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,304 @@
...
+type testServer struct {
+	testgrpc.UnimplementedTestServiceServer
+}
+
+func (s *testServer) UnaryCall(ctx context.Context, req *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+	return &testpb.SimpleResponse{}, nil
+}
+
+type statAuditLogger struct {
+	Stat         map[bool]int
```
gtcooke94, line 33: Let's name this something more meaningful - from the impl of `Log` it looks like it is a count of the Authorized vs. Unauthorized calls, mapped to `true` and `false` respectively?

Also, I think the map entries would be more clear if they were just string keys like `allow` and `deny` instead of true and false

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,304 @@
...
+	Stat      map[bool]int
+	SpiffeIds []string
+}
+
+func (loggerBuilder) Name() string {
+	return "stat_logger"
+}
+func (lb *loggerBuilder) Build(audit.LoggerConfig) audit.Logger {
+	return &statAuditLogger{
+		Stat:         lb.Stat,
```
gtcooke94, line 58: Do we need to pass through pointers/addresses here?

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,304 @@
...
+				Certificates: []tls.Certificate{cert},
+				ClientCAs:    certPool,
+			})
+			s := grpc.NewServer(
+				grpc.Creds(creds),
+				grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+				grpc.ChainStreamInterceptor(i.StreamInterceptor))
+			defer s.Stop()
+			testgrpc.RegisterTestServiceServer(s, &testServer{})
+			lis, err := net.Listen("tcp", "localhost:0")
```
gtcooke94, line 239: Is it standard in these tests to run on `0`? Usually the standards are somewhere around `8xxx`

gtcooke94 (commented):
A few small comments

authz/grpc_audit_end2end_test.go
rockspore: By putting 0 here, the system will automatically bind a free port. This allows multiple tests running in parallel.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (s *testServer) UnaryCall(ctx context.Context, req *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+	return &testpb.SimpleResponse{}, nil
+}
+
+type statAuditLogger struct {
+	Stat         map[bool]int
+	stdoutLogger audit.Logger
+	SpiffeIds    []string
```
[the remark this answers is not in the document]
erm-g: We're checking all the fields during unit testing so I don't think we need to duplicate it here. I try to concentrate on more global checks for this test. To be honest, I don't even think that SPIFFE belongs here but since there is a lack of tests for it I included it.

authz/grpc_audit_end2end_test.go
erm-g: I have the following 2 thoughts 

1. Making sure that `stdoutLogger` can be referenced from outside it's package and usable externally (unit tests don't cover it)
2. It execution doesn't fail while external usage (details of execution are covered by unit tests). 

So for a second theoretical xDS logger I'd also like to include it here

authz/grpc_audit_end2end_test.go
erm-g: IIUC we don't need but it's a kind of good default. @rockspore WDYT?

rockspore: Go maps are passed by references already. We don't need to worry about it being copied here.

authz/grpc_audit_end2end_test.go
rockspore: In that case, can we simply add `stdout_logger` as another logger in the policy so that we are also testing multiple loggers. Because embedding a logger is the implementation detail of a custom logger that we don't really need to cover in our tests.

authz/grpc_audit_end2end_test.go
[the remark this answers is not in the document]
rockspore: Technically I understand they don't belong here but I still think it's beneficial to verify all the fields since we mocked them in our unit tests. We would thank ourselves if some change in the future altered the behavior and the was the only failing test.

We don't have to be that exhaustive so if you could at least verify all fields in one test case I think that'd be decent.

authz/grpc_audit_end2end_test.go
erm-g: I renamed it plus added comments but I'd prefer to keep it as map[bool]. It just fits more naturally to Event.Authorized field

authz/grpc_audit_end2end_test.go
```diff
@@ -30,33 +41,38 @@ func (s *testServer) UnaryCall(ctx context.Context, req *testpb.SimpleRequest) (
 }
 
 type statAuditLogger struct {
-	Stat         map[bool]int
-	stdoutLogger audit.Logger
-	SpiffeIds    []string
+	AuthzDescisionStat map[bool]int      //Map to hold the counts of authorization decisions
```
gtcooke94, line 44: Golang will want periods at the end of comments IIRC

authz/grpc_audit_end2end_test.go
```diff
@@ -66,7 +82,7 @@ func (*loggerBuilder) ParseLoggerConfig(config json.RawMessage) (audit.LoggerCon
 
 const spiffeId = "spiffe://foo.bar.com/client/workload/1"
 
-func TestAuditLogger(t *testing.T) {
+func (s) TestAuditLogger(t *testing.T) {
```
gtcooke94, line 85: What is the `s` here?

authz/grpc_audit_end2end_test.go
```diff
@@ -287,18 +308,37 @@ func TestAuditLogger(t *testing.T) {
...
+				}
+			}
 		})
 	}
 }
+
+// generateEvent produces an map contaning audit.Event fields.
+// It's used to compare captured audit.Event with the matched rule during
+// `Allow All Deny Streaming - Audit All` scenario (authz_deny_all rule)
+func generateEventAsMap() map[string]string {
```
gtcooke94, line 336: Two alternatives to this
1. Just hard code it, you don't really need a function that returns a constant hard-coded map
2. Add `events map[string]string` as part of the table tests, then it's a part of test construction. Code it in as part of the test case, ignore it if it doesn't matter. Then you can do the `diff := cmp.Diff(...)` for this if it exists, rather than checking for a specific test name

authz/grpc_audit_end2end_test.go
gtcooke94: This is good with me now

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+type statAuditLogger struct {
+	AuthzDescisionStat map[bool]int      //Map to hold the counts of authorization decisions
+	EventContent       map[string]string //Map to hold event fields in key:value fashion
+	SpiffeIds          []string          //Slice to hold collected SPIFFE IDs
+}
+
+func (s *statAuditLogger) Log(event *audit.Event) {
+	if event.Authorized {
+		s.AuthzDescisionStat[true]++
```
rockspore, line 51: nit: Could do `s.AuthzDecisionStat[event.Authorized]++`. Also there is a typo for `Decision` here.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,344 @@
...
+	} else {
+		s.AuthzDescisionStat[false]++
+	}
+	s.SpiffeIds = append(s.SpiffeIds, event.Principal)
```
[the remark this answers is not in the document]
erm-g: I'm checking EventContent only once but I'd still prefer to check SPIFFE every time

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,344 @@
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/internal/grpctest"
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+	testpb "google.golang.org/grpc/interop/grpc_testing"
+	"google.golang.org/grpc/testdata"
+
+	_ "google.golang.org/grpc/authz/audit/stdout"
+)
+
+type s struct {
```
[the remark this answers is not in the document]
erm-g: That's an interesting one. I just realized that authz related tests belong to `authz_test` package but not to `authz` one. 
Refactored

authz/grpc_audit_end2end_test.go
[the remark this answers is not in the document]
rockspore: Is there any specific reason behind that? I thought if Spiffe ID is populated in one case, we can safely assume it's already there.

authz/grpc_audit_end2end_test.go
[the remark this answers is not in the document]
rockspore: jFYI, it's common to see `..._test` particularly for e2e test because we want the test target to have access to only exposed APIs.

rockspore (commented):
Please fix the vet testing failure.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
+/*
+ *
+ * Copyright 2021 gRPC authors.
```
rockspore, line 3: micronit: 2023

rockspore (approved):
LGTM. Thanks!

authz/grpc_audit_end2end_test.go
```diff
@@ -8,69 +26,67 @@ import (
...
+	s.EventContent["principal"] = event.Principal
+	s.EventContent["policy_name"] = event.PolicyName
+	s.EventContent["matched_rule"] = event.MatchedRule
+	s.EventContent["authorized"] = strconv.FormatBool(event.Authorized)
 }
 
 type loggerBuilder struct {
-	Stat      map[bool]int
-	SpiffeIds []string
+	AuthDecisionStat map[bool]int
```
gtcooke94, line 66: nit: `AuthzDecisionStat` to be consistent with everything else

gtcooke94 (approved):
Saw one typo and LGTM otherwise

dfawley:
> RELEASE NOTES:
> - TBD

Please fill in release notes when creating the PR (or asap after noticing they're missing due to failing checker).  This should be part of what is reviewed as part of the PR review.  In this case, it's just for testing, so we use `RELEASE NOTES: none` (or `n/a`) since our users don't really care much about our tests (they just assume we do a good enough job with them).

dfawley:
Also: only one `Type` label should be applied (pick the best one).  In this case, `Testing` makes the most sense since you aren't addressing a security-related issue, but it's not actually all that important if there are no release notes.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
+/*
```
dfawley, line 1: Nit: let's see if we can rename this file to something shorter.  At least remove the `grpc_` prefix since this is the grpc repo and it's unnecessary.  I'm not sure `end2end` is all that necessary, either, especially since these are the only audit logging tests.

Is this feature "auditing" or "audit logging"?  If the former: `audit_test.go` seems fine.  If the latter: `audit_logging_test.go` or `audit_log_test.go` or something like that works.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+	SpiffeIds          []string          // Slice to hold collected SPIFFE IDs
+}
+
+func (s *statAuditLogger) Log(event *audit.Event) {
+	s.AuthzDescisionStat[event.Authorized]++
+	s.EventContent["rpc_method"] = event.FullMethodName
+	s.EventContent["principal"] = event.Principal
+	s.EventContent["policy_name"] = event.PolicyName
+	s.EventContent["matched_rule"] = event.MatchedRule
+	s.EventContent["authorized"] = strconv.FormatBool(event.Authorized)
```
dfawley, line 62: Why not `s.EventContent = event` and make this field type `*audit.Event`?  (Also consider a rename to `lastEvent`?)

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+)
+
+func TestAudit(t *testing.T) {
+	grpctest.RunSubTests(t, s{})
+}
+
+type statAuditLogger struct {
+	AuthzDescisionStat map[bool]int      // Map to hold the counts of authorization decisions
+	EventContent       map[string]string // Map to hold event fields in key:value fashion
+	SpiffeIds          []string          // Slice to hold collected SPIFFE IDs
```
dfawley, line 53: Are these fields exported for a reason?

Also for Go style this needs to be: `SPIFFEIDs` or `spiffeIDs` if unexported.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+	s.AuthzDescisionStat[event.Authorized]++
+	s.EventContent["rpc_method"] = event.FullMethodName
+	s.EventContent["principal"] = event.Principal
+	s.EventContent["policy_name"] = event.PolicyName
+	s.EventContent["matched_rule"] = event.MatchedRule
+	s.EventContent["authorized"] = strconv.FormatBool(event.Authorized)
+}
+
+type loggerBuilder struct {
+	AuthDecisionStat map[bool]int
```
dfawley, line 66: Same: unexport these fields, please.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+	EventContent     map[string]string
+}
+
+func (loggerBuilder) Name() string {
+	return "stat_logger"
+}
+func (lb *loggerBuilder) Build(audit.LoggerConfig) audit.Logger {
```
dfawley, line 73: Blank line between functions please.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+		AuthzDescisionStat: lb.AuthDecisionStat,
+		EventContent:       lb.EventContent,
+	}
+}
+
+func (*loggerBuilder) ParseLoggerConfig(config json.RawMessage) (audit.LoggerConfig, error) {
+	return nil, nil
+}
+
+func (s) TestAuditLogger(t *testing.T) {
```
dfawley, line 84: Please add a comment with a brief description of the test.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+						{
+							"name": "stat_logger",
+							"config": {},
+							"is_optional": false
+						}
+					]
+				}
+			}`,
+			wantAllows: 0,
+			wantDenies: 0,
```
dfawley, line 119: These definitions all a long way come before I have any idea what the test is doing.  And when I scroll down, it seems like the test does a lot (more than a page worth of code is there).  Can there be a high level description of what happens during a test case where the `tests` struct is defined?  Just as simple as "each test case makes 2 unary calls and one client-streaming call" is fine.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+		authzPolicy  string
+		wantAllows   int
+		wantDenies   int
+		eventContent map[string]string
+	}{
+		"No audit": {
+			authzPolicy: `{
+				"name": "authz",
+				"allow_rules":
+				[
```
dfawley, line 95: Can you use a consistent pattern for the newlines?  Sometimes things appear on the same line (`"audit_loggers": [`) and sometimes they get newlines (here).  I would definitely prefer things more like `audit_loggers` and `audit_logging_options` over putting a newline before the open brace/bracket.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+			certPool := x509.NewCertPool()
+			if !certPool.AppendCertsFromPEM(ca) {
+				t.Fatal("failed to append certificates")
+			}
+			creds := credentials.NewTLS(&tls.Config{
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+				Certificates: []tls.Certificate{cert},
+				ClientCAs:    certPool,
+			})
```
dfawley, line 260: This seems to be the same in every test case... I would move it outside the loop.  Optionally: same thing with the logger builder, unless you're concerned about state leaking.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+		EventContent:       lb.EventContent,
+	}
+}
+
+func (*loggerBuilder) ParseLoggerConfig(config json.RawMessage) (audit.LoggerConfig, error) {
+	return nil, nil
+}
+
+func (s) TestAuditLogger(t *testing.T) {
+	tests := map[string]struct {
```
dfawley, line 85: Nit: I'm not sure I like this pattern all that much (`map` instead of a slice), and I don't think I've seen it before, either.  It forces even more indentation than there already is, and that makes the test cases harder to read.  Typically we do:

```go
tests := []struct{
  name string
  .....
}{
    {
        name: "..."
    },
    {
        name: "..."
    }
}

// OR tighter spacing which I prefer but others may not.

tests := []struct{
  name string
  .....
}{{
    name: "..."
}, {
    name: "..."
}}
```

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+				Certificates: []tls.Certificate{cert},
+				ClientCAs:    certPool,
+			})
+			s := grpc.NewServer(
+				grpc.Creds(creds),
+				grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+				grpc.ChainStreamInterceptor(i.StreamInterceptor))
+			defer s.Stop()
+			testgrpc.RegisterTestServiceServer(s, &testServer{})
```
dfawley, line 266: Does `testServer` have some magic behavior you need?

I would greatly prefer the use of `grpc/internal/stubserver` instead.  See `grpc/test/*` for example usages.  It should be really easy to use and not couple the behavior of this test and the service's implementation which is in another location.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+			lis, err := net.Listen("tcp", "localhost:0")
+			if err != nil {
+				t.Fatalf("error listening: %v", err)
+			}
+			go s.Serve(lis)
+
+			// Setup gRPC test client with certificates containing a SPIFFE Id.
+			cert, err = tls.LoadX509KeyPair(testdata.Path("x509/client_with_spiffe_cert.pem"), testdata.Path("x509/client_with_spiffe_key.pem"))
+			if err != nil {
+				t.Fatalf("tls.LoadX509KeyPair(x509/client1_cert.pem, x509/client1_key.pem) failed: %v", err)
```
dfawley, line 276: This error message doesn't match the code.

Two options:  1. fix it, or 2. use a helper to fix this and everything else forever and always:

```go
func loadKeys(t *testing.T, cert, key string) tls.WhateverTypeThisIs {
  t.Helper()
  cert, err := tls.LoadX509KeyPair(testdata.Path(cert), testdata.Path(key))
  if err != nil {
    t.Fatalf("tls.LoadX509KeyPair(%q, %q) failed: %v", cert, key, err)
  }
  return cert
}

File: authz/grpc_audit_end2end_test.go [hunk 26]
```diff
@@ -0,0 +1,332 @@
...
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+
+			// Make 2 unary calls and 1 streaming call.
+			client.UnaryCall(ctx, &testpb.SimpleRequest{})
+			client.UnaryCall(ctx, &testpb.SimpleRequest{})
+			stream, err := client.StreamingInputCall(ctx)
+			if err != nil {
+				t.Fatalf("failed StreamingInputCall err: %v", err)
+			}
```
dfawley, line 306: Why do you check this error but not any error that could be returned by `UnaryCall`?

Also, starting a stream almost never fails.  The way to get _that_ error is to capture it from `CloseAndRecv`.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+			clientConn, err := grpc.Dial(lis.Addr().String(), grpc.WithTransportCredentials(creds))
+			if err != nil {
+				t.Fatalf("grpc.Dial(%v) failed: %v", lis.Addr().String(), err)
+			}
+			defer clientConn.Close()
+			client := testgrpc.NewTestServiceClient(clientConn)
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+
+			// Make 2 unary calls and 1 streaming call.
```
dfawley, line 300: By the time I got here reading the code, it's pretty obvious what is happening.  😄  IMO: delete this comment but put something like this up at the top of the function.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+			req := &testpb.StreamingInputCallRequest{
+				Payload: &testpb.Payload{
+					Body: []byte("hi"),
+				},
+			}
+			if err := stream.Send(req); err != nil && err != io.EOF {
+				t.Fatalf("failed stream.Send err: %v", err)
+			}
+			stream.CloseAndRecv()
```
dfawley, line 315: This needs to check the error.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+			// Compare expected number of allows/denies with content of internal map of statAuditLogger.
+			if lb.AuthDecisionStat[true] != test.wantAllows {
+				t.Errorf("Allow case failed, want %v got %v", test.wantAllows, lb.AuthDecisionStat[true])
+			}
+			if lb.AuthDecisionStat[false] != test.wantDenies {
+				t.Errorf("Deny case failed, want %v got %v", test.wantDenies, lb.AuthDecisionStat[false])
+			}
```
dfawley, line 323: Maybe define `map[bool]int` as a `want` in the testcase struct instead?  Or replace the map in the logger with 2 bools so it's clearer that this is comparing equivalent things.

authz/grpc_audit_end2end_test.go
```diff
@@ -0,0 +1,332 @@
...
+			// Compare event fields with expected values from authz policy.
+			if test.eventContent != nil {
```
dfawley, line 325: Only one test case actually covers this.  And it only captures the most recent event.  Is that sufficient?

testdata/x509/create.sh
```diff
@@ -128,5 +128,24 @@ openssl req -x509                                                         \
   -addext "subjectAltName = URI:spiffe://foo.bar.com/client/workload/1, URI:https://bar.baz.com/client" \
   -sha256
 
+# Generate a cert with SPIFFE ID using client_with_spiffe_openssl.cnf
+openssl req -new                                    \
```
dfawley, line 132: Nit: please line up your backslashes like in the rest of the file.

authz/grpc_audit_end2end_test.go
erm-g: Here EventContent is a map created at `loggerBuilder` but manipulated by `statAuditLogger`. At the end I'm using `loggerBuilder` to access it's content. I'm not sure how to achieve the same using `*audit.Event` - maybe only if I use a slice or map wrapping it?

authz/audit_logging_test.go
```diff
@@ -0,0 +1,335 @@
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/internal/grpctest"
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+	testpb "google.golang.org/grpc/interop/grpc_testing"
+	"google.golang.org/grpc/testdata"
+
+	_ "google.golang.org/grpc/authz/audit/stdout"
+)
+
+func TestAudit(t *testing.T) {
```
dfawley, line 46: Please just `Test`, which is our convention.

authz/grpc_audit_end2end_test.go
dfawley: I DON"T UNDERSTAND THIS>  Why can't you just change the type of this field (now `lastEventContent`) to `*audit.Event` and rename the field to `lastEvent`?  Are you saying the tricky bit is communicating between the test and the constructed logger?  If so, 2 options:

1. Use `*audit.Event` and `*s.lastEvent = *event`
2. Use `**audit.Event` and `*s.lastEvent = event`

authz/audit_logging_test.go
```diff
@@ -0,0 +1,335 @@
...
+	return nil, nil
+}
+
+// TestAuditLogger examines audit logging invocations using four different authorization policies.
+// It covers scenarios including a disabled audit, auditing both 'allow' and 'deny' outcomes,
+// and separately auditing 'allow' and 'deny' outcomes.
+// Additionally, it checks if SPIFFE ID from a certificate is propagated correctly.
+func (s) TestAuditLogger(t *testing.T) {
+	// Each test data entry contains an authz policy for a grpc server,
+	// how many 'allow' and 'deny' outcomes we expect (each test case makes 2 unary calls and one client-streaming call),
```
dfawley, line 90: Please wrap all comments at 80 columns (every file and always).  Thanks!

authz/audit_logging_test.go
```diff
@@ -0,0 +1,335 @@
...
+	}{
+		{
+			name: "No audit",
+			authzPolicy: `{
+				"name": "authz",
+				"allow_rules": [
+					{
+						"name": "allow_UnaryCall",
+						"request":
+						{
```
dfawley, line 106: Missed one

authz/audit_logging_test.go
```diff
@@ -0,0 +1,335 @@
...
+							]
+						}
+					}
+				],
+				"deny_rules": [
+					{
+						"name": "deny_all",
+						"request": {
+							"paths":
+							[
```
dfawley, line 145: Missed one

authz/audit_logging_test.go
```diff
@@ -0,0 +1,335 @@
...
+		},
+		{
+			name: "Allow Unary - Audit Allow",
+			authzPolicy: `{
+				"name": "authz",
+				"allow_rules": [
+					{
+						"name": "allow_UnaryCall",
+						"request":
+						{
```
dfawley, line 183: Missed one

authz/grpc_audit_end2end_test.go
erm-g: The idea is that we already tested event content propagation in unit tests. Here it's an extra check because of end to end scenario including certificate with a SPIFFE ID. That's why I'm doing it just once

authz/grpc_audit_end2end_test.go
erm-g: That's a good point - I think the nature of the test is literally testing audit part of authz policy so I think just making calls and ignoring errors is fine.

authz/grpc_audit_end2end_test.go
erm-g: I'm leaning towards ignoring the errors here - https://github.com/grpc/grpc-go/pull/6304#discussion_r1207428721

authz/grpc_audit_end2end_test.go
dfawley: Wow what happened with my last comment here??  Sometimes the shift key sticks on my keyboard, but I can't believe I didn't see it.  Sorry if that came off as rude, I didn't mean to type in all-caps.

erm-g: No worries!

authz/audit/audit_logging_test.go
```diff
@@ -229,31 +227,46 @@ func (s) TestAuditLogger(t *testing.T) {
 			wantAuthzOutcomes: map[bool]int{true: 0, false: 3},
 		},
 	}
-
+	//Construct the credentials for the tests and the stub server
```
dfawley, line 230: Space after `//` please.

dfawley (commented):
Missed a couple recent changes

authz/audit/audit_logging_test.go
```diff
@@ -275,23 +279,26 @@ func (s) TestAuditLogger(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 			defer cancel()
 
-			client.UnaryCall(ctx, &testpb.SimpleRequest{})
-			client.UnaryCall(ctx, &testpb.SimpleRequest{})
+			_, err = client.UnaryCall(ctx, &testpb.SimpleRequest{})
+			validateCallResult(t, err)
```
dfawley, line 283: Inline this please, since it's so simple:

```go
if _, err := client.UnaryCall(ctx, &testpb.SimpleRequest{}); err != nil {
	t.Fatalf("Unexpected error from UnaryCall: %v", err)
}
```

erm-g: I also need to check if the err is a `permission denied` one - it's a valid scenario for the test (we count `deny` as well).
I think I can rely on `code` only but still it'll be 2 LOCs

```go
_, err = client.UnaryCall(ctx, &testpb.SimpleRequest{})
if errCode := status.Code(err); errCode != permissionDeniedStatus.Code() && errCode != codes.OK {
	t.Errorf("Call failed:%v", err)
}
```
Or I can create a helper function which takes `SimpleResponse, error` and returns back error code - smth like
```go
if errCode := errorCodeExtractor(client.UnaryCall(ctx, &testpb.SimpleRequest{})); errCode != permissionDeniedStatus.Code() && errCode != codes.OK {
	t.Errorf("Call failed:%v", err)
}
```
WDYT?

dfawley: So this is more of a problem of your test design.. You're using a table driven test but things are sufficiently different between iterations that it leads to these sorts of problems.

If you want to not care which one should happen:

```go
if _, err := UnaryCall(); err != nil && status.Code(err) != codes.PermissionDenied {
	t.Fatalf()
}
```

Or if you want to require no error when unary should not be denied and vice-versa:

```go
_, err := UnaryCall()
if test.wantAuthzOutcomes[false] > 1 /* LOL */ && status.Code(err) != codes.PermissionDenied {
	t.Fatalf()
} else if test.wantAuthzOutcomes[false] < 2 /* also LOL */ && err != nil {
	t.Fatalf()
}
```

Realistically, the latter is how your code should behave, but it's ugly.  Maybe the test should be reworked a bit or split into multiple tests, as this seems wrong.

dfawley: > My idea is to have a set of authz policies with various audit_condition values. In order to test each audit_condition I need to make a few calls and then check if audit logger was invoked as expected. That's why I think that iterations are common enough to use a table driven approach. From my perspective

Maybe add a `codes.Code` for the expected result from the Unary and Streaming RPC calls to the table?

erm-g: Agreed, I think it makes the test more readable. PTAL

authz/audit/audit_logging_test.go
```diff
@@ -279,19 +288,24 @@ func (s) TestAuditLogger(t *testing.T) {
...
-			validateCallResult(t, err)
-			_, err = client.UnaryCall(ctx, &testpb.SimpleRequest{})
-			validateCallResult(t, err)
+			if _, err := client.UnaryCall(ctx, &testpb.SimpleRequest{}); status.Code(err) != test.wantUnaryCallCode {
+				t.Fatalf("Unexpected UnaryCall fail: got %v want %v", err, test.wantUnaryCallCode)
+			}
+			if _, err := client.UnaryCall(ctx, &testpb.SimpleRequest{}); status.Code(err) != test.wantUnaryCallCode {
+				t.Fatalf("Unexpected UnaryCall fail: got %v want %v", err, test.wantUnaryCallCode)
+			}
 			stream, _ := client.StreamingInputCall(ctx)
```
dfawley, line 297: Please check the error.

dfawley (approved):

authz/audit/audit_logging_test.go
```diff
@@ -289,12 +289,15 @@ func (s) TestAuditLogger(t *testing.T) {
...
-				t.Fatalf("Unexpected UnaryCall fail: got %v want %v", err, test.wantUnaryCallCode)
+				t.Errorf("Unexpected UnaryCall fail: got %v want %v", err, test.wantUnaryCallCode)
 			}
 			if _, err := client.UnaryCall(ctx, &testpb.SimpleRequest{}); status.Code(err) != test.wantUnaryCallCode {
-				t.Fatalf("Unexpected UnaryCall fail: got %v want %v", err, test.wantUnaryCallCode)
+				t.Errorf("Unexpected UnaryCall fail: got %v want %v", err, test.wantUnaryCallCode)
+			}
+			stream, err := client.StreamingInputCall(ctx)
+			if err != nil {
+				t.Errorf("StreamingInputCall failed:%v", err)
```
dfawley, line 299: This one needs to be `Fatal` because using the `stream` is illegal.

erm-g: I need some help here - go/go-style/decisions.md?cl=head#keep-going suggests to prefer `Error` over `Fatal`

>to print out all of the failed checks in a single run

So for this test even if stream is not usable I can still proceed with unary calls to get the test results for it.
However the same doc suggests

>Calling t.Fatal is primarily useful for reporting an unexpected error condition

and unusable stream is definitely something we don't expect. 
WDYT?

dfawley: True, `Error` is nicer, _generally_, but this is a precondition for (some of) the rest of the test.  The "ideal" would be to skip just the streaming calls and the checks that depend on them happening, but that really clutters the code and isn't worth it _at all_.  Just `Fatal` since a precondition failed and keep things simple.

erm-g: Ok, sounds good - I changed `Errorf` to be used for

- `cmp.Diff` cases
- errors which `status.Code` is not expected

dfawley (approved):

PR #6307, opened 2023-05-22
authz: use pointer to to structpb.Struct instead of value

arvindbr8:
GoProtoAPI tricorder reported this issue in our weekly import CL. Using the value as is in the struct's type field is discouraged because they inherently lend themselves to shallow copies of messages later on. This diff addresses that, and also needs additional logic to check for nil struct Config.

RELEASE NOTES: none

### DIFF
--- a/authz/rbac_translator.go
+++ b/authz/rbac_translator.go
@@ -62,9 +62,9 @@ type rule struct {
 }
 
 type auditLogger struct {
-	Name       string          `json:"name"`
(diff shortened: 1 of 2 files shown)


gtcooke94 (approved):
LGTM

gtcooke94:
@rockspore can you take a quick look as well

rockspore (approved):

zasweq (approved):
LGTM.

PR #6299, opened 2023-05-18
benchmark: Add sleepBetweenRPCs and connections parameters

s-matyukevich:
I am trying to fix https://github.com/grpc/grpc-go/issues/5759 and have a branch with some promising results. However, the problem is that our current benchmarks don't really test the case I am trying to address. The case is the following: we have some grpc servers with a lot of mostly idle connections (in our case those connections mostly come from healtchecks) In this case grpc-go wastes a lot of memory on transport buffers, that aren't shared between connections. 

This PR modifies benchmarks so that we can simulate this case. It adds a random delay between subsequent RPCs and a configuration parameter which controls the maximum value of this delay. It also adds a parameter that allows configuring the number of connections. 

RELEASE NOTES: none

PR #6309, opened 2023-05-23
Use sync.Pool to share per-connection transport write buffer. 

s-matyukevich:
This is an attempt to fix https://github.com/grpc/grpc-go/issues/5759.

The use-case where this fix could be especially useful is a server with a lot of mostly idle connections (like connections that handle only healthchecks) In our infrastructure this use-case is very common, as we have some servers that handle request from thousands of clients, all of which generate relatively small volume of requests but has an active healthcheck setup, which forces every client connect to every server.  Some kind of subsetting would be useful in such scenario, but gRPC doesn't really implement subsetting. Grpc-go isn't great at handling memory in such use-case. Right now a lot of our users put envoy in front of grpc server just because it handles memory per connection a lot more efficiently (in one case we were able to reduce the memory consumed by an app from 1.5 GB to 500 MB just by putting envoy in front of the grpc server) 

This PR adds sync.Pool to the transport layer to share the write buffer between connections (I plan to provide a similar fix for the read buffer if this PR will be accepted) The buffer is claimed in the Write method and released on Flush, with an additional optimization that preserves the buffer if a single Write results in multiple Flushes. 

**Testing**

I submitted https://github.com/grpc/grpc-go/pull/6299 so that I can reproduce the case described above with benchmarks. I used the following comnand to demonstrate the impact of this change

```
go run benchmark/benchmain/main.go -workloads all -kbps 0 -latency 0s -maxConcurrentCalls 1000 -reqSizeBytes 1 -respSizeBytes 1 -benchtime 1m -shareConnection=false -sleepBetweenRPCs=20s -resultFile /tmp/master1 -memProfile /tmp/master1-mem
```
As you can see, I used tiny requests with a combination of large delays between requests to minimize the impact of other allocations on memory stats. 

Here are the results that I got:

Without the optimization:
```
root@traffic-grpc-go-benchmark-manual:/tmp/grpc-go# go tool pprof -top -inuse_space ../master3-mem
File: main
Type: inuse_space
Time: May 23, 2023 at 12:36am (UTC)
Showing nodes accounting for 104.04MB, 95.85% of 108.55MB total
Dropped 24 nodes (cum <= 0.54MB)
      flat  flat%   sum%        cum   cum%
   59.04MB 54.39% 54.39%    59.04MB 54.39%  google.golang.org/grpc/internal/transport.newBufWriter (inline)
   32.49MB 29.94% 84.33%    32.49MB 29.94%  bufio.NewReaderSize (inline)
    5.50MB  5.07% 89.40%     5.50MB  5.07%  runtime.malg
       2MB  1.84% 91.24%        2MB  1.84%  google.golang.org/grpc/internal/transport.(*http2Server).operateHeaders
    1.50MB  1.38% 92.62%     1.50MB  1.38%  sync.(*Pool).pinSlow
       1MB  0.92% 93.55%        1MB  0.92%  bytes.growSlice
       1MB  0.92% 94.47%        1MB  0.92%  golang.org/x/net/http2/hpack.(*headerFieldTable).addEntry (inline)
    0.50MB  0.46% 94.93%    94.04MB 86.63%  google.golang.org/grpc/internal/transport.NewServerTransport
    0.50MB  0.46% 95.39%        1MB  0.92%  golang.org/x/net/http2/hpack.NewDecoder
    0.50MB  0.46% 95.85%     1.50MB  1.38%  bytes.(*Buffer).grow
         0     0% 95.85%        1MB  0.92%  bytes.(*Buffer).ReadFrom
         0     0% 95.85%        1MB  0.92%  golang.org/x/net/http2.(*Framer).ReadFrame
         0     0% 95.85%        1MB  0.92%  golang.org/x/net/http2.(*Framer).readMetaFrame
         0     0% 95.85%        1MB  0.92%  golang.org/x/net/http2/hpack.(*Decoder).Write
         0     0% 95.85%        1MB  0.92%  golang.org/x/net/http2/hpack.(*Decoder).parseFieldLiteral
         0     0% 95.85%        1MB  0.92%  golang.org/x/net/http2/hpack.(*Decoder).parseHeaderFieldRepr
         0     0% 95.85%        1MB  0.92%  golang.org/x/net/http2/hpack.(*dynamicTable).add
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc.(*Server).Serve
         0     0% 95.85%    94.04MB 86.63%  google.golang.org/grpc.(*Server).Serve.func3
         0     0% 95.85%    94.04MB 86.63%  google.golang.org/grpc.(*Server).handleRawConn
         0     0% 95.85%        5MB  4.61%  google.golang.org/grpc.(*Server).handleRawConn.func1
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc.(*Server).handleStream
         0     0% 95.85%    94.04MB 86.63%  google.golang.org/grpc.(*Server).newHTTP2Transport
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc.(*Server).processStreamingRPC
         0     0% 95.85%        5MB  4.61%  google.golang.org/grpc.(*Server).serveStreams
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc.(*Server).serveStreams.func1.1
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc/benchmark/latency.(*conn).Read
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc/benchmark/latency.(*listener).Accept
         0     0% 95.85%     1.50MB  1.38%  google.golang.org/grpc/internal/transport.(*bufferPool).get (inline)
         0     0% 95.85%        5MB  4.61%  google.golang.org/grpc/internal/transport.(*http2Server).HandleStreams
         0     0% 95.85%     1.50MB  1.38%  google.golang.org/grpc/internal/transport.(*http2Server).handleData
         0     0% 95.85%        1MB  0.92%  google.golang.org/grpc/internal/transport.NewServerTransport.func2
         0     0% 95.85%    92.54MB 85.25%  google.golang.org/grpc/internal/transport.newFramer
         0     0% 95.85%        1MB  0.92%  io.Copy (inline)
         0     0% 95.85%        1MB  0.92%  io.CopyN
         0     0% 95.85%        1MB  0.92%  io.ReadAtLeast
         0     0% 95.85%        1MB  0.92%  io.ReadFull (inline)
         0     0% 95.85%        1MB  0.92%  io.copyBuffer
         0     0% 95.85%        1MB  0.92%  net.(*TCPListener).Accept
         0     0% 95.85%        1MB  0.92%  net.(*TCPListener).accept
         0     0% 95.85%     5.50MB  5.07%  runtime.newproc.func1
         0     0% 95.85%     5.50MB  5.07%  runtime.newproc1
         0     0% 95.85%     5.50MB  5.07%  runtime.systemstack
         0     0% 95.85%     1.50MB  1.38%  sync.(*Pool).Get
         0     0% 95.85%     1.50MB  1.38%  sync.(*Pool).pin
```

With the optimization:
```
root@traffic-grpc-go-benchmark-manual:/tmp/grpc-go# go tool pprof -top -inuse_space ../fork3-mem
File: main
Type: inuse_space
Time: May 23, 2023 at 12:42am (UTC)
Showing nodes accounting for 55.53MB, 100% of 55.53MB total
      flat  flat%   sum%        cum   cum%
   31.98MB 57.59% 57.59%    31.98MB 57.59%  bufio.NewReaderSize (inline)
    8.50MB 15.31% 72.90%     8.50MB 15.31%  runtime.malg
    2.50MB  4.51% 77.41%     2.50MB  4.51%  sync.(*Pool).pinSlow
       2MB  3.60% 81.01%        2MB  3.60%  golang.org/x/net/http2/hpack.(*headerFieldTable).addEntry (inline)
    1.50MB  2.70% 83.71%        3MB  5.40%  google.golang.org/grpc/internal/transport.(*http2Server).operateHeaders
       1MB  1.80% 85.52%        1MB  1.80%  bytes.growSlice
       1MB  1.80% 87.32%        1MB  1.80%  context.(*cancelCtx).Done
       1MB  1.80% 89.12%    35.48MB 63.89%  google.golang.org/grpc/internal/transport.NewServerTransport
       1MB  1.80% 90.92%        1MB  1.80%  google.golang.org/grpc/internal/transport.newControlBuffer (inline)
    0.53MB  0.96% 91.88%     0.53MB  0.96%  google.golang.org/grpc/internal/transport.newFramer.func1
    0.50MB  0.91% 92.79%     0.50MB  0.91%  runtime.doaddtimer
    0.50MB  0.91% 93.70%     0.50MB  0.91%  google.golang.org/protobuf/reflect/protoregistry.(*Files).RegisterFile.func2
    0.50MB   0.9% 94.60%     0.50MB   0.9%  google.golang.org/grpc.(*Server).processStreamingRPC
    0.50MB   0.9% 95.50%     0.50MB   0.9%  golang.org/x/net/http2.NewFramer
    0.50MB   0.9% 96.40%     0.50MB   0.9%  golang.org/x/net/http2.NewFramer.func2
    0.50MB   0.9% 97.30%     0.50MB   0.9%  net.newFD (inline)
    0.50MB   0.9% 98.20%     0.50MB   0.9%  google.golang.org/grpc/internal/transport.(*loopyWriter).registerStreamHandler (inline)
    0.50MB   0.9% 99.10%     0.50MB   0.9%  google.golang.org/grpc/internal/transport.newLoopyWriter
    0.50MB   0.9%   100%     0.50MB   0.9%  context.WithValue
         0     0%   100%        1MB  1.80%  bytes.(*Buffer).ReadFrom
         0     0%   100%        1MB  1.80%  bytes.(*Buffer).grow
         0     0%   100%     1.50MB  2.70%  golang.org/x/net/http2.(*Framer).ReadFrame
         0     0%   100%     0.53MB  0.96%  golang.org/x/net/http2.(*Framer).WriteHeaders
         0     0%   100%     0.53MB  0.96%  golang.org/x/net/http2.(*Framer).endWrite
         0     0%   100%        1MB  1.80%  golang.org/x/net/http2.(*Framer).readMetaFrame
         0     0%   100%        1MB  1.80%  golang.org/x/net/http2/hpack.(*Decoder).Write
         0     0%   100%        1MB  1.80%  golang.org/x/net/http2/hpack.(*Decoder).parseFieldLiteral
         0     0%   100%        1MB  1.80%  golang.org/x/net/http2/hpack.(*Decoder).parseHeaderFieldRepr
         0     0%   100%        1MB  1.80%  golang.org/x/net/http2/hpack.(*Encoder).WriteField
         0     0%   100%        2MB  3.60%  golang.org/x/net/http2/hpack.(*dynamicTable).add
         0     0%   100%     0.50MB   0.9%  google.golang.org/grpc.(*Server).Serve
         0     0%   100%    35.48MB 63.89%  google.golang.org/grpc.(*Server).Serve.func3
         0     0%   100%    35.48MB 63.89%  google.golang.org/grpc.(*Server).handleRawConn
         0     0%   100%        6MB 10.81%  google.golang.org/grpc.(*Server).handleRawConn.func1
         0     0%   100%     0.50MB   0.9%  google.golang.org/grpc.(*Server).handleStream
         0     0%   100%    35.48MB 63.89%  google.golang.org/grpc.(*Server).newHTTP2Transport
         0     0%   100%        6MB 10.81%  google.golang.org/grpc.(*Server).serveStreams
         0     0%   100%     0.50MB   0.9%  google.golang.org/grpc.(*Server).serveStreams.func1.1
         0     0%   100%        1MB  1.80%  google.golang.org/grpc/benchmark/latency.(*conn).Read
         0     0%   100%     0.50MB   0.9%  google.golang.org/grpc/benchmark/latency.(*listener).Accept
         0     0%   100%     0.53MB  0.96%  google.golang.org/grpc/internal/transport.(*bufWriter).Write
         0     0%   100%     2.50MB  4.51%  google.golang.org/grpc/internal/transport.(*bufferPool).get (inline)
         0     0%   100%        1MB  1.80%  google.golang.org/grpc/internal/transport.(*http2Client).handleData
         0     0%   100%        1MB  1.80%  google.golang.org/grpc/internal/transport.(*http2Client).reader
         0     0%   100%        6MB 10.81%  google.golang.org/grpc/internal/transport.(*http2Server).HandleStreams
         0     0%   100%     1.50MB  2.70%  google.golang.org/grpc/internal/transport.(*http2Server).handleData
         0     0%   100%     2.03MB  3.66%  google.golang.org/grpc/internal/transport.(*loopyWriter).handle
         0     0%   100%     1.53MB  2.76%  google.golang.org/grpc/internal/transport.(*loopyWriter).headerHandler
         0     0%   100%     0.53MB  0.96%  google.golang.org/grpc/internal/transport.(*loopyWriter).originateStream
         0     0%   100%     2.03MB  3.66%  google.golang.org/grpc/internal/transport.(*loopyWriter).run
         0     0%   100%     1.53MB  2.76%  google.golang.org/grpc/internal/transport.(*loopyWriter).writeHeader
         0     0%   100%        2MB  3.60%  google.golang.org/grpc/internal/transport.NewServerTransport.func2
         0     0%   100%    32.48MB 58.49%  google.golang.org/grpc/internal/transport.newFramer
         0     0%   100%     0.53MB  0.96%  google.golang.org/grpc/internal/transport.newHTTP2Client.func6
         0     0%   100%     0.50MB  0.91%  google.golang.org/grpc/interop/grpc_testing.file_grpc_testing_control_proto_init
         0     0%   100%     0.50MB  0.91%  google.golang.org/grpc/interop/grpc_testing.init.1
         0     0%   100%     0.50MB   0.9%  google.golang.org/grpc/metadata.NewIncomingContext (inline)
         0     0%   100%     0.50MB  0.91%  google.golang.org/protobuf/internal/filedesc.Builder.Build
         0     0%   100%     0.50MB  0.91%  google.golang.org/protobuf/internal/filetype.Builder.Build
         0     0%   100%     0.50MB  0.91%  google.golang.org/protobuf/reflect/protoregistry.(*Files).RegisterFile
         0     0%   100%     0.50MB  0.91%  google.golang.org/protobuf/reflect/protoregistry.rangeTopLevelDescriptors
         0     0%   100%        1MB  1.80%  io.Copy (inline)
         0     0%   100%        1MB  1.80%  io.CopyN
         0     0%   100%        1MB  1.80%  io.ReadAtLeast
         0     0%   100%        1MB  1.80%  io.ReadFull (inline)
         0     0%   100%        1MB  1.80%  io.copyBuffer
         0     0%   100%     0.50MB   0.9%  net.(*TCPListener).Accept
         0     0%   100%     0.50MB   0.9%  net.(*TCPListener).accept
         0     0%   100%     0.50MB   0.9%  net.(*netFD).accept
         0     0%   100%     0.50MB  0.91%  runtime.doInit
         0     0%   100%     0.50MB  0.91%  runtime.main
         0     0%   100%     0.50MB  0.91%  runtime.mcall
         0     0%   100%     0.50MB  0.91%  runtime.modtimer
         0     0%   100%     8.50MB 15.31%  runtime.newproc.func1
         0     0%   100%     8.50MB 15.31%  runtime.newproc1
         0     0%   100%     0.50MB  0.91%  runtime.park_m
         0     0%   100%     0.50MB  0.91%  runtime.resetForSleep
         0     0%   100%     0.50MB  0.91%  runtime.resettimer (inline)
         0     0%   100%     8.50MB 15.31%  runtime.systemstack
         0     0%   100%     3.03MB  5.46%  sync.(*Pool).Get
         0     0%   100%     2.50MB  4.51%  sync.(*Pool).pin
```

As you can see, in the unoptimized benchmark run `newBufWriter` method allocates 59.04MB or  54.39% of the whole memory. In the optimized run we now see this line instead `0.53MB  0.96% 91.88%     0.53MB  0.96%  google.golang.org/grpc/internal/transport.newFramer.func1` which corresponds to the memory allocated by the sync.Pool that I added to the newFramer function. 

I also have some results that demonstrates that this change doesn't negatively impact performance in the more realistic benchmark setups, but I will provide those results a bit later. 

RELEASE NOTES:
- TBD


internal/transport/http_util.go
```diff
@@ -389,7 +408,20 @@ func newFramer(conn net.Conn, writeBufferSize, readBufferSize int, maxHeaderList
...
+	if !ok {
+		pool = &sync.Pool{
+			New: func() interface{} {
+				b := make([]byte, size)
+				return &b
+			},
+		}
+		writeBufferPoolMap[size] = pool
+	}
+	writeBufferMutex.Unlock()
```
s-matyukevich, line 423: In theory, we can get rid of this mutext and pool map if we store sync.Pool per ClientConnection and per server. However, this will require modifying a lot of internal interface to pass the pool all the way down to the transport layer. This also will prevent us from sharing the pool between different ClientConnections and servers, so I decided to take the mutex approach. 

As far as I understand, this mutex shouldn't really impact the performance, as it got called only on new connection creation, which happens not very often in http2.

internal/transport/http_util.go
```diff
@@ -332,19 +334,33 @@ func (w *bufWriter) Write(b []byte) (n int, err error) {
...
 		}
 	}
 	return n, err
 }
 
 func (w *bufWriter) Flush() error {
+	err := w.flush()
+	if w.buf != nil {
+		b := w.buf
+		w.pool.Put(&b)
```
s-matyukevich, line 357: `b := w.buf` line is important as otherwise we keep modifying the same buf after putting it to the pool

s-matyukevich:
I was trying to test whether this change causes any significant CPU/latency regressions and run benchmarks on a dedicated 8 CPU 32GB virtual machine. I used https://github.com/grpc/grpc-go/pull/6299 with the following parameters 
```
go run benchmark/benchmain/main.go -workloads all -kbps 0 -latency 0s -maxConcurrentCalls 50 -reqSizeBytes 1024,102400 -respSizeBytes 1024,102400 -benchtime 1m -connections=50 -sleepBetweenRPCs=0s,10s -resultFile /tmp/master1 -memProfile /tmp/master1-mem -memProfileRate 1024
```
[Here](https://gist.github.com/s-matyukevich/4e0075d76ebcd9f879d77d3dab13d82e) are the results.

s-matyukevich:
I tested this with a real app and got 40% memory reduction. Each instance of the app is handling ~1.2k connections and has low volume of requests. 
Screenshot 2023-05-30 at 3 23 51 PM

dfawley:
Unfortunately with @arvindbr8 and @easwars out of town we are a bit short staffed right now.  Will resume in ~2 weeks if that's okay, thanks!

internal/transport/http_util.go
```diff
@@ -332,19 +334,33 @@ func (w *bufWriter) Write(b []byte) (n int, err error) {
 		n, err = w.conn.Write(b)
 		return n, toIOError(err)
 	}
+	if w.buf == nil {
+		b := w.pool.Get().(*[]byte)
+		w.buf = *b
+	}
 	for len(b) > 0 {
-		nn := copy(w.buf[w.offset:], b)
+		nn := copy((w.buf)[w.offset:], b)
```
easwars, line 342: Do we need the parentheses around `w.buf`?

internal/transport/http_util.go
easwars: When do entries from this map ever get deleted? Can this map grow in a unbounded fashion?

easwars (commented):
What are your thoughts about making this an opt-in feature, i.e, we could have a dial option which the user would have to set for them to get this behavior. If the dial option is not set, the user would get the old behavior.

At a later point in time, we could get rid of the dial option completely and make this the default behavior, or continue to have the dial option (but make the use of the sync.Pool the default one, and the user would have to use the dial option to explicitly disable it).

internal/transport/http_util.go
s-matyukevich: Right now - never. In theory it can grow unbounded if the app creates multiple servers and/or ClientConnections with different buffer sizes. I can't think about any use-case where this could be useful, but it is possible. I am not sure if it worth adding more complicated logic to protect against this, because my understanding is that go runtime will eventually deallocate all memory from unused pools, so we will be waisting memory only on storing pointers to empty pools, which should be neglectable in practise. 

Still I can fix this if you prefer:
* I can add a goroutine that removes unused pools. I'll probably need to store something like lastModifedTime for each pool to do that.  The downsides are some additional complexity + the fact that we need to take the lock in the goroutine.
* I can store a pool per Server and ClientConnection and propagate it all the way down to the transport layer.  The downsides are the fact that we need to extend some internal interfaces + we won't be able to reuse pools between different connections that uses the same buffer size (which I assume is a very common use-case) But in this case we can get rid of the lock completely.

Do you have any other ideas? Let me know what I should do here.

s-matyukevich:
> What are your thoughts about making this an opt-in feature, i.e, we could have a dial option which the user would have to set for them to get this behavior. If the dial option is not set, the user would get the old behavior.
> 
> At a later point in time, we could get rid of the dial option completely and make this the default behavior, or continue to have the dial option (but make the use of the sync.Pool the default one, and the user would have to use the dial option to explicitly disable it).

Sure, will do that. If we are not enabling this by default, what do you think if I add a similar fix for the read buffer in the same PR? I can use a single option to enable both.

internal/transport/http_util.go
easwars: > In theory it can grow unbounded if the app creates multiple servers and/or ClientConnections with different buffer sizes. I can't think about any use-case where this could be useful, but it is possible.

I agree. I don't expect applications to create multiple `grpc.ClientConn`s or `grpc.Server`s and pass them different values for the send and receive buffer sizes. But given that we are going to have a dial option and a server option to enable the sync.Pool on the client and server respectively, we could document it there, just so no one complains saying a few bytes are leaked.

@dfawley : Thoughts/Objections?

easwars:
> Sure, will do that. If we are not enabling this by default, what do you think if I add a similar fix for the read buffer in the same PR? I can use a single option to enable both.

Yes, totally.

jzelinskie:
What's the motivation to putting this behind a flag? Is simply to build confidence before it becomes the default behavior?

s-matyukevich:
I spend a few hours today looking into optimizing read buffer. Unless I am missing something I think it is completely useless. Here is why:
* http2 framer interacts with read buffer in 2 places [here](https://cs.opensource.google/go/x/net/+/refs/tags/v0.11.0:http2/frame.go;l=237) it reads headers and [here](https://cs.opensource.google/go/x/net/+/refs/tags/v0.11.0:http2/frame.go;l=506) it reads the frame itself. In both cases it uses io.ReadFull. It can do this because in both cases it knows the exact size it needs to read (headers size is hardcoded in http2 and payload size is taken from the header)
* [getReadBuf](https://cs.opensource.google/go/x/net/+/refs/tags/v0.11.0:http2/frame.go;l=440-446) function is used to allocate the buffer before reading the payload. Framer always allocates the buffer large enough to read the whole frame. 
* bufio.Reader in such case will simply copy the whole frame first to its internal buffer and then from the internal buffer to the array provided by the framer. It happens [here](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/bufio/bufio.go;l=250)

Given that, I think that per-connection read buffer just wastes memory and slightly decrease performance - going to double-check that with benchmarks.

easwars:
> What's the motivation to putting this behind a flag? Is simply to build confidence before it becomes the default behavior?

Yes.

internal/transport/http_util.go
s-matyukevich: fixed

s-matyukevich:
Here are the results of running benchmarks with and without read buffer enabled. They are marginally the same. I think we should disable this buffer by default - let me know what we should do about it. 

```
unary-networkMode_none-bufConn_false-keepalive_false-benchTime_1m0s-trace_false-latency_0s-kbps_0-MTU_0-maxConcurrentCalls_50-reqSize_1024B-respSize_1024B-compressor_off-channelz_false-preloader_false-clientReadBufferSize_-1-clientWriteBufferSize_-1-serverReadBufferSize_-1-serverWriteBufferSize_-1-sleepBetweenRPCs_0s-connections_50-recvBufferPool_nil-shareWriteBuffer_false
               Title       Before        After Percentage
            TotalOps      2183043      2169883    -0.60%
             SendOps            0            0      NaN%
             RecvOps            0            0      NaN%
            Bytes/op     18393.98     18456.23     0.34%
           Allocs/op       169.23       169.26     0.00%
             ReqT/op 298058137.60 296261358.93    -0.60%
            RespT/op 298058137.60 296261358.93    -0.60%
            50th-Lat  55.124406ms  55.386083ms     0.47%
            90th-Lat 140.256403ms 142.849927ms     1.85%
            99th-Lat 241.708681ms 246.698658ms     2.06%
             Avg-Lat  68.495813ms  68.904223ms     0.60%
           GoVersion     go1.20.4     go1.20.4
         GrpcVersion   1.57.0-dev   1.57.0-dev

streaming-networkMode_none-bufConn_false-keepalive_false-benchTime_1m0s-trace_false-latency_0s-kbps_0-MTU_0-maxConcurrentCalls_50-reqSize_1024B-respSize_1024B-compressor_off-channelz_false-preloader_false-clientReadBufferSize_-1-clientWriteBufferSize_-1-serverReadBufferSize_-1-serverWriteBufferSize_-1-sleepBetweenRPCs_0s-connections_50-recvBufferPool_nil-shareWriteBuffer_false
               Title       Before        After Percentage
            TotalOps      3640797      3742486     2.79%
             SendOps            0            0      NaN%
             RecvOps            0            0      NaN%
            Bytes/op     10008.73     10043.94     0.35%
           Allocs/op        38.40        38.43     0.00%
             ReqT/op 497090150.40 510974088.53     2.79%
            RespT/op 497090150.40 510974088.53     2.79%
            50th-Lat  36.099464ms   33.67806ms    -6.71%
            90th-Lat  67.922543ms  71.600016ms     5.41%
            99th-Lat 187.936214ms 194.275101ms     3.37%
             Avg-Lat  41.112238ms  39.974572ms    -2.77%
           GoVersion     go1.20.4     go1.20.4
         GrpcVersion   1.57.0-dev   1.57.0-dev

unconstrained-networkMode_none-bufConn_false-keepalive_false-benchTime_1m0s-trace_false-latency_0s-kbps_0-MTU_0-maxConcurrentCalls_50-reqSize_1024B-respSize_1024B-compressor_off-channelz_false-preloader_false-clientReadBufferSize_-1-clientWriteBufferSize_-1-serverReadBufferSize_-1-serverWriteBufferSize_-1-sleepBetweenRPCs_0s-connections_50-recvBufferPool_nil-shareWriteBuffer_false
               Title       Before        After Percentage
            TotalOps            0            0      NaN%
             SendOps     10664146     11351236     6.44%
             RecvOps      8005583      7453543    -6.90%
            Bytes/op      8630.77      8710.33     0.93%
           Allocs/op        25.91        26.07     3.86%
             ReqT/op 1456011400.53 1549822088.53     6.44%
            RespT/op 1093028932.27 1017657070.93    -6.90%
            50th-Lat           0s           0s      NaN%
            90th-Lat           0s           0s      NaN%
            99th-Lat           0s           0s      NaN%
             Avg-Lat           0s           0s      NaN%
           GoVersion     go1.20.4     go1.20.4
         GrpcVersion   1.57.0-dev   1.57.0-dev
```

easwars:
@s-matyukevich : Thank you for running the benchmarks and comparing the results.

@dfawley : What are your thoughts on disabling the buffer on the read path, based on @s-matyukevich's findings?

s-matyukevich:
I did more testing today and I can take my words back: read buffer does actually reduce the number of syscalls. The part that I was missing is that it accumulates data from unrelated frames (as all streams use the same netwrok connection) However, I still question its usefulness as I don't see any improvements in benchmarks. 

My current theory is that visible performance improvements that comes from the usage of the write buffer comes from the fact that it avoids sending small TCP packets by buffering the data before sending it on the wire, and the number of syscall doesn't matter a lot.

s-matyukevich:
I finally was able to get a consistent 10-20% throughput improvement if I compare the results or running benchmarks with read buffer disabled and with read buffer set to 1MB (which is ridiculously large, but makes the impact more visible). So read buffer does slightly improve performance and completely disabling it is probably a bad idea. Will try to use sync.Pool to share it instead.

s-matyukevich:
I created another [branch](https://github.com/s-matyukevich/grpc-go/pull/2) on top of this one to test the impact of sharing read buffer. However, I wasn't able to demonstrate any performance improvements. I also run our internal benchmarks in real k8s infrastructure with read buffer disabled, and instead of performance regression it shows very small but consistent performance improvements. I am not sure hot to explain that, but I am going to give up attempts to share read buffer and instead we will simply disable it in our grpc wrapper lib for all our users. 

I think this PR should be ready for another review, as I don't plan to modify it further without addition comments.

internal/transport/http_util.go
```diff
@@ -403,6 +429,26 @@ func newFramer(conn net.Conn, writeBufferSize, readBufferSize int, maxHeaderList
...
+	if !ok {
+		pool = &sync.Pool{
+			New: func() interface{} {
+				b := make([]byte, size)
+				return &b
+			},
+		}
+		writeBufferPoolMap[size] = pool
+	}
+	return pool
```
easwars, line 449: Restructure like this:

```golang
	pool, ok := writeBufferPoolMap[size]
	if ok {
		return pool
	}
	writeBufferPoolMap[size] = &sync.Pool{
		New: func() interface{} {
			b := make([]byte, size)
			return &b
		},
	}
	return writeBufferPoolMap[size]
```

I'm also OK with assigning the newly created `sync.Pool` to the local var `pool` and writing that to the map and returning it (as you currently have). The crux of this comment though is https://google.github.io/styleguide/go/decisions#indent-error-flow.

dialoptions.go
```diff
@@ -139,6 +139,15 @@ func newJoinDialOption(opts ...DialOption) DialOption {
 	return &joinDialOption{opts: opts}
 }
 
+// WithShareWriteBuffer allows reusing per-connection transport write buffer.
+// If this option is set to true every connection will release the buffer after
+// flushing the data on the wire.
+func WithShareWriteBuffer(val bool) DialOption {
```
easwars, line 145: Maybe this name and the name of the `ServerOption` could do with some bikeshedding. Options that come to my mind are
- WithWriteBufferSharing
- WithSharedWriteBuffer

easwars (approved):
Changes look good to me except for the minor nits and the naming bikeshedding.

easwars:
@dfawley : Could you take the second review please. Thanks.

internal/transport/http_util.go
```diff
@@ -403,6 +429,26 @@ func newFramer(conn net.Conn, writeBufferSize, readBufferSize int, maxHeaderList
 	return f
 }
 
+func getWriteBufferPool(writeBufferSize int, shareWriteBuffer bool) *sync.Pool {
```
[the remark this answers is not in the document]
s-matyukevich: done.

internal/transport/http_util.go
s-matyukevich: done

dialoptions.go
s-matyukevich: done, I used WithSharedWriteBuffer

easwars (approved):
There is a small vet failure with the docstring not starting with the name of the function. Could you please take care of that. Thanks.

server.go
```diff
@@ -235,6 +236,15 @@ func newJoinServerOption(opts ...ServerOption) ServerOption {
 	return &joinServerOption{opts: opts}
 }
 
+// SharedWriteBuffer allows reusing per-connection transport write buffer.
+// If this option is set to true every connection will release the buffer after
+// flushing the data on the wire.
+func SharedWriteBuffer(val bool) ServerOption {
```
dfawley, line 242: Please add an experimental tag here:
```go
// # Experimental
//
// later release.
```

dialoptions.go
```diff
@@ -139,6 +139,15 @@ func newJoinDialOption(opts ...DialOption) DialOption {
 	return &joinDialOption{opts: opts}
 }
 
+// WithSharedWriteBuffer allows reusing per-connection transport write buffer.
+// If this option is set to true every connection will release the buffer after
+// flushing the data on the wire.
+func WithSharedWriteBuffer(val bool) DialOption {
```
dfawley, line 145: Please add an experimental tag here:
```go
// # Experimental
//
// later release.
```

internal/transport/http_util.go
```diff
@@ -332,19 +339,34 @@ func (w *bufWriter) Write(b []byte) (n int, err error) {
...
+	// Only release the buffer if we are in a "shared" mode
+	if w.buf != nil && w.pool != nil {
+		b := w.buf
+		w.pool.Put(&b)
+		w.buf = nil
+	}
+	return err
+}
+
+func (w *bufWriter) flush() error {
```
dfawley, line 369: Another name here might be nice to distinguish it from `Flush`.  Something like `flushKeepBuffer`?

Or rename the other one to `FlushAndReleaseBuffer`..  Either way is fine.

dfawley (changes requested):
LGTM, thanks!  Just a couple minor things before we can merge it.

dfawley (approved):

ukai:
https://pkg.go.dev/google.golang.org/grpc#WithSharedWriteBuffer
```
 If this option is set to true every connection will release the buffer after flushing the data on the wire.
```

not release, but reuse ?

s-matyukevich:
> https://pkg.go.dev/google.golang.org/grpc#WithSharedWriteBuffer
> 
> ```
>  If this option is set to true every connection will release the buffer after flushing the data on the wire.
> ```
> 
> not release, but reuse ?

release, so it can be reused by other connections.

internal/transport/http_util.go
```diff
@@ -403,6 +432,24 @@ func newFramer(conn net.Conn, writeBufferSize, readBufferSize int, maxHeaderList
 	return f
 }
 
+func getWriteBufferPool(writeBufferSize int) *sync.Pool {
+	writeBufferMutex.Lock()
+	defer writeBufferMutex.Unlock()
+	size := writeBufferSize * 2
```
raghav-stripe, line 438: What is the significant of the `* 2` here?

dfawley: IIRC this matches the legacy behavior, which always doubled for reasons nobody knows but we don't want to change now due to it affecting existing applications.

raghav-stripe: That's fair. Do you mind expanding on the legacy behavior you are describing? It seems like if there is no shared buffer pool, the buffer is [allocated per the input size](https://github.com/grpc/grpc-go/blob/master/internal/transport/http_util.go#L314-L316). It only seems to be in the case of the buffer pool that we're allocating 2x the size (and this seems to be the PR the write buffer pool was added)?

dfawley: You may be right.. The old code multiplied by 2 when doing this allocation, so it seems we have changed the behavior for legacy users already:

https://github.com/grpc/grpc-go/blame/d524b409462c601ef3f05a7e1fba19755a337c77/internal/transport/http_util.go#L321

raghav-stripe: Yeah - it just felt somewhat bizarre because this is the very PR that also removed that same ` * 2` in the default path (see line 321 above), but added the ` * 2` for the buffer pool path.

raghav-stripe: If you think it's worth updating, I created https://github.com/grpc/grpc-go/pull/6983

Issue #4896, opened 2021-10-21
HeaderMatcher.invert_match should invert the operation, not the result


ejona86:
Go inverts the match result directly:
https://github.com/grpc/grpc-go/blob/6f2a92265e75acc87c687c8462d8e22fa2c7614f/internal/xds/rbac/matchers.go#L288-L290

However, the Envoy behavior is to invert the operation, such that `==` becomes `!=` and both fail for a missing entry:
https://github.com/envoyproxy/envoy/blob/b7393b0bd221dab43bea0d697df0954a5b01d7d7/source/common/http/header_utility.cc#L140-L146

This was implemented appropriately in Java. I'll add some language to call this out in the gRFC.

PR #6313, opened 2023-05-24
xds/internal/xdsclient: Add support for String Matcher Header Matcher in RDS

### LINKED ISSUE #6226: Add support for String Matcher in Header Matching
```
core] [Channel #7] ccResolverWrapper: reporting error to cc: received route is invalid: route name:"istio-safe-stability.v1--testecho.0"  match:{prefix:"/"  case_sensitive:{value:true}  headers:{name:"x-istio-cluster"  string_match:{exact:"outbound|8080||v1--testecho.istio-safe-stability.svc.cluster.local"}}}  route:{cluster:"outbound|8080||v1--testecho.istio-safe-stability.svc.cluster.local"  cluster_not_found_response_code:INTERNAL_SERVER_ERROR  timeout:{}  retry_policy:{retry_on:"connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes"  num_retries:{value:2}  retry_host_predicate:{name:"envoy.retry_host_predicates.previous_hosts"  typed_config:{type_url:"type.googleapis.com/envoy.extensions.retry.host.previous_hosts.v3.PreviousHostsPredicate"}}  host_selection_retry_max_attempts:5  retriable_status_codes:503}  max_stream_duration:{max_stream_duration:{}}}  metadata:{filter_metadata:{key:"istio"  value:{fields:{key:"config"  value:{string_value:"/apis/networking.istio.io/v1alpha3/namespaces/istio-safe-stability/virtual-service/v1--testecho-0-istio-autogenerated-k8s-gateway"}}}}}  decorator:{operation:"v1--testecho.istio-safe-stability.svc.cluster.local:8080/*"} has an unrecognized header matcher: &{StringMatch:exact:"outbound|8080||v1--testecho.istio-safe-stability.svc.cluster.local"}
```

This is produced by a standard HttpRoute -

```
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: gateway
  - group: gateway.networking.k8s.io
    kind: Service
    name: echo
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: v1--testecho
      port: 8080
      weight: 1
    matches:
    - headers:
      - name: x-istio-cluster
        type: Exact
        value: outbound|8080||v1--testecho.istio-safe-stability.svc.cluster.local
      path:
        type: PathPrefix
        value: /
```

NOTE: if you are reporting is a potential security vulnerability or a crash,
please follow our CVE process at
https://github.com/grpc/proposal/blob/master/P4-grpc-cve-process.md instead of
filing an issue here.

Please see the FAQ in our main README.md, then answer the questions below
before submitting your issue.

v1.56.0-dev ( tried on older versions too )

### What did you do?
If possible, provide a recipe for reproducing the error.

zasweq:
Fixes #6226. This PR adds support for String Matching in Header Matching for routes. It converts and validates the proto StringMatcher received from the control plane in the xDS Client, then creates a HeaderMatcher for it in ServiceConfig. Tests at xDS Client layer and header matcher layer, anything further will be too specific for an e2e test (as discussed about testing Philosophy with Eric and Doug, leave more specific behavior checks to unit tests and e2e tests as sanity checks to not bloat number and time of e2e tests).

RELEASE NOTES:
* xds/internal/xdsclient: Add support for String Matcher Header Matcher in RDS

### DIFF
--- a/internal/xds/matcher/matcher_header.go
+++ b/internal/xds/matcher/matcher_header.go
@@ -241,3 +241,34 @@ func (hcm *HeaderContainsMatcher) Match(md metadata.MD) bool {
 func (hcm *HeaderContainsMatcher) String() string {
 	return fmt.Sprintf("headerContains:%v%v", hcm.key, hcm.contains)
 }
+
+// HeaderStringMatcher matches on whether the header value matches against the
+// StringMatcher specified.
+type HeaderStringMatcher struct {
+	key           string
+	stringMatcher StringMatcher
+	invert        bool
+}
+
+// NewHeaderStringMatcher returns a new HeaderStringMatcher.
+func NewHeaderStringMatcher(key string, sm StringMatcher, invert bool) *HeaderStringMatcher {
+	return &HeaderStringMatcher{
+		key:           key,
+		stringMatcher: sm,
+		invert:        invert,
+	}
+}
+
+// Match returns whether the passed in HTTP Headers match according to the
+// specified StringMatcher.
+func (hsm *HeaderStringMatcher) Match(md metadata.MD) bool {
+	v, ok := mdValuesFromOutgoingCtx(md, hsm.key)
+	if !ok {
+		return false
+	}
+	return hsm.stringMatcher.Match(v) != hsm.invert
+}
+
+func (hsm *HeaderStringMatcher) String() string {
+	return fmt.Sprintf("headerString:%v:%v", hsm.key, hsm.stringMatcher)
+}
--- a/internal/xds/matcher/matcher_header_test.go
+++ b/internal/xds/matcher/matcher_header_test.go
@@ -467,3 +467,53 @@ func TestHeaderSuffixMatcherMatch(t *testing.T) {
 		})
 	}
 }
+
+func TestHeaderStringMatch(t *testing.T) {
+	tests := []struct {
+		name   string
+		key    string
+		sm     StringMatcher
+		md     metadata.MD
+		invert bool
+		want   bool
+	}{
+		{
+			name: "should-match",
+			key:  "th",
+			sm: StringMatcher{
+				exactMatch: newStringP("tv"),
+			},
+			md:     metadata.Pairs("th", "tv"),
+			invert: false,
+			want:   true,
+		},
+		{
+			name: "not match",
+			key:  "th",
+			sm: StringMatcher{
+				containsMatch: newStringP("tv"),
+			},
+			md:     metadata.Pairs("th", "not-match"),
+			invert: false,
+			want:   false,
+		},
+		{
+			name: "invert string match",
+			key:  "th",
+			sm: StringMatcher{
+				containsMatch: newStringP("tv"),
+			},
+			md:     metadata.Pairs("th", "not-match"),
+			invert: true,
+			want:   true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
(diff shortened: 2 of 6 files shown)


internal/xds/matcher/matcher_header.go
```diff
@@ -241,3 +241,34 @@ func (hcm *HeaderContainsMatcher) Match(md metadata.MD) bool {
...
+}
+
+// Match returns whether the passed in HTTP Headers match according to the
+// specified StringMatcher.
+func (hsm *HeaderStringMatcher) Match(md metadata.MD) bool {
+	v, ok := mdValuesFromOutgoingCtx(md, hsm.key)
+	if !ok {
+		return false
+	}
+	return hsm.stringMatcher.Match(v) != hsm.invert
```
dfawley, line 269: The behavior here is unclear in some cases:

1. If a header is missing, should it always report `false`?  Or should it report `true` if `invert=true`?
2. If a header value is `""`, is that distinct from the header being missing, in any way?
3. Is the check immediately above designed to only detect an error in usage?  Because it doesn't consider (1) at least.

internal/xds/matcher/matcher_header_test.go
```diff
@@ -467,3 +467,53 @@ func TestHeaderSuffixMatcherMatch(t *testing.T) {
...
+		{
+			name: "invert string match",
+			key:  "th",
+			sm: StringMatcher{
+				containsMatch: newStringP("tv"),
+			},
+			md:     metadata.Pairs("th", "not-match"),
+			invert: true,
+			want:   true,
+		},
```
dfawley, line 509: Please add test cases that cover the unusual corner cases mentioned above once we determine the correct behavior.

internal/xds/matcher/matcher_header_test.go
```diff
@@ -467,3 +467,53 @@ func TestHeaderSuffixMatcherMatch(t *testing.T) {
...
 	}
 }
+
+func TestHeaderStringMatch(t *testing.T) {
+	tests := []struct {
+		name   string
+		key    string
+		sm     StringMatcher
+		md     metadata.MD
+		invert bool
```
dfawley, line 477: Super nit: please swap these two in order (here and in the test cases) since `invert` is part of the matcher and `md` is the input passed to the matcher.  OR: build the matcher in the testcase list and put it in there instead of the individual fields.

internal/xds/matcher/matcher_header.go
zasweq: See my chat thread. I think this is WAI.

internal/xds/matcher/matcher_header_test.go
zasweq: Done. Chose the first option and swapped.

internal/xds/matcher/matcher_header.go
zasweq: 1. See Eric's explanation here: https://github.com/grpc/grpc-go/issues/4896
2. Yes, a set empty header string is logically distinct from it not being presence
3. No, this is part of the logic for the header matcher (not an error in usage, a header matcher can be called with metadata not containing the header key set in the matcher), as per explanation in 1

internal/xds/matcher/matcher_header_test.go
zasweq: Done. Added 3 test cases, invert and not invert for missing, and invert for empty string wanting true.

dfawley (approved):
Thanks!

PR #6315, opened 2023-05-24
passthrough filter name for RBAC

gtcooke94:
Draft, please ignore for now


dfawley:
Please see https://github.com/grpc/grpc-go/pull/6297#issuecomment-1553712814 for why you shouldn't need to do this (and please let me know if I'm wrong!).

PR #6331, opened 2023-05-31
client: fix Connect to handle channel idleness properly

dfawley:
`ClientConn.Connect()` needs to fully exit idle mode, not just awaken the balancer wrapper.  Before channel idleness, the LB policy was the only thing that needed to be notified, since the name resolver was never stopped.

RELEASE NOTES: none


ginayeh (approved):

dfawley:
Mergeable is stuck & hit known flake..closing & reopening.

PR #6350, opened 2023-06-05
opencensus: stop overwriting ctx parameter in tests

dfawley:
Saw this in the last import.


zasweq (approved):
Weird this got left. I thought I deleted this and changed callsite solely to make sure tests context was used in all operations which requires timeout.

PR #6351, opened 2023-06-06
xds/ringhash: cache connectivity state of subchannels inside picker

dfawley:
I did not add a new test for this.  The sequence of operations required here is fairly long and specific and requires races that are too difficult to reliably stimulate.  Corresponding change in C++: [grpc/grpc@`98cdac3` (#32326)](https://github.com/grpc/grpc/pull/32326/commits/98cdac31a0196d23a972acd3de027a645daf66ce).  (gRPC-Java always kept a copy of the states in the picker.)

RELEASE NOTES:
* xds/ringhash: cache connectivity state of subchannels inside picker to avoid rare races

### DIFF
--- a/xds/internal/balancer/ringhash/ringhash.go
+++ b/xds/internal/balancer/ringhash/ringhash.go
@@ -359,11 +359,13 @@ func (b *ringhashBalancer) UpdateSubConnState(sc balancer.SubConn, state balance
 		delete(b.scStates, sc)
 	}
 
-	// Because the picker caches the state of the subconns, we always regnerate
-	// and update the picker.
-	b.regeneratePicker()
-	b.logger.Infof("Pushing new state %v and picker %p", b.state, b.picker)
-	b.cc.UpdateState(balancer.State{ConnectivityState: b.state, Picker: b.picker})
+	if oldSCState != newSCState {
+		// Because the picker caches the state of the subconns, we always regnerate
+		// and update the picker when the effective state changes.
+		b.regeneratePicker()
+		b.logger.Infof("Pushing new state %v and picker %p", b.state, b.picker)
+		b.cc.UpdateState(balancer.State{ConnectivityState: b.state, Picker: b.picker})
+	}
 
 	switch b.state {
 	case connectivity.Connecting, connectivity.TransientFailure:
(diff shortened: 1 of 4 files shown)


xds/internal/balancer/ringhash/ringhash.go
```diff
@@ -347,37 +347,22 @@ func (b *ringhashBalancer) UpdateSubConnState(sc balancer.SubConn, state balance
...
-		b.regeneratePicker()
-		sendUpdate = true
 	case connectivity.TransientFailure:
 		// Save error to be reported via picker.
 		b.connErr = state.ConnectionError
-		b.regeneratePicker()
 	case connectivity.Shutdown:
 		// When an address was removed by resolver, b called RemoveSubConn but
 		// kept the sc's state in scStates. Remove state for this sc here.
 		delete(b.scStates, sc)
```
zasweq, line 359: Nit: Not a part of this PR, but this scStates field I feel like is wrongly named. This is a map from balancer.SubConn to the Ring Hashes SubConn with extra info. I feel like states is only a part of the reason this map exists.

xds/internal/balancer/ringhash/ringhash.go
```diff
@@ -347,37 +347,22 @@ func (b *ringhashBalancer) UpdateSubConnState(sc balancer.SubConn, state balance
...
 	}
 
-	if sendUpdate {
+	if oldSCState != newSCState {
+		// Because the picker caches the state of the subconns, we always regnerate
```
zasweq, line 363: regenerate

xds/internal/balancer/ringhash/ringhash.go
```diff
@@ -347,37 +347,22 @@ func (b *ringhashBalancer) UpdateSubConnState(sc balancer.SubConn, state balance
...
+		// and update the picker when the effective state changes.
```
zasweq, line 364: Nit: "of the SubConn." Optional addition: "This is so the Client Conn gets the most updated state as a former picker has a static cache of SubConn states at it's build time."

zasweq (approved):
LGTM.

xds/internal/balancer/ringhash/ringhash.go
dfawley: Fixed

xds/internal/balancer/ringhash/ringhash.go
dfawley: Fixed

xds/internal/balancer/ringhash/ringhash.go
dfawley: Yeah it probably was changed during development.  99% of why there exists a map is state tracking.  It also does a `queueConnect` in rare cases.

PR #6354, opened 2023-06-07
client: handle empty address lists correctly in addrConn.updateAddrs

dfawley:
This fixes a bug caused by #6302, and will need to be cherry-picked into v1.56.x before the release.

grpclb can run into this situation, which the added test also stimulates.  It may not be "proper" to have a SubConn with no addresses, but the result is currently a nil pointer panic, which we should not allow, and needs to be fixed ASAP.

RELEASE NOTES: none


test/subconn_test.go
```diff
@@ -0,0 +1,125 @@
...
+		EmptyCallF: func(ctx context.Context, in *testpb.Empty) (*testpb.Empty, error) {
+			return &testpb.Empty{}, nil
+		},
+	}
+	if err := ss.Start(nil, grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"tscc":{}}]}`)); err != nil {
+		t.Fatalf("Error starting server: %v", err)
+	}
+	defer ss.Stop()
+	if _, err := ss.Client.EmptyCall(ctx, &testpb.Empty{}); err != nil {
+		t.Fatalf("UnaryCall RPC failed: %v", err)
```
zasweq, line 113: Sorry, super nit: UnaryCall is distinct from EmptyCall wrt StubServer. Technically it is a UnaryCall, but would prefer calling this EmptyCall.

test/subconn_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	t.Log("Removing addresses from resolver and SubConn")
+	ss.R.UpdateState(resolver.State{Addresses: []resolver.Address{}})
+	awaitState(ctx, t, ss.CC, connectivity.TransientFailure)
+
+	t.Log("Re-adding addresses to resolver and SubConn")
+	ss.R.UpdateState(resolver.State{Addresses: []resolver.Address{{Addr: ss.Address}}})
+	if _, err := ss.Client.EmptyCall(ctx, &testpb.Empty{}); err != nil {
+		t.Fatalf("UnaryCall RPC failed: %v", err)
```
zasweq, line 123: Same here.

zasweq (approved):
LGTM.

PR #6357, opened 2023-06-07
orca: add application utilization and range checking

dfawley:
Implement the new field described [here](https://github.com/grpc/proposal/pull/374), and add missing range checks and tests.

RELEASE NOTES:
* orca: add support for application_utilization field and missing range checks on several metrics setters

### DIFF
--- a/orca/producer_test.go
+++ b/orca/producer_test.go
@@ -158,12 +158,12 @@ func (s) TestProducer(t *testing.T) {
 
 	// Set a few metrics and wait for them on the client side.
 	smr.SetCPUUtilization(10)
-	smr.SetMemoryUtilization(100)
-	smr.SetNamedUtilization("bob", 555)
+	smr.SetMemoryUtilization(0.1)
+	smr.SetNamedUtilization("bob", 0.555)
 	loadReportWant := &v3orcapb.OrcaLoadReport{
 		CpuUtilization: 10,
-		MemUtilization: 100,
-		Utilization:    map[string]float64{"bob": 555},
+		MemUtilization: 0.1,
+		Utilization:    map[string]float64{"bob": 0.555},
 	}
 
 testReport:
@@ -181,13 +181,13 @@ testReport:
 	}
 
 	// Change and add metrics and wait for them on the client side.
-	smr.SetCPUUtilization(50)
-	smr.SetMemoryUtilization(200)
-	smr.SetNamedUtilization("mary", 321)
+	smr.SetCPUUtilization(0.5)
+	smr.SetMemoryUtilization(0.2)
+	smr.SetNamedUtilization("mary", 0.321)
 	loadReportWant = &v3orcapb.OrcaLoadReport{
-		CpuUtilization: 50,
-		MemUtilization: 200,
-		Utilization:    map[string]float64{"bob": 555, "mary": 321},
+		CpuUtilization: 0.5,
+		MemUtilization: 0.2,
+		Utilization:    map[string]float64{"bob": 0.555, "mary": 0.321},
 	}
 
 	for {
@@ -322,8 +322,8 @@ func (s) TestProducerBackoff(t *testing.T) {
 	// Define a load report to send and expect the client to see.
 	loadReportWant := &v3orcapb.OrcaLoadReport{
 		CpuUtilization: 10,
-		MemUtilization: 100,
-		Utilization:    map[string]float64{"bob": 555},
+		MemUtilization: 0.1,
+		Utilization:    map[string]float64{"bob": 0.555},
 	}
 
 	// Unblock the fake.
@@ -444,8 +444,8 @@ func (s) TestProducerMultipleListeners(t *testing.T) {
 	// Define a load report to send and expect the client to see.
 	loadReportWant := &v3orcapb.OrcaLoadReport{
 		CpuUtilization: 10,
-		MemUtilization: 100,
-		Utilization:    map[string]float64{"bob": 555},
+		MemUtilization: 0.1,
+		Utilization:    map[string]float64{"bob": 0.555},
 	}
 
 	// Receive reports and update counts for the three listeners.
--- a/orca/server_metrics.go
+++ b/orca/server_metrics.go
@@ -27,8 +27,9 @@ import (
 // ServerMetrics is the data returned from a server to a client to describe the
 // current state of the server and/or the cost of a request when used per-call.
 type ServerMetrics struct {
-	CPUUtilization float64 // CPU utilization: [0, 1.0]; unset=-1
+	CPUUtilization float64 // CPU utilization: [0, inf); unset=-1
 	MemUtilization float64 // Memory utilization: [0, 1.0]; unset=-1
+	AppUtilization float64 // Application utilization: [0, inf); unset=-1
 	QPS            float64 // queries per second: [0, inf); unset=-1
 	EPS            float64 // errors per second: [0, inf); unset=-1
 
@@ -52,6 +53,9 @@ func (sm *ServerMetrics) toLoadReportProto() *v3orcapb.OrcaLoadReport {
 	if sm.MemUtilization != -1 {
 		ret.MemUtilization = sm.MemUtilization
 	}
+	if sm.AppUtilization != -1 {
+		ret.ApplicationUtilization = sm.AppUtilization
+	}
 	if sm.QPS != -1 {
 		ret.RpsFractional = sm.QPS
 	}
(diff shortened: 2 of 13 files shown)


orca/server_metrics.go
```diff
@@ -194,36 +221,70 @@ func (s *serverMetricsRecorder) SetMemoryUtilization(val float64) {
 // DeleteMemoryUtilization deletes the relevant server metric to prevent it
 // from being sent.
```
zasweq, line 222: Nit on these docstrings and also the interface docstrings. Most of them define "the relevant server metric", but I feel like it should just be as specific as "for the memory utilization metric" as on line 211 for all docstrings because it maps 1:1 with the name.

orca/server_metrics.go
```diff
@@ -181,11 +203,16 @@ func (s *serverMetricsRecorder) SetCPUUtilization(val float64) {
...
-	s.SetCPUUtilization(-1)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.state.CPUUtilization = -1
 }
 
 // SetMemoryUtilization records a measurement for the memory utilization metric.
 func (s *serverMetricsRecorder) SetMemoryUtilization(val float64) {
+	if val < 0 || val > 1 {
+		return
```
zasweq, line 214: Here and all other places with a range validation on this serverMetricsRecorder: do we want to plumb a logger in here and log in a non spammy way that we're ignoring an out of range metric, similar to how xDS Client gives info if nack, or say CDS encounters an error in it's serial handling of Client Conn State that returns no error, it logs the reason why the Client Conn State wasn't processed?

orca/server_metrics.go
```diff
@@ -63,21 +67,24 @@ func (sm *ServerMetrics) toLoadReportProto() *v3orcapb.OrcaLoadReport {
 
 // merge merges o into sm, overwriting any values present in both.
 func (sm *ServerMetrics) merge(o *ServerMetrics) {
+	mergeMap(sm.Utilization, o.Utilization)
+	mergeMap(sm.RequestCost, o.RequestCost)
+	mergeMap(sm.NamedMetrics, o.NamedMetrics)
```
zasweq, line 72: What happens if one of the values of this map is also -1 like below? Do we still want to overwrite the value for the key in the map? Should we ignore as below?

orca/server_metrics.go
```diff
@@ -63,21 +67,24 @@ func (sm *ServerMetrics) toLoadReportProto() *v3orcapb.OrcaLoadReport {
 
 // merge merges o into sm, overwriting any values present in both.
```
zasweq, line 68: This isn't correct? You overwrite any values for keys in map unconditionally, but only overwrite sm's memory if it's set. I guess if an explicit -1 means unset 1:1 this is right. I just don't know all the invariants of this system/how this data is plumbed around.

orca/producer_test.go
```diff
@@ -444,8 +444,8 @@ func (s) TestProducerMultipleListeners(t *testing.T) {
 	// Define a load report to send and expect the client to see.
 	loadReportWant := &v3orcapb.OrcaLoadReport{
 		CpuUtilization: 10,
-		MemUtilization: 100,
-		Utilization:    map[string]float64{"bob": 555},
+		MemUtilization: 0.1,
```
zasweq, line 447: Now that you had to move these floats to be within 0 and 1, can you please add a test for the case it isn't within range and also set to -1 wrt overwriting behavior in merge(). This will also show what happens at this producer layer if these cases happen.

orca/server_metrics_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Negatives for all these fields should be ignored.
+	smr.SetCPUUtilization(-2)
+	smr.SetMemoryUtilization(-3)
+	smr.SetApplicationUtilization(-4)
+	smr.SetQPS(-0.1)
+	smr.SetEPS(-0.6)
+	smr.SetNamedUtilization("x", -2)
+
+	// Memory and named utilizations are max of 1.
```
zasweq, line 109: So this should be ignored.

orca/server_metrics_test.go
```diff
@@ -0,0 +1,128 @@
...
+		Utilization:    map[string]float64{"x": 0.6},
+		NamedMetrics:   map[string]float64{},
+		RequestCost:    map[string]float64{},
+	}
+
+	got := smr.ServerMetrics()
+	if d := cmp.Diff(got, want); d != "" {
+		t.Fatalf("unexpected server metrics: -got +want: %v", d)
+	}
+}
```
zasweq, line 128: After thinking about my comment on merge() wrt overwriting at each layer (maps and top level metrics), can you also add tests for that.

orca/service_test.go
```diff
@@ -60,16 +60,18 @@ func (t *testServiceImpl) UnaryCall(context.Context, *testpb.SimpleRequest) (*te
 	t.requests++
 	t.mu.Unlock()
 
-	t.smr.SetNamedUtilization(requestsMetricKey, float64(t.requests))
+	t.smr.SetNamedUtilization(requestsMetricKey, float64(t.requests)*0.01)
```
zasweq, line 63: Like my comment above, what happens in this layer if these metrics aren't within range? Can you please add a test to show what behavior is going to be.

zasweq (changes requested):

orca/server_metrics.go
dfawley: For maps, `-1` has no special meaning.  Deleting from the map is how you delete the value.  -1 is unique to the fields that are scalar values.

orca/server_metrics.go
dfawley: > You overwrite any values for keys in map unconditionally, but only overwrite sm's memory if it's set.

No?  I range over `o.<map field>` and add all those key/value pairs into `sm.<map field>`.  It's a merge, not a replacement, so the idea is to keep anything in `sm` that's not in `o`.

orca/producer_test.go
dfawley: I have range checks in the new tests.

orca/server_metrics_test.go
dfawley: Clarified.

orca/server_metrics_test.go
dfawley: Added unit tests for merge..  It is already covered implicitly by the call metrics tests that use it, though.

orca/service_test.go
dfawley: This is covered in the unit tests.  It's just calls into the same object so if that works right then it must work right here, too.

orca/server_metrics.go
zasweq: Ah ok.

orca/producer_test.go
zasweq: I know, but what happens at this layer? Just ignored? I want it tested at this layer too, or are you saying specific behaviors are fine to encapsulate to it's granular components units.

orca/service_test.go
zasweq: Fine.

zasweq: This relates to discussions we've had and I've had with Eric about specific functionality scoped to the granular component for testing, and higher layers just do sanity to avoid excessive higher level tests. I'm actually running into this problem with my od change, since ParseConfig gets invoked at every layer of unit test, whether it's cds, cluster resolver, and obv. on the od ParseConfig test itself.

dfawley: ...Which is why I was saying to test the proto->JSON conversion as a unit test so you can get all the corner cases in there and not worry about it when testing the higher level logic.

orca/server_metrics.go
```diff
@@ -193,6 +195,9 @@ func copyMap(m map[string]float64) map[string]float64 {
 // SetCPUUtilization records a measurement for the CPU utilization metric.
 func (s *serverMetricsRecorder) SetCPUUtilization(val float64) {
 	if val < 0 {
+		if logger.V(2) {
+			logger.Infof("Ignoring CPU Utilization value out of range: %v", val)
```
zasweq, line 199: Log valid range here and elsewhere?

orca/server_metrics_test.go
```diff
@@ -126,3 +134,42 @@ func (s) TestServerMetrics_Setters_Range(t *testing.T) {
...
+		QPS:            -1,
+		EPS:            0,
+		Utilization:    map[string]float64{"x": 0.6},
+		NamedMetrics:   map[string]float64{"y": 0.2},
+		RequestCost:    map[string]float64{"a": 0.1},
+	}
+
+	sm2 := &ServerMetrics{
+		CPUUtilization: -1,
+		AppUtilization: 0,
```
zasweq, line 152: Unfortunate that MemUtilization's zero value = set valid metric which overwrites, but can't do anything about that.

zasweq (approved):
LGTM.

orca/server_metrics.go
dfawley: Ehh, they'll figure it out :)

orca/server_metrics_test.go
dfawley: Yeah the only option would be to use pointers and I don't really want to do that either.  We basically control everything around this struct and the values of the fields, though, so it's pretty safe.

orca/server_metrics.go
zasweq: "[0, infy)" this is very few characters to add to logs lol

PR #6360, opened 2023-06-07
weightedroundrobin: cherry-pick 2 commits from master

dfawley:
RELEASE NOTES:
* orca: add support for application_utilization field and missing range checks on several metrics setters _and_ * balancer/weightedroundrobin: if the ORCA application_utilization field is set, prefer it over cpu_utilization


zasweq (approved):
LGTM.

PR #6371, opened 2023-06-14
encoding: clarify when it is safe to register compressors and codecs

mostynb:
Previously the docs suggested that it was only safe to register compressors and codecs inide init() functions, I presume because these encoding.RegisterCompressor and encoding.RegisterCodec are not thread-safe. If that is the case, then it should also be safe to register codecs in the main goroutine before any other goroutines have been spawned.

Registering compressors and codecs from the main goroutine can be useful, because it allows application developers to ensure that their own compressor/codec is being used, and that transitively imported packages haven't overridden their preferred implementation.

RELEASE NOTES: none

### DIFF
--- a/encoding/encoding.go
+++ b/encoding/encoding.go
@@ -70,9 +70,10 @@ var registeredCompressor = make(map[string]Compressor)
 // header.  Servers also use it to send a response with the same encoding as
 // the request.
 //
-// NOTE: this function must only be called during initialization time (i.e. in
-// an init() function), and is not thread-safe.  If multiple Compressors are
-// registered with the same name, the one registered last will take effect.
+// NOTE: this function is not thread-safe, and must only be called during
+// initialization time (e.g. in an init() function, or in the main goroutine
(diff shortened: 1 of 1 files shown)


mostynb:
> Sorry for the delay in the response.

No worries :)

> You are right. It is safe to register compressors and codecs in the main goroutine before any other goroutines have been spawned also. However we have multiple register APIs which are similar other than the compressor and codec register.

Thanks for clarifying (I can use this info confidently in downstream documentation now, regardless of this PR landing or not).

> But im not convinced that there is additional value in this diff, since I feel the comment already talks about the API being not thread safe and need to be called during initialization. Hence closing this PR.

I think this tiny part of the current text is incorrect, and the source of my original misunderstanding: `i.e. in an init() function` - doesn't "i.e." mean "in other words"? If so, this is saying that you can *only* do this in an init() function, which is incorrect.

A smaller diff would be to change "i.e." to "e.g." ("for example"), if you don't like the idea of adding the extra text- would you be interested in landing that change instead?

PR #6377, opened 2023-06-15
xds: E2E Test for Audit Logging

gtcooke94:
This PR adds testing for audit logging logic to the existing RBAC tests. It very closely matches the tests for the authz path - https://github.com/grpc/grpc-go/pull/6304

I've currently only added a very basic test of the audit logger - the purpose here is to make sure that we can construct an audit logger through the xds path and it works.
I'm open to adding test cases for the various combinations of audit logging and authorization policies, but I think that's not really what we are testing here, and it's a pretty explosive combination and pollutes the test (if it's not necessary). All these cases should be covered in other tests like the unit tests and the authz path tests. @dfawley @easwars what do you think? I see compelling reasons for either choice.

RELEASE NOTES: N/A

### DIFF
--- a/test/xds/xds_server_rbac_test.go
+++ b/test/xds/xds_server_rbac_test.go
@@ -979,7 +979,7 @@ func createXDSTypedStruct(t *testing.T, in map[string]interface{}, name string) 
 	t.Helper()
 	pb, err := structpb.NewStruct(in)
 	if err != nil {
-		t.Fatalf("createXDSTypedStruct Failed during structpb.NewStruct: %v", err)
+		t.Fatalf("createXDSTypedStruct failed during structpb.NewStruct: %v", err)
 	}
 	typedStruct := &v3xdsxdstypepb.TypedStruct{
 		TypeUrl: typeURLPrefix + name,
@@ -987,7 +987,7 @@ func createXDSTypedStruct(t *testing.T, in map[string]interface{}, name string) 
 	}
 	customConfig, err := anypb.New(typedStruct)
 	if err != nil {
-		t.Fatalf("createXDSTypedStruct Failed during anypb.New: %v", err)
+		t.Fatalf("createXDSTypedStruct failed during anypb.New: %v", err)
 	}
 	return customConfig
 }
--- a/test/xds/xds_server_rbac_test.go
+++ b/test/xds/xds_server_rbac_test.go
@@ -692,10 +692,13 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 				// Toggle RBAC back on for next iterations.
 				envconfig.XDSRBAC = true
 
-				if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
-					t.Errorf("Authorization decision do not match\ndiff (-got +want):\n%s", diff)
+				if test.wantAuthzOutcomes != nil {
+					if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
+						t.Errorf("Authorization decision do not match\ndiff (-got +want):\n%s", diff)
+					}
 				}
 				if test.eventContent != nil {
+
 					if diff := cmp.Diff(lb.lastEvent, test.eventContent); diff != "" {
 						t.Errorf("Unexpected Message\ndiff (-got + want):\n%s", diff)
 					}
--- a/test/xds/xds_server_rbac_test.go
+++ b/test/xds/xds_server_rbac_test.go
@@ -20,6 +20,7 @@ package xds_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strconv"
@@ -28,6 +29,7 @@ import (
 
 	v3routerpb "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/router/v3"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/authz/audit"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/internal"
@@ -894,4 +896,34 @@ func (s) TestRBACToggledOff_WithBadRouteConfiguration(t *testing.T) {
 	if _, err := client.UnaryCall(ctx, &testpb.SimpleRequest{}); status.Code(err) != codes.OK {
 		t.Fatalf("UnaryCall() returned err with status: %v, if RBAC is disabled all RPC's should proceed as normal", status.Code(err))
 	}
+}
+
+type statAuditLogger struct {
+	authzDecisionStat map[bool]int // Map to hold counts of authorization decisions
+	lastEvent         *audit.Event // Field to store last received event
+}
+
+func (s *statAuditLogger) Log(event *audit.Event) {
+	s.authzDecisionStat[event.Authorized]++
+	*s.lastEvent = *event
+}
+
+type loggerBuilder struct {
+	authzDecisionStat map[bool]int
+	lastEvent         *audit.Event
+}
+
+func (loggerBuilder) Name() string {
+	return "stat_logger"
+}
+
+func (lb *loggerBuilder) Build(audit.LoggerConfig) audit.Logger {
+	return &statAuditLogger{
+		authzDecisionStat: lb.authzDecisionStat,
+		lastEvent:         lb.lastEvent,
+	}
+}
+
+func (*loggerBuilder) ParseLoggerConfig(config json.RawMessage) (audit.LoggerConfig, error) {
+	return nil, nil
 }
--- a/test/xds/xds_server_rbac_test.go
+++ b/test/xds/xds_server_rbac_test.go
@@ -696,12 +696,12 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 
 				if test.wantAuthzOutcomes != nil {
 					if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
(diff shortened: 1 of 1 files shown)


test/xds/xds_server_rbac_test.go
```diff
@@ -605,6 +630,13 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			func() {
+				// TODO add check for these stats
+				lb := &loggerBuilder{
+					authzDecisionStat: map[bool]int{true: 2, false: 0},
+					lastEvent:         &audit.Event{},
```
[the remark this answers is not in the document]
gtcooke94: So we would just be changing around whether we are storing pointers/refs or just the actual object?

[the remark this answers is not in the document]
rockspore: No. It should still store pointers. But `cmp.Diff(nil, nil)` will be ok. So when neither `lastEvent` nor `test.eventContent` is populated the test should pass.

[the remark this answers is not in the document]
gtcooke94: Discussed offline, we can't use `nil` because for the pointer to correctly propagate from the logger to the logger builder we do `*s.lastEvent = *lastEvent`, setting the underlying content being pointed to.

If `s.lastEvent` is `nil`, this causes a segfault, so we have to set it to an `audit.Event{}`

test/xds/xds_server_rbac_test.go
```diff
@@ -605,6 +630,12 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			func() {
```
rockspore, line 632: This is not new but do you know why such a anonymous func is needed here?

test/xds/xds_server_rbac_test.go
```diff
@@ -660,6 +691,18 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
...
 				// Toggle RBAC back on for next iterations.
 				envconfig.XDSRBAC = true
+
+				if test.wantAuthzOutcomes != nil {
+					if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
+						t.Errorf("Authorization decision do not match\ndiff (-got +want):\n%s", diff)
+					}
+				}
+				if test.eventContent != nil {
+
```
rockspore, line 701: nit: remove extra empty line.

test/xds/xds_server_rbac_test.go
```diff
@@ -660,6 +691,18 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
...
+					if diff := cmp.Diff(lb.lastEvent, test.eventContent); diff != "" {
+						t.Errorf("Unexpected Message\ndiff (-got + want):\n%s", diff)
```
rockspore, line 703: Why is it called "unexpected message"? Maybe something like event content makes more sense?

test/xds/xds_server_rbac_test.go
```diff
@@ -895,3 +938,55 @@ func (s) TestRBACToggledOff_WithBadRouteConfiguration(t *testing.T) {
...
+// This is used when converting a custom config from raw JSON to a TypedStruct.
+// The TypeURL of the TypeStruct will be "grpc.authz.audit_logging/<name>".
+const typeURLPrefix = "grpc.authz.audit_logging/"
+
+// Builds custom configs for audit logger RBAC protos.
+func createXDSTypedStruct(t *testing.T, in map[string]interface{}, name string) *anypb.Any {
+	t.Helper()
+	pb, err := structpb.NewStruct(in)
+	if err != nil {
+		t.Fatalf("createXDSTypedStructFailed during structpb.NewStruct: %v", err)
```
rockspore, line 981: space before failed. same for the one below.

test/xds/xds_server_rbac_test.go
```diff
@@ -433,10 +440,28 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
...
+							},
+						},
+					},
 				},
 			},
 			wantStatusEmptyCall: codes.OK,
 			wantStatusUnaryCall: codes.OK,
+			wantAuthzOutcomes:   map[bool]int{true: 2, false: 0},
+			eventContent: &audit.Event{
+				FullMethodName: "/grpc.testing.TestService/UnaryCall",
```
rockspore, line 461: It's blocked by another PR but I'd prefer the policy name, which would be the RBAC filter name, can be verified to be populated here.

test/xds/xds_server_rbac_test.go
gtcooke94: No clue, I'm just working into the existing test

test/xds/xds_server_rbac_test.go
gtcooke94: Changed to `Unexpected event`

test/xds/xds_server_rbac_test.go
gtcooke94: Good catch!

test/xds/xds_server_rbac_test.go
gtcooke94: I'll add a TODO

test/xds/xds_server_rbac_test.go
```diff
@@ -660,6 +693,17 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 				}
 				// Toggle RBAC back on for next iterations.
 				envconfig.XDSRBAC = true
+
+				if test.wantAuthzOutcomes != nil {
+					if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
+						t.Errorf("Authorization decision do not match\ndiff (-got +want):\n%s", diff)
```
rockspore, line 699: Unless it's the name of the function, I don't think error messages should start with capital letters. Same below.

test/xds/xds_server_rbac_test.go
```diff
@@ -660,6 +693,17 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 				}
 				// Toggle RBAC back on for next iterations.
 				envconfig.XDSRBAC = true
```
rockspore, line 695: I think it's worth considering adding some check to verify audit logging is not active till RBAC is toggled back.

rockspore (commented):
Mostly LGTM. I'm also open to adding more test cases but I'll leave it to you and other reviewers.

test/xds/xds_server_rbac_test.go
gtcooke94: I think it ends up being somewhat implicit in the test - it makes 2 calls with RBAC on, then 2 calls with RBAC off, and the expected outcome for logging is `allowed: 2, denied: 0`

gtcooke94: Agree that this isn't the best as it requires holding some test state in your head as you look at it, but I hesitate to make large changes to this existing test

test/xds/xds_server_rbac_test.go
```diff
@@ -895,3 +939,55 @@ func (s) TestRBACToggledOff_WithBadRouteConfiguration(t *testing.T) {
...
+// This is used when converting a custom config from raw JSON to a TypedStruct.
+// The TypeURL of the TypeStruct will be "grpc.authz.audit_logging/<name>".
+const typeURLPrefix = "grpc.authz.audit_logging/"
+
+// Builds custom configs for audit logger RBAC protos.
+func createXDSTypedStruct(t *testing.T, in map[string]interface{}, name string) *anypb.Any {
+	t.Helper()
+	pb, err := structpb.NewStruct(in)
+	if err != nil {
+		t.Fatalf("createXDSTypedStruct Failed during structpb.NewStruct: %v", err)
```
rockspore, line 982: Micro nit: I think should be lowercase `failed`. Same below.

rockspore (approved):

arvindbr8 (changes requested):
Just one minor comment about the recoverable assertion check. LGTM otherwise.

test/xds/xds_server_rbac_test.go
```diff
@@ -660,6 +693,17 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
...
 				envconfig.XDSRBAC = true
+
+				if test.wantAuthzOutcomes != nil {
+					if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
+						t.Errorf("authorization decision do not match\ndiff (-got +want):\n%s", diff)
+					}
+				}
+				if test.eventContent != nil {
+					if diff := cmp.Diff(lb.lastEvent, test.eventContent); diff != "" {
+						t.Errorf("unexpected event\ndiff (-got + want):\n%s", diff)
```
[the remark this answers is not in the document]
gtcooke94: Fixed

test/xds/xds_server_rbac_test.go
```diff
@@ -660,6 +693,17 @@ func (s) TestRBACHTTPFilter(t *testing.T) {
 				}
 				// Toggle RBAC back on for next iterations.
 				envconfig.XDSRBAC = true
+
+				if test.wantAuthzOutcomes != nil {
+					if diff := cmp.Diff(lb.authzDecisionStat, test.wantAuthzOutcomes); diff != "" {
+						t.Errorf("authorization decision do not match\ndiff (-got +want):\n%s", diff)
```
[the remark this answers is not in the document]
gtcooke94: Changed them to `Fatalf` - if the authz decisions don't match it doesn't make a lot of sense to try to compare the authz event outcomes because they are very likely nonsensical (because at a higher level the decision was wrong).

arvindbr8 (approved):

PR #6404, opened 2023-06-23
[PSM interop] Don't fail target if sub-target already failed (v1.47.x backport)

XuanWang-Amos:
Backport of #6390 to v1.47.x.
---
We don't want file multiple bugs if any of the sub-tests of the `url_map` test fails.

The same change in core: https://github.com/grpc/grpc/pull/33520.

RELEASE NOTES: N/A


zasweq (approved):
LGTM.

PR #6421, opened 2023-06-28
Fix typo in SECURITY.md and add .gitignore for VScode

SpencerBrown:
RELEASE NOTES: none


easwars (commented):
I'm happy to accept the type fix in SECURITY.md, but we have historically maintained that the `.gitignore` should be a configuration on the user owning the git repo and not centrally here. We possibly can't have entries in the `.gitignore` to satisfy every developer using grpc-go.

PR #6447, opened 2023-07-14
xds: Passthrough filter name to RBAC filters

gtcooke94:
Our filters are named in the input configs, but the overall name was not getting passed through to the rbac config.

This PR makes that name get passed through via the `ServerInterceptorBuilder`. Everything else that uses `ServerInterceptorBuilder` had to have this `name` parameter added, but it is blank for them.

Each `engine` in the constructed `ChainEngine` is expected to have the same overall policy name. I exported a new function `SetName` on `ChainEngine` so that the rbac code could set this value.

RELEASE NOTES: N/A

### DIFF
--- a/internal/xds/rbac/rbac_engine.go
+++ b/internal/xds/rbac/rbac_engine.go
@@ -112,6 +112,7 @@ func (cre *ChainEngine) IsAuthorized(ctx context.Context) error {
 	return nil
 }
 
+// SetName sets the policyName on each engine in the chain to name.
 func (cre *ChainEngine) SetName(name string) {
 	for _, e := range cre.chainedEngines {
 		e.policyName = name
--- a/internal/xds/rbac/rbac_engine.go
+++ b/internal/xds/rbac/rbac_engine.go
@@ -112,6 +112,12 @@ func (cre *ChainEngine) IsAuthorized(ctx context.Context) error {
 	return nil
 }
 
+func (cre *ChainEngine) SetName(name string) {
+	for _, e := range cre.chainedEngines {
+		e.policyName = name
+	}
+}
+
 // engine is used for matching incoming RPCs to policies.
 type engine struct {
 	// TODO(gtcooke94) - differentiate between `policyName`, `policies`, and `rules`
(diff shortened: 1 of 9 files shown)


xds/internal/httpfilter/rbac/rbac.go
```diff
@@ -179,7 +179,7 @@ var _ httpfilter.ServerInterceptorBuilder = builder{}
 
 // BuildServerInterceptor is an optional interface builder implements in order
```
rockspore, line 180: Do you mind rewording this comment? It doesn't feel grammatically correct to me.

internal/xds/rbac/rbac_engine.go
```diff
@@ -112,6 +112,12 @@ func (cre *ChainEngine) IsAuthorized(ctx context.Context) error {
 	return nil
 }
 
+func (cre *ChainEngine) SetName(name string) {
```
rockspore, line 116: If we are to have such a public method, can we unify two cases by also using this in https://github.com/grpc/grpc-go/blob/b0842e59c6cfbbe728a01d8765b0cb6c378b93d6/authz/grpc_authz_server_interceptors.go? We then can remove the `policyName` passed into `NewChainEngine` in `internal/xds/rbac` and `newEngine` below.

gtcooke94: Making sure I'm interpreting right - you're suggesting removing `name` from `NewChainEngine` and `newEngine`, not setting the name at all there, then using the new `SetName` function in the same way after they are built?

I agree it's cleaner not to set the name two different ways

gtcooke94: Will change, I like that - makes more sense than having two different ways of setting the name

xds/internal/httpfilter/rbac/rbac.go
gtcooke94: `builder` in this case is a noun referencing the `builder` struct rather than reading as `an interface-builder`

Adding the word `that` between `interface` and `builder` to make it clear

internal/xds/rbac/rbac_engine.go
gtcooke94: PTAL

rockspore (approved):
LGTM

xds/internal/httpfilter/rbac/rbac.go
rockspore: Thanks!

gtcooke94:
Closing because we decided that we do not want to pass through the filter name in this way

Issue #6485, opened 2023-07-28
loopyWriter continuing to run after GracefulStop causing race with t.Log


fho:
Hello,

we are starting a grpc-server in a Go testserver.
On test finish we call `Server.GracefulStop()` via `t.Cleanup()`.

We forward all log messages from the grpc packages to the test logger, by setting `grpclog.SetLoggerV2` to a `t.Log` wrapper.

When we run the tests with the `-race` parameter, it happens from now and then that a race is detected because loopyWriter logs a message to `t.Log()` after the test finish.
As I understand `Server.GracefulStop()` does not wait until the loopyWriter go-routine has terminated.

The message that is logged after test termination from loopyWriter happens here: https://github.com/grpc/grpc-go/blob/b8d36caf8d036e1fb91898bf93ccdc7c4e5a40ea/internal/transport/controlbuf.go#L543) maybe there are also other log calls that can happen after `GracefulStop`.

Is there a way to wait in a testcase until loopyWriter terminated that I'm missing?
If not, shouldn't it be possible to do a termination that ensures no go-routines are running afterwards?
This would also e.g. work well together with https://github.com/uber-go/goleak, that checks if any go-routines are leaked in a testcase.

grpc-go version: 1.57

dfawley:
> We forward all log messages from the grpc packages to the test logger, by setting grpclog.SetLoggerV2 to a t.Log wrapper.

We do the same thing in our tests.  :smile: 

It looks like `GracefulStop` isn't waiting for all the resources to go away - it's just waiting for the in-flight streams to end.

`Stop` is really what you should be using instead for this purpose.  `GracefulStop` makes sure pending streams have a chance to end gracefully, but in a test, you shouldn't have streams running when you think your test is done.  If you do actually need the behavior of `GracefulStop`, you can always do both: `t.Cleanup(func() { s.GracefulStop(); s.Stop() })`  But as a cleanup, you really _only_ want the hard-shutdown behavior, because if a test has hung, `Stop` will quickly terminate things, whereas `GracefulStop` will just hang forever.

`GracefulStop` should probably be updated to ensure the transports are fully shut down as well, but this would be a minor issue that we probably wouldn't be able to prioritize for some time.

fho:
@dfawley thanks a lot for your help.

Yes, `s.Stop()` is also sufficient in the testcases, all clients are shutdown before anyways.

Why should using `s.Stop()` help? It also does not wait for the loopyWriter go-routine.
[...]

PR #6489, opened 2023-07-31
grpc: Wait until resources finish cleaning up in Stop() and GracefulStop()

fho:
It was not guaranteed that when Stop() and GracefulStop() returned, the
loopyWriter go-routines terminated.
This can lead to a panic or a testing.(*common).logDepth() race
condition in Go Tests because t.Log is used after or during the testcase
terminates.
This can happen when:
- a GRPC server is started in a Go test,
- the GRPC logs are forwarded to t.Log,
- loopyWriter.run logs an error after server.Stop() and the Test
  method returns.

grpc@v1.57.0/internal/leakcheck is unable to detect that the
loopyWriter go-routine continues to run after server.Stop() because it
waits up to 10s for go-routines to terminate after a test finishes.
The loopyWriter returns much faster after Stop() returns.

To ensure that Stop and GracefulStop only return after the loopyWriter
go-routine terminates:
- rename http2Server.writerDone to loopyWriterDone, the channel was only
  used in test and is closed when loopyWriter returns
- wait in http2Server.HandleStreams() until loopyWriterDone was closed
- wait in Server.Stop() until all connections were removed, this is
  already done in GracefulStop.
  To abort GracefulStop() when Stop() is called a local copy of Server.lis and
  Server.conns was created and both fields were set to nil and then
  s.csv.Broadcast() was called to wake up an eventually sleeping
  GracefulStop() execution. This does not work anymore when we need to
  wait for connections to be removed in Stop() we need to check the
  length of Server.conns.
  Instead GracefulStop now waits until the the procedure completed.

GracefulStop now also stops running stream workers. Shared code between Stop and GracefulStop is moved to helper functions.

This fixes: https://github.com/grpc/grpc-go/issues/6485

RELEASE NOTES:
* grpc: Wait until resources finish cleaning up in Stop() and GracefulStop()


fho:
I don't know a good way to add a testcase for it because the mentioned issue with leakcheck.
To reproduce the issue consistently it would also be needed to inject into loopyWriter.run to wait a bit longer on shutdown.
To test the PR, I modified leakcheck, inserted a sleep in loopyWriter.run and did not use `test.s`: https://github.com/fho/grpc-go/commit/f2dce22e524332464bebcd9a30a128d5b6ab9c07.

The testcase also detects another running go-routine of the grpc-client:
```
    leakcheck.go:116: Leaked goroutine: goroutine 13 [runnable]:
        runtime.Gosched()
        	/usr/lib/go/src/runtime/proc.go:321 +0x1a
        google.golang.org/grpc/internal/transport.(*loopyWriter).run(0xc000318930)
        	/home/fho/git/grpc-go/internal/transport/controlbuf.go:590 +0x34b
        google.golang.org/grpc/internal/transport.newHTTP2Client.func6()
        	/home/fho/git/grpc-go/internal/transport/http2_client.go:451 +0x165
        created by google.golang.org/grpc/internal/transport.newHTTP2Client
        	/home/fho/git/grpc-go/internal/transport/http2_client.go:449 +0x39d2
    leakcheck.go:116: Leaked goroutine: goroutine 20 [runnable]:
```
This looks like analogous to the loopyWriter issue. I only had a brief look, I guess can be fixed by changing `readerDone` in a similar manner. That is an issue for another PR.

fho:
@dfawley thanks for your feedback, I'll try to implement the suggested approach.

Do you maybe have an idea how to add a testcase for the issue that is compatible with the current test code?

dfawley:
> Do you maybe have an idea how to add a testcase for the issue that is compatible with the current test code?

The only thing I could think of that wouldn't be flaky is a version of `leakcheck` that fails immediately instead of waiting 10 seconds.  If you wanted it to be reliable, you'd have to inject some extra code to optionally `Sleep` as you mention.  And that would only help catch the _current_ goroutines, not any added in the future.

But 1. it's OK to not have a test for this behavior, and 2. if there is a test for it, it's fine if it only fails ~10% of the time when the bug is present.

fho:
I pushed a new version, were I wait in HandleStreams() for the loopyWriter to terminate and wait in Stop() until all connections are removed. This does not work yet though.
When running the tests, Stop() is hanging forever when waiting for the connections to be removed (see [CI Logs](https://github.com/grpc/grpc-go/actions/runs/5809924418/job/15749726038?pr=6489).
Does somebody has an idea what is wrong?

dfawley:
> Stop() is hanging forever when waiting for the connections to be removed (see CI Logs.
> Does somebody has an idea what is wrong?

I think I see what's happening.  `HandleStreams` is returning, but `(*grpc.Server).serveStreams`, which calls it, is not returning because the _method handlers_ are still running.  Because `serveStreams` is not returning, even though the connection is fully closed, `removeConn` is not executed, meaning the `grpc.Server` (and `Stop`) still thinks the connection is there.

Since `HandleStreams` now does not return until it has cleaned up the connection, I wonder if we can remove the waitgroup in `serveStreams` entirely now?  That seems plausible.

fho:
> Since HandleStreams now does not return until it has cleaned up the connection, I wonder if we can remove the waitgroup in serveStreams entirely now? That seems plausible.

The `wg` in `Server.serveStreams()` is used to wait is for a go-routine that runs `Server.handleStream()` as part of the `st.HandleStreams()` `handle` func.
I don't see how it would be ensured without the waitGroup that the `Server.handleStream()` go-routines return before Stop/GracefulStop would return.

fho:
I'm also wondering why `s.stopServerWorkers` is only called in `Stop()` but not in `GracefulStop`.

dfawley:
> The wg in Server.serveStreams() is used to wait is for a go-routine that runs Server.handleStream() as part of the st.HandleStreams() handle func.

This doesn't sound right.

`st.HandleStreams` calls the closure that `server.go` passes it for each RPC.  When an RPC starts, the first thing that does is increment the `wg`, and when it finishes, it calls `wg.Done()` on it.  This means all the server handlers have to return before `wg.Wait()` will ever unblock.

But you changed `st.HandleStreams` so that it blocked until the transport was closed.  At that point, it seems fine to dereference the `st` which is what happens when `serveStreams` returns ([ref](https://github.com/grpc/grpc-go/blob/61a1f7792387b2cce06869a49838741393b0e850/server.go#L928)).  We can't rely on the method handlers all returning when the stream ends (these tests that are failing are blocked in the method handlers, which presumably will be cleaned up later), because it's running code that isn't owned by gRPC at that point.

> I'm also wondering why s.stopServerWorkers is only called in Stop() but not in GracefulStop.

This just seems to be an oversight.

fho:
@dfawley that works! 
I still don't understand the code sufficiently to judge if my change is correct.
Should I keep the abortGracefulStop atomic.Bool or is there a better way to abort the GracefulStop?

Tests currently fail on go 1.18 because atomic.Bool does not exist in that version.
Because Golang 1.21 was released and the min. required go version will therefore be changed to 1.19 which supports it, I kept it.

dfawley (commented):
> Should I keep the abortGracefulStop atomic.Bool or is there a better way to abort the GracefulStop?

I'm not sure I understand that part of this change.  `GracefulStop()` should continue to block until all connections are closed, even if `Stop()` is called, shouldn't it?

fho:
I have been on vacation. I'll follow-up this week.

internal/transport/http2_server.go
```diff
@@ -637,6 +637,7 @@ func (t *http2Server) operateHeaders(frame *http2.MetaHeadersFrame, handle func(
 // traceCtx attaches trace to ctx and returns the new context.
 func (t *http2Server) HandleStreams(handle func(*Stream), traceCtx func(context.Context, string) context.Context) {
 	defer close(t.readerDone)
+	defer func() { <-t.loopyWriterDone }()
```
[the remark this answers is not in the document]
fho: done

fho:
> > I'm also wondering why s.stopServerWorkers is only called in Stop() but not in GracefulStop.
> 
> This just seems to be an oversight.

I also added the call to GracefulStop.


> > Should I keep the abortGracefulStop atomic.Bool or is there a better way to abort the GracefulStop?
> 
> I'm not sure I understand that part of this change. `GracefulStop()` should continue to block until all connections are closed, even if `Stop()` is called, shouldn't it?

I removed that. My initial reason was to keep the current behavior but it actually did not even do that.

I also moved code that was used in Stop and GracefulStop to private functions.
This reduced the size of the Stop and GracefulStop functions and makes it easier to compare them. 
If you shouldn't be on board with that or should be done i a separate PR let me know or feel free too revert that commit.

arvindbr8:
@fho Might be a good thing to rebase to master and resolve conflicts in the two files. Also Doug is out sick at the moment. Seems like he would be out the rest of the week. We would need to wait on him since he has the most context on this change.

fho:
@arvindbr8 thanks for the update. I rebased the PR.

dfawley (commented):
This looks really, really good.  Very clean.  Thank you!  A few comments in-line, but nothing major.

One request: can we do something to test this functionality?  Like create a server (you can use the `stubserver` package to do this in just a few lines) and keep an RPC open, ensure `GracefulStop()` blocks while a stream is open but unblocks when it ends (and ensure all `net.Conn`s are closed by the time `GracefulStop()` returns), and also ensure `Stop()` will hard terminate a stream and unblock `GracefulStop()`

(Sorry for the delay in review.  I was out for awhile and it took me longer than expected to catch up on everything.)

dfawley:
Friendly ping since our stale bot can be a little rude.

I see you reacted to my PR comments.  Let me know if you have any questions and/or if you need extra time updating the PR.  Thanks!

fho:
> I see you reacted to my PR comments. Let me know if you have any questions and/or if you need extra time updating the PR. Thanks!

Currently not, your feedback is clear and makes also sense to me.
I currently don't know when I'll find time to continue with it but it's on my list and not forgotten.
If somebody else has time to continue it, please feel free to leave a short comment and pick it up. :-)

fho:
please don't close it.

dfawley:
Yes, we'd like this PR, too.  Unfortunately it's been hard to find time to pick it up, so hopefully you'll be able to take care of the last couple things.

server.go
```diff
@@ -1868,16 +1850,47 @@ func (s *Server) GracefulStop() {
...
+func (s *Server) closeServerTransports() {
+	for _, conns := range s.conns {
+		for st := range conns {
+			st.Close(errors.New("Server.Stop called"))
+		}
+	}
+}
+
+// s.mu must be held by the caller.
+func (s *Server) drainAllServerTransports() {
```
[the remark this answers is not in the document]
fho: Server.drainServerTransports already exist

server.go
```diff
@@ -1868,16 +1850,47 @@ func (s *Server) GracefulStop() {
...
-	if s.conns == nil {
-		s.mu.Unlock()
-		return
+	s.closeListeners()
+	s.drainAllServerTransports()
+	s.mu.Unlock()
+
+	// Wait for serving threads to be ready to exit.  Only then can we be sure no
+	// new conns will be created.
+	s.serveWG.Wait()
```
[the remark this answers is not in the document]
fho: > Don't we need to do this before we drain the transports, then?

The order looks correct to me:
1. First we close all listeners to prevent that new TCP connections are accepted
2. Then we call `drainAllServerTransports`, inform the clients about the incoming shutdown by sending GOAWAY
3. Then wait until all connections were closed (`s.serveWG.Wait()`).

Does that make sense or am I missing?


> Can we unify them even more, and have a stop(drain bool) or stop(stopTransport(*transport.ServerTransport)) method that the existing GracefulStop() and Stop() call to do everything?

great idea, I'll have a look

server.go
```diff
@@ -615,7 +614,6 @@ func (s *Server) serverWorker() {
 }
 
 func (s *Server) handleSingleStream(data *serverWorkerData) {
```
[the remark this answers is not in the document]
fho: In the meantime it got already removed in f2180b4d5403d2210b30b93098eb7da31c05c721

server.go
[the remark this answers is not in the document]
fho: changed

fho:
@dfawley I addresses all your suggestions except the tests.

> One request: can we do something to test this functionality? Like create a server (you can use the stubserver package to do this in just a few lines) and keep an RPC open, ensure GracefulStop() blocks while a stream is open but unblocks when it ends (and ensure all net.Conns are closed by the time GracefulStop() returns), and also ensure Stop() will hard terminate a stream and unblock GracefulStop()

I put some hours into it and doesn't have anything working.
I have a testcase that checks  if GracefulStop() blocks  after a  runnig GRPC-Server UnaryCall returns, which it does not.
So blocking until all grpc method calls finished seems not to be working.
Do you have an idea what might be wrong?
The testcase is here: https://github.com/fho/grpc-go/blob/7bb2f79cf7be6fe693a02c244a04d2db75700c19/test/gracefulstop_test.go#L221

server.go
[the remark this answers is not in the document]
dfawley: >> Don't we need to do this before we drain the transports, then?
> The order looks correct to me:
> 1. First we close all listeners to prevent that new TCP connections are accepted
> 2. Then we call drainAllServerTransports, inform the clients about the incoming shutdown by sending GOAWAY

I think there's a misunderstanding on your part.  `serveWG.Wait()` unblocks when all `Serve` calls unblock.  This happens when the listeners are successfully closed and we're done accepting new connections.

Your call to `drain` is before the call to `serveWG.Wait()`, but if we don't do `serveWG.Wait()` first, then we might miss draining a new transport.

server.go
```diff
@@ -1864,6 +1847,10 @@ func (s *Server) GracefulStop() {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	if !graceful {
+		s.closeServerTransports()
+	}
```
dfawley, line 1852: I do believe this should be:

```go
if graceful {
	s.drainAllServerTransports()
} else {
	s.closeServerTransports()
}
```

dfawley (commented):
> I put some hours into it and doesn't have anything working.

Oh no, sorry!  I thought this was really close to being done before, with just the small tweak to unify the code paths that were otherwise the same.

If you take your PR out of draft mode, if you're ready, the CI will run and I can look at what errors you're getting.  Thanks again for your contribution.

server.go
fho: I think I understood it now:
`closeServerTransports` is only called on a graceful stop here because it will abruptly terminate the connections.
In graceful mode, the connections will be closed in `*http2Server.HandleStreams` instead, after `EOF` on the connection

server.go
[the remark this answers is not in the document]
fho: thanks, got it!

I changed the order

dfawley:
> I have in another branch a testcase that checks if GracefulStop blocks until the client GRPC finished, which does not.
> You can find the test and results here:

The problem is the test is gracefully stopping a different server than the one the call is going to.

`ss.Start` creates and starts a server.  That server can be accessed via `ss.S`.  You are creating another server here, but then the client doesn't use it, because there was no way it would know how to reach it (the client in `ss` uses the server in `ss`):

https://github.com/fho/grpc-go/blob/28b2e40bc9608013fabb02bf74855e1c588890f1/test/gracefulstop_test.go#L232-L233

Deleting that server and using `ss.S.GracefulStop()` should take care of things.

dfawley:
One other thing:

Your test races starting the unary RPC with calling `GracefulStop`.  You probably want a signal that the server handler has been invoked before the call to `GracefulStop`.

fho:
> > I have in another branch a testcase that checks if GracefulStop blocks until the client GRPC finished, which does not.
> > You can find the test and results here:
> 
> The problem is the test is gracefully stopping a different server than the one the call is going to.

omg :facepalm: thanks a lot!

test/gracefulstop_test.go
```diff
@@ -216,3 +217,90 @@ func (s) TestGracefulStopClosesConnAfterLastStream(t *testing.T) {
...
+		t.Error("GracefulStop returned before GRPC method call ended")
+	default:
+	}
+
+	unblockGRPCCall <- struct{}{}
+	<-grpcClientCallReturned
+	<-gracefulStopReturned
+}
+
+func (s) TestStopAbortsBlockingGAbortsBlockingGRPCCall(t *testing.T) {
```
fho, line 269: The test for `Stop()` is in a file called `GracefulStop()`.
Creating an extra file `stop_test.go` file for a single tests seems over the top me.

We could:
- keep  the Stop() testcase in the gracefulstop_test.go file for now
- rename the `gfracefulstop_test.go` file to `server_stop_test.go`
- move all tests from gracefulstop_test.go to server_test.go

test/gracefulstop_test.go
```diff
@@ -216,3 +217,90 @@ func (s) TestGracefulStopClosesConnAfterLastStream(t *testing.T) {
...
+		ss.S.GracefulStop()
+		close(gracefulStopReturned)
+	}()
+
+	time.Sleep(time.Second)
+	select {
+	case <-gracefulStopReturned:
+		t.Error("GracefulStop returned before GRPC method call ended")
+	default:
+	}
```
fho, line 262: I'm calling `Sleep` to catch the case that `GracefulStop` would not wait for the client to terminate and delay unblocking the grpc client, to make it likely that it is still running while `GracefulStop` is executing.
Is there are more reliable way to ensure that `GracefulStop` is waiting for the client call to finish here?

test/gracefulstop_test.go
dfawley: My preference would be to move these tests to the root directory and put them in the `grpc_test` package.  Directly in `server_ext_test.go` seems perfect.

If for some reason that doesn't work (like you need functionality that's in the `test/` package), then that's fine, you can leave them there.

Also, no need to move the existing tests (unless you want to and it turns out to be easy, in which case SGTM).

internal/transport/http2_server.go.orig
dfawley, line 1: This file needs to be reverted.

test/gracefulstop_test.go
fho: done

test/gracefulstop_test.go
fho: Ok, then I'll keep the tests where they are.
They use `isConnClosedErr()` and now also `defaultTestShortTimeout` from the test package.

test/gracefulstop_test.go
```diff
@@ -216,3 +217,90 @@ func (s) TestGracefulStopClosesConnAfterLastStream(t *testing.T) {
...
+	}()
+
+	gracefulStopReturned := make(chan struct{})
+	<-grpcCallExecuting
+	go func() {
+		ss.S.GracefulStop()
+		close(gracefulStopReturned)
+	}()
+
+	time.Sleep(time.Second)
```
dfawley, line 257: Please remove this sleep now.  The `<-time.After` is meant to replace it.

test/gracefulstop_test.go
```diff
@@ -216,3 +217,90 @@ func (s) TestGracefulStopClosesConnAfterLastStream(t *testing.T) {
...
+		t.Error("GracefulStop returned before GRPC method call ended")
+	case <-time.After(defaultTestShortTimeout):
+	}
+
+	unblockGRPCCall <- struct{}{}
+	<-grpcClientCallReturned
+	<-gracefulStopReturned
+}
+
+func (s) TestStopAbortsBlockingGAbortsBlockingGRPCCall(t *testing.T) {
```
dfawley, line 269: I think something weird happened with this test name.

dfawley (approved):
LGTM except for the last two small things.  Thank you!

test/gracefulstop_test.go
fho: Yes, of course, removed

test/gracefulstop_test.go
fho: fixed

fho:
@dfawley I addressed them. Thanks a lot for you patience and great support and sorry for oversights at the end causing this unnecessary back and forth.

test/gracefulstop_test.go
```diff
@@ -216,3 +217,89 @@ func (s) TestGracefulStopClosesConnAfterLastStream(t *testing.T) {
 		<-gracefulStopDone // Wait for GracefulStop to return.
 	})
 }
+
+func (s) TestGracefulStopBlocksUntilGRPCConnectionsTerminate(t *testing.T) {
+	// This tests ensures that GracefulStop() blocks until all ongoing
+	// client GRPC calls finished.
```
zasweq, line 223: Nit: can you move to above the function definition. also switch "client GRPC calls" to "rpcs" please.

test/gracefulstop_test.go
```diff
@@ -216,3 +217,89 @@ func (s) TestGracefulStopClosesConnAfterLastStream(t *testing.T) {
...
+	<-grpcClientCallReturned
+	<-gracefulStopReturned
+}
+
+func (s) TestStopAbortsBlockingGRPCCall(t *testing.T) {
+	// This tests ensures that when Stop() is called while an ongoing grpc
+	// call is blocking that:
+	// - Stop() returns
+	// - and the GRPC call on the client side fails with an connection
+	// closed error
```
zasweq, line 273: s/"GRPC call on the client side"/"rpc finished" please

zasweq (changes requested):
Code all LGTM. All comments are minor nits with respect to comments/naming.

server.go
```diff
@@ -1854,62 +1849,66 @@ func ServerTransportStreamFromContext(ctx context.Context) ServerTransportStream
...
+	for len(s.conns) != 0 {
+		s.cv.Wait()
 	}
+	s.conns = nil
+}
 
-	for lis := range s.lis {
-		lis.Close()
+// s.mu must be held by the caller.
+func (s *Server) closeServerTransports() {
```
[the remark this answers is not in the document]
fho: Not "Unlocked"? Using "Locked" as suffix sounds they hold locks, which they don't.

server.go
```diff
@@ -1854,62 +1849,66 @@ func ServerTransportStreamFromContext(ctx context.Context) ServerTransportStream
...
 	}
 
-	s.mu.Lock()
-	if s.events != nil {
-		s.events.Finish()
-		s.events = nil
-	}
-	s.mu.Unlock()
+	s.waitForServerConnRemoval()
+	s.finishEventLog()
```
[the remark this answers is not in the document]
fho: Do you mean both functions or only finishEventLog?

test/gracefulstop_test.go
fho: Ok, I'll also do it for the existing testcases in this file then.

fho: done

server.go
```diff
@@ -1854,62 +1849,66 @@ func ServerTransportStreamFromContext(ctx context.Context) ServerTransportStream
...
-	s.quit.Fire()
-	defer s.done.Fire()
-
-	s.channelzRemoveOnce.Do(func() { channelz.RemoveEntry(s.channelzID) })
-	s.mu.Lock()
-	if s.conns == nil {
-		s.mu.Unlock()
-		return
+// s.mu must be held by the caller.
+// waitForServerConnRemoval blocks until s.conns is empty.
```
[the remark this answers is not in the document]
fho: done

server.go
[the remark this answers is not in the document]
fho: I inlined both functions as you wanted.

Imho, having those as separate  functions made the code easier to understand.

server.go
```diff
@@ -1854,62 +1849,66 @@ func ServerTransportStreamFromContext(ctx context.Context) ServerTransportStream
...
+	for _, conns := range s.conns {
+		for st := range conns {
+			st.Close(errors.New("Server.Stop called"))
+		}
 	}
-	s.lis = nil
+}
+
+// s.mu must be held by the caller.
+func (s *Server) drainAllServerTransports() {
```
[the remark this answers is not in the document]
fho: Same as https://github.com/grpc/grpc-go/pull/6489#discussion_r1375211583
Please double check if you real want to name them "Locked". Imho naming functions that don't acquire locks `Locked` is confusing.
There are also no other occurrences in the code where this naming scheme is followed.

server.go
```diff
@@ -1918,22 +1917,22 @@ func (s *Server) GracefulStop() {
...
-	s.mu.Unlock()
-	s.serveWG.Wait()
-	s.mu.Lock()
-
-	for len(s.conns) != 0 {
-		s.cv.Wait()
-	}
-	s.conns = nil
+// s.mu must be held by the caller.
+func (s *Server) finishEventLog() {
```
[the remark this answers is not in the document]
fho: obsolete, the function was inlined instead (https://github.com/grpc/grpc-go/pull/6489#discussion_r1375078599)

server.go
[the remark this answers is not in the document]
dfawley: `Locked` is a convention we use for functions that require a lock to be held before calling it.  We don't use it in older code, which is why you probably didn't see it elsewhere.

zasweq:
For the non inlined functions that are called while holding the server mutex, can you please add the suffix Locked(). Then I will approve and go ahead and merge :).

fho:
@zasweq yes, it's done! Thanks for your review

zasweq (approved):
LGTM. Thank you for the contribution.

PR #6490, opened 2023-07-31
add .gitignore

ahrtr:
add a .gitignore file to ignore `.idea` and `.DS_Store` which are automatically generated by IDE or Finder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.idea
+.DS_Store
+


ahrtr:
The two files I mentioned are actually very common files automatically generated. It's worth adding them into a .gitignore tracked by the repo.

dfawley:
If you use tools that produce these files, then you should really set your own `core.excludesFile` to prevent them from being committed to every git repo you use, instead of distributing that exclusion to everyone in every repo.

PR #6500, opened 2023-08-03
outlierdetection: fix unconditional calls of child UpdateSubConnState

dfawley:
If a child policy had a state listener for SubConn updates, we should invoke it directly instead of calling UpdateSubConnState.  Change #6481 properly forwarded updates from the parent, but missed the synthesized calls into the child to eject/uneject that originated from the outlierdetection policy itself.

RELEASE NOTES: none

### DIFF
--- a/xds/internal/balancer/outlierdetection/balancer.go
+++ b/xds/internal/balancer/outlierdetection/balancer.go
@@ -142,7 +142,6 @@ func (bb) Name() string {
 type scUpdate struct {
 	scw   *subConnWrapper
 	state balancer.SubConnState
-	cb    func(balancer.SubConnState)
 }
 
 type ejectionUpdate struct {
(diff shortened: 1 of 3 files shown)


xds/internal/balancer/outlierdetection/balancer.go
```diff
@@ -362,12 +361,11 @@ func (b *outlierDetectionBalancer) updateSubConnState(sc balancer.SubConn, state
 	b.scUpdateCh.Put(&scUpdate{
 		scw:   scw,
 		state: state,
-		cb:    cb,
 	})
 }
 
 func (b *outlierDetectionBalancer) UpdateSubConnState(sc balancer.SubConn, state balancer.SubConnState) {
-	b.updateSubConnState(sc, state, nil)
+	b.logger.Errorf("UpdateSubConnState(%v, %+v) called unexpectedly", sc, state)
```
zasweq, line 368: I don't get how if this logs an error and doesn't do anything do we keep it backward compatible?

zasweq (approved):
Oh I see, so your old pr persisted the listener callback only in the update from parent flow. This persists it in the scw to forward it down and branch on both update types.

xds/internal/balancer/outlierdetection/balancer.go
zasweq: Why not just keep the old flow the same (UpdateSubConnState works normally), until we delete? What's the advantage of this?

zasweq: Oh, I guess this won't ever get called since the Channel now calls the listener callback.

dfawley: > Why not just keep the old flow the same (UpdateSubConnState works normally), until we delete? What's the advantage of this?

The goal is to catch all the places where we use the old API and update them to the new one so the old one can be deleted.  If we keep the old API code working then we might miss a place where we needed to convert it when we go to delete the API itself.

PR #6510, opened 2023-08-08
balancer/leastrequest: Add least request balancer

zasweq:
This PR adds the least request load balancer, speced in https://github.com/grpc/proposal/blob/master/A48-xds-least-request-lb-policy.md.

RELEASE NOTES:
* balancer/leastrequest: Add least request balancer

### DIFF
--- a/balancer/leastrequest/balancer_test.go
+++ b/balancer/leastrequest/balancer_test.go
@@ -120,13 +120,12 @@ func (s) TestParseConfig(t *testing.T) {
 // setupBackends spins up three test backends, each listening on a port on
 // localhost. The three backends always reply with an empty response with no
 // error, and for streaming receive until hitting an EOF error.
-func setupBackends(t *testing.T) ([]string, func()) {
+func setupBackends(t *testing.T) []string {
 	t.Helper()
-
-	backends := make([]*stubserver.StubServer, 4)
-	addresses := make([]string, 4)
-	// Construct and start 4 working backends.
-	for i := 0; i < 4; i++ {
+	const numBackends = 3
+	addresses := make([]string, numBackends)
+	// Construct and start three working backends.
+	for i := 0; i < numBackends; i++ {
 		backend := &stubserver.StubServer{
 			EmptyCallF: func(ctx context.Context, in *testpb.Empty) (*testpb.Empty, error) {
 				return &testpb.Empty{}, nil
@@ -144,15 +143,10 @@ func setupBackends(t *testing.T) ([]string, func()) {
 			t.Fatalf("Failed to start backend: %v", err)
 		}
 		t.Logf("Started good TestService backend at: %q", backend.Address)
-		backends[i] = backend
+		t.Cleanup(func() { backend.Stop() })
 		addresses[i] = backend.Address
 	}
-	cancel := func() {
-		for _, backend := range backends {
-			backend.Stop()
-		}
-	}
-	return addresses, cancel
+	return addresses
 }
 
 // checkRoundRobinRPCs verifies that EmptyCall RPCs on the given ClientConn,
@@ -217,8 +211,7 @@ func (s) TestLeastRequestE2E(t *testing.T) {
 		index++
 		return ret
 	}
-	addresses, cancel := setupBackends(t)
-	defer cancel()
+	addresses := setupBackends(t)
 
 	mr := manual.NewBuilderWithScheme("lr-e2e")
 	defer mr.Close()
@@ -287,55 +280,38 @@ func (s) TestLeastRequestE2E(t *testing.T) {
 	index = 0
 	indexes = []uint32{
 		0, 0, // Causes first stream to be on first address.
-		0, 1, // Compares first address (which already has a RPC) to second, so choose second.
-		1, 2, // Compares second address (which already has a RPC) to third, so choose third.
+		0, 1, // Compares first address (one RPC) to second (no RPCs), so choose second.
+		1, 2, // Compares second address (one RPC) to third (no RPCs), so choose third.
+		0, 3, // Causes another stream on first address.
+		1, 0, // Compares second address (one RPC) to first (two RPCs), so choose second.
+		2, 0, // Compares third address (one RPC) to first (two RPCs), so choose third.
+		0, 0, // Causes another stream on first address.
+		2, 2, // Causes a stream on third address.
+		2, 1, // Compares third address (three RPCs) to second (two RPCs), so choose third.
 	}
-	// Start a streaming call on first, but don't finish the stream.
-	var peer1 peer.Peer
-	stream1, err := testServiceClient.FullDuplexCall(ctx, grpc.Peer(&peer1))
-	if err != nil {
-		t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
-	}
+	wantIndex := []uint32{0, 1, 2, 0, 1, 2, 0, 2, 1}
 
-	// Start a second streaming call. From the indexes injected into random
-	// number generator, this should compare Address 1, which already has a RPC
-	// to Address 2, so thus should start the stream on Address 2.
-	var peer2 peer.Peer
-	stream2, err := testServiceClient.FullDuplexCall(ctx, grpc.Peer(&peer2))
-	if err != nil {
-		t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
-	}
-
-	// Start a third streaming call. From the indexes injected into random
-	// number generator, this should compare Address 2, which already has a RPC
-	// to Address 3, so thus should start the stream on Address 3.
-	var peer3 peer.Peer
-	stream3, err := testServiceClient.FullDuplexCall(ctx, grpc.Peer(&peer3))
-	if err != nil {
-		t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
-	}
-
-	// Finish the streams to collect peer information.
-	stream1.CloseSend()
-	if _, err = stream1.Recv(); err != io.EOF {
-		t.Fatalf("unexpected error: %v, expected an EOF error", err)
-	}
-	stream2.CloseSend()
-	if _, err = stream2.Recv(); err != io.EOF {
-		t.Fatalf("unexpected error: %v, expected an EOF error", err)
-	}
-	stream3.CloseSend()
-	if _, err = stream3.Recv(); err != io.EOF {
-		t.Fatalf("unexpected error: %v, expected an EOF error", err)
-	}
-	if peer1.Addr.String() != peerAtIndex[0] {
-		t.Fatalf("got: %v, want: %v", peer1.Addr.String(), peerAtIndex[0])
-	}
-	if peer2.Addr.String() != peerAtIndex[1] {
-		t.Fatalf("got: %v, want: %v", peer2.Addr.String(), peerAtIndex[1])
-	}
-	if peer3.Addr.String() != peerAtIndex[2] {
-		t.Fatalf("got: %v, want: %v", peer3.Addr.String(), peerAtIndex[2])
+	// Start streaming RPC's, but do not finish them. Each created stream should
+	// be started based on the least request algorithm and injected randomness
+	// (see indexes slice above for exact expectations).
+	for _, wantIndex := range wantIndex {
+		stream, err := testServiceClient.FullDuplexCall(ctx)
+		if err != nil {
+			t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
+		}
+		defer func() {
+			stream.CloseSend()
+			if _, err = stream.Recv(); err != io.EOF {
+				t.Fatalf("unexpected error: %v, expected an EOF error", err)
+			}
+		}()
+		p, ok := peer.FromContext(stream.Context())
+		if !ok {
+			t.Fatalf("testServiceClient.FullDuplexCall has no Peer")
+		}
+		if p.Addr.String() != peerAtIndex[wantIndex] {
+			t.Fatalf("testServiceClient.FullDuplexCall's Peer got: %v, want: %v", p.Addr.String(), peerAtIndex[wantIndex])
+		}
 	}
 }
 
@@ -357,8 +333,7 @@ func (s) TestLeastRequestPersistsCounts(t *testing.T) {
 		index++
 		return ret
 	}
-	addresses, cancel := setupBackends(t)
-	defer cancel()
+	addresses := setupBackends(t)
 
 	mr := manual.NewBuilderWithScheme("lr-e2e")
 	defer mr.Close()
@@ -423,6 +398,7 @@ func (s) TestLeastRequestPersistsCounts(t *testing.T) {
 	// SubConns. Thus, since address 3 is the new address and the first two
 	// addresses are populated with RPCs, once the picker update of all 3 READY
 	// SubConns takes effect, all new streams should be started on address 3.
+	index = 0
 	indexes = []uint32{
 		0, 1, 2, 3, 4, 5,
 	}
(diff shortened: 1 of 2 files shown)


balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+	"google.golang.org/grpc/serviceconfig"
+)
+
+// Global to stub out in tests.
+var grpcranduint32 = grpcrand.Uint32
+
+// Name is the name of the least request balancer.
+const Name = "least_request_experimental"
+
+var logger = grpclog.Component("least request")
```
dfawley, line 39: Nit: `least-request`.  One day components will be configurable, with env vars, and spaces will be unfriendly.

balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+
+func init() {
+	balancer.Register(bb{})
+}
+
+// LBConfig is the balancer config for least_request_experimental balancer.
+type LBConfig struct {
```
dfawley, line 46: Unexport?

balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+	serviceconfig.LoadBalancingConfig `json:"-"`
+
+	// ChoiceCount is the number of random SubConns to sample to try and find
+	// the one with the Least Request. If unset, defaults to 2. If set to < 2,
```
dfawley, line 50: "with the _fewest outstanding requests_"?

balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+
+func init() {
+	balancer.Register(bb{})
+}
+
+// LBConfig is the balancer config for least_request_experimental balancer.
+type LBConfig struct {
+	serviceconfig.LoadBalancingConfig `json:"-"`
+
+	// ChoiceCount is the number of random SubConns to sample to try and find
```
dfawley, line 49: Nit: s/try and find/find/ (or use or choose or select)

balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+	// the one with the Least Request. If unset, defaults to 2. If set to < 2,
+	// will become 2, and if set to > 10, will become 10.
```
dfawley, line 51: If < 2 it is rejected.

balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (bb) Name() string {
+	return Name
+}
+
+func (bb) Build(cc balancer.ClientConn, bOpts balancer.BuildOptions) balancer.Balancer {
+	b := &leastRequestBalancer{scRPCCounts: make(map[balancer.SubConn]*int32)}
+	baseBuilder := base.NewBalancerBuilder(Name, b, base.Config{HealthCheck: true})
+	baseBalancer := baseBuilder.Build(cc, bOpts)
+	b.Balancer = baseBalancer
```
dfawley, line 85: Nit: delete the local?

balancer/leastrequest/leastrequest.go
```diff
@@ -0,0 +1,182 @@
...
+	return b
+}
+
+type leastRequestBalancer struct {
+	// Embeds balancer.Balancer because needs to intercept UpdateClientConnState
+	// to learn about choiceCount.
+	balancer.Balancer
```
dfawley, line 92: I guess I don't like the base balancer to begin with, so I shouldn't suggest this, but... Is there some way to support configs with it more directly vs. needing this pattern of wrapping it in another balancer to do so?

dfawley (commented):
I'll wait on reviewing the tests until the e2e-style ones are done.

balancer/leastrequest/leastrequest.go
zasweq: Done.

balancer/leastrequest/leastrequest.go
zasweq: Exported to prepare the LB Config easily in separate packages (the way we prepare JSON). See https://github.com/grpc/grpc-go/pull/6517/files#diff-f803bcdef3adc813ffc29931c88d8374e0e161d06f75137d8762a1c3b16d332eR199.

balancer/leastrequest/leastrequest.go
zasweq: Switched. Much better language haha.

balancer/leastrequest/leastrequest.go
zasweq: Switched.

balancer/leastrequest/leastrequest.go
zasweq: Ah whoops, forgot to switch this comment when I changed it.

balancer/leastrequest/leastrequest.go
zasweq: Done :).

balancer/leastrequest/leastrequest.go
zasweq: Ummmm so give it a knob like the knob to it to build pickers? I don't like the suggestion either, but let's video call about this if you feel strongly. This feels like a clean way to reuse logic?

dfawley: So I hadn't really thought this through fully, but I think this could work and possibly make things simpler and not be all that difficult to do, either:

- `base.baseBuilder` implements `balancer.ConfigParser`.
- In `baseBuilder.ParseConfig()`, if the associated `base.PickerBuilder` implements `ConfigParser`, then its method is called; otherwise it is a nop.
- Add the parsed LB config (via `UpdateClientConnState()`) to `base.PickerBuildInfo` so that `PickerBuilder`'s `Build()` will have access to it.

This way, you won't need to instantiate a `PickerBuilder`, create a `baseBalancer` with a reference to that, then store (embed) the `baseBalancer` inside the `PickerBuilder`.  Right now there is a cycle (A->B->A), which generally makes things harder to reason about.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+		}
+		t.Logf("Started good TestService backend at: %q", backend.Address)
+		backends[i] = backend
+		addresses[i] = backend.Address
+	}
+	cancel := func() {
+		for _, backend := range backends {
+			backend.Stop()
+		}
+	}
```
dfawley, line 154: `t.Cleanup` handles this case a little more nicely.  Use it just like `defer` but it only runs at the end of the test.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+}
+
+// setupBackends spins up three test backends, each listening on a port on
+// localhost. The three backends always reply with an empty response with no
+// error, and for streaming receive until hitting an EOF error.
+func setupBackends(t *testing.T) ([]string, func()) {
+	t.Helper()
+
+	backends := make([]*stubserver.StubServer, 4)
```
dfawley, line 126: If you use `t.Cleanup` (see below), this slice can be removed.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	// Start a third streaming call. From the indexes injected into random
+	// number generator, this should compare Address 2, which already has a RPC
+	// to Address 3, so thus should start the stream on Address 3.
+	var peer3 peer.Peer
+	stream3, err := testServiceClient.FullDuplexCall(ctx, grpc.Peer(&peer3))
+	if err != nil {
+		t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
+	}
```
dfawley, line 316: Let's do these in a loop to avoid duplication:

```go
for range peerAtIndex {
  // Start streams & validate peer (using `peer.FromContext(stream.Context())`)
  // defer stream closure to keep it open
}
```

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+	// between the indexes provided.
+	index = 0
+	indexes = []uint32{
+		0, 0, // Causes first stream to be on first address.
+		0, 1, // Compares first address (which already has a RPC) to second, so choose second.
+		1, 2, // Compares second address (which already has a RPC) to third, so choose third.
+	}
+	// Start a streaming call on first, but don't finish the stream.
+	var peer1 peer.Peer
+	stream1, err := testServiceClient.FullDuplexCall(ctx, grpc.Peer(&peer1))
```
dfawley, line 295: You should use `peer.FromContext(stream.Context())` for streaming RPCs instead.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+
+	// Start streaming RPCs, but do not finish them. Each subsequent stream
+	// should be started according to the least request algorithm, and chosen
+	// between the indexes provided.
+	index = 0
+	indexes = []uint32{
+		0, 0, // Causes first stream to be on first address.
+		0, 1, // Compares first address (which already has a RPC) to second, so choose second.
+		1, 2, // Compares second address (which already has a RPC) to third, so choose third.
+	}
```
dfawley, line 292: Let's add some more, even?
```
...
0, 3, // want: 0
1, 0, // want: 1
2, 0, // want: 2
0, 0, // want: 0
2, 2, // want: 2
2, 1, // want: 1
```

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+		}()
+	}
+
+	// Update the least request balancer to choice count 3. Also update the
+	// address list adding a third address. Alongside the injected randomness,
+	// this should trigger the least request balancer to search all created
+	// SubConns. Thus, since address 3 is the new address and the first two
+	// addresses are populated with RPCs, once the picker update of all 3 READY
+	// SubConns takes effect, all new streams should be started on address 3.
+	indexes = []uint32{
```
dfawley, line 426: You don't reset `index = 0`...  So I think this only works because you have 3 addresses and 3 choice counts.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+// setupBackends spins up three test backends, each listening on a port on
+// localhost. The three backends always reply with an empty response with no
+// error, and for streaming receive until hitting an EOF error.
+func setupBackends(t *testing.T) ([]string, func()) {
+	t.Helper()
+
+	backends := make([]*stubserver.StubServer, 4)
+	addresses := make([]string, 4)
+	// Construct and start 4 working backends.
+	for i := 0; i < 4; i++ {
```
dfawley, line 129: You start 4 backends but only ever use 3.

Please use a const instead of repeating the number 3x.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,471 @@
...
+		t.Fatalf("error in expected round robin: %v", err)
+	}
+
+	for i := 0; i < 50; i++ {
+		var peer peer.Peer
+		stream, err := testServiceClient.FullDuplexCall(ctx, grpc.Peer(&peer))
+		if err != nil {
+			t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
+		}
+		stream.CloseSend() // Finish stream to populate peer.
```
dfawley, line 463: As above, `peer.FromContext(stream.Context())` is preferred.

Should this start 50 concurrent RPCs and ensure they all go to (2), and then expect the next one on (0)?

balancer/leastrequest/balancer_test.go
zasweq: Yeah, no matter what the index starts at, it'll hit all 3 due to mod behavior. I'll add the index = 0 set anyway though.

balancer/leastrequest/balancer_test.go
zasweq: Argh I meant to change this, I updated comment but not the actual code lol. Switched to const.

balancer/leastrequest/leastrequest.go
zasweq: Discussed offline; decided to keep due to this line in the gRFC: "Counters will be local to each instance of the least_request_experimental policy and thereby not shared in cases where subchannels are used across multiple LB instances." Thus, you need heap construction for a new object that instances this state, and can't use a global picker builder registered in registry.

balancer/leastrequest/balancer_test.go
zasweq: Deleted.

balancer/leastrequest/balancer_test.go
zasweq: Changed t.Cleanup() on a backend.Stop().

balancer/leastrequest/balancer_test.go
zasweq: Done. Thanks for the suggestion. Saved a lot of duplication :).

balancer/leastrequest/balancer_test.go
zasweq: Done. Noted for the future.

balancer/leastrequest/balancer_test.go
zasweq: Done.

balancer/leastrequest/balancer_test.go
zasweq: Yeah, sure, done. It's actually 25 though, because address 1 and address 2 have 25 each :).

zasweq: Actually, you can't make an assertion once 25 RPC's are created to the added address. Ready SCs comes in as a map, which is non deterministic, so you have no control over which address the next RPC is started for (just know it's at index 0). Thus, to make an equivalent assertion, I made two unary RPC's and made sure they are equal in value.

zasweq: Discussed offline; decided to do an assertion for round robin here since it has 25 RPCs on each backend address at this point.

balancer/leastrequest/balancer_test.go
```diff
@@ -287,55 +280,38 @@ func (s) TestLeastRequestE2E(t *testing.T) {
...
+		stream, err := testServiceClient.FullDuplexCall(ctx)
+		if err != nil {
+			t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
+		}
+		defer func() {
+			stream.CloseSend()
+			if _, err = stream.Recv(); err != io.EOF {
+				t.Fatalf("unexpected error: %v, expected an EOF error", err)
+			}
+		}()
```
dfawley, line 307: I think we can just delete all of this and rely on ctx's cancelation to clean up everything.

Same with the other test.  Just make the server's handler do a `<-stream.Context().Done()` instead of doing a `stream.Recv()` in a loop.

balancer/leastrequest/balancer_test.go
```diff
@@ -451,18 +430,59 @@ func (s) TestLeastRequestPersistsCounts(t *testing.T) {
...
+	// injection/choiceCount causes all 3 to be compared every iteration).
+	for i := 0; i < 25; i++ {
+		stream, err := testServiceClient.FullDuplexCall(ctx)
 		if err != nil {
 			t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
 		}
-		stream.CloseSend() // Finish stream to populate peer.
-		if _, err = stream.Recv(); err != io.EOF {
-			t.Fatalf("unexpected error: %v, expected an EOF error", err)
+		defer func() {
```
dfawley, line 441: Same

balancer/leastrequest/balancer_test.go
```diff
@@ -451,18 +430,59 @@ func (s) TestLeastRequestPersistsCounts(t *testing.T) {
...
+		addresses[1]: 1,
+		addresses[2]: 1,
+	}
+	gotAddrCount := make(map[string]int)
+	for i := 0; i < len(addresses); i++ {
+		stream, err := testServiceClient.FullDuplexCall(ctx)
+		if err != nil {
+			t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
+		}
+		defer func() {
```
dfawley, line 471: Same

balancer/leastrequest/balancer_test.go
```diff
@@ -451,18 +430,59 @@ func (s) TestLeastRequestPersistsCounts(t *testing.T) {
...
+		if p.Addr.String() != addresses[2] {
+			t.Fatalf("testServiceClient.FullDuplexCall's Peer got: %v, want: %v", p.Addr.String(), addresses[2])
 		}
 	}
+
+	// Now 25 RPC's are active on each address, the next three RPC's should
+	// round robin, since choiceCount is three and the injected random indexes
+	// cause it to search all three addresses for fewest outstanding requests on
+	// each iteration.
+	wantAddrCount := map[string]int{
```
dfawley, line 460: For all of this and below, can you not use `checkRoundRobin`?

zasweq: No :/. checkRoundRobin gives it slack of it performs the operation within the ctx timeout. This is a lot more specific, as the moment address 3 gets populated with 25 active RPC's, *the next three RPC's should immediately be round robin*, not within the context of 5 seconds, three iterations of the three addresses happen.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,459 @@
...
+	for i := 0; i < len(addresses); i++ {
+		stream, err := testServiceClient.FullDuplexCall(ctx)
+		if err != nil {
+			t.Fatalf("testServiceClient.FullDuplexCall failed: %v", err)
+		}
+		p, ok := peer.FromContext(stream.Context())
+		if !ok {
+			t.Fatalf("testServiceClient.FullDuplexCall has no Peer")
+		}
+		if p.Addr != nil {
```
dfawley, line 452: Optional: IMO it's cleaner to remove the defensive programming and just let things panic if they're breaking assertions:

```go
p, _ := peer.FromContext(stream.Context)
gotAddrCount[p.Addr.String()]++
```

But this is fine, too.

dfawley (approved):

balancer/leastrequest/balancer_test.go
zasweq: Done.

balancer/leastrequest/balancer_test.go
zasweq: Done.

balancer/leastrequest/balancer_test.go
zasweq: Done.

balancer/leastrequest/balancer_test.go
zasweq: Done.

balancer/leastrequest/balancer_test.go
```diff
@@ -0,0 +1,459 @@
...
+			gotAddrCount[p.Addr.String()]++
+		}
+	}
+	if diff := cmp.Diff(gotAddrCount, wantAddrCount); diff != "" {
```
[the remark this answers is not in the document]
zasweq: Printed diff here. In check round robin no need for diff so just used !cmp.Equal().

PR #6520, opened 2023-08-09
xds/outlierdetection: Stop handling UpdateSubConnState forwarding

dfawley:
All children use `StateListener`, so this fallback path is no longer needed.  Update tests accordingly, too.

RELEASE NOTES: none

### DIFF
--- a/xds/internal/balancer/outlierdetection/balancer_test.go
+++ b/xds/internal/balancer/outlierdetection/balancer_test.go
@@ -685,11 +685,15 @@ func (s) TestUpdateAddresses(t *testing.T) {
 	var err error
 	stub.Register(t.Name(), stub.BalancerFuncs{
 		UpdateClientConnState: func(bd *stub.BalancerData, _ balancer.ClientConnState) error {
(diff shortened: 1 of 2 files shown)


xds/internal/balancer/outlierdetection/balancer_test.go
```diff
@@ -1387,11 +1385,6 @@ func (s) TestConcurrentOperations(t *testing.T) {
 				t.Error("ResolverError was called after Close(), which breaks the balancer API")
 			}
 		},
-		UpdateSubConnState: func(*stub.BalancerData, balancer.SubConn, balancer.SubConnState) {
-			if closed.HasFired() {
-				t.Error("UpdateSubConnState was called after Close(), which breaks the balancer API")
```
zasweq, line 1392: I'm assuming state listeners still follow this API (can't be called after close)?

zasweq (approved):
LGTM.

xds/internal/balancer/outlierdetection/balancer_test.go
dfawley: Correct!

PR #6521, opened 2023-08-09
xds/priority: stop forwarding UpdateSubConnState calls

dfawley:
balancergroup installs its own StateListener, so this is unnecessary.

RELEASE NOTES: none


arvindbr8 (approved):
lgtm

PR #6526, opened 2023-08-09
xds/clusterresolver: stop forwarding UpdateSubConnState calls

dfawley:
All children use `StateListener` so this is unnecessary now.  Also, delete associated test.

RELEASE NOTES: none

### DIFF
--- a/xds/internal/balancer/clusterresolver/clusterresolver.go
+++ b/xds/internal/balancer/clusterresolver/clusterresolver.go
@@ -17,7 +17,7 @@
  */
 
 // Package clusterresolver contains the implementation of the
-// xds_cluster_resolver_experimental LB policy which resolves endpoint addresses
+// cluster_resolver_experimental LB policy which resolves endpoint addresses
 // using a list of one or more discovery mechanisms.
 package clusterresolver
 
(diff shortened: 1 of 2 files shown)


xds/internal/balancer/clusterresolver/clusterresolver.go
```diff
@@ -421,11 +402,7 @@ func (c *ccWrapper) ResolveNow(resolver.ResolveNowOptions) {
...
 
 func (c *ccWrapper) NewSubConn(addrs []resolver.Address, opts balancer.NewSubConnOptions) (sc balancer.SubConn, err error) {
-	if opts.StateListener == nil {
-		// If already set, just allow updates to be sent directly to the
-		// child's listener.  Otherwise, we are responsible for forwarding the
-		// update we'll receive to the proper child.
-		opts.StateListener = func(state balancer.SubConnState) { c.b.UpdateSubConnState(sc, state) }
-	}
+	// No need to override opts.StateListener; just forward all calls to the
+	// child that created the SubConn.
```
zasweq, line 406: Optional nit: I don't think you need this comment anymore. If we have this comment we'd have to restate this in every balancer since we switched over.

zasweq (approved):
LGTM.

xds/internal/balancer/clusterresolver/clusterresolver.go
dfawley: Yeah, this should have just been deleted entirely since ccWrapper embeds ClientConn.  Done.

PR #6527, opened 2023-08-09
tests: stop using UpdateSubConnState

dfawley:
A future change will delete `stub.BalancerFuncs.UpdateSubConnState`... This prepares for that and uses `StateListener` instead.

RELEASE NOTES: none


arvindbr8 (approved):
lgtm

PR #6534, opened 2023-08-09
interop/grpc_testing: regenerate protos

arvindbr8:
RELEASE NOTES: none


arvindbr8:
There is a typo in a recent commit to gRPC-proto which is breaking vet. Posted a PR to fix it https://github.com/grpc/grpc-proto/pull/130

dfawley:
Make sure you're using the right version of protoc and protoc-gen-go.  `vet.sh` has the commands you can use to download and update.

dfawley:
Your Go version won't affect the output of the proto files, so don't worry about that.

dfawley (approved):

PR #6541, opened 2023-08-11
update CI jobs to use go 1.21

fho:
- update the github CI jobs to test with go 1.19, 1.20 and 1.21

Normally I would also update the min. version in the go.mod files, but they haven't been updated for a while and are still set to 1.17. 
I moved that therefore to a separate PR: https://github.com/grpc/grpc-go/pull/6542

RELEASE NOTES: none


arvindbr8:
thanks for the PR. We updated the test here https://github.com/grpc/grpc-go/pull/6543 without noticing this PR.

PR #6542, opened 2023-08-11
go: set min. required go version to 1.19

fho:
Update all go.mod files to require as min. go version 1.19.
grpc-go officially only supports the last 3 go versions.

RELEASE NOTES: none


dfawley:
I took care of this and some other stuff with #6544, but thanks for the PR!

PR #6544, opened 2023-08-11
*: update `interface{}` to `any` and `go.mod` version to `go 1.19`

dfawley:
```
gofmt -w -r 'interface{} -> any' .
find . -name go.mod | xargs sed -i 's/go 1\.1./go 1.19/'
```

And then fix one place where we had a local variable named `any` before this change. 

RELEASE NOTES: none


balancer/balancer.go
```diff
@@ -414,9 +414,9 @@ var ErrBadResolverState = errors.New("bad resolver state")
 type ProducerBuilder interface {
 	// Build creates a Producer.  The first parameter is always a
 	// grpc.ClientConnInterface (a type to allow creating RPCs/streams on the
-	// associated SubConn), but is declared as interface{} to avoid a
-	// dependency cycle.  Should also return a close function that will be
-	// called when all references to the Producer have been given up.
+	// associated SubConn), but is declared as any to avoid a dependency cycle.
```
easwars, line 417: Should we backquote this `any`?

easwars (approved):

easwars (commented):
Realized that i only reviewed one commit. Going through the whole list now :(

easwars (approved):

balancer/balancer.go
dfawley: Done.

Does the Go stdlib use much/any markdown in its formatting? I'm not sure, but I don't think the do..  It seems like we could do it in a lot more places and it might look nicer. Or maybe all the text decoration might make things distracting?

dfawley: > And it would be nice if backticks actually took effect on the godoc. Then it will actually serve a purpose and not be distracting

Oh.. Yeah, they aren't supported unfortunately.  And even less so in interface methods or struct field definitions, which are literal unformatted code font:

image

I'll keep this since "...declared as any..." is potentially confusing.

jizhuozhi:
Why does changing `interface{}` to `any` require go1.19? The smallest version that supports `any` should be 1.18.

dfawley:
> Why does changing `interface{}` to `any` require go1.19?

It doesn't.  This is just when we decided to make the switch.

PR #6556, opened 2023-08-15
test: improve and speed up channelz keepalive test

dfawley:
RELEASE NOTES: none


arvindbr8 (approved):
Noice! 
seems like this is should shave off ~3.5 sec. Maybe we dont have to end up splitting tests from `test` package.

PR #6557, opened 2023-08-16
client: fix ClientStream.Header() behavior

### LINKED ISSUE #6524: docs: Do nil headers from stream.Header() indicate an error?
I have an `rpc Test(TestRequest) returns (stream TestMessage)` with an implementation that just returns an error immediately.

Then the client does this:

```go
	stream, err := testClient.Test(ctx, &TestRequest{})
	if err != nil {...}    // don't expect an error here
	header, err := stream.Header()
        // but this waited for headers... the error headers were received... where did they go?
	if err != nil {...}  
```

What is flummoxing me is that most of the time, but not always, `stream.Header()` returns `nil, nil`. Occasionally, running in a debugger, it returns the error. A calls to `stream.Recv()` returns the error reliably, but IDK if there is an error or not, and could end up blocking indefinitely waiting for it, which isn’t good because I have an upstream waiting to know the stream was setup successfully.

What I noticed is that if there was an error, I get `nil` headers instead of `content-type: ...`. Can I reliably use this as a signal to invoke `Recv()` and get the underlying error? 

Why can't `Header()` return the error? Inspection in the debugger shows that it reads the error code and message headers out of the frame, but short-circuits because the stream has terminated and internal assigns `noHeaders`. I just want access to the state that is buried in there (there was an error) without having to resort to trying to recv with a timeout.

dfawley:
This fixes a regression introduced in #5763.  Fixes #6524

This change causes us to reliably report the stream status from `ClientStream.Header()`, including `io.EOF` in case the stream ended successfully with a trailers-only response.  To do this required moving the binary logging of trailers into `clientStream.finish()`.

By doing this, retry is also fixed in cases where an error is encountered on the stream before calling `Header()`.  Previously, if `Header()` was called first, we would report `ErrNoHeaders` from the transport which would be converted into a non-error by `clientStream.Header()`, which would then commit the RPC and prevent retries.  A test case was added to cover this scenario.

RELEASE NOTES:
* client: fix a bug that prevented `ClientStream.Header` from returning an error for trailers-only RPC responses, and as a side-effect, prevented retry of the RPC.

### DIFF
--- a/binarylog/binarylog_end2end_test.go
+++ b/binarylog/binarylog_end2end_test.go
@@ -31,10 +31,12 @@ import (
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/binarylog"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/grpclog"
 	iblog "google.golang.org/grpc/internal/binarylog"
 	"google.golang.org/grpc/internal/grpctest"
+	"google.golang.org/grpc/internal/stubserver"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
@@ -1057,5 +1059,41 @@ func (s) TestServerBinaryLogFullDuplexError(t *testing.T) {
 	count := 5
 	if err := testServerBinaryLog(t, &rpcConfig{count: count, success: false, callType: fullDuplexStreamRPC}); err != nil {
 		t.Fatal(err)
+	}
+}
+
+// TestCanceledStatus ensures a server that responds with a Canceled status has
+// its trailers logged appropriately and is not treated as a canceled RPC.
+func (s) TestCanceledStatus(t *testing.T) {
+	defer testSink.clear()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	const statusMsgWant = "server returned Canceled"
+	ss := &stubserver.StubServer{
+		UnaryCallF: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+			grpc.SetTrailer(ctx, metadata.Pairs("key", "value"))
+			return nil, status.Error(codes.Canceled, statusMsgWant)
+		},
+	}
+	if err := ss.Start(nil); err != nil {
+		t.Fatalf("Error starting endpoint server: %v", err)
+	}
+	defer ss.Stop()
+
+	if _, err := ss.Client.UnaryCall(ctx, &testpb.SimpleRequest{}); status.Code(err) != codes.Canceled {
+		t.Fatalf("Received unexpected error from UnaryCall: %v; want Canceled", err)
+	}
+
+	got := testSink.logEntries(true)
+	last := got[len(got)-1]
+	if last.Type != binlogpb.GrpcLogEntry_EVENT_TYPE_SERVER_TRAILER ||
+		last.GetTrailer().GetStatusCode() != uint32(codes.Canceled) ||
+		last.GetTrailer().GetStatusMessage() != statusMsgWant ||
+		len(last.GetTrailer().GetMetadata().GetEntry()) != 1 ||
+		last.GetTrailer().GetMetadata().GetEntry()[0].GetKey() != "key" ||
+		string(last.GetTrailer().GetMetadata().GetEntry()[0].GetValue()) != "value" {
+		t.Fatalf("Got binary log: %+v; want last entry is server trailing with status Canceled", got)
 	}
 }
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -1507,12 +1507,14 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 
 	isHeader := false
 
-	// If headerChan hasn't been closed yet
-	if atomic.CompareAndSwapUint32(&s.headerChanClosed, 0, 1) {
-		s.headerValid = true
-		if !endStream {
-			// HEADERS frame block carries a Response-Headers.
-			isHeader = true
+	if !endStream {
+		// HEADERS frame block carries a Response-Headers.
+		isHeader = true
+		// If headerChan hasn't been closed yet (expected, given we checked it
+		// above, but something else could have potentially closed the whole
+		// stream).
+		if atomic.CompareAndSwapUint32(&s.headerChanClosed, 0, 1) {
+			s.headerValid = true
 			// These values can be set without any synchronization because
 			// stream goroutine will read it only after seeing a closed
 			// headerChan which we'll close after setting this.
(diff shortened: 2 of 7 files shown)


internal/transport/http2_client.go
```diff
@@ -1505,30 +1505,25 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
 
-	isHeader := false
-
-	// If headerChan hasn't been closed yet
-	if atomic.CompareAndSwapUint32(&s.headerChanClosed, 0, 1) {
-		s.headerValid = true
-		if !endStream {
-			// HEADERS frame block carries a Response-Headers.
-			isHeader = true
+	if !endStream {
```
zasweq, line 1508: Optional (use some/all/none): mention this is headers flow, and that for trailers flow, header chan gets closed in closeStream() which writes status before unblocking the header chan, letting client stream read the status after it's block on header chan (in csAttempt).

internal/transport/http2_client.go
```diff
@@ -1505,30 +1505,25 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		return
 	}
 
-	isHeader := false
-
-	// If headerChan hasn't been closed yet
-	if atomic.CompareAndSwapUint32(&s.headerChanClosed, 0, 1) {
-		s.headerValid = true
-		if !endStream {
-			// HEADERS frame block carries a Response-Headers.
```
zasweq, line 1514: Optional: keep this comment?

stream.go
```diff
@@ -1002,18 +985,33 @@ func (cs *clientStream) finish(err error) {
...
 		}
 	}
+
 	cs.mu.Unlock()
 	// For binary logging. only log cancel in finish (could be caused by RPC ctx
-	// canceled or ClientConn closed). Trailer will be logged in RecvMsg.
+	// canceled or ClientConn closed).
 	//
 	// Only one of cancel or trailer needs to be logged. In the cases where
 	// users don't call RecvMsg, users must have already canceled the RPC.
```
zasweq, line 994: Optional: delete this comment, no longer applies since we don't log both events in different places: "In the cases where users don't call RecvMsg, users must have already canceled the RPC."

stream.go
```diff
@@ -1002,18 +985,33 @@ func (cs *clientStream) finish(err error) {
 			}
 		}
 	}
+
 	cs.mu.Unlock()
 	// For binary logging. only log cancel in finish (could be caused by RPC ctx
```
zasweq, line 990: Now it logs cancel and server trailers :).

test/end2end_test.go
```diff
@@ -6339,12 +6339,11 @@ func (s) TestGlobalBinaryLoggingOptions(t *testing.T) {
...
-			for {
-				_, err := stream.Recv()
-				if err == io.EOF {
-					return nil
-				}
+			_, err := stream.Recv()
+			if err == io.EOF {
+				return nil
 			}
+			return status.Errorf(codes.Unknown, "expected client to CloseSend")
```
zasweq, line 6346: Optional: s/CloseSend/call CloseSend().

zasweq (approved):
LGTM.

stream.go
dfawley: Hmm, I think the logic here is a bit wrong, since the server could return a `Canceled` status with trailers which is distinct from the client initiating cancelation.  I'll have to work on this and add a test since I think it worked properly before.

test/end2end_test.go
dfawley: done

stream.go
dfawley: Fixed

internal/transport/http2_client.go
dfawley: Added a comment

dfawley (commented):
PTAL at the recent commit.  Thanks!

stream.go
```diff
@@ -1002,18 +985,31 @@ func (cs *clientStream) finish(err error) {
...
-		c := &binarylog.Cancel{
-			OnClientSide: true,
-		}
-		for _, binlog := range cs.binlogs {
-			binlog.Log(cs.ctx, c)
+	// Only one of cancel or trailer needs to be logged.
+	if len(cs.binlogs) != 0 {
+		if err == errContextCanceled ||
+			err == errContextDeadline ||
+			err == ErrClientConnClosing {
```
zasweq, line 994: Optional style nit: I've never seen conditionals on separate lines like this in the codebase. Make one line

binarylog/binarylog_end2end_test.go
```diff
@@ -1059,3 +1061,39 @@ func (s) TestServerBinaryLogFullDuplexError(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestCanceledStatus ensures a server that responds with a Canceled status has
+// its trailers logged appropriately and is not treated as a canceled RPC.
+func (s) TestCanceledStatus(t *testing.T) {
```
zasweq, line 1067: This tests not hitting one of the three conditionals that cause cancel to be logged in finish(), and cause trailers to be logged. Are the other three conditionals that trigger a cancel to be logged tested already?

binarylog/binarylog_end2end_test.go
```diff
@@ -1059,3 +1061,39 @@ func (s) TestServerBinaryLogFullDuplexError(t *testing.T) {
...
+	}
+
+	got := testSink.logEntries(true)
+	last := got[len(got)-1]
+	if last.Type != binlogpb.GrpcLogEntry_EVENT_TYPE_SERVER_TRAILER ||
+		last.GetTrailer().GetStatusCode() != uint32(codes.Canceled) ||
+		last.GetTrailer().GetStatusMessage() != statusMsgWant ||
+		len(last.GetTrailer().GetMetadata().GetEntry()) != 1 ||
+		last.GetTrailer().GetMetadata().GetEntry()[0].GetKey() != "key" ||
+		string(last.GetTrailer().GetMetadata().GetEntry()[0].GetValue()) != "value" {
```
zasweq, line 1096: Just a thought, could you use the eventual proto.Equal call in this helper as other parts of the test: https://github.com/grpc/grpc-go/blob/11a7cd2e2ea21627f06b91869805fd18716debc0/binarylog/binarylog_end2end_test.go#L843?

binarylog/binarylog_end2end_test.go
```diff
@@ -1059,3 +1061,39 @@ func (s) TestServerBinaryLogFullDuplexError(t *testing.T) {
...
+		t.Fatalf("Got binary log: %+v; want last entry is server trailing with status Canceled", got)
```
zasweq, line 1097: Can you define a want entry instead and log that here "want: %+v"? As other parts of the file (which get the want out of https://github.com/grpc/grpc-go/blob/11a7cd2e2ea21627f06b91869805fd18716debc0/binarylog/binarylog_end2end_test.go#L873 expect.To(Client||Server)LogEntries (could just define you want for Server Trailers).

zasweq (changes requested):
Some minor nits. I approved previously I think.

binarylog/binarylog_end2end_test.go
dfawley: There is an existing cancel test here: https://github.com/grpc/grpc-go/blob/7f66074c371e4718407f94251720f8e12f211cb7/binarylog/binarylog_end2end_test.go#L961

binarylog/binarylog_end2end_test.go
dfawley: This isn't great but seems fine.  I don't really want to have to clear out a bunch of fields in the received entry to compare against a `want`, and if we don't use a `want` for the comparison it doesn't make sense to log it in the error.  This is a pretty simple test, and there should be enough here to debug if an error ever occurs.

stream.go
dfawley: Changed

binarylog/binarylog_end2end_test.go
zasweq: Sounds good.

zasweq (approved):
LGTM. Looks like vet proto is failing, but seems unrelated to this change.

PR #6586, opened 2023-08-25
stream: swallow Header errors as we used to; RecvMsg can still return it

### LINKED ISSUE #6524: docs: Do nil headers from stream.Header() indicate an error?

dfawley:
Fixes #6524 

This reverts the behavior change part of #6557 but allows retry to keep working, because the transport still returns the error through `withRetry()` via the attempt stream's `Header()` method.

This will need to be backported to 1.58.x.

RELEASE NOTES: none

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -802,7 +802,8 @@ func (cs *clientStream) Header() (metadata.MD, error) {
 
 	if err != nil {
 		cs.finish(err)
-		return nil, err
+		// Do not return the error.  The user should get it by calling Recv().
+		return nil, nil
 	}
 
 	if len(cs.binlogs) != 0 && !cs.serverHeaderBinlogged && m != nil {
--- a/stream.go
+++ b/stream.go
@@ -91,7 +91,9 @@ type Stream interface {
 // status package.
 type ClientStream interface {
 	// Header returns the header metadata received from the server if there
-	// is any. It blocks if the metadata is not ready to read.
+	// is any. It blocks if the metadata is not ready to read.  If the metadata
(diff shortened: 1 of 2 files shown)


arvindbr8 (approved):

arvindbr8 (approved):

arvindbr8:
LGTM, I guess proto needs another regenerate?

arvindbr8:
@dfawley do we need a release note on this one?

PR #6602, opened 2023-08-31
balancer/leastrequest: Cache atomic load and also add concurrent rpc test

zasweq:
This PR caches the atomic load of the picked SubConn in the picker algorithm in least request picker for efficiency reasons, rather than doing it per iteration. It also adds an e2e test which performs concurrent RPCs to check for a race condition.

Continuation of #6587. Needs to be backported to 1.58 alongside that bug fix PR.

RELEASE NOTES: N/A

### DIFF
--- a/balancer/leastrequest/balancer_test.go
+++ b/balancer/leastrequest/balancer_test.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


balancer/leastrequest/leastrequest.go
```diff
@@ -155,15 +155,18 @@ type picker struct {
...
 	for i := 0; i < int(p.choiceCount); i++ {
 		index := grpcranduint32() % uint32(len(p.subConns))
 		sc := p.subConns[index]
 		if pickedSC == nil {
 			pickedSC = &sc
+			pickedSCNumRPCs = pickedSC.numRPCs.Load()
 			continue
 		}
-		if sc.numRPCs.Load() < pickedSC.numRPCs.Load() {
+		if sc.numRPCs.Load() < pickedSCNumRPCs {
```
huangchong94, line 167: Should `sc.numRPCS.Load()` be cached too? to save one more `Load`

zasweq: Oh good point. Switched.

bitcoinfinancier (approved):

balancer/leastrequest/leastrequest.go
```diff
@@ -155,15 +155,19 @@ type picker struct {
...
 		index := grpcranduint32() % uint32(len(p.subConns))
 		sc := p.subConns[index]
 		if pickedSC == nil {
 			pickedSC = &sc
+			pickedSCNumRPCs = pickedSC.numRPCs.Load()
 			continue
 		}
-		if sc.numRPCs.Load() < pickedSC.numRPCs.Load() {
+		scNumRPCs := sc.numRPCs.Load()
+		if scNumRPCs < pickedSCNumRPCs {
```
dfawley, line 168: Optional: these can be merged into a compound if if desired.  Or even simpler:

```go
sc := p.subConns[index]
n := sc.numRPCs.Load()
if pickedSC == nil || n < pickedSCNumRPCs {
	pickedSC = &sc
	pickedSCNumRPCs = n
}
```

balancer/leastrequest/balancer_test.go
```diff
@@ -455,3 +456,55 @@ func (s) TestLeastRequestPersistsCounts(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	testServiceClient := testgrpc.NewTestServiceClient(cc)
+
+	var wg sync.WaitGroup
+	for i := 0; i < 100; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			testServiceClient.EmptyCall(ctx, &testpb.Empty{})
```
dfawley, line 505: Should these do several calls each, maybe, to make concurrent pick calls more likely?  Just make sure it doesn't run too long.

dfawley (approved):

balancer/leastrequest/balancer_test.go
zasweq: Good point. Made each goroutine perform 5 calls.

balancer/leastrequest/leastrequest.go
zasweq: Ah, beautiful suggestion :D. I tried breaking it but it works and is much cleaner. Thanks.

Issue #2817, opened 2019-05-15
stream: reuse codec input buffers OR avoid new allocations with Unmarshalv


dfawley:

ginayeh:
Bring tags over to #6619 and close this down.

Issue #2816, opened 2019-05-15
stream: allow codecs to reuse output buffers


dfawley:
This would require a codec API change/extension to recycle the memory once grpc is done writing it to the wire (or compressing it).

dfawley:
The PR that implemented this ultimately needed to be rolled back; this should have been reopened at that time.

menghanl:
There's another attempt to further reuse the buffer, but the team didn't have enough time to make more progress to review the PR (https://github.com/grpc/grpc-go/pull/3220#issuecomment-608655702). Future work can reuse some of the change from it.

PR #6608, opened 2023-09-06
Reuse server send buffers

PapaCharlie:
Under significant load, the server concurrently allocates a significant amount of buffers to serialize and compress the responses before sending them over the wire. This causes significant GC pressure and can cause large spikes in memory allocations.

Tests are benchmark results are pending.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -31,8 +31,12 @@ type baseCodec interface {
 	Unmarshal(data []byte, v any) error
 }
 
+type appendCodec interface {
+	MarshalAppend(buf []byte, v interface{}) ([]byte, error)
+}
+
 var _ baseCodec = Codec(nil)
-var _ baseCodec = encoding.Codec(nil)
+var _ baseCodec = encoding.AppendCodec(nil)
 
 // Codec defines the interface gRPC uses to encode and decode messages.
 // Note that implementations of this interface must be thread safe;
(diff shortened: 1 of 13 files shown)


arvindbr8:
@PapaCharlie, seems like your github account has an issue with CLA. Please fix using this [Help Article](https://confluence.linuxfoundation.org/pages/viewpage.action?pageId=86641302) as mentioned in the comment above.

arvindbr8:
@PapaCharlie Also please consider filing an issue with us before sending the PR. Changes like this require more investigation and proof - which needs to happen in an issue. 
Please open an issue with us [here](https://github.com/grpc/grpc-go/issues/new?assignees=&labels=Type%3A+Bug&projects=&template=bug.md). 
I'm closing this for now and we can reopen this once we have a solid idea on what the issue is and set in stone the fix (if required).

PapaCharlie:
Hey @arvindbr8, I didn't get a chance to link these issues when I opened the PR but here you go:
https://github.com/grpc/grpc-go/issues/2817 and https://github.com/grpc/grpc-go/issues/2816
This is also in response to pretty significant performance testing. I took some CPU and heap profiles for the same server code running on v1.58.0 and my fork under significant load (of the nature where the server is concurrently sending very large responses to many clients). Here are the results:

# v1.58.0 CPU profile
old_cpu
# v1.58.0 Heap profile (alloc_space)
old_heap

# Fork CPU profile
new_cpu
# Fork Heap profile (alloc_space)
new_heap

If you want me to repro this using simpler code (and/or a benchmark) I can give it a shot, it shouldn't be difficult. I just had these profile results handy. As you can see, the allocation overhead of not reusing the send buffers means the server is spending a lot its time collecting them instead of doing meaningful work. By reusing the send buffers, the performance drastically increases

PR #6609, opened 2023-09-07
xds: fix hash policy header to skip bin headers and use extra metadata

dfawley:
A42 update PR: https://github.com/grpc/proposal/pull/392

Also a minor performance optimization to not read metadata from context more than once if not required, as this operation allocates a map (https://github.com/grpc/grpc-go/blob/c2b0797a5353b29ca02869629c83151c980a6e31/metadata/metadata.go#L271).

RELEASE NOTES:
* xds: fix hash policy header to skip "-bin" headers and read content-type header as expected

### DIFF
--- a/xds/internal/resolver/serviceconfig.go
+++ b/xds/internal/resolver/serviceconfig.go
@@ -247,7 +247,8 @@ func (cs *configSelector) generateHash(rpcInfo iresolver.RPCInfo, hashPolicies [
 			}
 			values := emd.Get(policy.HeaderName)
 			if len(values) == 0 {
-				// Extra metadata takes precedence over the user's metadata.
+				// Extra metadata (e.g. the "content-type" header) takes
+				// precedence over the user's metadata.
 				values = md.Get(policy.HeaderName)
 				if len(values) == 0 {
 					// If the header isn't present at all, this policy is a no-op.
--- a/xds/internal/resolver/serviceconfig.go
+++ b/xds/internal/resolver/serviceconfig.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/internal/envconfig"
 	"google.golang.org/grpc/internal/grpcrand"
+	"google.golang.org/grpc/internal/grpcutil"
 	iresolver "google.golang.org/grpc/internal/resolver"
 	"google.golang.org/grpc/internal/serviceconfig"
 	"google.golang.org/grpc/internal/wrr"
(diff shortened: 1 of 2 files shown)


xds/internal/resolver/serviceconfig.go
```diff
@@ -229,19 +230,29 @@ func retryConfigToPolicy(config *xdsresource.RetryConfig) *serviceconfig.RetryPo
...
+			if !mdRead {
+				md, _ = metadata.FromOutgoingContext(rpcInfo.Context)
+				emd, _ = grpcutil.ExtraMetadata(rpcInfo.Context)
+				mdRead = true
+			}
+			values := emd.Get(policy.HeaderName)
 			if len(values) == 0 {
-				continue
+				// Extra metadata takes precedence over the user's metadata.
+				values = md.Get(policy.HeaderName)
```
zasweq, line 251: So at this point in the RPC lifecycle, the only overwriting extra metadata is "content-type"? "However, we will support matching against the content-type header; if a gRPC implementation does not add that header until after xDS route selection is done, then xDS route selection will assume a hard-coded value of application/grpc."

xds/internal/resolver/serviceconfig_test.go
```diff
@@ -106,6 +107,35 @@ func (s) TestGenerateRequestHash(t *testing.T) {
...
+				HeaderName:     "something-bin",
+			}, {
+				HashPolicyType: xdsresource.HashPolicyTypeChannelID,
+			}},
+			requestHashWant: channelID,
+			rpcInfo: iresolver.RPCInfo{
+				Context: metadata.NewOutgoingContext(context.Background(), metadata.Pairs("something-bin", "xyz")),
+			},
+		},
+		// Tests that extra metadata is used first.
```
zasweq, line 124: Nit: "used first" I don't think is semantically correct. I think a better phrase would be "takes precedence over a users metadata".

xds/internal/resolver/serviceconfig.go
```diff
@@ -229,19 +230,29 @@ func retryConfigToPolicy(config *xdsresource.RetryConfig) *serviceconfig.RetryPo
...
-			// If the header isn't present, no-op.
+			if !mdRead {
+				md, _ = metadata.FromOutgoingContext(rpcInfo.Context)
+				emd, _ = grpcutil.ExtraMetadata(rpcInfo.Context)
+				mdRead = true
+			}
+			values := emd.Get(policy.HeaderName)
 			if len(values) == 0 {
-				continue
+				// Extra metadata takes precedence over the user's metadata.
```
zasweq, line 250: Optional: Can you add a comment from the gRFC about content type? Or explain content type?

zasweq (approved):
LGTM.

xds/internal/resolver/serviceconfig.go
dfawley: Yes, the "extra metadata" is "content-type"... I'm not sure I like how this all was done (originally for route matching), but it works at least.

xds/internal/resolver/serviceconfig_test.go
dfawley: Changed

xds/internal/resolver/serviceconfig.go
dfawley: Mentioned content-type header

zasweq:
Thanks, feel free to merge.

PR #6624, opened 2023-09-12
grpc: Move some stats handler callouts from transport to gRPC layer server side

zasweq:
This PR moves the Conn Begin and InHeader calls from the transport layer to gRPC layer. The other three (OutHeader, OutTrailer, ConnEnd) will come in a future PR which will delete the need for the setContext() function. The OutHeader and OutTrailer will require some reworking of the ownership of logic between the layers of gRPC and Transport.

RELEASE NOTES: N/A

### DIFF
--- a/internal/transport/handler_server.go
+++ b/internal/transport/handler_server.go
@@ -167,6 +167,8 @@ func (ht *serverHandlerTransport) Close(err error) {
 
 func (ht *serverHandlerTransport) RemoteAddr() net.Addr { return strAddr(ht.req.RemoteAddr) }
 
+func (ht *serverHandlerTransport) LocalAddr() net.Addr { return nil }
+
 // strAddr is a net.Addr backed by either a TCP "ip:port" string, or
 // the empty string if unknown.
 type strAddr string
--- a/internal/transport/transport.go
+++ b/internal/transport/transport.go
@@ -287,8 +287,9 @@ type Stream struct {
 	// On server-side it is unused.
 	status *status.Status
 
-	bytesReceived uint32 // indicates whether any bytes have been received on this stream
-	unprocessed   uint32 // set if the server sends a refused stream or GOAWAY including this stream
+	bytesReceived    uint32 // indicates whether any bytes have been received on this stream
+	unprocessed      uint32 // set if the server sends a refused stream or GOAWAY including this stream
+	headerWireLength int
 
 	// contentSubtype is the content-subtype for requests.
 	// this must be lowercase or the behavior is undefined.
@@ -425,6 +426,12 @@ func (s *Stream) Context() context.Context {
 	return s.ctx
 }
 
+// SetContext sets the context of the stream. This will be deleted once the
+// stats handler callouts all move to gRPC layer.
+func (s *Stream) SetContext(ctx context.Context) {
+	s.ctx = ctx
+}
+
 // Method returns the method for the stream.
 func (s *Stream) Method() string {
 	return s.method
@@ -435,6 +442,12 @@ func (s *Stream) Method() string {
 // that is, after Done() is closed.
 func (s *Stream) Status() *status.Status {
 	return s.status
+}
+
+// HeaderWireLength returns the size of theheaders of the stream as received
+// from the wire.
+func (s *Stream) HeaderWireLength() int {
+	return s.headerWireLength
 }
 
 // SetHeader sets the header metadata. This can be called multiple times.
@@ -719,6 +732,9 @@ type ServerTransport interface {
 
 	// RemoteAddr returns the remote network address.
 	RemoteAddr() net.Addr
+
+	// LocalAddr returns the local network address.
+	LocalAddr() net.Addr
 
 	// Drain notifies the client this ServerTransport stops accepting new RPCs.
 	Drain(debugData string)
--- a/server.go
+++ b/server.go
@@ -912,9 +912,17 @@ func (s *Server) handleRawConn(lisAddr string, rawConn net.Conn) {
 		return
 	}
 	rawConn.SetDeadline(time.Now().Add(s.opts.connectionTimeout))
+	ctx := context.Background()
+	for _, sh := range s.opts.statsHandlers {
+		ctx = sh.TagConn(ctx, &stats.ConnTagInfo{
+			RemoteAddr: rawConn.RemoteAddr(),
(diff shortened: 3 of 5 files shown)


server.go
```diff
@@ -1707,6 +1715,20 @@ func (s *Server) handleStream(t transport.ServerTransport, stream *transport.Str
 		}
 	}
 
+	md, _ := metadata.FromIncomingContext(ctx)
+	for _, sh := range s.opts.statsHandlers {
+		ctx = sh.TagRPC(ctx, &stats.RPCTagInfo{FullMethodName: stream.Method()})
+		stream.SetContext(ctx) // To have calls in stream callouts work. Will delete once all stats handler calls come from the gRPC layer.
```
dfawley, line 1721: You can move this to after the loop so it only happens once.

internal/transport/handler_server.go
```diff
@@ -167,6 +167,8 @@ func (ht *serverHandlerTransport) Close(err error) {
 
 func (ht *serverHandlerTransport) RemoteAddr() net.Addr { return strAddr(ht.req.RemoteAddr) }
 
+func (ht *serverHandlerTransport) LocalAddr() net.Addr { return nil }
```
dfawley, line 170: Can you fix this?

internal/transport/transport.go
```diff
@@ -287,8 +287,9 @@ type Stream struct {
 	// On server-side it is unused.
 	status *status.Status
 
-	bytesReceived uint32 // indicates whether any bytes have been received on this stream
-	unprocessed   uint32 // set if the server sends a refused stream or GOAWAY including this stream
+	bytesReceived    uint32 // indicates whether any bytes have been received on this stream
+	unprocessed      uint32 // set if the server sends a refused stream or GOAWAY including this stream
+	headerWireLength int
```
dfawley, line 292: This should be grouped with other header stuff probably. And it needs a comment that it's only set by the server.

internal/transport/transport.go
```diff
@@ -437,6 +444,12 @@ func (s *Stream) Status() *status.Status {
 	return s.status
 }
 
+// HeaderWireLength returns the size of theheaders of the stream as received
```
dfawley, line 447: Addaspace pls

internal/transport/transport.go
```diff
@@ -437,6 +444,12 @@ func (s *Stream) Status() *status.Status {
...
+// from the wire.
```
dfawley, line 448: .... "Valid only on the server"?

server.go
```diff
@@ -912,9 +912,17 @@ func (s *Server) handleRawConn(lisAddr string, rawConn net.Conn) {
 		return
 	}
 	rawConn.SetDeadline(time.Now().Add(s.opts.connectionTimeout))
+	ctx := context.Background()
+	for _, sh := range s.opts.statsHandlers {
+		ctx = sh.TagConn(ctx, &stats.ConnTagInfo{
+			RemoteAddr: rawConn.RemoteAddr(),
+			LocalAddr:  rawConn.LocalAddr(),
+		})
+		sh.HandleConn(ctx, &stats.ConnBegin{})
```
dfawley, line 921: This is a change in behavior, but it's subtle.

If the handshaker replaces the `conn` with something connected to another remote/local address, then that's not reflected here.

More importantly it doesn't match what we do later for remote/local addr, which is use the server transports addresses.  We probably want to move this after `NewServerTransport` and use `st.LocalAddr()` and `st.RemoteAddr()` instead.

(Also this change means we call the stats handler for every connection even if it doesn't finish handshaking correctly, which is probably also not what we want.)

dfawley (commented):
The  handler_server also calls `TagRPC`...so now it will happen 2x.  It seems we might need a test for the handler server + stats handlers to make sure the events are being produced properly.

server.go
zasweq: Hmmmm ok I was wondering whether I should do this after handshaking, I noticed it changed to log unconditionally. I remember you mentioned either passing conn around or just using the wrapper of conn to get addr info, and we chose latter.

zasweq: https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_client.go#L372 moving it after transport will slightly change behavior as well, as previously TagConn was still called if there's a connection error in say writing intial http/2 preface/settings frame. For me, this is dependent on semantically what does TagConn represent? Accepting a TCP connection? Getting an HTTP/2 connection to a point where it finishes the initial frames?

internal/transport/transport.go
zasweq: Done.

internal/transport/transport.go
zasweq: Whoops, done :).

internal/transport/transport.go
zasweq: Sure. Added. Does it not know/need to know the client side eventually?

server.go
zasweq: Good point. Done.

internal/transport/transport.go
dfawley: Maybe, but at that time we can fix the comment.

Honestly one of the next cleanups I'd love to do is to split the `transport.Stream` type into "client" and "server" versions instead of sharing code awkwardly where most of it needs to check a boolean to figure out whether it's a client or server stream and do radically different behavior depending on which.

server.go
dfawley: As discussed offline, waiting for handshaking before calling `TagConn` seems fine.

internal/transport/handler_server.go
zasweq: This server handler transport has no access to local addr unfortunately. It was actually calling into stats handler with a nil local addr set in InHeader since it has no access to this. I read LocalAddr() (method I added to server transport) in my new PR, so this will keep the same functionality as in master and not log LocalAddr().

internal/transport/transport.go
zasweq: Hmmm ok sounds good.

server.go
zasweq: Did this after newHTTP2Transport in new PR (still comes before first stats handler call after TagConn, so as we discussed offline, this is fine).

zasweq:
Continued in #6716. Closing this PR, but I did get to all the comments here.

PR #6648, opened 2023-09-19
examples: add an example of flow control behavior

dfawley:
RELEASE NOTES:
* examples: add an example of flow control behavior


arvindbr8 (approved):
LGTM. 

(expect that im not convinced that a period should be followed by 2 whitespaces :P)

PR #6649, opened 2023-09-20
examples/features/loadbalancing: Add custom lb example

zasweq:
This PR adds a custom lb example, with a custom lb deployed as the top level balancer of the channel through service config. This scales up a previous example

RELEASE NOTES:
* examples/features/loadbalancing: Add custom lb example

### DIFF
--- a/examples/features/load_balancing/README.md
+++ b/examples/features/load_balancing/README.md
@@ -81,3 +81,34 @@ Note that it's possible to see two continues RPC sent to the same backend.
 That's because `round_robin` only picks the connections ready for RPCs. So if
 one of the two connections is not ready for some reason, all RPCs will be sent
 to the ready connection.
+
+# custom_round_robin
+
+```
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50051)
+this is examples/load_balancing (from :50052)
+this is examples/load_balancing (from :50051)
+```
+
+It is also possible to configure a custom load balancing policy through the
+service config. Specify the balancer type through the service config which you
+register in the registry. In this example, a custom balancer which routes to the
+first SubConn it receives except every n times, where n is configurable, in
+which is chooses the second SubConn it receives.
--- a/examples/features/load_balancing/client/main.go
+++ b/examples/features/load_balancing/client/main.go
@@ -21,14 +21,19 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
+	"sync/atomic"
 	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/balancer"
+	"google.golang.org/grpc/balancer/base"
 	"google.golang.org/grpc/credentials/insecure"
 	ecpb "google.golang.org/grpc/examples/features/proto/echo"
 	"google.golang.org/grpc/resolver"
+	"google.golang.org/grpc/serviceconfig"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


examples/features/load_balancing/client/main.go
```diff
@@ -123,4 +141,87 @@ func (*exampleResolver) Close()                                  {}
...
+type customRoundRobinBuilder struct{}
+
+func (customRoundRobinBuilder) Name() string {
+	return customRRName
+}
+
+func (customRoundRobinBuilder) Build(cc balancer.ClientConn, bOpts balancer.BuildOptions) balancer.Balancer {
+	crr := &customRoundRobin{}
+	baseBuilder := base.NewBalancerBuilder(customRRName, crr, base.Config{HealthCheck: true})
+	crr.Balancer = baseBuilder.Build(cc, bOpts)
```
dfawley, line 176: I'm really not a fan of the base balancer...are we sure we want users using it as a template for how to do their own LB policy?  I think it would be better to show the actual connection management APIs here instead.

examples/features/load_balancing/client/main.go
```diff
@@ -84,11 +89,24 @@ func main() {
...
+	}
+	defer customroundrobinConn.Close()
+	fmt.Println("--- calling helloworld.Greeter/SayHello with custom_round_robin ---")
+	makeRPCs(customroundrobinConn, 20)
 }
 
 // Following is an example name resolver implementation. Read the name
 // resolution example to learn more about it.
-
 type exampleResolverBuilder struct{}
```
dfawley, line 110: I don't believe this example intends to show off custom resolvers - can this be replaced with the manual resolver instead?  We may not have had it when this was created.

examples/features/load_balancing/client/main.go
```diff
@@ -123,4 +141,87 @@ func (*exampleResolver) Close()                                  {}
 
 func init() {
 	resolver.Register(&exampleResolverBuilder{})
+	balancer.Register(&customRoundRobinBuilder{})
```
dfawley, line 144: Maybe put the custom LB policy into its own file (or even its own sub-package), since that's how we expect it to be used typically?

examples/features/load_balancing/client/main.go
```diff
@@ -123,4 +141,87 @@ func (*exampleResolver) Close()                                  {}
...
+	crrCfg, ok := state.BalancerConfig.(*customRRConfig)
+	if !ok {
+		return balancer.ErrBadResolverState
+	}
+	crr.n = crrCfg.N
+	return crr.Balancer.UpdateClientConnState(state)
+}
+
+func (crr *customRoundRobin) Build(info base.PickerBuildInfo) balancer.Picker {
+	if len(info.ReadySCs) == 0 {
```
dfawley, line 197: If you want to make this deterministic, then also do this until `len(info.ReadySCs) == 2`?

examples/features/load_balancing/client/main.go
```diff
@@ -84,11 +89,24 @@ func main() {
 
 	fmt.Println("--- calling helloworld.Greeter/SayHello with round_robin ---")
 	makeRPCs(roundrobinConn, 10)
+	// You can also plug in your own custom lb policy, which needs to be
+	// configurable. This n is configurable. Try changing it and see how the
+	// behavior changes.
+	customroundrobinConn, err := grpc.Dial(
+		fmt.Sprintf("%s:///%s", exampleScheme, exampleServiceName),
```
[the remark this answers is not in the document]
zasweq: Done. Good point.

examples/features/load_balancing/client/main.go
```diff
@@ -123,4 +141,87 @@ func (*exampleResolver) Close()                                  {}
...
+	balancer.Register(&customRoundRobinBuilder{})
+}
+
+const customRRName = "custom_round_robin"
+
+type customRRConfig struct {
+	serviceconfig.LoadBalancingConfig `json:"-"`
+
+	// N represents how often pick iterations chose the second SubConn
+	// in the list. Defaults to 3.
```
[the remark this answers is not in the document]
zasweq: Ah good catch. Switched the default in ParseConfig() to 3.

examples/features/load_balancing/README.md
```diff
@@ -81,3 +81,34 @@ Note that it's possible to see two continues RPC sent to the same backend.
 That's because `round_robin` only picks the connections ready for RPCs. So if
 one of the two connections is not ready for some reason, all RPCs will be sent
 to the ready connection.
+
+# custom_round_robin
+
+```
```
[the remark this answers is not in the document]
zasweq: Done.

examples/features/load_balancing/client/main.go
zasweq: Switched.

examples/features/load_balancing/client/main.go
zasweq: Switched to petiole policy wrapping pick first after offline discussion.

examples/features/load_balancing/client/main.go
zasweq: Done.

examples/features/load_balancing/client/main.go
zasweq: Made deterministic in new PR based off child balancer connectivity states.

zasweq:
Opened a new PR #6691 with all the offline discussion. Made a whole new example because this example is outdated and shows concepts like custom resolvers, and new example needs to use endpoints.

Issue #1805, opened 2018-01-17
Ability to use a single interceptor for both unary and streaming RPCs


jhump:
After #1801 is completed, it becomes possible for a stream interceptor to be used when processing unary RPCs.

This issue is a feature request for some new API that allows for exactly this. The existing API, for backwards compatibility, will continue to register a stream interceptor that is only used for streaming RPCs. But a new method/dial option could be added that allows for registering a stream interceptor that is used for unary RPCs and/or for both streaming and unary RPCs.

dfawley:
I'd like to redesign our interceptor API.  Please add your list of requirements/nice-to-haves in this bug and I'll compile them here:

Requirements:
- Single interceptor type for unary and streaming
  - If it's difficult to implement, provide an optional convenience wrapper to generate a stream-style interceptor from a single function (like how unary interceptors work today).
- Include access to `ServiceDesc`/`FileDescriptorProto` (#1526)
- Chainable / able to easily register multiple handlers

Nice to Haves:
- Single API for stats handlers and interceptors
- Global registration pattern
- Able to specify via `CallOption`s
- Same/similar type signature as RPC handler itself (Java does this)

This is currently just a quick list based on my memory at the moment.  I intend to add to this as I run into more things.

devpro108:
@dfawley wondering any news on new interceptor design, when it will be available, looking forward to single interceptor for both unary and streams..

devpro108:
Thank you @dfawley

jhump:
@dfawley, I think stale bot got it wrong on this one, too. I see this is "on hold", but probably should not be closed -- AFAICT, it is not pending any further input from the reporter(s).

dfawley:
Yeah, it hit a lot of issues this morning.  It doesn't seem to be listening to the "onlyLabels" setting (or we set it wrong).

PR #6652, opened 2023-09-21
transport: Pass Header metadata to tap handle.

pstibrany:
This PR extends `tap.Info` struct with metadata, and modifies `http2_server.go` to pass available metadata to `ServerInHandle` function.

This gives tap handle more details that it can use to reject request early.

RELEASE NOTES:
* tap (experimental): Add Header metadata to tap handler

### DIFF
--- a/internal/transport/http2_server.go
+++ b/internal/transport/http2_server.go
@@ -35,7 +35,6 @@ import (
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
-
 	"google.golang.org/grpc/internal/grpclog"
 	"google.golang.org/grpc/internal/grpcutil"
 	"google.golang.org/grpc/internal/pretty"
--- a/internal/transport/http2_server.go
+++ b/internal/transport/http2_server.go
@@ -35,6 +35,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
+
 	"google.golang.org/grpc/internal/grpclog"
 	"google.golang.org/grpc/internal/grpcutil"
 	"google.golang.org/grpc/internal/pretty"
@@ -566,7 +567,7 @@ func (t *http2Server) operateHeaders(frame *http2.MetaHeadersFrame, handle func(
 	}
 	if t.inTapHandle != nil {
 		var err error
-		if s.ctx, err = t.inTapHandle(s.ctx, &tap.Info{FullMethodName: s.method}); err != nil {
+		if s.ctx, err = t.inTapHandle(s.ctx, &tap.Info{FullMethodName: s.method, Header: mdata}); err != nil {
 			t.mu.Unlock()
 			if t.logger.V(logLevel) {
 				t.logger.Infof("Aborting the stream early due to InTapHandle failure: %v", err)
--- a/tap/tap.go
+++ b/tap/tap.go
@@ -27,6 +27,8 @@ package tap
 
 import (
 	"context"
+
+	"google.golang.org/grpc/metadata"
 )
 
 // Info defines the relevant information needed by the handles.
@@ -34,6 +36,10 @@ type Info struct {
 	// FullMethodName is the string of grpc method (in the format of
 	// /package.service/method).
 	FullMethodName string
+
+	// Header contains the header metadata received.
+	Header metadata.MD
+
 	// TODO: More to be added.
 }
 
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -44,7 +44,6 @@ import (
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
-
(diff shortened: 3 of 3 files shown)


dfawley:
May I ask your use case for the `tap` package?  Have you compared this to using interceptors for the same purpose, by any chance?

The other grpc languages don't have this construct (and seem to be doing just fine without it), and we recently noticed that it's excluded from metrics & tracing, which is not very desirable from an observability perspective, and we may not want to introduce metrics callouts specifically for it, either, given its sensitive location.  So I was actually considering just calling it deprecated and not recommending anyone to use it.

pstibrany:
> May I ask your use case for the `tap` package? Have you compared this to using interceptors for the same purpose, by any chance?

Hi. Our use case is to drop requests early, before spending too much resources on it. We use this when server is already unable to accept more messages (eg. too many inflight requests). We have tried to use interceptors before, but that did not provide adequate protection.

From what I understand, gRPC-go is processing the request in following steps (discussing unary case):

1. Read request headers.
2. Call tap handle, if configured. Tap handle can reject the request at this point.
3. Request is then handed to one of worker goroutines (if configured, and any is free), or new goroutine is spawned.
4. Generated method handler is called. This method handler receives function to read the request into memory. After reading request, generated method handler then then unmarshals the message.
6. Unmarshaled message is passed to interceptors and eventually registered handler.

As can be seen in this description, when interceptors run, quite a lot of resources have already been used: possibly new goroutine spawned, request is in memory twice (once as bytes buffer, once as unmarshaled message). We want to avoid all that when we know that server can't possibly handle any more requests at the moment.

When we tried to use interceptors, we couldn't implement the protection effectively, and we could still crash the server under the load, for example by sending enough big messages -- then the server will just OOM.

> The other grpc languages don't have this construct (and seem to be doing just fine without it), and we recently noticed that it's excluded from metrics & tracing, which is not very desirable from an observability perspective, and we may not want to introduce metrics callouts specifically for it, either, given its sensitive location. So I was actually considering just calling it deprecated and not recommending anyone to use it.

We've also noticed exclusion from metrics, because our metrics get updated in stats handler and interceptors. However that's easy to fix: we can increment relevant metric when rejecting request in tap handle. As for tracing, in our setup client would log returned grpc error to its trace span. (But we use grpc only in internal communication)

I think that even with these limitations, tap handle is useful feature and gives us much-needed protection against overloading server with too many messages.

dfawley:
Thanks for all the detailed info - it was really helpful!

Do you happen to work with gRPC in other languages besides Go (on the server side)?  I'm not sure, but I believe the recommended way of dealing with these kinds of issues would be to use a [LimitingListener](https://pkg.go.dev/golang.org/x/net/netutil#LimitListener) and [MaxConcurrentStreams](https://pkg.go.dev/google.golang.org/grpc#MaxConcurrentStreams) instead.  Have you tried these, by any chance?

It's also possible the other languages tend to provide interceptors at a lower level.  Especially if you're looking at the message payload size in unary RPCs, I believe the other languages provide interceptors that are more like our streaming interceptors, but for all RPCs, that do not require the payload to be received and decoded before calling them, as our unary interceptors do.  Unary interceptors can be nice for usability, but otherwise have some undesirable properties (e.g. they require implementing things twice if you have both unary and streaming RPCs that want the same interceptor behavior).  To that end, we do have #1805 open to provide a streaming-style interceptor even for unary RPCs.  It hasn't been prioritized, but might allow for sufficiently early termination of RPCs - it would cut out steps 4 and 5, saving 2 copies of the message payload.  Could something like that potentially work for you in place of the tap handle feature?

dfawley (commented):
I think we can take this PR, but be aware that use of the tap handler feature is not recommended.

Also, can you please add a test for this.  There is an existing `TestTap` in `test/end2end_test.go` that you can probably pretty easily extend:

https://github.com/grpc/grpc-go/blob/fd9ef7263ae34b82b6200fee951af1b2edcb16df/test/end2end_test.go#L2111

pstibrany:
> I think we can take this PR, but be aware that use of the tap handler feature is not recommended.

👍 Understood. The best I can hope for is that it won't be removed anytime soon (before usable alternative). Thank you!

> Also, can you please add a test for this. There is an existing `TestTap` in `test/end2end_test.go` that you can probably pretty easily extend:
> 
> https://github.com/grpc/grpc-go/blob/fd9ef7263ae34b82b6200fee951af1b2edcb16df/test/end2end_test.go#L2111

Thank you for the pointer. I was wondering how to add test for this change and missed the end2end test!

pstibrany:
> Also, can you please add a test for this. There is an existing `TestTap` in `test/end2end_test.go` that you can probably pretty easily extend:

I've added test for this change. Please take a look again.

dfawley (approved):

dfawley:
+zasweq for a second look

zasweq (approved):
LGTM.

pstibrany:
Thank you!

bboreham:
> We have experimented with MaxConcurrentStreams (which is per-connection)

Not any more: as of #6703 it is also implemented to limit the whole server.

bboreham:
Apologies, this was a fantasy that entered my mind during intensive debugging.

PR #6655, opened 2023-09-21
balancer/weightedroundrobin: fix ticker leak on update

ginayeh:
cherry-pick #6643 

RELEASE NOTES:
- balancer/weightedroundrobin: fix a memory leak caused by not stopping a `time.Ticker`


dfawley (approved):

dfawley:
FYI: I moved the release notes here because we don't need them in the next minor release, just in this patch release.

PR #6670, opened 2023-09-29
crl provider: Static and FileWatcher provider implementations

erm-g:
Part of CRL Provider effort. The functionality in this PR includes a built in Static and FileWatcher provider implementations. It also gives users a tool to implement their own providers

RELEASE NOTES: N/A


security/advancedtls/advancedtls_test.go
```diff
@@ -594,14 +619,14 @@ func (s) TestClientServerHandshake(t *testing.T) {
 		// server custom check fails
 		{
 			desc:                       "Client sets peer cert, reload root function with verifyFuncGood; Server sets bad custom check; mutualTLS",
```
gtcooke94, line 621: I see you changed this test case and it's not a new one - just curious as to why you changed it?

security/advancedtls/advancedtls_test.go
```diff
@@ -704,6 +729,37 @@ func (s) TestClientServerHandshake(t *testing.T) {
 				Cache:             cache,
 			},
 		},
+		// Client: set valid credentials with the revocation config
+		// Server: set valid credentials with the revocation config
+		// Expected Behavior: success, because non of the certificate chains sent in the connection are revoked
```
gtcooke94, line 734: nit: `non` -> `none`

security/advancedtls/advancedtls_test.go
```diff
@@ -704,6 +729,37 @@ func (s) TestClientServerHandshake(t *testing.T) {
...
+			desc:                   "Client sets peer cert, reload root function with verifyFuncGood; Server sets peer cert, reload root function; Client uses CRL; mutualTLS",
+			clientCert:             []tls.Certificate{cs.ClientCert3},
+			clientGetRoot:          getRootCAsForClientCRL,
+			clientVerifyFunc:       clientVerifyFuncGood,
+			clientVType:            CertVerification,
+			clientRevocationConfig: makeStaticCRLProvider(false),
+			serverMutualTLS:        true,
+			serverCert:             []tls.Certificate{cs.ServerCert3},
+			serverGetRoot:          getRootCAsForServerCRL,
+			serverVType:            CertVerification,
```
gtcooke94, line 745: Consider adding `serverExpectError: false` here to make the test case definition easier to parse at a glance.

security/advancedtls/crl.go
```diff
@@ -83,15 +88,48 @@ func (s RevocationStatus) String() string {
...
+		return nil, fmt.Errorf("parseCRLExtensions() failed err = %v", err)
+	}
+	crlExt.RawIssuer, err = extractCRLIssuer(b)
+	if err != nil {
+		return nil, fmt.Errorf("extractCRLIssuer() failed err= %v", err)
+	}
+	return crlExt, nil
+}
+
+// ReadCRLFile reads a file from the provided path, and returns constructed
```
gtcooke94, line 119: suggest something like "ReadCRLFile reads a crl file from the provided path and returns a constructed CRL struct."

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,229 @@
...
+	"os"
+	"time"
+)
+
+const defaultCRLRefreshDuration = 1 * time.Hour
+
+// CRLProvider is the interface to be implemented to enable custom CRL provider
+// behavior.
+//
+// The interface defines how the data is read, but doesn't prescribe a way
```
gtcooke94, line 34: Suggest something more along the lines of
"The interface defines how gRPC gets CRLs from the provider during handshakes, but doesn't prescribe a specific way to load and store CRLs."

Also suggest adding info about this being directly on the path of connection establishment, so implementations of the `CRL` function need to be fast, and slow things such as file IO should be done asynchronously.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,229 @@
...
+	// CRL accepts x509 Cert and returns back related CRL struct. The CRL struct
+	// can be nil, can contain empty or non-empty list of revkoed certificates.
+	// Callers are expected to use the returned value as read-only.
+	CRL(cert *x509.Certificate) (*CRL, error)
+}
+
+// StaticCRLProvider implements CRLProvider interface by accepting CRL structs
+// and storing them in-memory.
+type StaticCRLProvider struct {
+	// TODO CRL is sort of our internal representation - provide an API for
```
gtcooke94, line 51: I think that the way this has been structured makes it easy for users to construct `CRL` objects?

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,229 @@
...
+
+// MakeStaticCRLProvider returns a new instance of the StaticCRLProvider.
+func MakeStaticCRLProvider() *StaticCRLProvider {
+	p := StaticCRLProvider{}
+	p.crls = make(map[string]*CRL)
+	return &p
+}
+
+// AddCRL adds/updates provided CRL to in-memory storage.
+func (p *StaticCRLProvider) AddCRL(crl *CRL) {
```
gtcooke94, line 65: This impl is significantly different than C-core -  there we accept a std::vector<std::string> where each string is the raw crl file content, then read those in. There's o way to set/change CRLs after creation. To me, that's the correct behavior of something called `StaticCrlProvider`. The implementation here is more of a dynamic map the user can change and update.

I think I'm actually the one that wrote this sell a while ago? But I don't think it's what should be done anymore.
WDYT? Can you change the behavior to be consistent with what I described from C++ above?

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,229 @@
...
+
+// CRL returns CRL struct if it was previously loaded by calling AddCRL.
+func (p *StaticCRLProvider) CRL(cert *x509.Certificate) (*CRL, error) {
+	key := cert.Issuer.ToRDNSequence().String()
+	return p.crls[key], nil
+}
+
+// Options represents a data structure holding a
+// configuration for FileWatcherCRLProvider.
+type Options struct {
```
gtcooke94, line 78: I think we should name this `FileWatcherOptions`

gtcooke94: Alternatively (and this is a go style thing so I'm not 100% sure) we could just eschew the `Options` struct and have the `MakeFileWatcherCrlProvider` just take these arguments directly

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,229 @@
...
+		if p.opts.cRLReloadingFailedCallback != nil {
+			p.opts.cRLReloadingFailedCallback(err)
+		}
+	}
+
+	successCounter := 0
+	failCounter := 0
+	for _, file := range files {
+		filePath := fmt.Sprintf("%s/%s", p.opts.CRLDirectory, file.Name())
+		err := p.addCRL(filePath)
```
gtcooke94, line 186: I think we need some thread safety around these updates, or we need to use a backing map type that is thread safe

security/advancedtls/crl_provider_test.go
```diff
@@ -0,0 +1,181 @@
...
+			desc:        "Chain with no CRL for issuer",
+			certs:       makeChain(t, testdata.Path("client_cert_1.pem")),
+			expectNoCRL: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			for _, c := range tt.certs {
+				crl, err := p.CRL(c)
```
gtcooke94, line 71: Can we also check that the correct CRL is returned, rather than just that a non-nil value is returned?

security/advancedtls/crl_provider_test.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		})
+	}
+}
+
+// TestFileWatcherCRLProviderConfig checks creation of FileWatcherCRLProvider,
+// and the validation of Options configuration. The configurations include empty
+// one, non existing CRLDirectory, invalid RefreshDuration, and the correct one.
+func TestFileWatcherCRLProviderConfig(t *testing.T) {
+	if _, err := MakeFileWatcherCRLProvider(Options{}); err == nil {
```
gtcooke94, line 87: I think we can split the `if` cases here into separate tests, or make this a Table Test?

gtcooke94, line 87: And let's check that the `errors` being returned are the expected errors

security/advancedtls/crl_provider_test.go
```diff
@@ -0,0 +1,181 @@
...
+	regularProvider.Close()
+}
+
+// TestFileWatcherCRLProvider tests how FileWatcherCRLProvider handles the major
+// four cases for CRL checks. It scans the CRLs under crl directory to populate
+// the in-memory storage. Then we construct unrevoked, revoked leaf, and revoked
+// intermediate chains, as well as a chain without CRL for issuer, and check
+// that it’s correctly processed. Additionally, we also check if number of
+// invocations of custom callback is correct.
+func TestFileWatcherCRLProvider(t *testing.T) {
```
gtcooke94, line 122: Do we have a test verifying the async refreshing goroutine works?

gtcooke94 (commented):
Looks really good! Have a few questions/comments

security/advancedtls/advancedtls_test.go
erm-g: That's a good catch, I was experimenting with some chains and accidentally submitted (it doesn't influence the outputs). Will revert it back

security/advancedtls/crl_provider_test.go
erm-g: Not yet, it'll require playing with the timers so I'm planning to have a separate PR for that (like e2e test)

dfawley (commented):
I got about halfway through yesterday, but there are some meta things that should probably be addressed (primarily: it seems there are a lot of existing, unaddressed comments) before I continue.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// MakeFileWatcherCRLProvider returns a new instance of the
+// FileWatcherCRLProvider. It uses Options to validate and apply configuration
+// required for creating a new instance.
+func MakeFileWatcherCRLProvider(o Options) (*FileWatcherCRLProvider, error) {
+	if err := o.validate(); err != nil {
+		return nil, err
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
erm-g: I feel I need some help here.
I followed a pattern from [watcher.go](https://github.com/grpc/grpc-go/blob/59f57b160e79284dee90b26e4532a57542eaf056/credentials/tls/certprovider/pemfile/watcher.go#L116)
After looking through the codebase it looks to me that for background tasks a new context is being used: [balancers](https://github.com/grpc/grpc-go/blob/59f57b160e79284dee90b26e4532a57542eaf056/balancer/weightedroundrobin/balancer.go#L326), [authz filewatcher](https://github.com/grpc/grpc-go/blob/59f57b160e79284dee90b26e4532a57542eaf056/authz/grpc_authz_server_interceptors.go#L116), [dns resolver](https://github.com/grpc/grpc-go/blob/59f57b160e79284dee90b26e4532a57542eaf056/internal/resolver/dns/dns_resolver.go#L134)
Passing context across functions is a pattern for a call flow like [clientconn](https://github.com/grpc/grpc-go/blob/59f57b160e79284dee90b26e4532a57542eaf056/clientconn.go#L949), [streaming](https://github.com/grpc/grpc-go/blob/59f57b160e79284dee90b26e4532a57542eaf056/stream.go#L1210)

Did I get the patterns right? Or is it more 'old' vs 'new' way of using the context?

security/advancedtls/crl.go
```diff
@@ -253,16 +291,35 @@ func fetchIssuerCRL(rawIssuer []byte, crlVerifyCrt []*x509.Certificate, cfg Revo
...
 }
 
+func fetchCRL(c *x509.Certificate, crlVerifyCrt []*x509.Certificate, cfg RevocationConfig) (*CRL, error) {
+	if cfg.CRLProvider != nil {
+		crl, err := cfg.CRLProvider.CRL(c)
+		if err != nil {
+			return nil, fmt.Errorf("CrlProvider failed err = %v", err)
+		}
+		if crl == nil {
+			// TODO print out cert info here? What cert contents are okay to have in a log?
```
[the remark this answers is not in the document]
erm-g: Yeah we had an offline discussion with @gtcooke94 and decided not to print any for both C++ and Go

security/advancedtls/crl.go
```diff
@@ -83,15 +88,48 @@ func (s RevocationStatus) String() string {
 	return [...]string{"RevocationUndetermined", "RevocationUnrevoked", "RevocationRevoked"}[s]
 }
 
-// certificateListExt contains a pkix.CertificateList and parsed
+// CRL contains a pkix.CertificateList and parsed
 // extensions that aren't provided by the golang CRL parser.
-type certificateListExt struct {
+// GRPC requires certain specifics for CRLs - all CRLs should be loaded using
+// NewCRL() for bytes directly or ReadCRLFile() to read directly from a filepath
+type CRL struct {
```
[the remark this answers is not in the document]
erm-g: After an offline discussion with @gtcooke94 we made this info private for both C++ and Go

security/advancedtls/crl_deprecated.go
```diff
@@ -83,9 +83,9 @@ func (s RevocationStatus) String() string {
 	return [...]string{"RevocationUndetermined", "RevocationUnrevoked", "RevocationRevoked"}[s]
 }
 
-// certificateListExt contains a pkix.CertificateList and parsed
+// CRL contains a pkix.CertificateList and parsed
 // extensions that aren't provided by the golang CRL parser.
-type certificateListExt struct {
+type CRL struct {
```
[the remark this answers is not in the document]
erm-g: IIUC we need support 3 latest releases of Go and it's 1.21, 1.20, and 1.19 (https://go.dev/dl/) - so yes, it's safe to delete it (1.19+ is a prereq for deletion)
Do you prefer me to do it here or via a separate PR?

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
erm-g: Refactored to channel

security/advancedtls/crl_deprecated.go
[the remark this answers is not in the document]
dfawley: I don't mind either way, but I don't think we should make any of the changes in this PR, since the file should be deleted instead.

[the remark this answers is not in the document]
erm-g: Ok I just created a separate PR and added you as a reviewer - https://github.com/grpc/grpc-go/pull/6721

security/advancedtls/advancedtls_test.go
```diff
@@ -594,14 +619,14 @@ func (s) TestClientServerHandshake(t *testing.T) {
 		// server custom check fails
 		{
 			desc:                       "Client sets peer cert, reload root function with verifyFuncGood; Server sets bad custom check; mutualTLS",
-			clientCert:                 []tls.Certificate{cs.ClientCert1},
-			clientGetRoot:              getRootCAsForClient,
+			clientCert:                 []tls.Certificate{cs.ClientCert3},
```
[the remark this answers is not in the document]
erm-g: I'd prefer to leave it as is for this PR since it fits the current pattern - `security/advancedtls/internal/testutils/testutils.go`
Definitely something to address during 'Unifying API' effort (will be a separate PR)

security/advancedtls/advancedtls_test.go
```diff
@@ -594,14 +619,14 @@ func (s) TestClientServerHandshake(t *testing.T) {
...
+			clientGetRoot:              getRootCAsForClientCRL,
 			clientVerifyFunc:           clientVerifyFuncGood,
 			clientVType:                CertVerification,
 			clientExpectHandshakeError: true,
 			serverMutualTLS:            true,
-			serverCert:                 []tls.Certificate{cs.ServerCert1},
-			serverGetRoot:              getRootCAsForServer,
+			serverCert:                 []tls.Certificate{cs.ServerCert3},
```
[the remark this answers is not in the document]
erm-g: I'd prefer to leave it as is for this PR since it fits the current pattern - security/advancedtls/internal/testutils/testutils.go
Definitely something to address during 'Unifying API' effort (will be a separate PR)

security/advancedtls/advancedtls_test.go
erm-g: Done

security/advancedtls/advancedtls_test.go
erm-g: It fits the current pattern, if `serverExpectError` is missing from `test struct` it's `false`.

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider_test.go
erm-g: That's a good point but I'd prefer to keep it very simple here and do detailed checks on integration test level

security/advancedtls/crl_provider_test.go
erm-g: I'd prefer to have 3 separate test methods - to check config, CRL api, and directory scanning. I think config one can also be fit to Table test pattern (like the other 2) but I think it's a little bit easier to read it in the current format

erm-g: Looking at how tests are organized for `credentials/tls/certprovider/pemfile/watcher_test.go` I think it's an overkill for the config tests.

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl.go
erm-g: Done

security/advancedtls/advancedtls_test.go
```diff
@@ -371,6 +372,30 @@ func (s) TestClientServerHandshake(t *testing.T) {
...
+	getRootCAsForClientCRL := func(params *GetRootCAsParams) (*GetRootCAsResults, error) {
+		return &GetRootCAsResults{TrustCerts: cs.ClientTrust3}, nil
+	}
+
+	getRootCAsForServerCRL := func(params *GetRootCAsParams) (*GetRootCAsResults, error) {
+		return &GetRootCAsResults{TrustCerts: cs.ServerTrust3}, nil
+	}
+
+	makeStaticCRLProvider := func(crlPath string) *RevocationConfig {
+
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/advancedtls_test.go
```diff
@@ -371,6 +372,30 @@ func (s) TestClientServerHandshake(t *testing.T) {
...
+		rawCRLs := make([][]byte, 0)
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl.go
```diff
@@ -83,15 +88,48 @@ func (s RevocationStatus) String() string {
 	return [...]string{"RevocationUndetermined", "RevocationUnrevoked", "RevocationRevoked"}[s]
 }
 
-// certificateListExt contains a pkix.CertificateList and parsed
+// CRL contains a pkix.CertificateList and parsed
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl.go
```diff
@@ -83,15 +88,48 @@ func (s RevocationStatus) String() string {
...
 // extensions that aren't provided by the golang CRL parser.
-type certificateListExt struct {
+// GRPC requires certain specifics for CRLs - all CRLs should be loaded using
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl.go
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl.go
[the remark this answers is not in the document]
erm-g: Done - unexported

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,242 @@
...
+
+const defaultCRLRefreshDuration = 1 * time.Hour
+
+// CRLProvider is the interface to be implemented to enable custom CRL provider
+// behavior.
+//
+// The interface defines how the data is read, but doesn't prescribe a way
+// CRL are loaded and stored. Such implementations can be used in
+// RevocationConfig of advancedtls.ClientOptions and/or
+// advancedtls.ServerOptions .
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+//
+// The interface defines how gRPC gets CRLs from the provider during handshakes,
+// but doesn't prescribe a specific way to load and store CRLs. Such
+// implementations can be used in RevocationConfig of advancedtls.ClientOptions
+// and/or advancedtls.ServerOptions.
+// Please note that checking CRLs is being directly on the path of connection
+// establishment, so implementations of the CRL function need to be fast, and
+// slow things such as file IO should be done asynchronously.
+// TODO(erm-g): Add link to related gRFC once it's ready.
+// Please refer to https://github.com/grpc/proposal/ for more details.
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+type CRLProvider interface {
+	// CRL accepts x509 Cert and returns back related CRL struct. The CRL struct
+	// can be nil, can contain empty or non-empty list of revkoed certificates.
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+	// Callers are expected to use the returned value as read-only.
```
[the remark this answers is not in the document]
erm-g: Removed

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
erm-g: In the case it’s nil it’s up to the caller how to treat it - Scenario 1 (and pseudocode for suggested behavior) under https://github.com/grpc/proposal/blob/dddf32d0116376dd0c48adee7b0071a20bc82b5b/A69-crl-enhancements.md#map-content-and-api-outcomes
Some additional context - we don’t require PKI to distribute CRL files for all the certs, so we expect nils to happen

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("advancedtls: CRLDirectory %v does not exist", o.CRLDirectory)
+		}
+		return err
+	}
+	if !fileInfo.IsDir() {
+		return fmt.Errorf("advancedtls: CRLDirectory %v is not a directory", o.CRLDirectory)
+	}
+	_, err = os.Open(o.CRLDirectory)
```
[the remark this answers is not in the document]
erm-g: Yeah I think it makes sense. Instead of throwing a number of custom error messages now I just have one but include original error from os.Read
Yes, the validation that directory exists happens at creation time (background scan also kicks out automatically). 'Lazy' part of the scenario is - the directory can be empty and CRL files can be shipped later.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		if os.IsPermission(err) {
+			return fmt.Errorf("advancedtls: CRLDirectory %v is not readable", o.CRLDirectory)
+		}
+		return err
+	}
+	// Checks related to RefreshDuration.
+	if o.RefreshDuration <= 0 || o.RefreshDuration < time.Second {
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+
+// StaticCRLProvider implements CRLProvider interface by accepting raw content
+// of CRL files at creation time and storing parsed CRL structs  in-memory.
+type StaticCRLProvider struct {
+	crls map[string]*CRL
+}
+
+// MakeStaticCRLProvider processes raw content of CRL files, adds parsed CRL
+// structs into in-memory, and returns a new instance of the StaticCRLProvider.
+func MakeStaticCRLProvider(rawCRLs [][]byte) *StaticCRLProvider {
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+// Please refer to https://github.com/grpc/proposal/ for more details.
+type CRLProvider interface {
+	// CRL accepts x509 Cert and returns back related CRL struct. The CRL struct
+	// can be nil, can contain empty or non-empty list of revkoed certificates.
+	// Callers are expected to use the returned value as read-only.
+	CRL(cert *x509.Certificate) (*CRL, error)
+}
+
+// StaticCRLProvider implements CRLProvider interface by accepting raw content
+// of CRL files at creation time and storing parsed CRL structs  in-memory.
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+// Options represents a data structure holding a
+// configuration for FileWatcherCRLProvider.
+type Options struct {
+	CRLDirectory               string          // Path of the directory containing CRL files
+	RefreshDuration            time.Duration   // Time interval between CRLDirectory scans
+	crlReloadingFailedCallback func(err error) // Custom callback executed when a CRL file can’t be processed
+}
+
+// FileWatcherCRLProvider implements the CRLProvider interface by periodically scanning
+// CRLDirectory (see Options) and storing CRL structs in-memory
```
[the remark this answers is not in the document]
erm-g: Done for both

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+	crls map[string]*CRL
+	opts Options
+	mu   sync.Mutex
+	done chan struct{}
+}
+
+// MakeFileWatcherCRLProvider returns a new instance of the
+// FileWatcherCRLProvider. It uses Options to validate and apply configuration
+// required for creating a new instance.
+func MakeFileWatcherCRLProvider(o Options) (*FileWatcherCRLProvider, error) {
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+type FileWatcherCRLProvider struct {
+	crls map[string]*CRL
+	opts Options
+	mu   sync.Mutex
+	done chan struct{}
+}
+
+// MakeFileWatcherCRLProvider returns a new instance of the
+// FileWatcherCRLProvider. It uses Options to validate and apply configuration
+// required for creating a new instance.
```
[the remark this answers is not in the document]
erm-g: I agree that this comment look ugly but moving `FileWatcherCRLProvider.` to the first line makes it 84. After refactoring Make->New it'd be still 83 if I move it there.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+			return
+		case <-ticker.C:
+			p.ScanCRLDirectory()
+		}
+	}
+}
+
+// Close stops the background refresh of CRLDirectory of FileWatcherCRLProvider
+func (p *FileWatcherCRLProvider) Close() {
+	close(p.done)
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// ScanCRLDirectory starts the process of scanning Options.CRLDirectory and
+// updating in-memory storage of CRL structs. Please note that the same method is
+// called periodically by run goroutine.
+// TODO(erm-g): Add link to related gRFC once it's ready.
+// Please refer to https://github.com/grpc/proposal/ for edge case details.
+func (p *FileWatcherCRLProvider) ScanCRLDirectory() {
```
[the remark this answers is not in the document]
erm-g: Yes, I have the following scenario in my mind - users configure the refresh interval to be an hour (or a day) but if they receive an urgent notice to refresh all the CRLs they need an API for that. If it's not exposed they'd need to recreate the providers.
I think it's a great idea to add locks - at least some protection against infinite loops calling ScanCRLDirectory()
Done

security/advancedtls/crl_provider_test.go
```diff
@@ -0,0 +1,297 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc/security/advancedtls/testdata"
+)
+
+const nonCRLFilesUnderCRLDirectory = 5
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider_test.go
```diff
@@ -0,0 +1,297 @@
...
+		t.Fatalf("CRLDirectory must exist")
+	}
+	if defaultProvider, err := MakeFileWatcherCRLProvider(Options{CRLDirectory: testdata.Path("crl/provider")}); err == nil {
+		if defaultProvider.opts.RefreshDuration != defaultCRLRefreshDuration {
+			t.Fatalf("RefreshDuration is not properly updated by validate() func")
+		}
+		defaultProvider.Close()
+	} else {
+		t.Fatal("Unexpected error:", err)
+	}
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/advancedtls_test.go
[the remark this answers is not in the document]
gtcooke94: I don't think we should continue to add bad code that we plan to refactor later just to keep with a bad existing pattern - let's go ahead and name these variables with something meaningful rather than making more work for ourselves down the road.

In fact, I'd argue the current pattern isn't much of a pattern - it's just lazy variable naming. For there to be a pattern there would need to be some meaning to `Cert1`, `Cert2`, etc

security/advancedtls/advancedtls_test.go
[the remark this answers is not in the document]
gtcooke94: See other thread

security/advancedtls/crl_deprecated.go
dfawley, line 1: Please revert/re-delete these.

security/advancedtls/testdata/crl/provider/README.md
dfawley, line 1: Optional: this directory should be moved under `internal`, probably.

security/advancedtls/testdata/crl/provider/README.md
```diff
@@ -0,0 +1,73 @@
...
+keyUsage = digitalSignature, keyEncipherment
+```
+
+* Generate CRLs
+For CRL generation we need 2 more files called `index.txt` and `crlnumber.txt`:
+```
+$ echo "1000" > crlnumber.txt
+$ touch index.txt
+```
+Also we need anothe config `crl.cnf` -
```
dfawley, line 48: another*

security/advancedtls/testdata/crl/provider/README.md
```diff
@@ -0,0 +1,73 @@
+About This Directory
+-------------
+The directory test data (certificates and CRLs) used for testing CRL providers 
+functionality.
+
+How to Generate Test Data Using OpenSSL
```
dfawley, line 6: Optional: it's nice that you show how to regenerate everything, but it would be even nicer if there was just a script that could be run that does it instead.

security/advancedtls/testdata/crl/provider/filewatcher/.gitignore
dfawley, line 1: Please don't submit these kinds of files.

But, also, why does this directory exist?  If it's for during-testing use, then make a temp directory instead.

security/advancedtls/testdata/crl/provider/extensions.conf
```diff
@@ -0,0 +1,4 @@
+subjectKeyIdentifier = hash
+authorityKeyIdentifier = keyid,issuer
+basicConstraints = CA:FALSE
+keyUsage = digitalSignature, keyEncipherment
```
dfawley, line 4: Nit: you may want a newline at the end of this file?

security/advancedtls/internal/testutils/testutils.go
```diff
@@ -35,12 +35,18 @@ type CertStore struct {
 	// ClientCert2 is the certificate sent by client to prove its identity.
 	// It is trusted by ServerTrust2.
 	ClientCert2 tls.Certificate
+	// ClientCert3 is the certificate sent by client to prove its identity.
+	// It is trusted by ServerTrust3. Used in CRL tests
+	ClientCert3 tls.Certificate
```
dfawley, line 40: At what point do we use an array instead of enumerating each one in its own variable?

security/advancedtls/internal/testutils/testutils.go
```diff
@@ -79,12 +89,18 @@ func (cs *CertStore) LoadCerts() error {
 	if cs.ClientCert2, err = tls.LoadX509KeyPair(testdata.Path("client_cert_2.pem"), testdata.Path("client_key_2.pem")); err != nil {
 		return err
 	}
+	if cs.ClientCert3, err = tls.LoadX509KeyPair(testdata.Path("crl/provider/client_cert.pem"), testdata.Path("crl/provider/client_cert.key")); err != nil {
+		return err
+	}
```
dfawley, line 94: Similar, but... loops.

Also why does this break the existing pattern?

Is it possible to keep using the old one?  Also note that a new directory for new files means more work when we import into g3, so it would be nice to keep everything together if possible.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+const defaultCRLRefreshDuration = 1 * time.Hour
+
+// CRLProvider is the interface to be implemented to enable custom CRL provider
+// behavior.
+//
+// The interface defines how gRPC gets CRLs from the provider during handshakes,
+// but doesn't prescribe a specific way to load and store CRLs. Such
+// implementations can be used in RevocationConfig of advancedtls.ClientOptions
+// and/or advancedtls.ServerOptions.
+// Please note that checking CRLs is being directly on the path of connection
```
dfawley, line 38: is being directly -> is

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+// establishment, so implementations of the CRL function need to be fast, and
+// slow things such as file IO should be done asynchronously.
+// TODO(erm-g): Add link to related gRFC once it's ready.
+//
+// [A69: CRL Enhancements]: https://github.com/grpc/proposal/pull/382
```
dfawley, line 43: Nothing references this tag though?  So I believe this doesn't get rendered...

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+	<-p.done
+}
+
+// ScanCRLDirectory starts the process of scanning
+// FileWatcherOptions.CRLDirectory and updating in-memory storage of CRL
+// structs. Please note that the same method is called periodically by run
+// goroutine.
+// TODO(erm-g): Add link to related gRFC once it's ready.
+//
+// [A69: CRL Enhancements]: https://github.com/grpc/proposal/pull/382
```
dfawley, line 170: As above.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+func (p *FileWatcherCRLProvider) run() {
+	defer close(p.done)
+	ticker := time.NewTicker(p.opts.RefreshDuration)
+	defer ticker.Stop()
+	p.ScanCRLDirectory()
+
+	for {
+		select {
+		case <-p.stop:
+			ticker.Stop()
```
dfawley, line 150: This isn't needed because of the `defer`.

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+func (p *StaticCRLProvider) CRL(cert *x509.Certificate) (*CRL, error) {
+	return p.crls[cert.Issuer.ToRDNSequence().String()], nil
+}
+
+// FileWatcherOptions represents a data structure holding a configuration for
+// FileWatcherCRLProvider.
+type FileWatcherOptions struct {
+	CRLDirectory               string          // Path of the directory containing CRL files
+	RefreshDuration            time.Duration   // Time interval between CRLDirectory scans
+	crlReloadingFailedCallback func(err error) // Custom callback executed when a CRL file can’t be processed
```
dfawley, line 88: Please name or document that it's for testing only.

security/advancedtls/testdata/crl/provider/README.md
erm-g: Done

security/advancedtls/testdata/crl/provider/filewatcher/.gitignore
erm-g: Yes, it's for testing only - refactored to `os.MkdirTemp`

security/advancedtls/testdata/crl/provider/extensions.conf
erm-g: Done

security/advancedtls/internal/testutils/testutils.go
erm-g: This is a great q. In general I find 'testutils.go' a little bit confusing - we load everything there and then use variables named like `cs.ServerCert1` and `cs.ClientCert1` across the codebase. So when I open a test it's tough to understand what exactly the certs mean.
Since you were the approver of [AuditLogger test](https://github.com/grpc/grpc-go/pull/6304) can I use the same pattern (test has properly named helper methods which load needed certs for it) instead?
https://github.com/grpc/grpc-go/blob/e88e8498c6df39703d1992bc4b677e7ef477145d/authz/audit/audit_logging_test.go#L329

security/advancedtls/internal/testutils/testutils.go
erm-g: There is the same problem with existing `testdata` and `testdata/crl` dirs - Readme files don't contain the exact commands and don't cover all the files. That's why I created a clean subdir with better documentation inside.
If import into g3 is tough I can utilize an existing dir but add suffix/prefix for new files - WDYT?

security/advancedtls/crl_provider.go
erm-g: done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+// behavior.
+//
+// The interface defines how gRPC gets CRLs from the provider during handshakes,
+// but doesn't prescribe a specific way to load and store CRLs. Such
+// implementations can be used in RevocationConfig of advancedtls.ClientOptions
+// and/or advancedtls.ServerOptions.
+// Please note that checking CRLs is being directly on the path of connection
+// establishment, so implementations of the CRL function need to be fast, and
+// slow things such as file IO should be done asynchronously.
+// TODO(erm-g): Add link to related gRFC once it's ready.
```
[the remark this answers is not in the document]
erm-g: done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+// Close stops the background refresh of CRLDirectory of FileWatcherCRLProvider.
+func (p *FileWatcherCRLProvider) Close() {
+	close(p.stop)
+	<-p.done
+}
+
+// ScanCRLDirectory starts the process of scanning
+// FileWatcherOptions.CRLDirectory and updating in-memory storage of CRL
+// structs. Please note that the same method is called periodically by run
+// goroutine.
```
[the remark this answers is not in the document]
erm-g: Rephrased

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+			if p.opts.crlReloadingFailedCallback != nil {
+				p.opts.crlReloadingFailedCallback(err)
+			}
+			continue
+		}
+		tempCRLs[crl.certList.Issuer.ToRDNSequence().String()] = crl
+		successCounter++
+	}
+	// Only if all the files are processed successful we can swap maps (there
+	// might be deletions of entries in this case
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+	if len(files) == successCounter {
+		p.mu.Lock()
+		defer p.mu.Unlock()
+		p.crls = make(map[string]*CRL)
+		for key, value := range tempCRLs {
+			p.crls[key] = value
+		}
+		grpclogLogger.Infof("Scan of CRLDirectory %v completed, %v files found and processed successfully, in-memory CRL storage flushed and repopulated", p.opts.CRLDirectory, len(files))
+	} else {
+		// Since some of the files failed we can only add/update entries in the map
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,235 @@
...
+func (o *FileWatcherOptions) validate() error {
+	// Checks relates to CRLDirectory.
+	if o.CRLDirectory == "" {
+		return fmt.Errorf("advancedtls: CRLDirectory needs to be specified")
+	}
+	if _, err := os.ReadDir(o.CRLDirectory); err != nil {
+		return fmt.Errorf("advancedtls: CRLDirectory %v is not readable: %v", o.CRLDirectory, err)
+	}
+	// Checks related to RefreshDuration.
+	if o.RefreshDuration < time.Second {
```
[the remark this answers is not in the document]
erm-g: Yes, it's cross-language. Some context - we faced an issue in C++ for another IO reloaders when users were setting 0 (no delays between reads) and logs got polluted and unreadable. Now we have a 1 second check for such functionality (if users want faster reload they should implement their own `MyUberFastWatcher`).
However I can hardly imagine PKI distributing CRL files with that frequency so I'm fine with a minute as a boundary here - WDYT?

security/advancedtls/crl_provider.go
erm-g: `crlReloadingFailedCallback` is not for testing - https://github.com/grpc/proposal/blob/dddf32d0116376dd0c48adee7b0071a20bc82b5b/A69-crl-enhancements.md#provider-implementation---directory-provider
`Particularly in the case of updating CRLs, we don't want to error and crash the server if there is a bad update. However, we still want this error to be known, so we will have an optional and overridable error callback when CRLs fail to be read. This will let users tie in whatever alerting/monitoring they may want in these failure cases. This callback will be for notifications/alerting, it will not be a decision-making callback`

security/advancedtls/internal/testutils/testutils.go
dfawley: I'm fine with whatever.  Even this is acceptable.  I'm just trying to point out things I see that don't look as good as they could be.

security/advancedtls/internal/testutils/testutils.go
dfawley: It would be even nicer to add documentation and/or scripts where they don't exist, rather than create a new directory where things are better.  Unless you're taking an AI to do that as a cleanup.

Different filenames is fine.  I just know creating new directories for data files incurs an extra cost on our import process, and using data files is different in OSS vs. internally.  So I'd rather avoid it.

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
dfawley: 1m seems reasonable to me, but this should be a cross-language decision.  Note that the frequency of distribution is not the same as the tolerance for staleness.

security/advancedtls/crl_provider.go
dfawley: It's not exported.

erm-g: Done

security/advancedtls/internal/testutils/testutils.go
erm-g: Moved everything under existing `testdata/crl` directory, each file has a `provider_` prefix, merged Readme files

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
erm-g: At the end of the day - yes, gRPC code is the caller. Provider is a new part of [RevocationConfig struct](https://github.com/gtcooke94/grpc-go/blob/340757d28e4713ac7dab71cb45376584b6ca67db/security/advancedtls/crl.go#L59), which is used by [checkCert function](https://github.com/gtcooke94/grpc-go/blob/340757d28e4713ac7dab71cb45376584b6ca67db/security/advancedtls/crl.go#L313) and underlying [fetchCRL](https://github.com/gtcooke94/grpc-go/blob/340757d28e4713ac7dab71cb45376584b6ca67db/security/advancedtls/crl.go#L294)
So if CRL is nil, fetchCRL throws an error which is handled by checkCert and it returns [RevocationUndetermined](https://github.com/gtcooke94/grpc-go/blob/340757d28e4713ac7dab71cb45376584b6ca67db/security/advancedtls/crl.go#L322)
I think the documentation you're looking for is present on checkCert [level](https://github.com/gtcooke94/grpc-go/blob/340757d28e4713ac7dab71cb45376584b6ca67db/security/advancedtls/crl.go#L310) 
Do you want to keep it as is or to add few lines to the CRLProvider interface?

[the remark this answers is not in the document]
dfawley: > I think the documentation you're looking for is present on checkCert level

That is a downstream comment on an internal function.

If we expect users to implement it, we should document it fully here such that they'll be able to understand what happens for all the possible return values.  The godoc comments should be comprehensive, or if it's just too complicated, refer them to a full document that contains all the information they need as a user of the API (not the gRFC).

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl.go
```diff
@@ -83,13 +88,46 @@ func (s RevocationStatus) String() string {
...
+	}
+	return crlExt, nil
+}
+
+// ReadCRLFile reads a file from the provided path, and returns constructed CRL
+// struct from it.
+func ReadCRLFile(path string) (*CRL, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("readFile(%v) failed err = %v", path, err)
```
dfawley, line 124: `os.ReadFile`?

Also I don't think this is the right format for errors in Go.  This is more the style for test failures, as it appears to be suited for programmers and not users.  Errors would usually look more like:

```go
return nil, fmt.Errorf("failed to read %q: %v", path, err)
```

(or `%w` instead of `%v` if you want users to be able to unwrap the read error, and then this becomes part of your API.)

You can see examples from the stdlib:

https://github.com/search?q=repo%3Agolang%2Fgo%20fmt.Errorf&type=code

security/advancedtls/crl_provider.go
```diff
@@ -0,0 +1,238 @@
...
+// NewFileWatcherCRLProvider returns a new instance of the
+// FileWatcherCRLProvider. It uses FileWatcherOptions to validate and apply
+// configuration required for creating a new instance. Users should call Close
+// to stop the background refresh of CRLDirectory.
+func NewFileWatcherCRLProvider(o FileWatcherOptions) (*FileWatcherCRLProvider, error) {
+	if err := o.validate(); err != nil {
+		return nil, err
+	}
+	done := make(chan struct{})
+	stop := make(chan struct{})
```
dfawley, line 119: Nit: delete these locals and initialize them in place.

security/advancedtls/crl_provider.go
```diff
@@ -129,9 +129,9 @@ func (o *FileWatcherOptions) validate() error {
 		return fmt.Errorf("advancedtls: CRLDirectory %v is not readable: %v", o.CRLDirectory, err)
 	}
 	// Checks related to RefreshDuration.
-	if o.RefreshDuration < time.Second {
+	if o.RefreshDuration < time.Minute {
 		o.RefreshDuration = defaultCRLRefreshDuration
-		grpclogLogger.Warningf("RefreshDuration must larger then 1 second: provided value %v, default value will be used %v", o.RefreshDuration, defaultCRLRefreshDuration)
+		grpclogLogger.Warningf("RefreshDuration must be larger than 1 minute: provided value %v, default value %v will be used.", o.RefreshDuration, defaultCRLRefreshDuration)
```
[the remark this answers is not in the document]
erm-g: Done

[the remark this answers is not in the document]
erm-g: Yeah that's a good idea - I split into 2 separate cases to enable slightly different log messages

security/advancedtls/crl_provider.go
```diff
@@ -39,7 +39,7 @@ const defaultCRLRefreshDuration = 1 * time.Hour
 // establishment, so implementations of the CRL function need to be fast, and
 // slow things such as file IO should be done asynchronously.
 //
-// [A69: CRL Enhancements]: https://github.com/grpc/proposal/pull/382
+// [gRFC A69]: https://github.com/grpc/proposal/blob/dddf32d0116376dd0c48adee7b0071a20bc82b5b/A69-crl-enhancements.md
```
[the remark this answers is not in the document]
erm-g: I don't think so. This is how it was previously handled for AuditLogger -
https://github.com/grpc/grpc-go/commit/1f73ed5fcf75b3e464c83ab17fc73144e2161620

So originally a link to PR itself was used, and then it was changed to the permanent .md file

security/advancedtls/crl.go
dfawley: That original line was about what you were printing.  `readFile()` wasn't the thing that failed (and it doesn't exist); it was `os.ReadFile` that failed.  So it would be better to use the proper name of the thing that failed and not a function that doesn't exist.  But it's even better to not use debugging-style messages as errors, so then I suggested that.

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
dfawley: >> Will this link remain if the owner of the fork deletes their branch?
> I don't think so.

If that's the case, then please don't use an ephemeral link for a godoc comment.  The old `.../pull/382` link seemed fine?

security/advancedtls/crl_provider.go
```diff
@@ -136,9 +135,13 @@ func (o *FileWatcherOptions) validate() error {
 		return fmt.Errorf("advancedtls: CRLDirectory %v is not readable: %v", o.CRLDirectory, err)
 	}
 	// Checks related to RefreshDuration.
-	if o.RefreshDuration < time.Minute {
+	if o.RefreshDuration <= 0 {
+		grpclogLogger.Warningf("RefreshDuration is not set or negative: provided value %v, default value %v will be used.", o.RefreshDuration, defaultCRLRefreshDuration)
```
dfawley, line 139: It's probably not a warning to use the default value...

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl.go
erm-g: Oh I see, sorry - GitHub always highlights few lines so I thought there might be a better than `os.ReadFile` function to use here.

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_deprecated.go
erm-g: Done

security/advancedtls/testdata/crl/provider/README.md
erm-g: Removed 'provider' directory - https://github.com/grpc/grpc-go/pull/6670#discussion_r1367408760

security/advancedtls/testdata/crl/provider/filewatcher/.gitignore
erm-g: Cleaned up the directory as well - https://github.com/grpc/grpc-go/pull/6670#discussion_r1370812412

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider.go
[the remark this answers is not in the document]
erm-g: Done - 1 minute will be used in C as well

security/advancedtls/crl_provider.go
```diff
@@ -136,7 +136,7 @@ func (o *FileWatcherOptions) validate() error {
 	}
 	// Checks related to RefreshDuration.
 	if o.RefreshDuration <= 0 {
-		grpclogLogger.Warningf("RefreshDuration is not set or negative: provided value %v, default value %v will be used.", o.RefreshDuration, defaultCRLRefreshDuration)
+		grpclogLogger.Infof("RefreshDuration is not set or negative: provided value %v, default value %v will be used.", o.RefreshDuration, defaultCRLRefreshDuration)
```
dfawley, line 139: Actually I don't think this necessarily needs to be logged at all?  Especially in the case where you're just using the default, anything being logged would be spam.  Maybe if you had a negative number?  But who would even do such a thing?  IMO:

```go
if dur == 0 {
	// use default, silently
} else {
	// use 1m and info or warning that we're using 1m instead of the illegal value they used
}
```

See https://github.com/grpc/grpc-go/blob/master/Documentation/log_levels.md for some guidance we try to follow regarding this.

dfawley (commented):
LGTM except the last nit.  And also investigating the race test failure.

security/advancedtls/crl_provider.go
erm-g: Done

security/advancedtls/crl_provider_test.go
```diff
@@ -191,9 +197,11 @@ func (s) TestFileWatcherCRLProvider(t *testing.T) {
 			}
 		})
 	}
+	nonCRLMutex.Lock()
```
dfawley, line 200: Does calling `p.Close` here, before the check, not remove the need for the mutex?  If not, why not?  I would think that would eliminate the concurrency concerns.

erm-g: Yeah that's a very good point - current `p.Close` impl waits for gouroutines to end. I tested locally and `-race` is green

security/advancedtls/crl_provider_test.go
```diff
@@ -137,9 +138,14 @@ func (s) TestFileWatcherCRLProviderConfig(t *testing.T) {
 // that it’s correctly processed. Additionally, we also check if number of
 // invocations of custom callback is correct.
 func (s) TestFileWatcherCRLProvider(t *testing.T) {
+	t.Parallel()
 	const nonCRLFilesUnderCRLDirectory = 5
 	nonCRLFilesSet := make(map[string]struct{})
+	//var callbackMutex sync.Mutex
```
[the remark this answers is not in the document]
erm-g: Done

security/advancedtls/crl_provider_test.go
```diff
@@ -141,7 +141,6 @@ func (s) TestFileWatcherCRLProvider(t *testing.T) {
 	t.Parallel()
 	const nonCRLFilesUnderCRLDirectory = 5
 	nonCRLFilesSet := make(map[string]struct{})
-	//var callbackMutex sync.Mutex
 	var nonCRLMutex sync.Mutex
```
dfawley, line 144: So can you delete this now?

security/advancedtls/crl_provider_test.go
```diff
@@ -258,6 +264,8 @@ func (s) TestFileWatcherCRLProviderDirectoryScan(t *testing.T) {
 		t.Run(tt.desc, func(t *testing.T) {
 			copyFiles(sourcePath, targetPath, tt.fileNames, t)
 			p.ScanCRLDirectory()
+			p.scanMutex.Lock()
+			defer p.scanMutex.Unlock()
```
dfawley, line 268: Anything similar here?  We really shouldn't need mutexes while testing conditions, or else they might be flaky if we lose a race.

erm-g: Here I'm not sure - I need the same instance of Provider for all the table tests, and I need to make the check `cmp.Diff(len(p.crls), tt.expectedEntries)` before the next test case scan, that's why I use mutex from `p.ScanCRLDirectory`

security/advancedtls/crl_provider_test.go
erm-g: Yeah, there is no need to protect this local to the test map anymore. We do the read once provider's goroutine is done

security/advancedtls/crl_provider_test.go
dfawley: Oh, I see, you're taking the mutex from the thing you're testing.

Can you call `p.CRL()` instead of directly accessing `p.crls` then?

erm-g: I think it's doable - I can provide a list of filenames to be scanned (`[]string{"2.crl", "3.crl"}` and list of CRLs to be found in the map by calling `p.CRL()` 
Some downside - the current approach ensures the map contains 'exactly x' entries (`len(p.crls)`) but the new one will be about 'at least x' entries. Is that ok?

dfawley: Oh right, `CRL()` takes a parameter and you'd need to call it multiple times.

This is OK, then, though I'd prefer stricter validation, i.e. checking the contents of the map, or at least the keys, and not just the length.

dfawley (approved):
@gtcooke94 might want to take another pass after all the changes, but LGTM for me.

security/advancedtls/crl_provider_test.go
erm-g: I did the refactoring to the stricter validation and I think it looks better now (plus no more calling mutexes inside tests)

security/advancedtls/crl_provider_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+// TestFileWatcherCRLProviderDirectoryScan tests how FileWatcherCRLProvider
+// handles different contents of FileWatcherOptions.CRLDirectory.
+// We update the content with various (correct and incorrect) CRL files and
+// check if in-memory storage was properly updated. Please note that the same
+// instance of FileWatcherCRLProvider is used for the whole test so test cases
+// are not independent from each other.
+func (s) TestFileWatcherCRLProviderDirectoryScan(t *testing.T) {
+	sourcePath := testdata.Path("crl")
+	targetPath := createTmpDir(t)
```
rockspore, line 210: Shouldn't we remove this directory once the tests are finished? Or are all the Go versions we support recent enough for us to use https://godoc.corp.google.com/pkg/testing#T.TempDir?

erm-g: Good point, added the 'remove dir' approach.

PR #6676, opened 2023-10-03
github: add code coverage with codecov.io

arvindbr8:
Fixes: #1676 

Codecov Reach graph - https://app.codecov.io/github/grpc/grpc-go/tree/code_coverage

RELEASE NOTES: none


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,27 @@
...
+      with:
+        go-version: 'stable'
+    - name: Run coverage
+      run: |
+        # See -coverprofile format here. https://cs.opensource.google/go/x/tools/+/refs/tags/v0.13.0:cover/profile.go;l=55-58;drc=81efdbcac4736176ac97c60577b0069f76414c44
+        echo "mode: set" > ${GITHUB_WORKSPACE}/coverage.out
+        append_coverage() {
+          [ -f $1 ] && (sed '1d' $1 >> ${GITHUB_WORKSPACE}/coverage.out)
+        }
+        go test -coverprofile=cov.out -covermode=set google.golang.org/grpc/...
```
dfawley, line 18: I think to fix the cumulative coverage problem, we can just add `-coverpkg=./...` here?

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,27 @@
...
+        append_coverage cov.out
+        for MOD_FILE in $(find . -name 'go.mod' | grep -Ev '^\./go\.mod'); do
```
dfawley, line 20: Can we remove the `| grep -v` and not special case the root directory above?

arvindbr8:
Discussed offline. I've made a few changes since the last push. 

We had a cumulative problem of `go test --cover` only covering unit tests. It did not include coverage from e2e style tests. For this adding `coverpkg=./...` worked which includes coverage of tests 

Since most of the changes are to the main module, it makes sense to limit the scope of presubmit check to only run `go test --cover` for the main module.

Ignored the following paths in codecov.
  - "test/"
  - "benchmark/"
  - "internal/testutils"
  - ".*.pb.go"

dfawley (approved):

PR #6679, opened 2023-10-03
resolver: Add an endpoint map type

zasweq:
This PR adds an endpoint map type for use by Petiole policies for pick first children and also balancers like Outlier Detection which store state unique to individual endpoints. The endpoints are unique based off the unordered set of address string counts within an endpoint.

RELEASE NOTES:
* resolver: Add an endpoint map type

### DIFF
--- a/resolver/map.go
+++ b/resolver/map.go
@@ -230,9 +230,9 @@ func (em *EndpointMap) Values() []any {
 	return ret
 }
 
-// find returns the pointer to endpoint node in the EndpointMap endpoints map if
-// the endpoint node is already present. If not found, nil is returned. The
-// comparisons are done on the unordered set of addresses within an endpoint.
+// find returns a pointer to the endpoint node in em if the endpoint node is
+// already present. If not found, nil is returned. The comparisons are done on
+// the unordered set of addresses within an endpoint.
 func (em EndpointMap) find(e endpointNode) *endpointNode {
 	for endpoint := range em.endpoints {
 		if e.Equal(endpoint) {
--- a/resolver/map.go
+++ b/resolver/map.go
@@ -137,8 +137,6 @@ func (a *AddressMap) Values() []any {
 	return ret
 }
 
-type endpointNodeList []endpointNode
-
 type endpointNode struct {
 	addrs map[string]int
 	val   any
@@ -161,10 +159,7 @@ func (en endpointNode) Equal(en2 endpointNode) bool {
 func toEndpointNode(endpoint Endpoint) endpointNode {
 	en := make(map[string]int)
 	for _, addr := range endpoint.Addresses {
-		if count, ok := en[addr.Addr]; ok {
-			en[addr.Addr] = count + 1
-		}
-		en[addr.Addr] = 1
+		en[addr.Addr]++
 	}
 	return endpointNode{
 		addrs: en,
@@ -172,26 +167,23 @@ func toEndpointNode(endpoint Endpoint) endpointNode {
 }
 
 // EndpointMap is a map of endpoints to arbitrary values keyed on only the
-// unordered set of addresses string within an endpoint. The reason this uses a
-// slice instead of a map is because slices or maps cannot be used as map keys.
-// Thus, for efficiency store as a map and compare to others so each node
-// comparison is o(n), where n is the number of unique addresses in the endpoint
-// rather than having to sort a list on each comparison which is n(log(n))
-// (cannot use slices for the latter case or maps for the former case as map
-// keys, so needs to be a list).
+// unordered set of address strings within an endpoint. This map is not thread
+// safe, cannot access multiple times concurrently. The zero value for an
+// EndpointMap is an empty EndpointMap ready to use.
 type EndpointMap struct {
-	endpoints endpointNodeList // Cannot use maps as a map key. Thus, keep a list, and o(n) search. Not on RPC fast path so efficiency not as important.
-}
-
-// NewEndpointMap returns a new EndpointMap.
-func NewEndpointMap() *EndpointMap {
-	return &EndpointMap{}
+	// Use a slice instead of a map is because slices or maps cannot be used as
+	// map keys. Thus, for efficiency store as a map and compare to others so
+	// each node comparison is o(n), where n is the number of unique addresses
+	// in the endpoint rather than having to sort a list on each comparison
+	// which is n(log(n)) (cannot use slices for the latter case or maps for the
+	// former case as map keys, so needs to be a list).
+	endpoints []endpointNode // Cannot use maps as a map key. Thus, keep a list, and o(n) search. Not on RPC fast path so efficiency not as important.
 }
 
 // Get returns the value for the address in the map, if present.
 func (em *EndpointMap) Get(e Endpoint) (value any, ok bool) {
 	en := toEndpointNode(e)
-	if i := em.endpoints.find(en); i != -1 {
+	if i := em.find(en); i != -1 {
 		return em.endpoints[i].val, true
 	}
 	return nil, false
@@ -200,7 +192,7 @@ func (em *EndpointMap) Get(e Endpoint) (value any, ok bool) {
 // Set updates or adds the value to the address in the map.
 func (em *EndpointMap) Set(e Endpoint, value any) {
 	en := toEndpointNode(e)
-	if i := em.endpoints.find(en); i != -1 {
+	if i := em.find(en); i != -1 {
 		em.endpoints[i].val = value
 		return
 	}
@@ -242,11 +234,11 @@ func (em *EndpointMap) Values() []any {
 	return ret
 }
 
-// find returns the index of endpoint in the endpointNodeList slice, or -1 if
-// not present. The comparisons are done on the unordered set of the addresses
-// within endpoint.
-func (l endpointNodeList) find(e endpointNode) int {
-	for i, endpoint := range l {
+// find returns the index of endpoint in the EndpointMap endpoints slice, or -1
+// if not present. The comparisons are done on the unordered set of the
+// addresses within an endpoint.
+func (em EndpointMap) find(e endpointNode) int {
+	for i, endpoint := range em.endpoints {
 		if endpoint.Equal(e) {
 			return i
 		}
@@ -257,7 +249,7 @@ func (l endpointNodeList) find(e endpointNode) int {
 // Delete removes the specified endpoint from the map.
 func (em *EndpointMap) Delete(e Endpoint) {
 	en := toEndpointNode(e)
-	entry := em.endpoints.find(en)
+	entry := em.find(en)
 	if entry == -1 {
 		return
 	}
--- a/resolver/map.go
+++ b/resolver/map.go
@@ -167,8 +167,8 @@ func toEndpointNode(endpoint Endpoint) endpointNode {
 
 // EndpointMap is a map of endpoints to arbitrary values keyed on only the
 // unordered set of address strings within an endpoint. This map is not thread
-// safe, cannot access multiple times concurrently. The zero value for an
-// EndpointMap is an empty EndpointMap ready to use.
+// safe, thus it is unsafe to access concurrently. Must be created via
+// NewEndpointMap; do not construct directly.
 type EndpointMap struct {
 	endpoints map[*endpointNode]any
 }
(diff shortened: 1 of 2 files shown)


resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+
+type endpointNodeList []endpointNode
+
+type endpointNode struct {
+	addrs map[string]int
+	val   any
+}
+
+// Equal returns whether the unordered set of addrs counts are the same between
+// the endpoint nodes.
```
dfawley, line 148: Why use the counts?  I don't think an endpoint should include the same address multiple times, and if it does, it seems like we'd just ignore the duplicates.

resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+		en[addr.Addr] = 1
+	}
+	return endpointNode{
+		addrs: en,
+	}
+}
+
+// EndpointMap is a map of endpoints to arbitrary values keyed on only the
+// unordered set of addresses string within an endpoint. The reason this uses a
+// slice instead of a map is because slices or maps cannot be used as map keys.
```
dfawley, line 176: External documentation should not include implementation details or rational for implementation details unless it's important to the user of the type.

resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+		}
+		en[addr.Addr] = 1
+	}
+	return endpointNode{
+		addrs: en,
+	}
+}
+
+// EndpointMap is a map of endpoints to arbitrary values keyed on only the
+// unordered set of addresses string within an endpoint. The reason this uses a
```
dfawley, line 175: Nit: "address stringS"

resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+	for _, en := range em.endpoints {
+		ret = append(ret, en.val)
+	}
+	return ret
+}
+
+// find returns the index of endpoint in the endpointNodeList slice, or -1 if
+// not present. The comparisons are done on the unordered set of the addresses
+// within endpoint.
+func (l endpointNodeList) find(e endpointNode) int {
```
dfawley, line 248: Suggestion: remove this type as it obfuscates that it's a slice.  Implement this method on `EndpointMap` instead.

resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+// (cannot use slices for the latter case or maps for the former case as map
+// keys, so needs to be a list).
+type EndpointMap struct {
+	endpoints endpointNodeList // Cannot use maps as a map key. Thus, keep a list, and o(n) search. Not on RPC fast path so efficiency not as important.
+}
+
+// NewEndpointMap returns a new EndpointMap.
+func NewEndpointMap() *EndpointMap {
+	return &EndpointMap{}
+}
```
dfawley, line 189: Generally if the zero value of the struct is safe to use, we'd just avoid having a constructor and say the zero value is safe to use as an empty map.  E.g. https://pkg.go.dev/bytes@go1.21.1#Buffer

The exception is if you think it might need an initialization later, especially if it's an API that we can't ever change.

resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+	return true
+}
+
+func toEndpointNode(endpoint Endpoint) endpointNode {
+	en := make(map[string]int)
+	for _, addr := range endpoint.Addresses {
+		if count, ok := en[addr.Addr]; ok {
+			en[addr.Addr] = count + 1
+		}
+		en[addr.Addr] = 1
```
dfawley, line 167: `en[addr.Addr]++`

resolver/map.go
```diff
@@ -136,3 +136,134 @@ func (a *AddressMap) Values() []any {
...
+	en := toEndpointNode(e)
+	entry := em.endpoints.find(en)
+	if entry == -1 {
+		return
+	}
+	if len(em.endpoints) == 1 {
+		em.endpoints = nil
+	}
+	copy(em.endpoints[entry:], em.endpoints[entry+1:])
+	em.endpoints = em.endpoints[:len(em.endpoints)-1]
```
dfawley, line 268: Another option would be to make `endpoints` a `map[*endpointNode]struct{}`.  `find` would return the `*endpointNode` if the endpoint can be found.  `Delete` can be a simple `delete(endpoints, entry)`.

resolver/map.go
zasweq: I use the counts because it feels like a defensive way with no cost incured. I.e. I either do a presence check with a struct{} or bool or I have the value be an int. I know it's not an intended use case of an endpoint to have the same address (by string) multiple times but I added this since it seems to be similar to add it vs. presence.

resolver/map.go
zasweq: Ah, good catch. Switched.

resolver/map.go
zasweq: Interesting, switched.

resolver/map.go
zasweq: Ah, ok noted. Moved the comment to the field (which is unexported and users of this package won't see it).

resolver/map.go
zasweq: Hmmmm good idea. Switched.

resolver/map.go
zasweq: The issue I see with using a map of unordered addresses in a map itself is it's not allowed (similar to how you exclude attributes in your address map, since map[key] underlying using something similar to cmp.Equal, which panics on a map key, as you mentioned offline). Thus, I made it a list, with o(n) operations not on the RPC path, so I can do comparisons on each node. I don't think this would work with a map, unless you can overwrite the _, ok := map[key] operation, which I have not have found a way to do.

resolver/map.go
zasweq: As per comment below, chose not to go map route so deleted this constructor. If we do need to change this, it's internal and we won't break users :).

resolver/map.go
dfawley: Maybe you weren't understanding my suggestion.  Pointers are valid for map keys, and are compared using pointer equality:

https://go.dev/play/p/VWhAQUEueKR

We wouldn't do lookups using the key, though, even though a map was used.  But this simplifies deletions to: 1. find the entry to be deleted by scanning all the entries; 2. `delete(endpoints, entryPtr)`

resolver/map.go
dfawley: This is not internal.  This is in the `grpc/resolver` package, which is experimental and intended to be accessed by users, but subject to change. But breaking changes will still break users.

resolver/map.go
zasweq: Ah, I see. Switched to that.

resolver/map.go
```diff
@@ -136,3 +136,120 @@ func (a *AddressMap) Values() []any {
...
+		en[addr.Addr]++
+	}
+	return endpointNode{
+		addrs: en,
+	}
+}
+
+// EndpointMap is a map of endpoints to arbitrary values keyed on only the
+// unordered set of address strings within an endpoint. This map is not thread
+// safe, cannot access multiple times concurrently. The zero value for an
```
dfawley, line 170: "This map is not thread safe, cannot access multiple times concurrently."

Please fix the grammar.

resolver/map.go
```diff
@@ -136,3 +136,120 @@ func (a *AddressMap) Values() []any {
...
+// EndpointMap is an empty EndpointMap ready to use.
```
dfawley, line 171: The zero value is no longer safe to use, so change comment to say to use NewEndpointMap to create.

resolver/map.go
dfawley: Ping. Let's make this a set, which is how the docstring below indicates it behaves as well.

resolver/map.go
```diff
@@ -136,3 +136,120 @@ func (a *AddressMap) Values() []any {
...
+// Len returns the number of entries in the map.
+func (em *EndpointMap) Len() int {
+	return len(em.endpoints)
+}
+
+// Keys returns a slice of all current map keys, as endpoints specifying the
+// addresses present in the endpoint keys, in which uniqueness is determined by
+// the unordered set of addresses. Thus, endpoint information returned is not
+// the full endpoint data but can be used for EndpointMap accesses.
+func (em *EndpointMap) Keys() []Endpoint { // TODO: Persist the whole endpoint data to return in nodes? No use case now, but one could come up in future.
```
dfawley, line 211: Are you sure we don't already need it?  You can't create/update a child policy without the full endpoint data.

resolver/map.go
```diff
@@ -136,3 +136,120 @@ func (a *AddressMap) Values() []any {
...
+				})
+			}
+		}
+		ret = append(ret, endpoint)
+	}
+	return ret
+}
+
+// Values returns a slice of all current map values.
+func (em *EndpointMap) Values() []any {
```
dfawley, line 228: Is there a use for this?  I don't think AddressMap.Values is even used, and I don't see where it would be useful.

resolver/map.go
```diff
@@ -136,3 +136,120 @@ func (a *AddressMap) Values() []any {
...
+	ret := make([]any, 0, len(em.endpoints))
+	for _, val := range em.endpoints {
+		ret = append(ret, val)
+	}
+	return ret
+}
+
+// find returns the pointer to endpoint node in the EndpointMap endpoints map if
+// the endpoint node is already present. If not found, nil is returned. The
```
dfawley, line 237: `find returns a pointer to the endpoint node`?

resolver/map.go
```diff
@@ -136,3 +136,120 @@ func (a *AddressMap) Values() []any {
...
+		}
+	}
+	return nil
+}
+
+// Delete removes the specified endpoint from the map.
+func (em *EndpointMap) Delete(e Endpoint) {
+	en := toEndpointNode(e)
+	entry := em.find(en)
+	if entry != nil {
```
dfawley, line 252: Merge lines and reduce variable scope: `if entry := em.find(en); entry != nil {`

resolver/map.go
zasweq: Sorry, I don't get what is wrong with this statement.

resolver/map.go
zasweq: Alright sounds good. Done.

resolver/map.go
zasweq: Yeah. This usage of Keys() for resolver map is simply a helper to index back into the map: https://github.com/grpc/grpc-go/blob/b0e55408084062ac4120518f8a6cb1a224c66f27/examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go#L139 (this is how AddressMap.Keys() is used as well, to build a set and use to immediately access map. So thus, this comment is there because we do not persist any attributes to return in this function, and simply build out a new endpoint with the unordered set of addresses to return. Also, the number of addresses will be lost here if I implement your count suggestion above.

zasweq: https://github.com/grpc/grpc-go/blob/master/balancer/base/balancer.go#L124

resolver/map.go
zasweq: Ah, ok sounds good.

resolver/map.go
zasweq: This is still technically a set, right? I don't see any downside to doing it this way, and I think it is more technically correct to take into account address counts (still unordered). Technically, what do you see as the pro of AABC == ABC vs AABC != ABC? I feel like it makes more sense that they're not equivalent (and this case will never happen in practice anyway). It allows the resolver to encode more information in the endpoint (250 ms delay fail address a, then try address a again)...hmmm. I don't think happy eyeballs makes sense wrt doing aa, but right now unless we use keys() to pass down to child the duplicated addresses in an endpoint will be conveyed to the pick first child. So I think whether we account for address counts here is orthogonal to whether we want to pass down AABC to pick first child and have it do happy eyeballs algorithm on AA (relates to that discussion we had monday about pick first never reporting TF).

zasweq: I put this in the dualstack design chat. Will wait for more discussion before choosing to implement this suggestion or not.

resolver/map.go
zasweq: Ah whoops yeah forgot to change this. Switched comment to what you said for address map switched to endpoint map: "Must be created via NewEndpointMap; do not construct directly."

resolver/map.go
zasweq: Mark said to strip counts, so will implement this suggestion.

resolver/map.go
zasweq: Deleted this TODO and added information to top level docstring.

resolver/map.go
zasweq: Done.

dfawley (approved):

PR #6685, opened 2023-10-04
Explicitly specify the `dns` schema for the ALTS hand-shaker.

mjamaloney:
Before this change applications that override the default resolver may not be able to talk to the metadata server to start the ALTS Handshake, resulting in DirectPath not being used.


mjamaloney:
I just read over CLA guidance, I need to resend from my @google.com.

PR #6691, opened 2023-10-05
examples: Add custom load balancer example

zasweq:
Contains #6679. This PR adds a custom load balancer example, deployed as the top level balancer of the channel through a service config. This custom load balancer is a petiole policy wrapping pick first, which is the intended way we want users to write their own custom load balancers, as users get features such as Sticky Transient Failure, Health Checking, and the Happy Eyeballs algorithm out of the box.

As part of this PR, pick_first is changed to work on the endpoints list and handle the endpoints list as defined in A61. This change works within system because the Channel is already emitting a Slice of Endpoints.

A helper interface between the petiole policy and the base pick first is also added. This lets petiole policy focus on the business logic, as seen by the simplicitly of custom_round_robin.

RELEASE NOTES:
* examples: Add custom lb example


ginayeh:
Please rebase your comment before merging.

dfawley (changes requested):
I have a lot of comments.  Maybe we should just simplify this for now and make something small that doesn't need to worry about being a petiole policy just so we can get it done with a reasonable amount of effort.

examples/features/customloadbalancer/README.md
```diff
@@ -0,0 +1,57 @@
+# Custom Load Balancer
+
+This examples shows how to deploy a custom load balancer in a `ClientConn`.
```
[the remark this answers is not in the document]
zasweq: Done.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+	balancer.Register(customRoundRobinBuilder{})
+}
+
+const customRRName = "custom_round_robin"
+
+type customRRConfig struct {
+	serviceconfig.LoadBalancingConfig `json:"-"`
+
+	// N represents how often pick iterations chose the second SubConn in the
+	// list. Defaults to 3. If 0 never choses second SubConn.
```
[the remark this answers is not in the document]
zasweq: Done. "choose" is a strange word in the English language haha.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+		pfs:              resolver.NewEndpointMap(),
+		pickFirstBuilder: pfBuilder,
+	}
+}
+
+var logger = grpclog.Component("example")
+
+type customRoundRobin struct {
+	// All state and operations on this balancer are either initialized at build
+	// time and read only after, or are only accessed as part of it's
```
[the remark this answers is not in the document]
zasweq: Ah, whoops. Done.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+		pickFirst := ep.(balancer.Balancer)
+		// pick first was removed by resolver (unique endpoint logically
+		// corresponding to pick first child was removed).
+		if _, ok := endpointSet.Get(e); !ok {
+			pickFirst.Close()
+			crr.pfs.Delete(e)
+		}
+	}
+	crr.inhibitPickerUpdates = false
+	crr.regeneratePicker() // one synchronous picker update per Update Client Conn State operation.
```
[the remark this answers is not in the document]
zasweq: Done.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+	pfs              *resolver.EndpointMap
+
+	n                    uint32
+	inhibitPickerUpdates bool
+}
+
+func (crr *customRoundRobin) UpdateClientConnState(state balancer.ClientConnState) error {
+	if logger.V(2) {
+		logger.Info("custom_round_robin: got new ClientConn state: ", state)
+	}
```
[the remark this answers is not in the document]
zasweq: Deleted this one. Left the error log from UpdateSubConnState since it signals to the reader that that is deprecated and shouldn't be called. Should I leave that one too?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+	bOpts balancer.BuildOptions
+	// Note that this balancer is a petiole policy which wraps pick first (see
+	// gRFC A61). This is the intended way a user written custom lb should be
+	// specified, as pick first will contain a lot of useful functionality, such
+	// as Sticky Transient Failure, Happy Eyeballs, and Health Checking.
+	pickFirstBuilder balancer.Builder
+	pfs              *resolver.EndpointMap
+
+	n                    uint32
+	inhibitPickerUpdates bool
```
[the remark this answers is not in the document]
zasweq: Done.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+	cc    balancer.ClientConn
+	bOpts balancer.BuildOptions
+	// Note that this balancer is a petiole policy which wraps pick first (see
+	// gRFC A61). This is the intended way a user written custom lb should be
+	// specified, as pick first will contain a lot of useful functionality, such
+	// as Sticky Transient Failure, Happy Eyeballs, and Health Checking.
+	pickFirstBuilder balancer.Builder
+	pfs              *resolver.EndpointMap
+
+	n                    uint32
```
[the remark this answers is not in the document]
zasweq: Done. Switched to persist a customRRConfig.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	crr.n = crrCfg.N
+
+	endpointSet := resolver.NewEndpointMap()
+	crr.inhibitPickerUpdates = true
+	for _, endpoint := range state.ResolverState.Endpoints {
+		endpointSet.Set(endpoint, nil)
+		var pickFirst *balancerWrapper
+		pf, ok := crr.pfs.Get(endpoint)
+		if ok {
```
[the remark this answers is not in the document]
zasweq: Ah, good point. Done.

pickfirst.go
```diff
@@ -125,22 +123,51 @@ func (b *pickfirstBalancer) UpdateClientConnState(state balancer.ClientConnState
...
-		grpcrand.Shuffle(len(addrs), func(i, j int) { addrs[i], addrs[j] = addrs[j], addrs[i] })
-	}
 
 	if b.logger.V(2) {
 		b.logger.Infof("Received new config %s, resolver state %s", pretty.ToJSON(cfg), pretty.ToJSON(state.ResolverState))
 	}
 
+	var addrs []resolver.Address
+	if len(state.ResolverState.Endpoints) != 0 {
+		endpoints := state.ResolverState.Endpoints
```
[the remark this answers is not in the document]
zasweq: Ah, good point. Done.

pickfirst.go
```diff
@@ -125,22 +123,51 @@ func (b *pickfirstBalancer) UpdateClientConnState(state balancer.ClientConnState
...
+			grpcrand.Shuffle(len(endpoints), func(i, j int) { endpoints[i], endpoints[j] = endpoints[j], endpoints[i] })
+		}
+
+		// "Flatten the list by concatenating the ordered list of addresses for each
+		// of the endpoints, in order." - A61
+		for _, endpoint := range endpoints {
+			// "In the flattened list, interleave addresses from the two address
+			// families, as per RFC-8304 section 4." - A61
+			// This language is handled by this iteration through endpoints, as ipv4
+			// and ipv6 are specified as part of the same endpoint.
```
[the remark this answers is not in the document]
zasweq: I see what you're saying. How would this be communicated to the balancer, through an attribute of some kind? Or do we expect this balancer to parse the number of bits of the IP address and determine?

pickfirst.go
```diff
@@ -125,22 +123,51 @@ func (b *pickfirstBalancer) UpdateClientConnState(state balancer.ClientConnState
...
+			addrs = append(addrs, endpoint.Addresses...)
+		}
+		// Endpoints not set, process addresses until migrate resolver emissions
+		// fully to Endpoints. The top channel does wrap emitted addresses with
+		// endpoints, however some balancers such as weighted target do not forwarrd
+		// the corresponding correct endpoints down/split endpoints properly. Once
+		// all balancers correctly forward endpoints down, can delete this else
+		// conditional.
```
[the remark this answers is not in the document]
zasweq: Ah, good point. Switched.

examples/features/customloadbalancer/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+A client is created, to connect to both of these servers (they get both
+server addresses from the name resolver in two separate endpoints). The client is configured with the load
+balancer specified in the service config, which in this case is custom_round_robin.
+
+### custom_round_robin
+
+The client is configured to use `custom_round_robin`. `custom_round_robin` is a petiole policy,
+which creates a pick first child for every endpoint it receives. It waits until both pick first children
+become ready, then defers to the first pick first child's picker, choosing the connection to localhost:20000, except
```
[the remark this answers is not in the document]
zasweq: Done.

examples/features/customloadbalancer/README.md
```diff
@@ -0,0 +1,57 @@
...
+serving address in the response. So the server on "localhost:20001" will reply to the RPC
+with `this is examples/customloadbalancing (from localhost:20001)`.
+
+A client is created, to connect to both of these servers (they get both
+server addresses from the name resolver in two separate endpoints). The client is configured with the load
+balancer specified in the service config, which in this case is custom_round_robin.
+
+### custom_round_robin
+
+The client is configured to use `custom_round_robin`. `custom_round_robin` is a petiole policy,
```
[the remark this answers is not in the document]
zasweq: Removed language.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,249 @@
...
+func init() {
+	balancer.Register(customRoundRobinBuilder{})
+}
+
+const customRRName = "custom_round_robin"
+
+type customRRConfig struct {
+	serviceconfig.LoadBalancingConfig `json:"-"`
+
+	// N represents how often pick iterations chose the second SubConn in the
```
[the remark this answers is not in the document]
zasweq: Switched to "chooseSecond".

zasweq (commented):
Got to all comments. Thanks for the pass.

zasweq:
Added BalancerAggregator utility we discussed offline :). This is ready to review now.

internal/balancer/balanceraggregator/balanceraggregator.go
dfawley, line 1: This needs to be available outside `internal` for users to use it.

internal/balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,168 @@
...
+		var bal *balancerWrapper
+		if child, ok := ba.children.Get(endpoint); ok {
+			bal = child.(*balancerWrapper)
+		} else {
+			bal = &balancerWrapper{
+				endpoint:   endpoint,
+				ClientConn: ba.parent,
+				ba:         ba,
+			}
+			bal.Balancer = gracefulswitch.NewBalancer(bal, ba.bOpts)
```
dfawley, line 82: Pull into the above?

internal/balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,168 @@
...
+type BalancerAggregator struct {
+	parent Parent
+	bOpts  balancer.BuildOptions
+
+	children *resolver.EndpointMap
+
+	inhibitChildUpdates bool
+}
+
+func (ba *BalancerAggregator) UpdateClientConnState(state balancer.ClientConnState) error {
```
dfawley, line 67: These methods will need comments.

internal/balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,168 @@
...
+	for _, e := range ba.children.Keys() {
+		child, _ := ba.children.Get(e)
+		bal := child.(balancer.Balancer)
+		if _, ok := endpointSet.Get(e); !ok {
+			bal.Close()
+			ba.children.Delete(e)
+		}
+	}
+	ba.inhibitChildUpdates = false
+	ba.updateChildStates()
```
dfawley, line 105: Maybe `defer` these (in `ResolverError` too)?

Otherwise note the bug that exists when an error occurs and you return early.

internal/balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,168 @@
...
+		bal.Close()
+	}
+}
+
+func (ba *BalancerAggregator) updateChildStates() {
+	if ba.inhibitChildUpdates {
+		return
+	}
+
+	childUpdates := make([]ChildState, ba.children.Len())
```
dfawley, line 135: I think you want `make([]ChildState, 0, ba.children.Len())` otherwise you end up with a bunch of empty children.

internal/balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,168 @@
...
+	for _, child := range ba.children.Values() {
+		bw := child.(*balancerWrapper)
+		childUpdates = append(childUpdates, ChildState{
+			Endpoint: bw.endpoint,
+			State:    bw.state,
+		})
```
dfawley, line 141: Maybe store a ChildState instead of the fields separately?

pickfirst.go
```diff
@@ -107,22 +106,49 @@ func (b *pickfirstBalancer) UpdateClientConnState(state balancer.ClientConnState
...
+		// "Flatten the list by concatenating the ordered list of addresses for each
+		// of the endpoints, in order." - A61
+		for _, endpoint := range endpoints {
+			// "In the flattened list, interleave addresses from the two address
+			// families, as per RFC-8304 section 4." - A61
+			// TODO: support the above language.
+			addrs = append(addrs, endpoint.Addresses...)
+		}
+	} else {
+		// Endpoints not set, process addresses until migrate resolver emissions
```
dfawley, line 139: until "we" migrate ... ?  Or "until all resolvers produce Endpoints <and the Addresses field is removed>"?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,178 @@
...
+	// SubConn in the list. Defaults to 3. If 0 never choose the second SubConn.
+	ChooseSecond uint32 `json:"chooseSecond,omitempty"`
+}
+
+type customRoundRobinBuilder struct{}
+
+func (customRoundRobinBuilder) ParseConfig(s json.RawMessage) (serviceconfig.LoadBalancingConfig, error) {
+	// Hardcode a pick first with no shuffling, since this is a petiole, and
+	// that is what petiole policies will interact with.
+	gspf, err := balanceraggregator.ParseConfig(json.RawMessage(balanceraggregator.PickFirstConfig))
```
dfawley, line 57: Oh..since this is a const you can actually do it at `init()` time, and it doesn't need to be in the `customRRConfig` either.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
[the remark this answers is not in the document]
dfawley: Remove it now, though, since it isn't needed?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,178 @@
...
+	crr.cfg = crrCfg
+	return crr.balancerAggregator.UpdateClientConnState(balancer.ClientConnState{
+		BalancerConfig: crrCfg.ChildPolicy,
+		ResolverState:  state.ResolverState,
+	})
+}
+
+func (crr *customRoundRobin) ResolverError(err error) {
+	crr.balancerAggregator.ResolverError(err)
+}
```
dfawley, line 122: Should this embed the BalancerAggregator instead?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,178 @@
...
+// This function is deprecated. SubConn state updates now come through listener
+// callbacks. This balancer does not deal with SubConns directly and has no need
+// to intercept listener callbacks.
+func (crr *customRoundRobin) UpdateSubConnState(sc balancer.SubConn, state balancer.SubConnState) {
+	logger.Errorf("custom_round_robin: UpdateSubConnState(%v, %+v) called unexpectedly", sc, state)
+}
+
+func (crr *customRoundRobin) Close() {
+	crr.balancerAggregator.Close()
+}
```
dfawley, line 133: These, too....

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,178 @@
...
+	inhibitPickerUpdates bool
+}
+
+func (crr *customRoundRobin) UpdateClientConnState(state balancer.ClientConnState) error {
+	crrCfg, ok := state.BalancerConfig.(*customRRConfig)
+	if !ok {
+		return balancer.ErrBadResolverState
+	}
+	crr.cfg = crrCfg
+	return crr.balancerAggregator.UpdateClientConnState(balancer.ClientConnState{
```
dfawley, line 114: Comment?

```go
// A call to UpdateClientConnState should always produce a new Picker.  That
// is guaranteed to happen since the aggregator will always call UpdateChildState
// in its UpdateClientConnState.
```

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+
+	// For determinism, this balancer only updates it's picker when both
+	// backends of the example are ready. Thus, no need to keep track of
+	// aggregated state and can simply specify this balancer is READY once it
+	// has two ready children. Other balancers can keep track of aggregated
+	// state and interact with errors as part of picker.
+	if len(readyPickers) != 2 {
+		return
```
dfawley, line 151: This needs to send a Picker.

I think we need the aggregated state here, too.

zasweq: Agreed for most petiole policies we want to incorporate errors/reflect most recent state to clientconn, but don't you want the picker in this example to be deterministic (i.e. two ready SubConn's which give you the pattern in the README?)

internal/balancer/balanceraggregator/balanceraggregator.go
zasweq: I don't think you can do that, since the graceful switch needs the ref to the bal, so I declare it first and then set this field.

internal/balancer/balanceraggregator/balanceraggregator.go
zasweq: Hmm right I don't set it to false. Added defer in both operations. Thinking about it though, this question applies to the endpoint map too. If I return early, I'm midway through adding (and haven't yet deleted the children that need to be cleaned up), and thus have this weird interspliced state where the children are partially updated and no longer viable ones no longer stick around. Any thoughts here?

internal/balancer/balanceraggregator/balanceraggregator.go
zasweq: Ah right haha. I think this was getting covered up with my for iteration in custom_round_robin (just checking for two ready pickers, so ignoring first two empty ones).

internal/balancer/balanceraggregator/balanceraggregator.go
zasweq: Done. Feels a little weird since I set 50% of it in constructor and the rest in UpdateState, but it cleans this line up, and I guess it's doing that here anyway, just adds a nesting layer to the data structure that aligns to what it'll send upward anyway. Thanks.

pickfirst.go
zasweq: I think I meant we migrate. That seems more blanket term than resolvers producing Endpoints (I think you have to migrate more of the codebase in the future).

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Done. I thought you didn't like globals though? Maybe since it applies in this case since it's a constant scoped to this balancer package. All Petioles will use this, so maybe a top level var in every petiole package?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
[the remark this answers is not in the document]
zasweq: Ah, whoops, forgot to delete. Deleted.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Good point. Done.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Switched to Embedded. Good point.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Sure, added.

internal/balancer/balanceraggregator/balanceraggregator.go
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+	balancer.Balancer   // Simply forward balancer.Balancer operations.
+	balancer.ClientConn // embed to intercept UpdateState, doesn't deal with SubConns
+
+	ba *BalancerAggregator
+
+	childState ChildState
+}
+
+func (bw *balancerWrapper) UpdateState(state balancer.State) {
+	bw.childState.State = state
```
dfawley, line 256: This is racy with updateState.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package balanceraggregator implements a BalancerAggregator helper.
+package balanceraggregator
```
dfawley, line 20: Bikeshedding is needed, I think.

How about `endpointmanager'?  `endpointshard[ing]`?

Also let's mark this as experimental for now.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+	"google.golang.org/grpc/balancer/base"
+	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/internal/balancer/gracefulswitch"
+	"google.golang.org/grpc/internal/grpcrand"
+	"google.golang.org/grpc/resolver"
+	"google.golang.org/grpc/serviceconfig"
+)
+
+// ChildState is the balancer state of the child along with the
+// endpoint which ID's the child balancer.
```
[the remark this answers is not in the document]
zasweq: Ah good point about each vs a. Done for both.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+func Build(cc balancer.ClientConn, opts balancer.BuildOptions) *BalancerAggregator {
+	return &BalancerAggregator{
+		cc:       cc,
+		bOpts:    opts,
+		children: resolver.NewEndpointMap(),
+	}
+}
+
+// BalancerAggregator is a balancer that wraps child balancers. It creates a
+// child balancer with child config for every Endpoint received. It updates the
```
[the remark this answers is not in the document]
zasweq: Good point. Added unique clause.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+// child states on any update from parent or child.
+type BalancerAggregator struct {
+	cc    balancer.ClientConn
+	bOpts balancer.BuildOptions
+
+	children *resolver.EndpointMap
+
+	inhibitChildUpdates bool
```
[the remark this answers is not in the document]
zasweq: Added: // inhibitChildUpdates is set during UpdateClientConnState/ResolverError
	// calls (calls to children will each produce an update, only want one
	// update).

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// UpdateClientConnState creates a child for new endpoints, deletes children for
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+// endpoints that are no longer present. It also updates all the children, and
+// sends a single synchronous update of the children's aggregate state at the
+// end of the UpdateClientConnState operation. If an endpoint has no addresses,
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+// returns error. Otherwise returns first error found from a child, but fully
+// processes the new update.
+func (ba *BalancerAggregator) UpdateClientConnState(state balancer.ClientConnState) error {
+	if len(state.ResolverState.Endpoints) == 0 {
+		return errors.New("endpoints list is empty")
+	}
+
+	// Check/return early if any endpoints have no addresses.
+	// TODO: eventually make this configurable.
```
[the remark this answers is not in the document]
zasweq: Good point, done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+	for _, endpoint := range state.ResolverState.Endpoints {
+		if len(endpoint.Addresses) == 0 {
+			return errors.New("endpoint has empty addresses")
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+	endpointSet := resolver.NewEndpointMap()
+	ba.inhibitChildUpdates = true
+	defer func() {
+		ba.inhibitChildUpdates = false
+		ba.updateState()
+	}()
+	var ret error
+	// Update/Create new children.
+	for _, endpoint := range state.ResolverState.Endpoints {
+		endpointSet.Set(endpoint, nil)
```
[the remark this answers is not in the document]
zasweq: Hmm good point...although thinking about this, you could get a endpoint {a, b, c} with attribute x, and endpoint {b, c, a} with attribute y...what do we want the behavior to be here (edge case that I don't think should come out of any of our maintained resolvers). The code as written uses the last positional endpoint to determine the attributes. This change makes it use the first positional endpoint to determine the attributes.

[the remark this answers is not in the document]
zasweq: I changed it to what you had to avoid duplicate updates, but question here about attributes still stands.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+		if _, ok := endpointSet.Get(e); !ok {
+			bal.Close()
+			ba.children.Delete(e)
+		}
+	}
+	return ret
+}
+
+// ResolverError forwards the resolver error to all of the BalancerAggregator's
+// children, sends a single synchronous update of the childStates at the end of
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (ba *BalancerAggregator) Close() {
+	for _, child := range ba.children.Values() {
+		bal := child.(balancer.Balancer)
+		bal.Close()
+	}
+}
+
+// updateState updates this components state. It sends the aggregated state, and
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (p *pickerWithChildStates) Pick(info balancer.PickInfo) (balancer.PickResult, error) {
+	nextIndex := atomic.AddUint32(&p.next, 1)
+	picker := p.pickers[nextIndex%uint32(len(p.pickers))]
+	return picker.Pick(info)
+}
+
+// ChildStatesFromPicker returns the persisted child states from the picker if
+// picker is of type pickerWithChildStates.
```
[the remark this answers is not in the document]
zasweq: Done, added.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Added:
//
// # Experimental
//
// Notice: This package is EXPERIMENTAL and may be changed or removed in a
// later release.

The more terse comment I saw was "// All APIs in this package are experimental." Let me know if you would prefer that instead.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,265 @@
...
+	for _, child := range ba.children.Values() {
+		bal := child.(balancer.Balancer)
+		bal.Close()
+	}
+}
+
+// updateState updates this components state. It sends the aggregated state, and
+// a picker with round robin behavior with all the child states present if
+// needed.
+func (ba *BalancerAggregator) updateState() {
```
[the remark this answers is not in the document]
zasweq: Added a lock on the child, and hold it in UpdateState() write, and updateState() read. I think no deadlocks occur, and I think this is correct.

The reason I did it per child is to avoid unnecessary blocking since the read is scoped to a child. This works if the goal is eventual consistency thinking through the scenarios.

Say you get an UpdateCCS for endpoint A, B, writes current state, calls update state (lock not held) at the end. endpoint B sends a new update before UpdateState() on parent. First Update will be endpointAState, and endpointBNewState, and then updateState() from endpointB update will trigger a resend. An extra duplicate update I think from previous experience is acceptable, and not technically incorrect. The initial singular updateState() from UpdateCCS() could get a new update from a child after calling updateState(), but this feels fine.

In the case of getting numerous interleaved UpdateState() from numerous, the updateState() *after* unlocking in UpdateState() makes the last update trigger the updateState() operation which will eventually reflect the most recent update states to the parent.

Let me know what you think.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Responded with technical explanation for my fix on other comment. I made the race condition fix a separate commit so it's easier to see. I copied the whole childState into another local var on the stack, I could switch to just holding the lock and then reading from the child, but the readability made me feel like it's worth the extra copy/memory.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+type BalancerAggregator struct {
+	cc    balancer.ClientConn
+	bOpts balancer.BuildOptions
+
+	children *resolver.EndpointMap
+
+	// inhibitChildUpdates is set during UpdateClientConnState/ResolverError
+	// calls (calls to children will each produce an update, only want one
+	// update).
+	inhibitChildUpdates bool
```
dfawley, line 71: This also needs to be protected by a mutex, right?  It's set by calls coming in from the parent and read by concurrent updates from children.

dfawley: Or actually it's probably fine if it's atomic instead.

The reason we want to quell the updates from children during the update is that it's churn.  Having an async call come in from a child that has an update come out during the UpdateCCS call should be fine, unless it's hard to make the parent LB policy able to handle that (but I doubt that?).

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+		cc:       cc,
+		bOpts:    opts,
+		children: resolver.NewEndpointMap(),
+	}
+}
+
+// BalancerAggregator is a balancer that wraps child balancers. It creates a
+// child balancer with child config for every unique Endpoint received. It
+// updates the child states on any update from parent or child.
+type BalancerAggregator struct {
```
dfawley, line 62: Nit: unexport.  Optional: rename to `balancer`?

balancer/balanceraggregator/balanceraggregator.go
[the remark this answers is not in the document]
dfawley: This is what the EndpointMap is for, right?  If they have different attributes but the same addresses, then they should be treated as unique.  (This applies for address attributes at least.  Not sure about endpoint attributes, but that seems reasonable too, since it could affect the behavior of the LB policy that handles them.)

balancer/balanceraggregator/balanceraggregator.go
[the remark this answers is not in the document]
dfawley: There's far too many "update"s and "state"s in that for me to make sense of it, sorry.

I'm pretty sure we want to prevent multiple concurrent calls to the parent LB policy's `UpdateState` method.  Our channel implementation handles it just fine, but ideally we can make this serialize those calls to simplify the implementation of the parent LB policy.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Yeah I'll make it atomic. The update during UpdateCCS call is part of my middle scenario I wrote in my reply to your synchronization comment. Thanks.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+	next        uint32
+}
+
+func (p *pickerWithChildStates) Pick(info balancer.PickInfo) (balancer.PickResult, error) {
+	nextIndex := atomic.AddUint32(&p.next, 1)
+	picker := p.pickers[nextIndex%uint32(len(p.pickers))]
+	return picker.Pick(info)
+}
+
+// ChildStatesFromPicker returns the persisted child states from the picker if
```
dfawley, line 252: Nit: "persisted" seems to add no benefit to this sentence.  And "picker if picker" reads awkwardly.  How about just:

```go
// ChildStatesFromPicker returns the most recent state all children managed by the
// BalancerAggregator <name TBD> balancer that produced this picker.
```

balancer/balanceraggregator/balanceraggregator.go
dfawley: The terse comment is older, I think.  The bigger one formats more nicely in godoc.

balancer/balanceraggregator/balanceraggregator.go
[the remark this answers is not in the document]
dfawley: ...in which case we can probably just use the same lock to guard `inhibitChildUpdates` instead of worrying about making it atomic...

balancer/balanceraggregator/balanceraggregator.go
zasweq: Alright I'll leave it as is then.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Deleted persisted. added the i.e. picker if the picker. I think we need to leave the rest as is though, because your comment implies it returns the most recent state unconditionally, where it is conditional on the picker type.

balancer/balanceraggregator/balanceraggregator.go
zasweq: I renamed to balancer, but it collided with import. I'll unexport for now and once we figure out name we can rename to that (and continue unexport).

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	readyPickers := make([]balancer.Picker, 0)
+	connectingPickers := make([]balancer.Picker, 0)
+	idlePickers := make([]balancer.Picker, 0)
+	transientFailurePickers := make([]balancer.Picker, 0)
+
+	childStates := make([]ChildState, 0, ba.children.Len())
+	ba.mu.Lock()
+	defer ba.mu.Unlock()
+	for _, child := range ba.children.Values() {
```
dfawley, line 187: I think you still have races here.

You don't take this lock in UpdateClientConnState, which mutates `ba.children`, and now you're reading the same map here.

Let's have the mutex just protect `ba.children` and `ba.children.childState` and make sure we don't make any calls across levels with it held?

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+// and a picker with round robin behavior with all the child states present if
+// needed.
+func (ba *balancerAggregator) updateState() {
+	if ba.inhibitChildUpdates.Load() {
+		return
+	}
+	readyPickers := make([]balancer.Picker, 0)
+	connectingPickers := make([]balancer.Picker, 0)
+	idlePickers := make([]balancer.Picker, 0)
+	transientFailurePickers := make([]balancer.Picker, 0)
```
dfawley, line 182: The "0" here is not necessary.  Actually the `make` isn't, either.  You could try:

```go
var readyPickers, connectingPickers, idlePickers, transientFailurePickers []balancer.Picker
```

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+		pickers = connectingPickers
+	} else if len(idlePickers) >= 1 {
+		aggState = connectivity.Idle
+		pickers = idlePickers
+	} else if len(transientFailurePickers) >= 1 {
+		aggState = connectivity.TransientFailure
+		pickers = transientFailurePickers
+	} else {
+		aggState = connectivity.TransientFailure
+		pickers = append(pickers, base.NewErrPicker(errors.New("no children to pick from")))
```
dfawley, line 224: Remove the `append`?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,151 @@
...
+			if childState.State.ConnectivityState == connectivity.Ready {
+				readyPickers = append(readyPickers, childState.State.Picker)
+			}
+		}
+		// If both children are ready, pick using the custom round robin
+		// algorithm.
+		if len(readyPickers) == 2 {
+			picker := &customRoundRobinPicker{
+				pickers:      readyPickers,
+				chooseSecond: crr.cfg.ChooseSecond,
```
dfawley, line 123: This is also a race between setting and reading.  Atomic is probably fine?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -0,0 +1,151 @@
...
+func (crr *customRoundRobin) UpdateClientConnState(state balancer.ClientConnState) error {
+	crrCfg, ok := state.BalancerConfig.(*customRRConfig)
+	if !ok {
+		return balancer.ErrBadResolverState
+	}
+	crr.cfg = crrCfg
+	// A call to UpdateClientConnState should always produce a new Picker.  That
+	// is guaranteed to happen since the aggregator will always call
+	// UpdateChildState in its UpdateClientConnState.
+	return crr.Balancer.UpdateClientConnState(balancer.ClientConnState{
```
dfawley, line 103: Should we error unless we get 2 endpoints, instead of doing UpdateCCS on the child?

zasweq: Sure. Added.

balancer/balanceraggregator/balanceraggregator.go
zasweq: You could theoretically get more than one child in TF right? In the Java case that we discussed, it round robins over the error messages of the children in that case...with some intricacies as to whether the overall resolver error takes precedence or not.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Ok sounds good since you can len(nil) and append to nil. Thanks

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+	// calls (calls to children will each produce an update, only want one
+	// update).
+	inhibitChildUpdates atomic.Bool
+
+	mu sync.Mutex // Sync updateState callouts and childState recent state updates
+}
+
+// UpdateClientConnState creates a child for new endpoints and deletes children
+// for endpoints that are no longer present. It also updates all the children,
+// and sends a single synchronous update of the children's aggregate state at
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -0,0 +1,286 @@
...
+// the end of the UpdateClientConnState operation. If any endpoint has no
+// addresses, returns error. Otherwise returns first error found from a child,
```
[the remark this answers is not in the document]
zasweq: Added.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Yeah, it does, but only if the provided picker is the right type (it takes a picker interface, as that is the type provided to parent in UpdateState, not the concrete type of the picker. Otherwise there is nothing to look into for child states.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Oh ok so you don't like unnecessary globals *that are mutated*.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Sorry for all these synchronization correctness issues. I just took a look at Outlier Detection, and I guard the core data structures on updates from parent, updates from child, and the interval timer algorithm. I was initially worried about littering all the operations with mutexes but thinking about it these operations on the balancer type aren't on the RPC fast path, and don't happen at too high a frequency anyway where I think lock contention would be a problem.

zasweq: Let me think about weird behavioral splicing things that can happen here if I guard the children map in UpdateClientConnState.

zasweq: Fixed race as discussed offline.

balancer/balanceraggregator/balanceraggregator.go
dfawley: Isn't that what the above `pickers = transientFailurePickers` does?  This is with zero children, so there's only ever one, and pickers is set by exactly one of these if/else clauses.

balancer/balanceraggregator/balanceraggregator.go
```diff
@@ -49,11 +49,12 @@ type ChildState struct {
 
 // NewBalancer returns a new balancerAggregator.
 func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer {
-	return &balancerAggregator{
+	ba := &balancerAggregator{
 		cc:       cc,
 		bOpts:    opts,
-		children: resolver.NewEndpointMap(),
 	}
+	ba.children.Store(resolver.NewEndpointMap())
```
dfawley, line 56: https://github.com/golang/go/issues/63289 :(

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -88,15 +88,18 @@ type customRoundRobin struct {
...
+	cfg atomic.Pointer[customRRConfig]
 }
 
 func (crr *customRoundRobin) UpdateClientConnState(state balancer.ClientConnState) error {
 	crrCfg, ok := state.BalancerConfig.(*customRRConfig)
 	if !ok {
 		return balancer.ErrBadResolverState
 	}
-	crr.cfg = crrCfg
+	if el := len(state.ResolverState.Endpoints); el != 2 {
```
dfawley, line 99: Nit: slightly nicer is `if el := state.ResolverState.Endpoints; len(el) != 2 { ... }` then the error will include the actual list instead of just the size.

balancer/balanceraggregator/balanceraggregator.go
dfawley: Please add "managed by this balancer" or something like that here as discussed offline.

balancer/balanceraggregator/balanceraggregator.go
dfawley: "blah blah blah the EndpointSharding balancer blah blah" maybe?  Seems better than "EndpointShard" and "EndpointSharder"  I'm also okay with EndpointManager but I like what "sharding" implies better.  Also open to other ideas.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Ah :(. Yeah it felt weird having to declare an local var to store a pointer instead of inlining it at struct creation time.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Good point. Done.

balancer/balanceraggregator/balanceraggregator.go
zasweq: Switched comment to "ChildStatesFromPicker returns the child states from the picker that are managed by this balancer."

balancer/balanceraggregator/balanceraggregator.go
zasweq: Oh whoops for some reason my eyes went to the else if above. I see what you're saying now. Switched.

balancer/balanceraggregator/balanceraggregator.go
zasweq: I think EndpointsSharding or EndpointListSharding. EndpointSharding singular implies that you're sharding a single endpoint?

zasweq: Or even EndpointListBalancer, as C++ does. I'll switch it to EndpointSharding for the time being. I also kind of see that the Endpoint can be referring to what is the thing that is the output of the sharding (the atom). We can discuss this more, I'm fine with whatever.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -258,8 +258,8 @@ func (p *pickerWithChildStates) Pick(info balancer.PickInfo) (balancer.PickResul
 	return picker.Pick(info)
 }
 
-// ChildStatesFromPicker returns the child states from the picker if the picker
-// is one provided in a call to UpdateState from this balancer.
+// ChildStatesFromPicker returns the child states from the picker that are
+// managed by this balancer.
```
dfawley, line 262: This reads weirdly now.  "returns the child states from the picker that are ___" -- to me "that are" applies to "the picker" and so doesn't disagree.

Maybe "returns the state of all the children managed by the endpoint sharding balancer that created this picker" or something like that?

dfawley (commented):
Will you be adding to examples_test.sh in this PR?

balancer/endpointsharding/endpointsharding.go
```diff
@@ -16,13 +16,13 @@
  *
  */
 
-// Package balanceraggregator implements a balancerAggregator helper.
+// Package endpointsharding implements a endpointSharding helper.
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -47,20 +47,20 @@ type ChildState struct {
 	State    balancer.State
 }
 
-// NewBalancer returns a new balancerAggregator.
+// NewBalancer returns a new endpointSharding.
```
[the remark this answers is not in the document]
zasweq: Had the same description as your top level package comment: "NewBalancer returns a load balancing policy that manages homogenous child policies each owning a single endpoint."

balancer/endpointsharding/endpointsharding.go
zasweq: Switched to your suggestion.

zasweq:
Added to examples test as discussed offline.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -23,20 +23,16 @@ import (
...
 	"google.golang.org/grpc/serviceconfig"
 )
 
-var gracefulSwitchPickFirst serviceconfig.LoadBalancingConfig
-
 func init() {
 	balancer.Register(customRoundRobinBuilder{})
-	// Hardcode a pick first with no shuffling, since this is a petiole, and
-	// that is what petiole policies will interact with.
-	gracefulSwitchPickFirst, _ = endpointsharding.ParseConfig(json.RawMessage(endpointsharding.PickFirstConfig))
```
dfawley: Wasn't this better done here to avoid a potential runtime failure?  I think with the blank import of `grpc` it should be fixed.

Also this should `panic` (or `log.Fatal`) if it returns an error.

examples/examples_test.sh
```diff
@@ -110,7 +110,7 @@ declare -A EXPECTED_SERVER_OUTPUT=(
     ["features/authz"]="unary echoing message \"hello world\""
     ["features/cancellation"]="server: error receiving from stream: rpc error: code = Canceled desc = context canceled"
     ["features/compression"]="UnaryEcho called with message \"compress\""
-    ["features/customloadbalancer"]="serving on localhost:20000"
+    ["features/customloadbalancer"]="serving on localhost:50051"
```
dfawley, line 113: I would like to see an `EXPECTED_CLIENT_OUTPUT` entry for this example in this file or else we aren't actually testing it.  It also needs to be added to `EXAMPLES`, right?

dfawley: Oops I thought I was looking at the full diffs, but I was just looking at one of many recent commits.  Sorry.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: For some reason when I do this blank import it still registers (i.e. grpc package init) after this module gets registered. Everything depends on gRPC and not vice versa so I don't know why this init was hitting first but it was...I think it has something to do with setting a global var, which runs before other inits: https://yourbasic.org/golang/package-init-function-main-execution-order/ but it should initialize to just nil so I don't know what's happening here.

zasweq: Oh it fixed. I think the issue was this depended on balancer aggregator, which depended on graceful switch but no dependency on ...grpc, so this init would run pre grpc init since it wasn't down the dependency tree, and top level main hit grpc eventually in the depth first search.

balancer/balanceraggregator/balanceraggregator.go
dfawley: Yeah, it's "sharded _by endpoint_".

I like this name.  I'm also okay with `balancer/endpointlist.NewEndpointListBalancer`.  `EndpointSharded` could also work.

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -31,8 +31,11 @@ import (
 	"google.golang.org/grpc/serviceconfig"
 )
 
+var gracefulSwitchPickFirst serviceconfig.LoadBalancingConfig
+
 func init() {
 	balancer.Register(customRoundRobinBuilder{})
+	gracefulSwitchPickFirst, _ = endpointsharding.ParseConfig(json.RawMessage(endpointsharding.PickFirstConfig))
```
dfawley, line 38: Please add `err` capturing and make sure no error was returned, or `logger.Fatal` if so.

dfawley (commented):
Can you please add a simple test or 2 into the endpointsharding (or whatever we call it) package, please?

examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
zasweq: Done.

zasweq:
Added e2e test that tested the main behaviors I view endpoint sharding to have (a child per endpoint, default behavior of round robin over pickers in aggregated state, access to child picker in parent UpdateState).

dfawley (approved):
:tada: 

Now to rewrite round_robin :laughing:

zasweq:
And move pick first :).

PR #6701, opened 2023-10-06
codes: update docstring to indicate expected usage

dfawley:
Related to https://github.com/grpc/grpc/pull/34588; specifically document that users should use ONLY the values defined by us.

RELEASE NOTES:
* codes: clarify that only codes defined by this package are valid and that users should not cast other values to `codes.Code`

### DIFF
--- a/codes/codes.go
+++ b/codes/codes.go
@@ -25,7 +25,13 @@ import (
 	"strconv"
 )
 
-// A Code is an unsigned 32-bit error code as defined in the gRPC spec.
+// A Code is a status code according to the gRPC spec:
+//
+// https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+//
+// Only the codes defined as consts in this package are valid codes. Do not use
+// other code values.  Behavior and interopability are implementation-specific
+// and not guaranteed.
 type Code uint32
 
 const (
--- a/codes/codes.go
+++ b/codes/codes.go
@@ -25,13 +25,13 @@ import (
 	"strconv"
 )
 
-// A Code is a status code according to the gRPC spec:
-//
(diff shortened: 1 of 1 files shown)


codes/codes.go
```diff
@@ -25,7 +25,13 @@ import (
...
 )
 
-// A Code is an unsigned 32-bit error code as defined in the gRPC spec.
+// A Code is a status code according to the gRPC spec:
+//
+// https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+//
+// Only the codes defined as consts in this package are valid codes. Do not use
+// other code values.  Behavior and interopability are implementation-specific
+// and not guaranteed.
```
zasweq, line 34: Can you clarify the scope of the language of "behavior and interoperability" here? What is not guaranteed? Any permutation of languages with defined status codes? Only status codes undefined?

codes/codes.go
```diff
@@ -25,7 +25,13 @@ import (
 	"strconv"
 )
 
-// A Code is an unsigned 32-bit error code as defined in the gRPC spec.
+// A Code is a status code according to the gRPC spec:
+//
+// https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
```
zasweq, line 30: Easwar made me switch urls to a certain format: https://github.com/grpc/grpc-go/blob/master/xds/internal/balancer/wrrlocality/balancer.go#L19 (I think this is a strong heuristic of .go files). Can you please switch this?

zasweq (approved):
LGTM.

codes/codes.go
dfawley: Clarified.

codes/codes.go
dfawley: Done

PR #6702, opened 2023-10-08
Add LocalAddr to peer.Peer (#6577)

### LINKED ISSUE #6577: add LocalAddr to peer.Peer client side

We want to be able to log the local address in addition to the remote address from our server-side interceptors to aid in debugging.

### Proposed Solution

Add `LocalAddr net.Addr` to `peer.Peer`.

### Alternatives Considered

This information could also be made available via some other context attribute. It doesn't really matter as long as there is a way to get it.

### Additional Context

This was already added to the Java implementation several years ago. https://github.com/grpc/grpc-java/pull/4906

prestonvasquez:
Resolves #6577 , requesting the ability to log the local address in addition to the remote address for server-side interceptors for debugging. 

It isn't obvious to me if we want to also include the local address in the server handler, [here](https://github.com/grpc/grpc-go/blob/master/internal/transport/handler_server.go#L386) . This would require parsing the information form the http request, likely using [X-Forward-For ](https://en.wikipedia.org/wiki/X-Forwarded-For).

There may be the necessity for a followup ticket to extend the binarylog package to include the local address in gRPC binary logging. 

Hope this helps!

### DIFF
--- a/balancer/leastrequest/balancer_test.go
+++ b/balancer/leastrequest/balancer_test.go
@@ -161,14 +161,18 @@ func checkRoundRobinRPCs(ctx context.Context, client testgrpc.TestServiceClient,
 		gotAddrCount = make(map[string]int)
 		// Perform 3 iterations.
 		var iterations [][]string
+		var localIterations [][]string
 		for i := 0; i < 3; i++ {
 			iteration := make([]string, len(addrs))
+			localIteration := make([]string, len(addrs))
 			for c := 0; c < len(addrs); c++ {
 				var peer peer.Peer
 				client.EmptyCall(ctx, &testpb.Empty{}, grpc.Peer(&peer))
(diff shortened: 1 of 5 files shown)


balancer/leastrequest/balancer_test.go
```diff
@@ -161,14 +161,18 @@ func checkRoundRobinRPCs(ctx context.Context, client testgrpc.TestServiceClient,
 		gotAddrCount = make(map[string]int)
 		// Perform 3 iterations.
 		var iterations [][]string
+		var localIterations [][]string
```
zasweq, line 164: Please revert this file. We don't need to add these assertions here.

zasweq (changes requested):

zasweq:
Closing this because this is part of #6716.

PR #6704, opened 2023-10-10

dfawley:
RELEASE NOTES:
* server: prohibit more than MaxConcurrentStreams handlers from running at once


zasweq (approved):
LGTM.

PR #6706, opened 2023-10-10

dfawley:


zasweq (approved):
LGTM.

internal/transport/http2_server.go
```diff
@@ -171,15 +171,10 @@ func NewServerTransport(conn net.Conn, config *ServerConfig) (_ ServerTransport,
 		ID:  http2.SettingMaxFrameSize,
 		Val: http2MaxFrameLen,
 	}}
-	// TODO(zhaoq): Have a better way to signal "no limit" because 0 is
-	// permitted in the HTTP2 spec.
-	maxStreams := config.MaxStreams
-	if maxStreams == 0 {
-		maxStreams = math.MaxUint32
-	} else {
+	if config.MaxStreams != math.MaxUint32 {
```
ahrtr, line 174: Does it mean that it will NOT take effect any more if applications set MaxConcurrentStreams as `math.MaxUint32`?

Note the default value set by etcd is `math.MaxUint32`, reference:
- https://github.com/etcd-io/etcd/blob/a404ee5f6b8f0e3590b9abcc540d8813f09872aa/server/embed/serve.go#L264
- https://github.com/etcd-io/etcd/blob/a404ee5f6b8f0e3590b9abcc540d8813f09872aa/server/embed/config.go#L66

ahrtr: It seems OK, since the [default maxConcurrentStreams is math.MaxUint32](https://github.com/grpc/grpc-go/blob/32e3ef1ed1c494e3d748b93d67a0b7ec5d089a49/server.go#L176)

server.go
```diff
@@ -2075,3 +2073,32 @@ func validateSendCompressor(name, clientCompressors string) error {
...
+		<-q.wait
+	}
+}
+
+func (q *atomicSemaphore) release() {
+	// N.B. the "<= 0" check below should allow for this to work with multiple
+	// concurrent calls to acquire, but also note that with synchronous calls to
+	// acquire, as our system does, n will never be less than -1.  There are
+	// fairness issues (queuing) to consider if this was to be generalized.
+	if atomic.AddInt64(&q.n, -1) <= 0 {
```
bboreham, line 2096: Suspect this `-1` should be `1`.
Spotted by @colega

PR #6716, opened 2023-10-10
grpc: Move some stats handler calls to gRPC layer, and add local address to peer.Peer

zasweq:
This PR moves ConnBegin, ConnEnd, and InHeader from the transport layer to the gRPC layer. The other two callouts still in the transport, OutHeader and OutTrailer will require a reworking of the ownership of streaming code and logic with respect to functionality in gRPC layer and transport layer. Continuation of #6624.

RELEASE NOTES:
* peer: Add local address to peer.Peer


internal/transport/handler_server.go
```diff
@@ -371,16 +379,16 @@ func (ht *serverHandlerTransport) HandleStreams(startStream func(*Stream)) {
...
-		contentSubtype: ht.contentSubtype,
+		id:               0, // irrelevant
+		requestRead:      func(int) {},
+		cancel:           cancel,
+		buf:              newRecvBuffer(),
+		st:               ht,
+		method:           req.URL.Path,
+		recvCompress:     req.Header.Get("grpc-encoding"),
+		contentSubtype:   ht.contentSubtype,
+		headerWireLength: 0, // doesn't know header wire length, will call into stats handler as 0.
```
dfawley, line 391: This one doesn't seem to be available.  I think we'll need:

https://github.com/golang/go/issues/18997

Please leave a comment in the code referencing the issue.

internal/transport/handler_server.go
```diff
@@ -371,16 +379,16 @@ func (ht *serverHandlerTransport) HandleStreams(startStream func(*Stream)) {
...
 	}
 	pr := &peer.Peer{
```
dfawley, line 393: Reuse `Peer()`?

internal/transport/handler_server.go
```diff
@@ -347,7 +355,7 @@ func (ht *serverHandlerTransport) WriteHeader(s *Stream, md metadata.MD) error {
 	return err
 }
 
-func (ht *serverHandlerTransport) HandleStreams(startStream func(*Stream)) {
+func (ht *serverHandlerTransport) HandleStreams(_ context.Context, startStream func(*Stream)) {
```
dfawley, line 358: Doesn't ignoring the context mess everything up?

Can grpc not pass the `http.Request.Context()` (with things like Peer added) so that it doesn't need to be ignored here?

internal/transport/transport.go
```diff
@@ -698,7 +712,7 @@ type ClientTransport interface {
 // Write methods for a given Stream will be called serially.
 type ServerTransport interface {
 	// HandleStreams receives incoming streams using the given handler.
```
dfawley, line 714: "the Context given is used as the base context for all streams started on this transport."?

internal/transport/transport.go
```diff
@@ -717,9 +731,15 @@ type ServerTransport interface {
...
 	Close(err error)
 
+	// LocalAddr returns the local network address.
+	LocalAddr() net.Addr
+
 	// RemoteAddr returns the remote network address.
 	RemoteAddr() net.Addr
 
+	// Peer returns the peer of the server transport.
+	Peer() *peer.Peer
```
dfawley, line 741: Should we delete Local/RemoteAddr and only have this?

server.go
```diff
@@ -971,12 +971,42 @@ func (s *Server) newHTTP2Transport(c net.Conn) transport.ServerTransport {
 	return st
 }
 
-func (s *Server) serveStreams(st transport.ServerTransport) {
-	defer st.Close(errors.New("finished serving streams for the server transport"))
-	var wg sync.WaitGroup
+type connectionKey struct{}
+
+// GetConnection gets the connection from the context.
+func GetConnection(ctx context.Context) net.Conn {
```
dfawley, line 977: I don't think we want this to be public.

zasweq: Hmmm ok I'll plumb it through /internal good point

internal/transport/handler_server.go
```diff
@@ -167,6 +167,14 @@ func (ht *serverHandlerTransport) Close(err error) {
 
 func (ht *serverHandlerTransport) RemoteAddr() net.Addr { return strAddr(ht.req.RemoteAddr) }
 
+func (ht *serverHandlerTransport) LocalAddr() net.Addr { return nil } // Server Handler transport has no access to local addr (was simply not calling sh with local addr).
```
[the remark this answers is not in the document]
zasweq: Ah ok, done. Since you only want peer though will be added to the Peer() method and this will be deleted.

internal/transport/transport.go
zasweq: Sure, done. There's an in flight PR for this but the contributor isn't getting back to my comments so I'll go ahead and just add it to this PR.

internal/transport/handler_server.go
zasweq: Done. Thanks.

internal/transport/handler_server.go
zasweq: Ah, good point. Done.

internal/transport/handler_server.go
zasweq: It was being ignored in master already, and it uses the requests Context (ctx := ht.req.Context()). I think this is the same functionality, and I don't think I want to add something to gRPC layer using the http.Request since it's a server handler transport specific concept.

internal/transport/transport.go
zasweq: As per comment above, will wait until further discussion on the server handler transport to add this docstring.

zasweq (commented):
Thanks for the pass. Got to all comments.

arvindbr8:
@zasweq -- seems like there is a conflict. Adding you to the assignees to take care of that.

zasweq:
Ah thanks resolved

internal/internal.go
```diff
@@ -62,6 +62,8 @@ var (
 	// gRPC server. An xDS-enabled server needs to know what type of credentials
 	// is configured on the underlying gRPC server. This is set by server.go.
 	GetServerCredentials any // func (*grpc.Server) credentials.TransportCredentials
+	// GetConnection gets the connection from the context.
+	GetConnection any // func (context.Context) net.Conn
```
dfawley, line 66: It should be fine to give this a type to avoid the casts, since it's only things from the stdlib.

But, it should also be fine to implement it fully within the internal package.  But then.. why not just leave it where it is and export the setter?  Or find another place in internal for it?

internal/transport/handler_server.go
zasweq: Discussed offline; decided to move peer creation to construction time and read it in gRPC before calling this.

internal/transport/transport.go
zasweq: Decided to do this, added this comment.

internal/internal.go
zasweq: Thought about this, decided to move back to transport. I don't really think any of the current internal directories make semantical sense, and as per offline discussion I think there's complications about trying to keep file paths consistent.

dfawley (approved):
LGTM modulo the last comment.

internal/transport/handler_server.go
```diff
@@ -75,11 +75,25 @@ func NewServerHandlerTransport(w http.ResponseWriter, r *http.Request, stats []s
 		return nil, errors.New(msg)
 	}
 
+	var localAddr net.Addr
+	if la := r.Context().Value(http.LocalAddrContextKey); la != nil {
+		localAddr = la.(net.Addr)
```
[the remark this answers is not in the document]
zasweq: Done.

PR #6724, opened 2023-10-12
grpc: Add a pointer of server to ctx passed into stats handler

zasweq:
This PR adds a pointer to the grpc.Server into the context passed into the stats handler. It also adds an internal only helper on the server that returns whether a given method is registered or not. This will be used in OpenTelemetry instrumentation for security purposes with respect to the cardinality of the method attribute. Contains and is based on (and technically requires) https://github.com/grpc/grpc-go/pull/6716. The next step after these two PRs is to move the rest of the stats handler callouts to the gRPC layer, get registered methods plumbed client side, canoncalize Target() without breaking it and then the OpenTelemetry instrumentation itself :).

RELEASE NOTES: N/A

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -73,6 +73,11 @@ var (
 	// xDS-enabled server invokes this method on a grpc.Server when a particular
 	// listener moves to "not-serving" mode.
 	DrainServerTransports any // func(*grpc.Server, string)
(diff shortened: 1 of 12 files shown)


dfawley:
I believe this will need to be rebased when you merge #6716?

zasweq:
Opened a new PR #6750 :).

Issue #6667, opened 2023-09-28
Panic (nil pointer dereference) in processStreamingRPC


alexmt:
### What version of gRPC are you using?

v1.58.0

### What version of Go are you using (`go version`)?

go1.21.1

### What operating system (Linux, Windows, …) and version?

MacOS

### What did you do?

The example below demonstrates a simple gRPC server that returns an error which triggers nil pointer dereference in `processStreamingRPC`:

```golang
package main

import (
	"net"

	"google.golang.org/grpc"
	"google.golang.org/grpc/status"
)

type grpcstatus interface {
	GRPCStatus() *status.Status
	Error() string
}

type wrapper struct {
	grpcstatus
}

func main() {
	srv := grpc.NewServer(
		grpc.UnknownServiceHandler(func(srv interface{}, srcStream grpc.ServerStream) error {
			var fakeErr wrapper
			return &fakeErr
		}),
	)
	ln, err := net.Listen("tcp", "0.0.0.0:8080")
	if err != nil {
		panic(err)
	}
	if err := srv.Serve(ln); err != nil {
		panic(err)
	}
}
```

### What did you expect to see?

The server should return `codes.Unknown` error.

### What did you see instead?

Server crashes due to missing nil check in `google.golang.org/grpc/status.FromError` [method](https://github.com/grpc/grpc-go/blob/fd9ef7263ae34b82b6200fee951af1b2edcb16df/status/status.go#L101):

```golang
	type grpcstatus interface{ GRPCStatus() *Status }
	if gs, ok := err.(grpcstatus); ok {
		// gs still might be nil even if err is converted to grpcstatus
		grpcStatus := gs.GRPCStatus()
		...
```

dfawley:
Seems like we'd also need the same check here:

https://github.com/grpc/grpc-go/blob/dd9270d663c346b82341f7f51c4cded2bba5a653/status/status.go#L113-L115

nidhi-singh02:
I would like to fix this :)

dfawley:
Sounds good, please send a PR!

dfawley:
@nidhi-singh02 we unassigned you due to a lack of response / PR, but just let us know if you'd still like to work on this!

nidhi-singh02:
Hey, I saw label on this issue saying that clarity is required from the
reporter.  So I assumed we may need some inputs from the reporter. Is that
not the case ?

Yes, I would like to work on this.

dfawley:
[...]

PR #6730, opened 2023-10-15
status: fix panic on nil pointer dereference

JalajGoswami:
**Description:**
This pull request resolves issue #6667 "Panic (nil pointer dereference) in processStreamingRPC". This change ensures that the program does not crash due to a nil pointer dereference.

**Changes Made:**
- Added nil checks in FromError function of status package.

**Testing Done:**
- To test this change, I ran the status_test.go.

RELEASE NOTES: N/A

### DIFF
--- a/resolver/map.go
+++ b/resolver/map.go
@@ -136,3 +136,116 @@ func (a *AddressMap) Values() []any {
 	}
 	return ret
 }
+
+type endpointNode struct {
+	addrs map[string]struct{}
+}
+
+// Equal returns whether the unordered set of addrs are the same between the
+// endpoint nodes.
+func (en *endpointNode) Equal(en2 *endpointNode) bool {
+	if len(en.addrs) != len(en2.addrs) {
+		return false
+	}
+	for addr := range en.addrs {
+		if _, ok := en2.addrs[addr]; !ok {
+			return false
+		}
+	}
+	return true
+}
+
+func toEndpointNode(endpoint Endpoint) endpointNode {
+	en := make(map[string]struct{})
+	for _, addr := range endpoint.Addresses {
+		en[addr.Addr] = struct{}{}
+	}
+	return endpointNode{
+		addrs: en,
+	}
+}
+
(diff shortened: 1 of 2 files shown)


status/status.go
```diff
@@ -99,6 +99,11 @@ func FromError(err error) (s *Status, ok bool) {
 	}
 	type grpcstatus interface{ GRPCStatus() *Status }
 	if gs, ok := err.(grpcstatus); ok {
+		if gs == nil {
+			// gs might be nil even after succesful assertion
+			// returning error status
+			return New(codes.Unknown, err.Error()), false
+		}
 		grpcStatus := gs.GRPCStatus()
 		if grpcStatus == nil {
```
arvindbr8, line 108: I would squash this into a short circuit conditional check like this
```suggestion
		if gs == nil || gs.GRPCStatus() == nil {
```

status/status.go
```diff
@@ -111,6 +116,11 @@ func FromError(err error) (s *Status, ok bool) {
 	}
 	var gs grpcstatus
 	if errors.As(err, &gs) {
+		if gs == nil {
+			// gs might be nil even after succesful assertion
+			// returning error status
+			return New(codes.Unknown, err.Error()), false
+		}
 		grpcStatus := gs.GRPCStatus()
 		if grpcStatus == nil {
```
arvindbr8, line 125: Same here 
```suggestion
		if gs == nil || gs.GRPCStatus() == nil {
```

status/status.go
arvindbr8, line 1: If you go with my comment, `codecov` might not complain about diff coverage. However, I would recommend adding a condition that forces the condition where `gs == nil`

arvindbr8 (changes requested):

arvindbr8:
`Testing / tests: vet` failed due to the typo.

JalajGoswami:
Thanks @arvindbr8 for your help, I'm applying the changes you suggested & creating a fresh pr

PR #6733, opened 2023-10-17
status: fix issue grpc#6667 Panic on nil pointer dereference

JalajGoswami:
**Description:**
This pull request resolves issue grpc#6667 "Panic (nil pointer dereference) in processStreamingRPC". This change ensures that the program does not crash due to a nil pointer dereference.

**Changes Made:**
- Added nil checks in FromError function of status package.

**Testing Done:**
- To test this change, I ran the status_test.go.

RELEASE NOTES: N/A

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -99,7 +99,7 @@ func FromError(err error) (s *Status, ok bool) {
 	}
 	type grpcstatus interface{ GRPCStatus() *Status }
 	if gs, ok := err.(grpcstatus); ok && gs != nil {
-	if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
+		if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
 			return grpcStatus, true
 		}
 	} else if errors.As(err, &gs) && gs != nil {
--- a/status/status.go
+++ b/status/status.go
@@ -97,6 +97,13 @@ func FromError(err error) (s *Status, ok bool) {
 	if err == nil {
 		return nil, true
 	}
+	defer func() {
+		if r := recover(); r != nil {
+			// type assertion passes but GRPCStatus method is still nil
+			s = New(codes.Unknown, err.Error())
+			ok = false
+		}
+	}()
 	type grpcstatus interface{ GRPCStatus() *Status }
 	if gs, ok := err.(grpcstatus); ok && gs != nil {
 		if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
--- a/status/status.go
+++ b/status/status.go
@@ -98,31 +98,21 @@ func FromError(err error) (s *Status, ok bool) {
 		return nil, true
 	}
 	type grpcstatus interface{ GRPCStatus() *Status }
-	if gs, ok := err.(grpcstatus); ok {
-		if gs == nil || gs.GRPCStatus() == nil {
-			// Error has status nil, which maps to codes.OK. There
-			// is no sensible behavior for this, so we turn it into
-			// an error with codes.Unknown and discard the existing
-			// status.
-			return New(codes.Unknown, err.Error()), false
+	if gs, ok := err.(grpcstatus); ok && gs != nil {
+	if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
+			return grpcStatus, true
 		}
-		grpcStatus := gs.GRPCStatus()
-		return grpcStatus, true
-	}
-	var gs grpcstatus
-	if errors.As(err, &gs) {
-		if gs == nil || gs.GRPCStatus() == nil {
-			// Error wraps an error that has status nil, which maps
-			// to codes.OK.  There is no sensible behavior for this,
-			// so we turn it into an error with codes.Unknown and
-			// discard the existing status.
-			return New(codes.Unknown, err.Error()), false
+	} else if errors.As(err, &gs) && gs != nil {
+		if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
+			p := grpcStatus.Proto()
+			p.Message = err.Error()
+			return status.FromProto(p), true
 		}
-		grpcStatus := gs.GRPCStatus()
-		p := grpcStatus.Proto()
-		p.Message = err.Error()
-		return status.FromProto(p), true
 	}
+	// Error is or wraps an error that has status nil, which maps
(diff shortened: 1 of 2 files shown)


status/status.go
```diff
@@ -99,26 +99,26 @@ func FromError(err error) (s *Status, ok bool) {
 	}
 	type grpcstatus interface{ GRPCStatus() *Status }
 	if gs, ok := err.(grpcstatus); ok {
-		grpcStatus := gs.GRPCStatus()
-		if grpcStatus == nil {
+		if gs == nil || gs.GRPCStatus() == nil {
```
dfawley, line 102: Sorry to contradict the previous reviewer.  There was some effort taken in the past to avoid calling `gs.GRPCStatus()` multiple times.  Let's try to keep that condition true.  E.g.


```go
type grpcstatus interface { GRPCStatus() *Status }
var gs grpcstatus
if gs, ok := err.(grpcstatus); ok && gs != nil {
	if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
		return grpcStatus, true
	}
} else if errors.As(err, &gs) && gs != nil {
	if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
		p := grpcStatus.Proto()
		p.Message = err.Error()
		return status.FromProto(p), true
	}
}
// Error is or wraps an error that has status nil, which maps
// to codes.OK..................
return New(codes.Unknown, err.Error()), false
```

JalajGoswami: Thanks @dfawley for letting me know about the constraint for `GRPCStatus` , I will update the PR as per your suggestions

status/status.go
```diff
@@ -98,31 +98,21 @@ func FromError(err error) (s *Status, ok bool) {
...
-	if gs, ok := err.(grpcstatus); ok {
-		grpcStatus := gs.GRPCStatus()
-		if grpcStatus == nil {
-			// Error has status nil, which maps to codes.OK. There
-			// is no sensible behavior for this, so we turn it into
-			// an error with codes.Unknown and discard the existing
-			// status.
-			return New(codes.Unknown, err.Error()), false
+	if gs, ok := err.(grpcstatus); ok && gs != nil {
+	if grpcStatus := gs.GRPCStatus(); grpcStatus != nil {
```
dfawley, line 102: I believe the formatting here is wrong.

dfawley (approved):
LGTM, but please extend the tests around here to cover the case where the GRPCStatus() function itself isn't implemented:

https://github.com/grpc/grpc-go/blob/e14d5831b59b0f46853d076640ae85f964164b37/status/status_test.go#L216

Thanks!

JalajGoswami:
Sorry for the delay, actually I got sick that's why didn't able to commit sooner. Please check the Test & naming of test function and struct.

status/status_test.go
```diff
@@ -202,6 +202,21 @@ func (s) TestFromErrorWrapped(t *testing.T) {
 	}
 }
 
+type customErrorNotImplementsInterface struct {
```
dfawley, line 205: Sorry; I believe this is 100% equivalent to `TestFromErrorUnknownError`.


Looking back at my suggestion, I believe it is nonsensical.  I was thinking more like:

```go
type grpcStatus interface {
	GRPCStatus() *grpcstatus.Status
}

type customErrorIsNilImpl struct {
	grpcStatus // embed; don't implement, so the function is `nil`
}
```

But in this case, even this type assertion in the code should succeed _and_ be non-nil:

```go
	if gs, ok := err.(grpcstatus); ok && gs != nil {
```

It will then panic when it's called, but that's fine.  We're already very deep in the "the programmer did something very wrong" territory; we don't also need to detect and prevent panics because they've fully failed to correctly implement the interface.

So I think you can just delete this test case.

Thanks!

dfawley (commented):
> Sorry for the delay, actually I got sick that's why didn't able to commit sooner. Please check the Test & naming of test function and struct.

No worries at all.  Thanks for getting back to us & hope you are feeling better!

JalajGoswami:
I have removed the test

dfawley (dismissed):

status/status_test.go
dfawley: Wait.  Isn't this test case the same as the code in the bug report?  So does this fix the code in the bug report?

Please add a test case that fails without your changes but passes with them.  Thanks!

arvindbr8:
Just for clarity: @JalajGoswami -- please take a look at Doug's comment [here](https://github.com/grpc/grpc-go/pull/6733#discussion_r1377994047)

JalajGoswami:
Please check, I tried different things but only recover block worked, now this change will fix the issue of bug report i.e. it will not panic and will return status with Unknown code.
I have wrote the test with interface embedded but not implemented case (passed), also let me know if there are any trade-offs of using recover().

status/status.go
```diff
@@ -97,32 +97,29 @@ func FromError(err error) (s *Status, ok bool) {
...
 	type grpcstatus interface{ GRPCStatus() *Status }
-	if gs, ok := err.(grpcstatus); ok {
-		grpcStatus := gs.GRPCStatus()
-		if grpcStatus == nil {
-			// Error has status nil, which maps to codes.OK. There
-			// is no sensible behavior for this, so we turn it into
-			// an error with codes.Unknown and discard the existing
-			// status.
-			return New(codes.Unknown, err.Error()), false
+	if gs, ok := err.(grpcstatus); ok && gs != nil {
```
dfawley, line 108: Isn't the `gs != nil` unnecessary here now, then?

I'm not sure this actually catches anything.  So maybe we can keep the rest of the code as is was before and just add the defer/recover?

dfawley (commented):
After thinking about this further, I don't think we should make this change.

It looks like, in your production code, you created a special error type that embeds our interface but doesn't actually implement it.  I'm not fond of making code more complex than it needs to be to account for what amounts to an avoidable programmer error.  You should instead change your code so that this situation is not possible.

Note that it's still possible to make the code that you wrote here panic, by embedding `error` and leaving it unset, as `err.Error()` would panic at that point.

For a similar example to what's happening here, consider this, which makes a similar panic happen in the standard library:

https://go.dev/play/p/TMSPgFX3Ht8

```go
type someReader struct {
	io.Reader
}

func main() {
	a, err := io.ReadAll(&someReader{})
	fmt.Println(a, err)
}
```

PR #6736, opened 2023-10-18
move from github.com/golang/protobuf to google.golang.org/protobuf/proto

Clement-Jean:
This is the first step of solving the issue #5316

RELEASE NOTES: none


Clement-Jean:
@dfawley I'm not sure I understand the test failing. Could you guide me through it?

arvindbr8:
@Clement-Jean another push after `go mod tidy` should make `vet` happy. The codecov/patch is not a blocking check, so I wouldn't worry about that too much.

arvindbr8 (changes requested):
Thanks for the PR! One question tho: A quick grep showed the following as the only remaining non `.pb.go` usages of `github.com/golang/protobuf`. 

```
channelz/service/service_test.go
credentials/credentials.go
internal/pretty/pretty.go
```

Was the plan to remove the usages in a future PR? The first 2 in the list seem pretty small enough to add to this PR -- which makes us more confident to remove `protov1` from `pretty/pretty.go`

balancer/rls/config.go
```diff
@@ -308,5 +307,5 @@ func convertDuration(d *durationpb.Duration) (time.Duration, error) {
 	if d == nil {
 		return 0, nil
 	}
-	return ptypes.Duration(d)
+	return d.AsDuration(), d.CheckValid()
```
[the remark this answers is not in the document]
Clement-Jean: After checking the code (https://github.com/protocolbuffers/protobuf-go/blob/68463f0e96c93bc19ef36ccd3adfe690bfdb568c/types/known/durationpb/duration.pb.go#L200) and the documentation (https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/types/known/durationpb#Duration.CheckValid) it seems to do a check for a range within -10000 years to +10000 years inclusive.

xds/internal/xdsclient/transport/loadreport.go
```diff
@@ -170,11 +170,14 @@ func (t *Transport) recvFirstLoadStatsResponse(stream lrsStream) ([]string, time
 		t.logger.Infof("Received first LoadStatsResponse: %s", pretty.ToJSON(resp))
 	}
 
-	interval, err := ptypes.Duration(resp.GetLoadReportingInterval())
-	if err != nil {
+	rInterval := resp.GetLoadReportingInterval()
+
```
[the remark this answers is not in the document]
Clement-Jean: Is this the same as the following comment?

Clement-Jean:
For now, as per @dfawley, the plan is to keep service_test.go as is (issue #5316). This means that credentials.go cannot be updated, and I made the judgment call to keep pretty.go as is because of this. Hope that's ok.

Finally, `go mod tidy` doesn't seem to do anything so the vet.sh is still not working.

arvindbr8:
Sorry, i should have been more specific. `vet` failed in the `examples/` dir, so you'd have prolly have to run `go mod tidy` in the directory. I've run the command and pushed to our branch. 

You can also do `VET_SKIP_PROTO=1 ./vet.sh` to run vet in local. 

_____

@dfawley: Could you please take another look at the changes here?

arvindbr8 (approved):

channelz/service/func_linux.go
dfawley, line 1: Is it possible to revert this directory, please?  I'm making some pretty major changes and would like to avoid the conflict unless this is necessary for the rest of the PR.

internal/status/status.go
```diff
@@ -31,10 +31,10 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/ptypes"
 	spb "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/proto"
```
dfawley, line 36: I believe this causes an API change, right?  Now our status package only works with new messages, and breaks with the old ones?  IIUC we should be using this instead in `WithDetails` and `Details`: https://pkg.go.dev/google.golang.org/protobuf/protoadapt#MessageV1

internal/status/status.go
```diff
@@ -154,12 +154,12 @@ func (s *Status) Details() []any {
...
 	for _, any := range s.s.Details {
-		detail := &ptypes.DynamicAny{}
-		if err := ptypes.UnmarshalAny(any, detail); err != nil {
+		detail, err := any.UnmarshalNew()
+		if err != nil {
 			details = append(details, err)
 			continue
 		}
-		details = append(details, detail.Message)
+		details = append(details, detail)
```
dfawley, line 162: Are these new-style protos compatible with users running with `github.com/golang/proto` and `ptypes`?  If not, we can't make this part of this change.

Clement-Jean: How would I go about testing this? Should I write a `status_test.go` and check that both versions work?

Clement-Jean: For example, I wrote this:

```go
func (s) TestWithDetailsVersionsCompat(t *testing.T) {
	details := []protoreflect.ProtoMessage{
		ptypes.TimestampNow(),
		ptypes.DurationProto(1 * time.Hour),
		&timestamppb.Timestamp{},
		&durationpb.Duration{},
	}
	status := New(codes.InvalidArgument, "an error")

	if _, err := status.WithDetails(details...); err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
}
```

and it passes.

Clement-Jean: Not sure about how to test DynamicAny though, if you have any idea let me know.

Clement-Jean: @dfawley

dfawley: > How would I go about testing this? Should I write a status_test.go and check that both versions work?

Yes, that sounds good.

> For example, I wrote this [...] and it passes.

This doesn't look like what I was thinking (because it's testing `WithDetails`, not `Details`), but it's still a useful test.  What I had in mind was something that called `Details` on a status containing details, and does type assertions using protov1 messages:

```go
st := status.New().WithDetails(ptypes.TimestampNow(), timestamppb.Now()) // Create with old ptypes and new tspb
x := st.Details()[0]
if _, ok := x.(protov1.Message); !ok {
	t.Fatal()
}
if _, ok := x.(timestamppb.Timestamp); !ok {
	t.Fatalf()
}
```

I thought that `anypb.UnmarshalNew()` produced _new_ `protov2.Message`s which were incompatible with _old_ `protov1.Message`s.  But if the new `protov2.Message` is a superset of the old `protov1.Message` then I think this should be safe to change, especially because it's returned as `any` instead of `protov1.Message`.

internal/status/status.go
dfawley: @Clement-Jean This changes the API of WithDetails, because `proto.Message` is now different.

dfawley:
> @dfawley anything left to do?

Yes; pinged the thread above..

internal/status/status.go
Clement-Jean: did you see the tests ([status_test.go](https://github.com/Clement-Jean/grpc-go/blob/remove_old_proto_pkg/internal/status/status_test.go))? You mean that this isn't enough to prove the API didn't change? If that's not enough maybe I can revert the changes or add more tests to make sure it doesn't break

Clement-Jean:
@dfawley just checking. I replied in the comments, did you see?

dfawley:
@Clement-Jean do you still have this change?  I didn't see this comment in time.

Clement-Jean:
@dfawley yes, still have it

dfawley:
Great!  I'm unable to reopen this right now, because your fork was deleted.

dfawley:
I'll take a look tomorrow!

dfawley:
I'll take a look tomorrow!  (Sorry for the spam.)

dfawley:
Wait, something is wrong with github.. It let me click "reopen", but it's still showing closed with the same error.

Clement-Jean:
I think I accidentally removed the fork from Github, but I still have it locally. I'll clean it up and push it back, it should be ready tomorrow for you

dfawley:
@Clement-Jean It's still not letting me reopen this, though.. Would you mind creating a new PR, please?

PR #6699, opened 2023-10-06
alts: Reduce ALTS counter overflow length from 5 to 4.

matthewstevenson88:
This change is being made for consistency with the changes we're making to our internal implementation.

RELEASE NOTES: none

PR #6746, opened 2023-10-24
Revert "alts: Reduce ALTS counter overflow length from 5 to 4."

matthewstevenson88:
Reverts grpc/grpc-go#6699

This will break all ALTS Go connections as soon as they attempt to rekey.

RELEASE NOTES: none


arvindbr8:
@matthewstevenson88 -- could you add a note in the description for the reason for the revert?

arvindbr8 (approved):

PR #6749, opened 2023-10-25
envconfig: remove env vars for on-by-default features

dfawley:
Unfortunately the diffs are quite large, but it's very mechanical and should be quick to remove since it's all red code.

RELEASE NOTES: none

### DIFF
--- a/internal/testutils/xds/bootstrap/bootstrap.go
+++ b/internal/testutils/xds/bootstrap/bootstrap.go
@@ -116,12 +116,9 @@ func Contents(opts Options) ([]byte, error) {
 		cfg.XdsServers[0].ServerFeatures = append(cfg.XdsServers[0].ServerFeatures, "ignore_resource_deletion")
 	}
 
-	auths := make(map[string]authority)
-	if envconfig.XDSFederation {
-		// This will end up using the top-level server list for new style
-		// resources with empty authority.
-		auths[""] = authority{}
-	}
+	// This will end up using the top-level server list for new style
+	// resources with empty authority.
+	auths := map[string]authority{"": {}}
 	for n, auURI := range opts.Authorities {
 		auths[n] = authority{XdsServers: []server{{
 			ServerURI:      auURI,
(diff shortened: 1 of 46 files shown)


xds/internal/balancer/clusterresolver/configbuilder.go
```diff
@@ -98,47 +97,27 @@ func buildPriorityConfig(priorities []priorityConfig, xdsLBPolicy *internalservi
...
-					}
-				}
-				continue
-			}
-			for n, c := range configs {
+			odCfgs := convertClusterImplMapToOutlierDetection(configs, p.mechanism.outlierDetection)
+			for n, c := range odCfgs {
 				retConfig.Children[n] = &priority.Child{
-					Config: &internalserviceconfig.BalancerConfig{Name: clusterimpl.Name, Config: c},
+					Config: &internalserviceconfig.BalancerConfig{Name: outlierdetection.Name, Config: c},
```
arvindbr8, line 103: Its not obvious to me why we are changing this to `outlierdetection.Name` here. Could you lmk

arvindbr8 (approved):
LGTM modulo one comment

internal/testutils/xds/bootstrap/bootstrap.go
```diff
@@ -116,12 +116,9 @@ func Contents(opts Options) ([]byte, error) {
...
 
-	auths := make(map[string]authority)
-	if envconfig.XDSFederation {
-		// This will end up using the top-level server list for new style
-		// resources with empty authority.
-		auths[""] = authority{}
-	}
+	// This will end up using the top-level server list for new style
+	// resources with empty authority.
+	auths := map[string]authority{"": {}}
```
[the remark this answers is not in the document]
dfawley: I will leave this here to avoid a behavior change in this PR, but we should look into this.

xds/internal/balancer/clusterresolver/configbuilder.go
dfawley: This is how the diffs rendered.

The remaining code is the code from inside the `if envconfig.XDSOutlierDetection`, and that ends with a `continue`.  The code on the left from 113-120 was dead code if `XDSOutlierDetection` was always true.  But the diffs combined them since there were similar elements.

arvindbr8: oh i see. SG

PR #6750, opened 2023-10-25
grpc: Add a pointer of server to ctx passed into stats handler

zasweq:
This PR adds a pointer to the grpc.Server into the context passed into the stats handler. It also adds an internal only helper on the server that returns whether a given method is registered or not. This will be used in OpenTelemetry instrumentation for security purposes with respect to the cardinality of the method attribute. The next step after these two PRs is to move the rest of the stats handler callouts to the gRPC layer, get registered methods plumbed client side, canoncalize Target() without breaking it and then the OpenTelemetry instrumentation itself :).

RELEASE NOTES: N/A

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -73,6 +73,11 @@ var (
 	// xDS-enabled server invokes this method on a grpc.Server when a particular
 	// listener moves to "not-serving" mode.
 	DrainServerTransports any // func(*grpc.Server, string)
+	// IsRegisteredMethod returns whether the passed in method is registered as
+	// a method on the server.
+	IsRegisteredMethod any // func(*grpc.Server, string)
+	// ServerFromContext returns the server from the context.
+	ServerFromContext any // func(context.Context) *grpc.Server
 	// AddGlobalServerOptions adds an array of ServerOption that will be
 	// effective globally for newly created servers. The priority will be: 1.
 	// user-provided; 2. this method; 3. default values.
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -75,7 +75,7 @@ var (
 	DrainServerTransports any // func(*grpc.Server, string)
 	// IsRegisteredMethod returns whether the passed in method is registered as
 	// a method on the server.
-	IsRegisteredMethod any // func(*grpc.Server, string)
+	IsRegisteredMethod any // func(*grpc.Server, string) bool
 	// ServerFromContext returns the server from the context.
 	ServerFromContext any // func(context.Context) *grpc.Server
 	// AddGlobalServerOptions adds an array of ServerOption that will be
--- a/internal/stubstatshandler/stubstatshandler.go
+++ b/internal/stubstatshandler/stubstatshandler.go
@@ -1,0 +1,72 @@
+/*
+ *
+ * Copyright 2023 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package stubstatshandler is a stubbable implementation of
+// google.golang.org/grpc/stats.Handler for testing purposes.
+package stubstatshandler
+
+import (
+	"context"
+
+	"google.golang.org/grpc/stats"
+
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+)
+
+// StubStatsHandler is a stats handler that is easy to customize within
+// individual test cases.
+type StubStatsHandler struct {
+	// Guarantees we satisfy this interface; panics if unimplemented methods are
+	// called.
+	testgrpc.TestServiceServer
+
+	TagRPCF     func(ctx context.Context, info *stats.RPCTagInfo) context.Context
+	HandleRPCF  func(ctx context.Context, info stats.RPCStats)
+	TagConnF    func(ctx context.Context, info *stats.ConnTagInfo) context.Context
+	HandleConnF func(ctx context.Context, info stats.ConnStats)
+}
+
+// TagRPC calls the StubStatsHandler's TagRPCF, if set.
+func (ssh *StubStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	if ssh.TagRPCF != nil {
+		return ssh.TagRPCF(ctx, info)
+	}
+	return ctx
+}
+
+// HandleRPC calls the StubStatsHandler's HandleRPCF, if set.
+func (ssh *StubStatsHandler) HandleRPC(ctx context.Context, rs stats.RPCStats) {
+	if ssh.HandleRPCF != nil {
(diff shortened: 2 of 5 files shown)


internal/internal.go
```diff
@@ -73,6 +73,11 @@ var (
 	// xDS-enabled server invokes this method on a grpc.Server when a particular
 	// listener moves to "not-serving" mode.
 	DrainServerTransports any // func(*grpc.Server, string)
+	// IsRegisteredMethod returns whether the passed in method is registered as
+	// a method on the server.
+	IsRegisteredMethod any // func(*grpc.Server, string)
```
dfawley, line 78: This returns a `bool` right?

internal/stubstatshandler/stubstatshandler.go
```diff
@@ -0,0 +1,72 @@
...
+
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+)
+
+// StubStatsHandler is a stats handler that is easy to customize within
+// individual test cases.
+type StubStatsHandler struct {
+	// Guarantees we satisfy this interface; panics if unimplemented methods are
+	// called.
+	testgrpc.TestServiceServer
```
dfawley, line 36: Why are you bundling a stats handler with a service implementation?  That seems undesirable unless it's somehow important to keep the two things together.  (And now that I've look at the rest of the code, it seems unused and unnecessary.)

Also, we should put this under `testutils` (either in that package or a subdirectory of it).

server.go
```diff
@@ -1953,6 +1958,44 @@ func (s *Server) getCodec(contentSubtype string) baseCodec {
...
+
+// contextWithServer sets the Server in the context.
+func contextWithServer(ctx context.Context, server *Server) context.Context {
+	return context.WithValue(ctx, serverKey{}, server)
+}
+
+// isRegisteredMethod returns whether the passed in method is registered as a
+// method on the server. /service/method and service/method will match if the
+// service and method are registered on the server.
+func (s *Server) isRegisteredMethod(serviceMethod string) bool {
```
dfawley, line 1977: Isn't this very similar to the code that looks up the handler for a method?  Can we reuse code somehow, or even just call it?

stats/stats_test.go
```diff
@@ -1457,3 +1462,77 @@ func (s) TestMultipleServerStatsHandler(t *testing.T) {
...
+			}
+			errorCh.Send(nil)
+			return ctx
+		},
+	}
+	ss := &stubserver.StubServer{
+		UnaryCallF: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+			return &testpb.SimpleResponse{}, nil
+		},
+		FullDuplexCallF: func(stream testpb.TestService_FullDuplexCallServer) error {
```
dfawley, line 1513: You never call this; either delete it or test it please.

stats/stats_test.go
```diff
@@ -1457,3 +1462,77 @@ func (s) TestMultipleServerStatsHandler(t *testing.T) {
...
+			// after. These internal calls will be how the OpenTelemetry instrumentation
+			// component accesses this server and the subsequent helper on the server.
+			server := internal.ServerFromContext.(func(context.Context) *grpc.Server)(ctx)
+			if server == nil {
+				errorCh.Send("stats handler received ctx has no server present")
+			}
+			isRegisteredMethod := internal.IsRegisteredMethod.(func(*grpc.Server, string) bool)
+			// /s/m and s/m are valid.
+			if !isRegisteredMethod(server, "/grpc.testing.TestService/UnaryCall") {
+				errorCh.Send(errors.New("UnaryCall should be a registered method according to server"))
```
dfawley, line 1490: You don't actually need the error channel for this.  You can just do `t.Errorf` here.  Then all you need to do is ensure the code actually was called.  You can do this by closing a `done`, channel or with a waitgroup, etc.  One benefit of this is that you'll see all the test case failures in a single run if multiple of these checks don't work.

internal/internal.go
zasweq: Ah, right, whoops. Added.

internal/stubstatshandler/stubstatshandler.go
zasweq: Ah, whoops, this is leftover from the stubserver copy paste. Changed to embedding a stats.Handler.

stats/stats_test.go
zasweq: I do test this. See line 1493. However, you are right that I never initiate a bidirectional streaming RPC from the test. Would you like me to add that?

stats/stats_test.go
zasweq: Oh, right, this was when it was a defined stats handler rather than a stub with methods declared in test. Thus, before I couldn't close on t *testing.T, and now I can :). Switched.

zasweq: Chose a waitGroup.

internal/stubstatshandler/stubstatshandler.go
zasweq: Moved file to within testutils as internal/testutils/stubstatshandler.go.

server.go
zasweq: Yes, it is very similar, but I chose not to do it because this line is very specific: https://github.com/grpc/grpc-go/blob/master/server.go#L1731, and I thought it would be hard to generalize what's shared into a helper. If you feel strongly about this after reading the handleStream() code, I can try and pull out logic into helper.

stats/stats_test.go
dfawley: This closure being set or not doesn't change the test at line 1494.  The method is registered because it's in the service descriptor.  I'm fine with deleting this; I don't think there's any need to test both streaming and unary RPCs unless there are different code paths that lead to setting the server in the context, which there aren't.

server.go
dfawley: Yeah please see what you can do.  I think a method that returns `(service string, method string, ok bool)` would be good.

stats/stats_test.go
zasweq: Ah, right, these closures != service descriptor, it's all there anyway. Deleting.

server.go
zasweq: I tried writing it, and it seems distinct enough logically to not warrant a helper. Your function signature doesn't distinguish between a. malformed method name (important here for tracing and logging: https://github.com/grpc/grpc-go/blob/master/server.go#L1732) b. registered unary rpc c. registered streaming name (important here: https://github.com/grpc/grpc-go/blob/master/server.go#L1770, to chose to go through processUnary or processStreaming). This helper's scope is not to log anything in traces or channelz, but to parse a :method header received from the wire and determine whether it's registered or not registered. I think it's sufficiently different to not warrant a new helper.

internal/testutils/stubstatshandler.go
```diff
@@ -16,24 +16,21 @@
...
 
 // StubStatsHandler is a stats handler that is easy to customize within
-// individual test cases.
+// individual test cases. It is a stubbable implementation of
+// google.golang.org/grpc/stats.Handler for testing purposes.
 type StubStatsHandler struct {
 	// Guarantees we satisfy this interface; panics if unimplemented methods are
 	// called.
-	testgrpc.TestServiceServer
+	stats.Handler
```
dfawley, line 33: If we're fully implementing it (which we must be given the earlier snapshot) then this is unnecessary; just delete it.  If you want to have a paranoid check that we're implementing it, you can add a `var _ stats.Handler = (*StubStatsHandler)(nil)`.  But that's not strictly necessary.

dfawley (approved):

internal/testutils/stubstatshandler.go
zasweq: Ah ok deleted.

PR #6760, opened 2023-11-02
security/advancedtls: FileWatcher CRL provider initialization enhancement

erm-g:
Part of CRL Provider effort. The functionality in this PR includes enhancement for FileWatcher provider initialization logic. It also adds minor godoc improvements.

RELEASE NOTES: N/A

### DIFF
--- a/security/advancedtls/crl_provider.go
+++ b/security/advancedtls/crl_provider.go
@@ -91,7 +91,7 @@ func (p *StaticCRLProvider) CRL(cert *x509.Certificate) (*CRL, error) {
 // FileWatcherCRLProvider.
 type FileWatcherOptions struct {
 	CRLDirectory               string          // Path of the directory containing CRL files
-	RefreshDuration            time.Duration   // Time interval between CRLDirectory scans, can't be smaller than 1 minute
+	RefreshDuration            time.Duration   // Time interval (default value is 1 hour) between CRLDirectory scans, can't be smaller than 1 minute
 	CRLReloadingFailedCallback func(err error) // Custom callback executed when a CRL file can’t be processed
 }
 
--- a/security/advancedtls/crl_provider.go
+++ b/security/advancedtls/crl_provider.go
@@ -52,6 +52,9 @@ type CRLProvider interface {
 	//
 	// [RFC5280 - Undetermined]: https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3
 	CRL(cert *x509.Certificate) (*CRL, error)
+
+	// Close cleans up resources allocated by the provider.
+	Close()
 }
 
 // StaticCRLProvider implements CRLProvider interface by accepting raw content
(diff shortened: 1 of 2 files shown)


dfawley (approved):
It seems like a new test (or test case) should be added to cover making sure the initial scan happens synchronously.

erm-g:
> It seems like a new test (or test case) should be added to cover making sure the initial scan happens synchronously.

It's covered by updated [TestFileWatcherCRLProvider](https://github.com/grpc/grpc-go/blob/d97b6a9d8a92dfa26a4e302cfd9019fc03bc90db/security/advancedtls/crl_provider_test.go#L143)
If the initial load is async, then [CRL check](https://github.com/grpc/grpc-go/blob/d97b6a9d8a92dfa26a4e302cfd9019fc03bc90db/security/advancedtls/crl_provider_test.go#L187) happens before the CRLDirectory content is loaded and the test fails, I double checked it locally.

dfawley:
> It's covered by updated TestFileWatcherCRLProvider

Oh, I see.  The old `Close` was done to block until the scan happened.  Very odd to still access `p` after calling `p.Close`, which should normally be considered illegal, but that's not happening anymore which is good.

gtcooke94 (approved):
LGTM

security/advancedtls/crl_provider.go
```diff
@@ -52,6 +52,9 @@ type CRLProvider interface {
 	//
 	// [RFC5280 - Undetermined]: https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3
 	CRL(cert *x509.Certificate) (*CRL, error)
+
+	// Close cleans up resources allocated by the provider.
+	Close()
```
dfawley, line 57: Why change this?  This forces the static CRL provider to implement it, when it doesn't need it.

erm-g: I was taking another look at gRFC and realized that `Close()` is specified at the interface level there - https://github.com/grpc/proposal/pull/382/#discussion_r1379510339
Then I checked the Go repo and realized that having `Close()' at interface level plus no-op impls (if needed) is an existing pattern - https://github.com/grpc/grpc-go/blob/70f1a4045da95b93f73b6dbdd7049f3f053c0680/balancer/balancer.go#L388
During internal discussion folks suggested to make that change - WDYT?

dfawley: We need to close the LB policy from gRPC; hence it's included in that interface.  You don't close the CRL provider from advancedTLS.  So the same reasoning doesn't apply in this case.  I would not add it here, since it isn't necessary.

gtcooke94: This was designed to be very close to the Credential Reloading feature which has `Close` in the interface
https://github.com/grpc/grpc-go/blob/6bed35367c88ff9587572ce58746beac6fafb612/credentials/tls/certprovider/provider.go#L75-L89

dfawley: If the cert provider doesn't need it (i.e. gRPC never calls it), then we should consider removing it from there.  I'm not sure whether that's the case. 
 But that's not justification to add it here.  It's not needed here, so we probably shouldn't add it.

Feel free to disregard my advice if you really want this.

erm-g: I agree with Doug here, for example `Close()` from provider.go is used in [xDS case](https://github.com/grpc/grpc-go/blob/482de22249427eca78db31ebb9937fbd30e81599/xds/internal/balancer/cdsbalancer/cdsbalancer.go#L182) by gRPC for recreation.
Since we don't have such cases for CRL, I'd prefer to drop it from the interface. @gtcooke94 WDYT?

PR #6762, opened 2023-11-03
stress: client logs total num of calls made

temawi:
Needed for an internal, DirectPath related, stress test.

RELEASE NOTES: none

### DIFF
--- a/stress/client/main.go
+++ b/stress/client/main.go
@@ -28,6 +28,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"google.golang.org/grpc"
@@ -55,7 +56,8 @@ var (
 	tlsServerName        = flag.String("server_host_override", "foo.test.google.fr", "The server name use to verify the hostname returned by TLS handshake if it is not empty. Otherwise, --server_host is used.")
 	caFile               = flag.String("ca_file", "", "The file containing the CA root cert file")
 
-	logger = grpclog.Component("stress")
+	totalNumCalls int32
+	logger        = grpclog.Component("stress")
 )
 
 // testCaseWithWeight contains the test case type and its weight.
@@ -206,7 +208,6 @@ func startServer(server *server, port int) {
 	s := grpc.NewServer()
 	metricspb.RegisterMetricsServiceServer(s, server)
 	s.Serve(lis)
-
 }
 
 // performRPCs uses weightedRandomTestSelector to select test case and runs the tests.
@@ -241,6 +242,7 @@ func performRPCs(gauge *gauge, conn *grpc.ClientConn, selector *weightedRandomTe
 			interop.DoCustomMetadata(client, grpc.WaitForReady(true))
 		}
 		numCalls++
+		atomic.AddInt32(&totalNumCalls, 1)
 		gauge.set(int64(float64(numCalls) / time.Since(startTime).Seconds()))
 
 		select {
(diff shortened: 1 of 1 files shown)


temawi:
@dfawley, who could review this?

stress/client/main.go
```diff
@@ -241,6 +242,7 @@ func performRPCs(gauge *gauge, conn *grpc.ClientConn, selector *weightedRandomTe
 			interop.DoCustomMetadata(client, grpc.WaitForReady(true))
```
dfawley, line 242: I don't think we want `WaitForReady` anymore.  There are probably hysterical raisins for why it was once there, but it can be removed now.  Or in a later PR; either way.

stress/client/main.go
dfawley, line 1: Can you move all of this stuff under `/interop` (i.e. `cd grpc-go; mv stress interop/`)?  I don't think it really wants to be a top-level thing.

stress/client/main.go
```diff
@@ -241,6 +242,7 @@ func performRPCs(gauge *gauge, conn *grpc.ClientConn, selector *weightedRandomTe
...
 		}
 		numCalls++
+		atomic.AddInt32(&totalNumCalls, 1)
```
dfawley, line 245: This works, but maybe, optionally, avoid so many atomic increments via:

```go
var numCalls int64
defer func() { atomic.AddInt64(&totalNumCalls, numCalls) }()
```

stress/client/main.go
```diff
@@ -55,7 +56,8 @@ var (
 	tlsServerName        = flag.String("server_host_override", "foo.test.google.fr", "The server name use to verify the hostname returned by TLS handshake if it is not empty. Otherwise, --server_host is used.")
 	caFile               = flag.String("ca_file", "", "The file containing the CA root cert file")
 
-	logger = grpclog.Component("stress")
+	totalNumCalls int32
```
dfawley, line 59: Should this be `int64`?  Or do we think we'll never manage 2B calls?

For reference, 2B microseconds is ~30 minutes.  So if we can do 1 RPC / microsecond (e.g. with parallelism), we'll overflow in half an hour of stress testing.  `int64` seems prudent.

stress/client/main.go
```diff
@@ -335,6 +337,6 @@ func main() {
 		close(stop)
```
dfawley, line 337: Optional for this PR but would be nice to fix up soon if we are reviving this:

More natural and simpler would be:

```go
ctx := context.Background
if *testDurationSecs > 0 {
	ctx, cancel = context.WithTimeout(ctx, *testDurationSecs * time.Second)
	defer cancel()
}
........
		performRPCs(ctx, g, conn, testSelector)
......

func performRPCs() {
....
	for ctx.Err() == nil {
		....
	}
}
```

Even better would be to _then_ pass `ctx` to `interop.DoBlah` but that probably affects other things too.  But it's following a bad practice and should be cleaned up one day.  I'll file an issue for that.

dfawley: (#6763 FYI)

stress/client/main.go
temawi: I can do that, but rather in a later PR so that this commit stays focused on one thing.

stress/client/main.go
temawi: Yes, it should be int64, i used a long in Java as well.

stress/client/main.go
temawi: Later PR... Same reasons as with moving to under `/interop`.

stress/client/main.go
temawi: Yes, that's better, done.

stress/client/main.go
temawi: I need to chat with you a bit more about this, not changing in this PR.

dfawley (approved):

PR #6767, opened 2023-11-06
feat: add protojson encoding for server

hongshengjie:


zasweq:
Hello, sorry, but we cannot take this PR. If you would like this feature to be in all gRPC, can you please open an issue in the grpc/grpc repo (https://github.com/grpc/grpc/issues). If we want to add a feature like this, it would have to be a cross language feature.

PR #6768, opened 2023-11-06
xds/resolver: extend test to re-add listener

dfawley:
This is just for peace of mind since a similar situation came up in interop tests, and I wanted to make sure it was handled correctly.

RELEASE NOTES: none


zasweq (approved):
LGTM.

PR #6769, opened 2023-11-06
stress: Move package under interop

temawi:
RELEASE NOTES: none


vet.sh
```diff
@@ -1,71 +1,75 @@
 #!/bin/bash
 
-set -ex  # Exit on error; debugging enabled.
-set -o pipefail  # Fail a pipe if any sub-command fails.
+set -ex         # Exit on error; debugging enabled.
```
dfawley, line 3: Sorry, but can you revert all the lines that didn't need to be changed? It will kill `blame`.

dfawley (approved):

PR #6771, opened 2023-11-07
reflection: don't serialize placeholders

### LINKED ISSUE #6770: reflection: if some files cannot be resolved, reflection will send invalid descriptors to client
When generated Protobuf code registers file descriptors with the global registry, whenever there are missing dependencies, they are silently replaced with "placeholder" descriptors.

This happens if a file uses the wrong import path. For example, a file is compiled with the path "validate/validate.proto", so it is registered in `protoregistry.GlobalFiles` with that string as its path. Another file that imports it uses an import statement of "proto/validate/validate.proto" instead. When this file is registered during program initialization, the dependency cannot be found (because its name doesn't match the way the file was actually registered). So the reflection library inserts a placeholder file descriptor.

When gRPC server reflection is used to download descriptors for the importing file, it will provide a junk file descriptor for the missing dependency, the one named "proto/validate/validate.proto". (Original grpcurl bug report: https://github.com/fullstorydev/grpcurl/issues/423).

This can result in confusing error messages because the reason the descriptor is invalid (aside from the fact that it is empty, so any symbols that the importing file references will be unresolvable) is that the `syntax` field of the file descriptor is the string `<unknown:0>`. When trying to process this descriptor, this can result in a very confusing error message to users.

A separate fix is needed in the Protobuf runtime to just _not_ set the `syntax` field of the descriptor if its value is unknown, instead of setting this string, That would help... a little. But without a change to the gRPC reflection implementation, the client will still get a descriptor that is not actually usable. The error message would then be a linking issue since the imported symbol still won't be available. IMO, the proper solution is for the reflection implementation to _elide_ placeholder descriptors. That will result in a better error message along the lines of `could not resolve file "proto/validate/validate.proto": not found`.

jhump:
Resolves #6770.

On a related note, I'm also trying to land a patch to the protobuf runtime:
https://go-review.googlesource.com/c/protobuf/+/540455

I think both that and this fix are useful. This one is likely more useful since a "not found" error for a particular file will be a better error message to users, given that the issue is in fact that an imported file is missing from the global registry.

RELEASE NOTES:
* reflection: do not send invalid descriptors to clients for files that cannot be fully resolved

### DIFF
--- a/reflection/serverreflection.go
+++ b/reflection/serverreflection.go
@@ -176,11 +176,20 @@ type serverReflectionServer struct {
 // wire format ([]byte). The fileDescriptors will include fd and all the
 // transitive dependencies of fd with names not in sentFileDescriptors.
 func (s *serverReflectionServer) fileDescWithDependencies(fd protoreflect.FileDescriptor, sentFileDescriptors map[string]bool) ([][]byte, error) {
+	if fd.IsPlaceholder() {
+		// If the given root file is a placeholder, treat it
+		// as missing instead of serializing it.
+		return nil, protoregistry.NotFound
+	}
 	var r [][]byte
 	queue := []protoreflect.FileDescriptor{fd}
 	for len(queue) > 0 {
 		currentfd := queue[0]
 		queue = queue[1:]
+		if currentfd.IsPlaceholder() {
+			// Skip any missing files in the dependency graph.
+			continue
+		}
 		if sent := sentFileDescriptors[currentfd.Path()]; len(r) == 0 || !sent {
 			sentFileDescriptors[currentfd.Path()] = true
 			fdProto := protodesc.ToFileDescriptorProto(currentfd)
--- a/reflection/serverreflection_test.go
+++ b/reflection/serverreflection_test.go
@@ -170,6 +170,206 @@ func (x) TestAllExtensionNumbersForTypeName(t *testing.T) {
 	}
 }
 
+func (x) TestFileDescWithDependencies(t *testing.T) {
+	depFile, err := protodesc.NewFile(
+		&descriptorpb.FileDescriptorProto{
+			Name: proto.String("dep.proto"),
+		}, nil,
+	)
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+
+	deps := &protoregistry.Files{}
+	if err := deps.RegisterFile(depFile); err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+
+	rootFileProto := &descriptorpb.FileDescriptorProto{
+		Name: proto.String("root.proto"),
+		Dependency: []string{
+			"google/protobuf/descriptor.proto",
+			"reflection/grpc_testing/proto2_ext2.proto",
+			"dep.proto",
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


dfawley (approved):
Thank you for the fix!

dfawley:
@jhump

If you have some idea how to explain "placeholder" better in a user-friendly changelog message (I said "not completely defined"???) then please update the `RELEASE NOTES` section I added to https://github.com/grpc/grpc-go/pull/6771#issue-1981434479 - thanks!

jhump:
> If you have some idea how to explain "placeholder" better in a user-friendly changelog message

@dfawley, another way is to frame it more like the title of the corresponding issue (#6770):

* reflection: do not send invalid descriptors to the client for files that cannot be fully resolved

WDYT? I went ahead and updated the description in case that looks good to you.

arvindbr8 (approved):

arvindbr8:
LGTM, thanks for the fix @jhump

nikitaksv:
@arvindbr8 Hello. This update broke all popular clients with Reflection API support. Postman, Insomnia, grpcurl can't read API via reflection due to missing IsPlaceholder() on line https://github.com/grpc/grpc-go/blob/master/reflection/serverreflection.go#L189.  

Add various imports (validation, openapiv2) to proto and see for yourself that reflection in clients will not be read.
I commented out the code at 
https://github.com/grpc/grpc-go/blob/master/reflection/serverreflection.go#L189-L191
and
https://github.com/grpc/grpc-go/blob/master/reflection/serverreflection.go#L179-L183

And reflection API in Postman, Insomnia, grpcurl worked.

jhump:
> Add various imports (validation, openapiv2) to proto and see for yourself that reflection in clients will not be read.

@nikitaksv, this is not true. I have plenty of examples of working protos with those imports, and reflection works fine with them. This only happens if you use the _wrong_ import path when importing those files. (See https://buf.build/docs/reference/protobuf-files-and-packages#imports for more info.) For example for openapiv2, you _must_ use the following path:
```proto
import "protoc-gen-openapiv2/options/annotations.proto";
```
If your import references that file using any other path, then the descriptors cannot be correctly linked at runtime.

The placeholder descriptors are junk -- they come across as empty files at best (as of [this fix](https://go-review.googlesource.com/c/protobuf/+/540455)) or as invalid, broken files (before that fix).

What library were you using in your reflection clients that they could successfully process these descriptors? I know that `grpcurl` used older versions of `github.com/jhump/protoreflect/grpcreflect`. Perhaps the others did, too, because that is likely the only library that happened to tolerate these files. That package (prior to v1.15) did not properly _validate_ the downloaded file descriptors, so it was a fluke that they _ever_ really worked. Using the official Protobuf runtimes/packages (such as [google.golang.org/protobuf/reflect/protoreflect](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect), but also languages other than Go), the placeholder descriptors downloaded from the server _never_ worked. These files were only tolerated because they only provide custom options, which weren't resolved by that `grpcreflect` package, which is why that package didn't "notice" that the files were junk. If the same symptom (incorrect import path) occurred with a file that contained a message or enum type that the importing file needed, it would have broken.

PR #6775, opened 2023-11-08
Set top-level read-only workflow permissions

### LINKED ISSUE #6774: Set coverage.yml workflow to run with read-only tokens
Almost all of gRPC-Go's GitHub workflows run with proper permissions. However, I've noticed that `coverage.yml` currently runs with the default write-all token.

This issue can be solved in two ways:

- add top-level read-only permissions to the workflow; and/or
- set the default token permissions to read-only in the repo settings.

I'll send a PR with the workflow changes, if you're interested.

If you instead (or also) wish to modify the default token permissions:

1. Open the repo settings
2. Go to [Actions > General](https://github.com/grpc/grpc-go/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

This setting can (or should) also be configured at the org level, but that might break workflows in other repositories that are "unwittingly" using the write-all token.

---

**Disclosure**: My name is Pedro and I work with Google and the Linux Foundation's [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem.

pnacht:
Fixes #6774.

As mentioned in the issue, this PR adds top-level read-only permissions to `coverage.yml`, ensuring it can't be used for supply-chain attacks on the repo.

I've also made a few similar changes to the other workflows, mostly just setting `write` permissions at job-level instead of top-level. This serves to future-proof the workflows in case new jobs (that don't need those permissions) are added to the workflows. However, this change has no immediate impact to those workflows' security: the tokens effectively used in those jobs are unchanged.

RELEASE NOTES: None

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -8,15 +8,17 @@ on:
 
 permissions:
   contents: read
-  security-events: write
-  pull-requests: read
-  actions: read
 
 jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
     timeout-minutes: 30
+
+    permissions:
+      security-events: write
+      pull-requests: read
+      actions: read
 
     strategy:
       fail-fast: false
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,5 +1,9 @@
 name: codecov
 on: [push, pull_request]
+
+permissions:
+  contents: read
+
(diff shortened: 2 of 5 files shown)


arvindbr8 (approved):
LGTM. 

----

@dfawley: Could you PTAL as well?

dfawley (approved):
Thank you for the fix!

PR #6776, opened 2023-11-08
credentials: if not set, restrict to TLS v1.2+ and CipherSuites per RFC7540

### LINKED ISSUE #6758: GRPC Server allows usage of TLS1.0 for HTTP/2 protocol
Hello, I want to ask some a question regarding implementation of RFC 7540.
[RFC 7540](https://datatracker.ietf.org/doc/html/rfc7540#section-9.2) enforces some restrictions on TLS MinVersion and cipher suites used for HTTP/2 connections. It states that HTTP/2 connections must use TLS version 1.2 or higher.
I've  tested example from grpc-go and it seems that server allows TLS1.0 connections.
Looks like both tls and advancedtls implementations use default Golang limit on server tls:
* https://github.com/grpc/grpc-go/blob/v1.59.0/credentials/tls.go#L143
* https://github.com/grpc/grpc-go/blob/b82468a3462e18361e6816e25842375f517255b1/security/advancedtls/advancedtls.go#L590
which is TLS 1.0 for server.
```
cd ./examples/features/encryption/TLS/
go run server/main.go

openssl-1.1 s_client -tls1 -connect localhost:50051
...
---
New, TLSv1.0, Cipher is ECDHE-RSA-AES128-SHA
Server public key is 4096 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1
    Cipher    : ECDHE-RSA-AES128-SHA
    Session-ID: 75DCDAD64A6B7B436E7583C5E49FEB11FF0CFA92DD58CA4B41A97D7AEC1E27BD
    Session-ID-ctx: 
    Master-Key: 2AB2840F395D5C80DD36494B1130416FFEAF53EB00A24A71816729099D773A8E8E4DC85AF844F6D91797990EDBAA5A41
    PSK identity: None
    PSK identity hint: None
    SRP username: None
...
```
Is it some unknown deviation from RFC, or I've got something wrong?

dfawley:
Fixes #6758

cc @ejona86 FYI

RELEASE NOTES:
* credentials: if not set, set TLS `MinVersion` to 1.2 and `CipherSuites` according to supported suites not forbidden by RFC7540.  This is a behavior change to bring us into better alignment with RFC 7540.

### DIFF
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -153,10 +153,39 @@ func (c *tlsCreds) OverrideServerName(serverNameOverride string) error {
 	return nil
 }
 
+// The following cipher suites are forbidden for use with HTTP/2 by
+// https://datatracker.ietf.org/doc/html/rfc7540#appendix-A
+var tls12ForbiddenCipherSuites = map[uint16]struct{}{
+	tls.TLS_RSA_WITH_AES_128_CBC_SHA:         {},
+	tls.TLS_RSA_WITH_AES_256_CBC_SHA:         {},
+	tls.TLS_RSA_WITH_AES_128_GCM_SHA256:      {},
+	tls.TLS_RSA_WITH_AES_256_GCM_SHA384:      {},
+	tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA: {},
+	tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA: {},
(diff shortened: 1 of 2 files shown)


ginayeh (approved):
you might want to check the spelling suggestion from vet before merging

PR #6779, opened 2023-11-09
Add LocalAddr to http2Client.getPeer()

### LINKED ISSUE #6577: add LocalAddr to peer.Peer client side

CarlosRDomin:
Fixes #6577.

Followup to #6716 that exposes `LocalAddr` also to grpc clients (#6716 only exposed it to servers). Release note in #6716 encapsulates this logic as well.

RELEASE NOTES: n/a

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -495,8 +495,9 @@ func (t *http2Client) newStream(ctx context.Context, callHdr *CallHdr) *Stream {
 
 func (t *http2Client) getPeer() *peer.Peer {
 	return &peer.Peer{
-		Addr:     t.remoteAddr,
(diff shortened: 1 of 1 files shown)


zasweq (approved):
LGTM.

dfawley (approved):

PR #6796, opened 2023-11-15
xDS: Atomically read and write xDS security configuration client side

zasweq:
This PR does a few things:
* Makes handshake info fields set at init time and read only after that, eliminating need for mutex
* Scale up newHandshakeInfo() constructor to all fields
* Atomically write and read handshake info for cds

This eliminates acting on partial updates due to synchronization only between setters on individual fields, so reads and writes (within the context of one Handshake and one new update) could arbitrarily intersplice and cause correctness issues due to partial reads of handshake info configuration. Security configuration still needs to be per cluster and not set by the top level cluster, but this fixes the race.

RELEASE NOTES:
* xDS: Atomically read and write xDS security configuration client side

### DIFF
--- a/credentials/xds/xds_client_test.go
+++ b/credentials/xds/xds_client_test.go
@@ -27,6 +27,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 	"unsafe"
@@ -220,11 +221,11 @@ func newTestContextWithHandshakeInfo(parent context.Context, root, identity cert
 	// Creating the HandshakeInfo and adding it to the attributes is very
 	// similar to what the CDS balancer would do when it intercepts calls to
 	// NewSubConn().
-	var sm []matcher.StringMatcher
+	var sms []matcher.StringMatcher
 	if sanExactMatch != "" {
-		sm = []matcher.StringMatcher{matcher.StringMatcherForTesting(newStringP(sanExactMatch), nil, nil, nil, nil, false)}
+		sms = []matcher.StringMatcher{matcher.StringMatcherForTesting(newStringP(sanExactMatch), nil, nil, nil, nil, false)}
 	}
-	info := xdsinternal.NewHandshakeInfo(root, identity, sm, false)
+	info := xdsinternal.NewHandshakeInfo(root, identity, sms, false)
 	uPtr := unsafe.Pointer(info)
 	addr := xdsinternal.SetHandshakeInfo(resolver.Address{}, &uPtr)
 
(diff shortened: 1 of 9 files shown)


credentials/xds/xds_client_test.go
```diff
@@ -219,11 +220,13 @@ func newTestContextWithHandshakeInfo(parent context.Context, root, identity cert
 	// Creating the HandshakeInfo and adding it to the attributes is very
 	// similar to what the CDS balancer would do when it intercepts calls to
 	// NewSubConn().
-	info := xdsinternal.NewHandshakeInfo(root, identity)
+	var sm []matcher.StringMatcher
```
dfawley, line 223: Nit/optional: `sms` to pluralize string matcher.

credentials/xds/xds_client_test.go
```diff
@@ -560,7 +562,10 @@ func (s) TestClientCredsProviderSwitch(t *testing.T) {
 	// Create a new root provider which uses the correct trust roots. And update
 	// the HandshakeInfo with the new provider.
 	root2 := makeRootProvider(t, "x509/server_ca_cert.pem")
-	handshakeInfo.SetRootCertProvider(root2)
+	handshakeInfo = xdsinternal.NewHandshakeInfo(root2, nil, []matcher.StringMatcher{matcher.StringMatcherForTesting(newStringP(defaultTestCertSAN), nil, nil, nil, nil, false)}, false)
+	uPtr = unsafe.Pointer(handshakeInfo)
+	addr = xdsinternal.SetHandshakeInfo(resolver.Address{}, &uPtr)
+	ctx = icredentials.NewClientHandshakeInfoContext(ctx, credentials.ClientHandshakeInfo{Attributes: addr.Attributes})
```
dfawley, line 568: Can we just update the existing pointer instead of creating a new entry in the context? That's how the production code will work.

internal/credentials/xds/handshake_info.go
```diff
@@ -66,16 +66,16 @@ func (hi *HandshakeInfo) Equal(other *HandshakeInfo) bool {
...
-// updated with hInfo.
-func SetHandshakeInfo(addr resolver.Address, hInfo *HandshakeInfo) resolver.Address {
-	addr.Attributes = addr.Attributes.WithValue(handshakeAttrKey{}, hInfo)
+// updated with hiPtr.
+func SetHandshakeInfo(addr resolver.Address, hiPtr *unsafe.Pointer) resolver.Address {
+	addr.Attributes = addr.Attributes.WithValue(handshakeAttrKey{}, hiPtr)
 	return addr
 }
 
 // GetHandshakeInfo returns a pointer to the HandshakeInfo stored in attr.
```
dfawley, line 75: Maybe "pointer to the *HandshakeInfo"?

internal/credentials/xds/handshake_info.go
```diff
@@ -85,40 +85,21 @@ func GetHandshakeInfo(attr *attributes.Attributes) *HandshakeInfo {
 //
 // Safe for concurrent access.
```
dfawley, line 86: Delete this now?  Or say it's immutable?

xds/internal/server/conn_wrapper.go
```diff
@@ -128,8 +128,7 @@ func (c *connWrapper) XDSHandshakeInfo() (*xdsinternal.HandshakeInfo, error) {
 	c.identityProvider = ip
 	c.rootProvider = rp
 
-	xdsHI := xdsinternal.NewHandshakeInfo(c.rootProvider, c.identityProvider)
-	xdsHI.SetRequireClientCert(secCfg.RequireClientCert)
+	xdsHI := xdsinternal.NewHandshakeInfo(c.rootProvider, c.identityProvider, nil, secCfg.RequireClientCert)
```
dfawley, line 131: Optional: return directly instead of using a temp var.

dfawley (approved):
Logic all LGTM...just a few mostly minor nits/suggestions/etc.

credentials/xds/xds_client_test.go
zasweq: Fair. Switched.

internal/credentials/xds/handshake_info.go
zasweq: I feel like technically it's fine to call any methods on it concurrently since it's all read only and set at init time, but since it's not really important to call out/doesn't apply I'll delete it.

credentials/xds/xds_client_test.go
zasweq: Ah, good point. Changed.

PR #6799, opened 2023-11-15
orca: use atomic pointer instead of mutex in server metrics recorder to improve performance

danielzhaotongliu:
PR included:
- Use atomic pointers instead of mutex to improve performance.
- When the server metrics recorder is modified/written (e.g. `SetCPUUtilization`) create a new snapshot and atomically store it. When the server metrics recorder is read, return the reference to the latest snapshot. Similar in implementation to C++.

Motivation:
- See internal email thread and benchmarks regarding performance

Tested:
- `go test -cpu 1,4 -timeout 7m ./...`
- `go test -race -cpu 1,4 -timeout 7m ./...`
- Verified new implementation improved internal benchmark performance

RELEASE NOTES:

- orca: use atomic pointer to improve performance in server metrics recorder

### DIFF
--- a/orca/call_metrics.go
+++ b/orca/call_metrics.go
@@ -91,7 +91,7 @@ func (rw *recorderWrapper) recorder() CallMetricsRecorder {
 func (rw *recorderWrapper) setTrailerMetadata(ctx context.Context) {
 	var sm *ServerMetrics
 	if rw.smp != nil {
-		sm = copyServerMetrics(rw.smp.ServerMetrics())
+		sm = rw.smp.ServerMetrics()
 		sm.merge(rw.r.ServerMetrics())
 	} else {
 		sm = rw.r.ServerMetrics()
--- a/orca/call_metrics.go
+++ b/orca/call_metrics.go
@@ -91,7 +91,7 @@ func (rw *recorderWrapper) recorder() CallMetricsRecorder {
 func (rw *recorderWrapper) setTrailerMetadata(ctx context.Context) {
 	var sm *ServerMetrics
 	if rw.smp != nil {
-		sm = rw.smp.ServerMetrics()
+		sm = copyServerMetrics(rw.smp.ServerMetrics())
 		sm.merge(rw.r.ServerMetrics())
 	} else {
 		sm = rw.r.ServerMetrics()
--- a/orca/server_metrics.go
+++ b/orca/server_metrics.go
@@ -19,7 +19,7 @@
 package orca
 
 import (
-	"sync"
+	"sync/atomic"
 
 	v3orcapb "github.com/cncf/xds/go/xds/data/orca/v3"
 )
@@ -142,8 +142,7 @@ type ServerMetricsRecorder interface {
 }
 
 type serverMetricsRecorder struct {
-	mu    sync.Mutex     // protects state
-	state *ServerMetrics // the current metrics
+	state atomic.Pointer[ServerMetrics] // latest snapshot of the current metrics
 }
 
 // NewServerMetricsRecorder returns an in-memory store for ServerMetrics and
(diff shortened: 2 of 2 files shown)


orca/call_metrics.go
```diff
@@ -91,7 +91,7 @@ func (rw *recorderWrapper) recorder() CallMetricsRecorder {
 func (rw *recorderWrapper) setTrailerMetadata(ctx context.Context) {
 	var sm *ServerMetrics
 	if rw.smp != nil {
-		sm = rw.smp.ServerMetrics()
+		sm = copyServerMetrics(rw.smp.ServerMetrics())
```
dfawley, line 94: This copy probably shouldn't ever be necessary, or else `ServerMetrics` isn't implemented according to its documentation (that it's immutable).  And it looks like we don't need to make a copy in the implementation, either, because it's never modified.  Is that right?  (I'm assuming that the caller of `ServerMetrics` doesn't mutate it, which should always be the case when we call it. And maybe we should also document that in the interface.)

yousukseung: In each response we need to add the union of per-server and per-call metrics where the latter takes a higher a precedence, so a copy is necessary per response I believe.

dfawley: The `ServerMetrics` method itself is supposed to be making the copy, if one is needed.

danielzhaotongliu: Done.
The `ServerMetrics()` method now returns _**A COPY**_ of the latest snapshot in the latest commit as per-contract. 

For context, the `merge` operation would union the per-server (which is now a copy) and the per-call metrics where the latter takes higher precedence. This maintains the same behaviour as before and preserves data integrity and thread safety.

dfawley: Sorry maybe I'm not understanding.  Why copy, still?  I think we should assume the return value of `ServerMetrics` is not to be modified (let's document that as I mentioned above if it isn't already), in which case we can save making a copy when reading.

danielzhaotongliu: The return value of `ServerMetrics` _**could be modified**_ in `setTrailerMetadata` if both the the per-server (`server_metrics.go`) and per-call metrics (`call_metrics.go`) exist for per-call load reporting, where we need to `merge` (union) the both of them where per-call metrics take precedence [1, 2]. Also, see gRFC A51 [3]

If we do NOT make a copy (and only return the reference to the latest snapshot), then the per-call metrics data can pollute the per-server metrics data (as per-call metrics take higher precedence). For example, say during one RPC call, both QPS exists in per-server and per-call metrics, after the merge operation, the per-server metrics would have the QPS from the per-call metrics. Since we only return the reference, thus for the next update operation (and subsequent RPC calls) for per-server metrics (say `SetCPUUtilization`), it would still have the QPS value from the per-call metrics. 

If the above scenario is the intended behaviour then we don't need the copy, but from my understanding of the gRFC A51, it is not and we should always return a copy of the `ServerMetrics` FYI: @yousukseung 

References
[1] https://github.com/grpc/grpc-go/blob/master/orca/call_metrics.go#L94-L95
[2] https://github.com/grpc/grpc-go/blob/master/orca/server_metrics.go#L68-L88
[3] https://github.com/grpc/proposal/blob/master/A51-custom-backend-metrics.md#the-server-api

dfawley: Ahh..  I was thinking that "merge" was merging two _protos_ for some reason.  But we merge and then create the proto.

So another option is to make it so `ServerMetrics` must be read-only, and merge the two sources into a third in (1), but I'm not sure this saves much.

Otherwise, I'm fine with this if you are.

dfawley (approved):

orca/call_metrics.go
danielzhaotongliu: I am fine with the current approach as is. Merging the per-call and per-server metrics into a third with [1] would essentially be equivalent to performing another copy operation (as we have to check and set every field if applicable) and thus would not offer any savings. Resolving the thread for now.  Thanks for your review and questions!

PR #6804, opened 2023-11-16
client: rework resolver and balancer wrappers to avoid deadlock

### LINKED ISSUE #6783: Race between calls from resolver and entering idle mode
From https://github.com/grpc/grpc-go/pull/6778#issuecomment-1804857680:

https://github.com/grpc/grpc-go/blob/be1d1c10a930fa0e91fc4bcd01963035011e2466/clientconn.go#L429

And `ccr.UpdateState`'s callback calls `updateResolverState` which needs `cc.mu`:

https://github.com/grpc/grpc-go/blob/be1d1c10a930fa0e91fc4bcd01963035011e2466/clientconn.go#L829

So if `UpdateState` (and all the things that schedule callbacks?) doesn't check whether the ccr is closing before scheduling callbacks, then we can deadlock.

dfawley:
- Balancer wrapper is now recreated for each trip out of idle mode (instead of supporting its own idle mode).
- Removed mutexes or limited as much as possible.
- All calls into resolver/balancer are done in serializer, with no mutexes held.
- Calls from old resolver instances are ignored.
- ExitIdleCond is no longer needed by closing the idlenessMgr at the start of Close.
- Constructors figure out their own parameters from cc instead of being passed in.
- Handle resolver.Build->cc.updateResolverState->balancer.UpdateState->resolver.ResolveNow properly
- Idle Manager starts in idle state.
- All channel idleness management goes through idle manager -- e.g. Connect().
- Add lock and closed state to balancer_wrapper to prevent a race where the
  balancer creates a subchannel and we enter idle mode during that process.
- Split resolver wrapper into constructor and start method so the serializer is
  available during resolver.Build.
- Add EnterIdleModeForTesting to idleness manager and use idlenessMgr methods
  in the idleness test instead of methods on the ClientConn which should not be
  called directly.

Fixes #6783 

RELEASE NOTES:
* client: fix race that could cause a deadlock while entering idle mode and receiving a name resolver update

### DIFF
--- a/balancer_wrapper.go
+++ b/balancer_wrapper.go
@@ -51,11 +51,13 @@ type ccBalancerWrapper struct {
 	// read-only afterwards, and therefore can be accessed without a mutex.
 	cc               *ClientConn
 	opts             balancer.BuildOptions
-	balancer         *gracefulswitch.Balancer
 	serializer       *grpcsync.CallbackSerializer
 	serializerCancel context.CancelFunc
 
-	curBalancerName string // only accessed within the serializer
+	// The following fields are only accessed within the serializer or during
+	// initialization.
+	curBalancerName string
+	balancer        *gracefulswitch.Balancer
 
 	// The following field is protected by mu.  Caller must take cc.mu before
 	// taking mu.
--- a/balancer_wrapper.go
+++ b/balancer_wrapper.go
@@ -109,20 +109,6 @@ func (ccb *ccBalancerWrapper) updateClientConnState(ccs *balancer.ClientConnStat
 	return <-errCh
 }
 
-// updateSubConnState is invoked by grpc to push a subConn state update to the
-// underlying balancer.
-func (ccb *ccBalancerWrapper) updateSubConnState(sc balancer.SubConn, s connectivity.State, err error) {
-	ccb.serializer.Schedule(func(ctx context.Context) {
-		if ctx.Err() != nil || ccb.balancer == nil {
-			return
-		}
-		// Even though it is optional for balancers, gracefulswitch ensures
-		// opts.StateListener is set, so this cannot ever be nil.
-		// TODO: delete this comment when UpdateSubConnState is removed.
-		sc.(*acBalancerWrapper).stateListener(balancer.SubConnState{ConnectivityState: s, ConnectionError: err})
-	})
-}
-
 // resolverError is invoked by grpc to push a resolver error to the underlying
 // balancer.  The call to the balancer is executed from the serializer.
 func (ccb *ccBalancerWrapper) resolverError(err error) {
@@ -301,6 +287,20 @@ type acBalancerWrapper struct {
 
 	mu        sync.Mutex
 	producers map[balancer.ProducerBuilder]*refCountedProducer
+}
+
+// updateState is invoked by grpc to push a subConn state update to the
+// underlying balancer.
+func (acbw *acBalancerWrapper) updateState(s connectivity.State, err error) {
+	acbw.ccb.serializer.Schedule(func(ctx context.Context) {
+		if ctx.Err() != nil || acbw.ccb.balancer == nil {
+			return
+		}
+		// Even though it is optional for balancers, gracefulswitch ensures
+		// opts.StateListener is set, so this cannot ever be nil.
+		// TODO: delete this comment when UpdateSubConnState is removed.
+		acbw.stateListener(balancer.SubConnState{ConnectivityState: s, ConnectionError: err})
+	})
 }
 
 func (acbw *acBalancerWrapper) String() string {
--- a/balancer_wrapper.go
+++ b/balancer_wrapper.go
@@ -217,7 +217,7 @@ func (ccb *ccBalancerWrapper) NewSubConn(addrs []resolver.Address, opts balancer
 	ccb.mu.Lock()
 	if ccb.closed {
 		ccb.mu.Unlock()
-		return nil, errConnIdling
+		return nil, fmt.Errorf("balancer is being closed; no new SubConns allowed.")
 	}
 	ccb.mu.Unlock()
 
--- a/balancer_wrapper.go
+++ b/balancer_wrapper.go
@@ -37,7 +37,8 @@ import (
 // ccBalancerWrapper implements methods corresponding to the ones on the
 // balancer.Balancer interface. The ClientConn is free to call these methods
 // concurrently and the ccBalancerWrapper ensures that calls from the ClientConn
-// to the Balancer happen synchronously and in order.
+// to the Balancer happen in order by performing them in the serializer, without
+// any mutexes held.
 //
 // ccBalancerWrapper also implements the balancer.ClientConn interface and is
 // passed to the Balancer implementations. It invokes unexported methods on the
@@ -54,10 +55,9 @@ type ccBalancerWrapper struct {
 	serializer       *grpcsync.CallbackSerializer
 	serializerCancel context.CancelFunc
 
-	// The following fields are only accessed within the serializer.
-	curBalancerName string
+	curBalancerName string // only accessed within the serializer
 
-	// The following fields are protected by mu.  Caller must take cc.mu before
+	// The following field is protected by mu.  Caller must take cc.mu before
 	// taking mu.
 	mu     sync.Mutex
 	closed bool
@@ -92,7 +92,7 @@ func (ccb *ccBalancerWrapper) updateClientConnState(ccs *balancer.ClientConnStat
 	errCh := make(chan error)
 	ok := ccb.serializer.Schedule(func(ctx context.Context) {
 		defer close(errCh)
-		if ctx.Err() != nil {
+		if ctx.Err() != nil || ccb.balancer == nil {
 			return
 		}
 		err := ccb.balancer.UpdateClientConnState(*ccs)
@@ -111,7 +111,7 @@ func (ccb *ccBalancerWrapper) updateClientConnState(ccs *balancer.ClientConnStat
 // underlying balancer.
 func (ccb *ccBalancerWrapper) updateSubConnState(sc balancer.SubConn, s connectivity.State, err error) {
 	ccb.serializer.Schedule(func(ctx context.Context) {
-		if ctx.Err() != nil {
+		if ctx.Err() != nil || ccb.balancer == nil {
 			return
 		}
 		// Even though it is optional for balancers, gracefulswitch ensures
@@ -122,11 +122,10 @@ func (ccb *ccBalancerWrapper) updateSubConnState(sc balancer.SubConn, s connecti
 }
 
 // resolverError is invoked by grpc to push a resolver error to the underlying
-// balancer.  This is always executed from the serializer, so it is safe to call
-// into the balancer here.
+// balancer.  The call to the balancer is executed from the serializer.
 func (ccb *ccBalancerWrapper) resolverError(err error) {
 	ccb.serializer.Schedule(func(ctx context.Context) {
-		if ctx.Err() != nil {
+		if ctx.Err() != nil || ccb.balancer == nil {
 			return
 		}
 		ccb.balancer.ResolverError(err)
@@ -145,7 +144,7 @@ func (ccb *ccBalancerWrapper) resolverError(err error) {
 // the graceful balancer switching process if the name does not change.
 func (ccb *ccBalancerWrapper) switchTo(name string) {
 	ccb.serializer.Schedule(func(ctx context.Context) {
-		if ctx.Err() != nil {
+		if ctx.Err() != nil || ccb.balancer == nil {
 			return
 		}
 		// TODO: Other languages use case-sensitive balancer registries. We should
@@ -193,10 +192,10 @@ func (ccb *ccBalancerWrapper) close() {
 	ccb.serializerCancel()
 }
 
-// exitIdle invokes the balancer's exitIdle method in the scheduler.
+// exitIdle invokes the balancer's exitIdle method in the serializer.
 func (ccb *ccBalancerWrapper) exitIdle() {
 	ccb.serializer.Schedule(func(ctx context.Context) {
-		if ctx.Err() != nil {
+		if ctx.Err() != nil || ccb.balancer == nil {
 			return
 		}
 		ccb.balancer.ExitIdle()
--- a/balancer_wrapper.go
+++ b/balancer_wrapper.go
@@ -32,15 +32,6 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
-type ccbMode int
-
-const (
-	ccbModeActive = iota
-	ccbModeIdle
-	ccbModeClosed
-	ccbModeExitingIdle
-)
-
 // ccBalancerWrapper sits between the ClientConn and the Balancer.
 //
 // ccBalancerWrapper implements methods corresponding to the ones on the
@@ -57,84 +48,85 @@ const (
 type ccBalancerWrapper struct {
 	// The following fields are initialized when the wrapper is created and are
 	// read-only afterwards, and therefore can be accessed without a mutex.
-	cc   *ClientConn
-	opts balancer.BuildOptions
+	cc       *ClientConn
+	opts     balancer.BuildOptions
+	balancer *gracefulswitch.Balancer
 
-	// Outgoing (gRPC --> balancer) calls are guaranteed to execute in a
-	// mutually exclusive manner as they are scheduled in the serializer. Fields
-	// accessed *only* in these serializer callbacks, can therefore be accessed
-	// without a mutex.
-	balancer        *gracefulswitch.Balancer
+	serializer       *grpcsync.CallbackSerializer
+	serializerCancel context.CancelFunc
+
+	// The following fields are only accessed within the serializer.
 	curBalancerName string
-
-	// mu guards access to the below fields. Access to the serializer and its
-	// cancel function needs to be mutex protected because they are overwritten
-	// when the wrapper exits idle mode.
-	mu               sync.Mutex
-	serializer       *grpcsync.CallbackSerializer // To serialize all outoing calls.
-	serializerCancel context.CancelFunc           // To close the seralizer at close/enterIdle time.
-	mode             ccbMode                      // Tracks the current mode of the wrapper.
 }
 
 // newCCBalancerWrapper creates a new balancer wrapper in idle state. The
 // underlying balancer is not created until the switchTo() method is invoked.
-func newCCBalancerWrapper(cc *ClientConn, bopts balancer.BuildOptions) *ccBalancerWrapper {
+func newCCBalancerWrapper(cc *ClientConn) *ccBalancerWrapper {
+	ctx, cancel := context.WithCancel(cc.ctx)
 	ccb := &ccBalancerWrapper{
-		cc:   cc,
-		opts: bopts,
-		mode: ccbModeIdle,
+		cc: cc,
+		opts: balancer.BuildOptions{
+			DialCreds:        cc.dopts.copts.TransportCredentials,
+			CredsBundle:      cc.dopts.copts.CredsBundle,
+			Dialer:           cc.dopts.copts.Dialer,
+			Authority:        cc.authority,
+			CustomUserAgent:  cc.dopts.copts.UserAgent,
+			ChannelzParentID: cc.channelzID,
+			Target:           cc.parsedTarget,
+		},
+		serializer:       grpcsync.NewCallbackSerializer(ctx),
+		serializerCancel: cancel,
 	}
+	ccb.balancer = gracefulswitch.NewBalancer(ccb, ccb.opts)
 	return ccb
 }
 
 // updateClientConnState is invoked by grpc to push a ClientConnState update to
-// the underlying balancer.
+// the underlying balancer.  This is always executed from the serializer, so
+// it is safe to call into the balancer here.
 func (ccb *ccBalancerWrapper) updateClientConnState(ccs *balancer.ClientConnState) error {
-	ccb.mu.Lock()
-	errCh := make(chan error, 1)
-	// Here and everywhere else where Schedule() is called, it is done with the
-	// lock held. But the lock guards only the scheduling part. The actual
-	// callback is called asynchronously without the lock being held.
-	ok := ccb.serializer.Schedule(func(_ context.Context) {
-		errCh <- ccb.balancer.UpdateClientConnState(*ccs)
+	errCh := make(chan error)
+	ok := ccb.serializer.Schedule(func(ctx context.Context) {
+		defer close(errCh)
+		if ctx.Err() != nil {
+			return
+		}
+		err := ccb.balancer.UpdateClientConnState(*ccs)
+		if logger.V(2) && err != nil {
+			logger.Infof("error from balancer.UpdateClientConnState: %v", err)
+		}
+		errCh <- err
 	})
 	if !ok {
-		// If we are unable to schedule a function with the serializer, it
-		// indicates that it has been closed. A serializer is only closed when
-		// the wrapper is closed or is in idle.
-		ccb.mu.Unlock()
-		return fmt.Errorf("grpc: cannot send state update to a closed or idle balancer")
+		return nil
 	}
-	ccb.mu.Unlock()
-
-	// We get here only if the above call to Schedule succeeds, in which case it
-	// is guaranteed that the scheduled function will run. Therefore it is safe
-	// to block on this channel.
-	err := <-errCh
-	if logger.V(2) && err != nil {
-		logger.Infof("error from balancer.UpdateClientConnState: %v", err)
-	}
-	return err
+	return <-errCh
 }
 
 // updateSubConnState is invoked by grpc to push a subConn state update to the
 // underlying balancer.
 func (ccb *ccBalancerWrapper) updateSubConnState(sc balancer.SubConn, s connectivity.State, err error) {
-	ccb.mu.Lock()
-	ccb.serializer.Schedule(func(_ context.Context) {
+	ccb.serializer.Schedule(func(ctx context.Context) {
+		if ctx.Err() != nil {
+			return
+		}
 		// Even though it is optional for balancers, gracefulswitch ensures
 		// opts.StateListener is set, so this cannot ever be nil.
+		// TODO: delete this comment when UpdateSubConnState is removed.
 		sc.(*acBalancerWrapper).stateListener(balancer.SubConnState{ConnectivityState: s, ConnectionError: err})
 	})
-	ccb.mu.Unlock()
 }
 
+// resolverError is invoked by grpc to push a resolver error to the underlying
+// balancer.  This is always executed from the serializer, so it is safe to call
+// into the balancer here.
 func (ccb *ccBalancerWrapper) resolverError(err error) {
-	ccb.mu.Lock()
-	ccb.serializer.Schedule(func(_ context.Context) {
+	ccb.serializer.Schedule(func(ctx context.Context) {
+		if ctx.Err() != nil {
+			return
+		}
 		ccb.balancer.ResolverError(err)
 	})
-	ccb.mu.Unlock()
 }
 
 // switchTo is invoked by grpc to instruct the balancer wrapper to switch to the
@@ -148,8 +140,10 @@ func (ccb *ccBalancerWrapper) resolverError(err error) {
 // the ccBalancerWrapper keeps track of the current LB policy name, and skips
 // the graceful balancer switching process if the name does not change.
 func (ccb *ccBalancerWrapper) switchTo(name string) {
-	ccb.mu.Lock()
-	ccb.serializer.Schedule(func(_ context.Context) {
+	ccb.serializer.Schedule(func(ctx context.Context) {
+		if ctx.Err() != nil {
+			return
+		}
 		// TODO: Other languages use case-sensitive balancer registries. We should
 		// switch as well. See: https://github.com/grpc/grpc-go/issues/5288.
 		if strings.EqualFold(ccb.curBalancerName, name) {
@@ -157,7 +151,6 @@ func (ccb *ccBalancerWrapper) switchTo(name string) {
 		}
 		ccb.buildLoadBalancingPolicy(name)
 	})
-	ccb.mu.Unlock()
 }
 
 // buildLoadBalancingPolicy performs the following:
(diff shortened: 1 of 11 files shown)


dfawley:
@zasweq I may have misunderstood, so if @arvindbr8 can review this tomorrow then feel free to ignore.  Otherwise if you could review this next week so we can have it done in time to cherry-pick it to the release branch, that would be great.  Thanks!

balancer_wrapper.go
```diff
@@ -57,84 +48,89 @@ const (
...
 		// opts.StateListener is set, so this cannot ever be nil.
+		// TODO: delete this comment when UpdateSubConnState is removed.
 		sc.(*acBalancerWrapper).stateListener(balancer.SubConnState{ConnectivityState: s, ConnectionError: err})
 	})
-	ccb.mu.Unlock()
 }
 
+// resolverError is invoked by grpc to push a resolver error to the underlying
+// balancer.  This is always executed from the serializer, so it is safe to call
+// into the balancer here.
```
zasweq, line 126: The grammar here and your other docstrings of this type feels weird. What is "this"? This overall function isn't run in the serializer, just the early return and call into balancer. Something like "ResolverError operation on balancer is done inside callback serializer, thus is safe." would be more appropriate/concise imo.

balancer_wrapper.go
```diff
@@ -186,113 +183,41 @@ func (ccb *ccBalancerWrapper) buildLoadBalancingPolicy(name string) {
...
-
-	<-done
+	ccb.serializerCancel()
 }
 
-func (ccb *ccBalancerWrapper) isIdleOrClosed() bool {
-	ccb.mu.Lock()
-	defer ccb.mu.Unlock()
-	return ccb.mode == ccbModeIdle || ccb.mode == ccbModeClosed
+// exitIdle invokes the balancer's exitIdle method in the scheduler.
```
zasweq, line 196: Nit: switch scheduler/serializer, more specific.

balancer_wrapper.go
```diff
@@ -186,113 +183,41 @@ func (ccb *ccBalancerWrapper) buildLoadBalancingPolicy(name string) {
...
+		}
+		ccb.balancer.ExitIdle()
+	})
 }
 
 func (ccb *ccBalancerWrapper) NewSubConn(addrs []resolver.Address, opts balancer.NewSubConnOptions) (balancer.SubConn, error) {
-	if ccb.isIdleOrClosed() {
-		return nil, fmt.Errorf("grpc: cannot create SubConn when balancer is closed or idle")
+	ccb.cc.mu.Lock()
+	defer ccb.cc.mu.Unlock()
```
zasweq, line 208: This protection of this whole operation and the race condition it fixes I will need to be onboarded too :).

balancer_wrapper.go
```diff
@@ -325,25 +246,39 @@ func (ccb *ccBalancerWrapper) UpdateAddresses(sc balancer.SubConn, addrs []resol
...
-
+	ccb.mu.Unlock()
 	// Update picker before updating state.  Even though the ordering here does
 	// not matter, it can lead to multiple calls of Pick in the common start-up
 	// case where we wait for ready and then perform an RPC.  If the picker is
 	// updated later, we could call the "connecting" picker when the state is
 	// updated, and then call the "ready" picker after the picker gets updated.
+
+	// Note that there is no need to check if the balancer wrapper was closed,
+	// as we know the graceful switch LB policy will not call cc if it has been
```
zasweq, line 265: Nit: switch cc/ccb, in this context cc can semantically refer to ccb.cc

resolver_wrapper.go
```diff
@@ -152,41 +123,42 @@ func (ccr *ccResolverWrapper) UpdateState(s resolver.State) error {
...
 
 // NewAddress is called by the resolver implementation to send addresses to
 // gRPC.
 func (ccr *ccResolverWrapper) NewAddress(addrs []resolver.Address) {
-	ccr.serializer.Schedule(func(_ context.Context) {
-		ccr.addChannelzTraceEvent(resolver.State{Addresses: addrs, ServiceConfig: ccr.curState.ServiceConfig})
-		ccr.curState.Addresses = addrs
-		ccr.cc.updateResolverState(ccr.curState, nil)
-	})
+	ccr.cc.mu.Lock()
```
zasweq, line 150: Oh I see, protects the operations on the Client Conn with this mutex, so operations from both resolver_wrapper and balancer_wrapper are sync.

clientconn.go
```diff
@@ -321,102 +307,79 @@ func (cc *ClientConn) addTraceEvent(msg string) {
...
-
-	cc.balancerWrapper.exitIdleMode()
-	cc.firstResolveEvent = grpcsync.NewEvent()
 	cc.mu.Unlock()
 
 	// This needs to be called without cc.mu because this builds a new resolver
-	// which might update state or report error inline which needs to be handled
-	// by cc.updateResolverState() which also grabs cc.mu.
-	if err := cc.initResolverWrapper(cc.dopts.copts.TransportCredentials); err != nil {
+	// which might update state or report error inline needs to acquire cc.mu.
```
zasweq, line 330: which needs

clientconn.go
```diff
@@ -321,102 +307,79 @@ func (cc *ClientConn) addTraceEvent(msg string) {
...
 	}
 
 	cc.addTraceEvent("exiting idle mode")
 	return nil
 }
 
+// initIdleStateLocked initializes common state to how it should be while idle.
+func (cc *ClientConn) initIdleStateLocked() {
+	cc.resolverWrapper = newCCResolverWrapper(cc)
+	cc.balancerWrapper = newCCBalancerWrapper(cc)
```
zasweq, line 342: Oh here's the callsite I talked about in an earlier comment. Again optional: briefly discuss why you need to recreate everything in this scenario.

internal/idle/idle.go
```diff
@@ -69,8 +53,7 @@ type manager struct {
 	// Can be accessed without atomics or mutex since these are set at creation
 	// time and read-only after that.
 	enforcer Enforcer // Functionality provided by grpc.ClientConn.
-	timeout  int64    // Idle timeout duration nanos stored as an int64.
-	logger   grpclog.LoggerV2
```
zasweq, line 73: Hmm guess you dropped the logger here too.

resolver_wrapper.go
```diff
@@ -23,127 +23,98 @@ import (
...
-}
-
 // ccResolverWrapper is a wrapper on top of cc for resolvers.
 // It implements resolver.ClientConn interface.
 type ccResolverWrapper struct {
-	// The following fields are initialized when the wrapper is created and are
-	// read-only afterwards, and therefore can be accessed without a mutex.
-	cc                  resolverStateUpdater
-	channelzID          *channelz.Identifier
+	// The following fields are read-only.
```
zasweq, line 36: s/"are initialized at build time, and read only after."?

resolver_wrapper.go
```diff
@@ -23,127 +23,98 @@ import (
...
-
-	// ccr.resolver field is set only after the call to Build() returns. But in
-	// the process of building, the resolver may send an error update which when
-	// propagated to the balancer may result in a re-resolution request.
-	if ccr.closed || ccr.resolver == nil {
-		return
-	}
-	ccr.resolver.ResolveNow(o)
+	ccr.serializer.Schedule(func(ctx context.Context) {
+		if ctx.Err() != nil || ccr.resolver == nil {
```
zasweq, line 85: Why closed bool in balancer wrapper and below and nil condtional/invariant here in resolver_wrapper?

zasweq (commented):
Did a style pass. Technically, I have a very minute understanding of the original Exit Idle code, and this fix. I think since Easwar is back next week he should be primary reviewer since he wrote the ExitIdle code. We should also have a technical sync where you briefly describe to us the original code, all the issues, the fix and why it overall fixes.

internal/idle/idle.go
```diff
@@ -93,52 +76,51 @@ type manager struct {
...
 		return
 	}
 
 	// It is safe to ignore the return value from Reset() because this method is
 	// only ever called from the timer callback, which means the timer has
 	// already fired.
-	m.timer.Reset(d)
+	if m.timer != nil {
+		m.timer.Stop()
+	}
```
easwars, line 104: Couple of things here:
- The comment at the top of this block still talks about `Reset()`, which is not called anymore
- Why do we need to `Stop` the timer instead of calling `Reset()`?
  - this method is only ever called from the timer callback or when we are exiting idle. In the former case, the timer has already fired and in the latter case, the timer is not running. Can we get away with not doing `Reset` or `Stop` and simply creating a new timer here? That should simplify the code and get rid of the comment.

internal/idle/idle.go
```diff
@@ -93,52 +76,51 @@ type manager struct {
...
 // resetIdleTimer resets the idle timer to the given duration. This method
 // should only be called from the timer callback.
-func (m *manager) resetIdleTimer(d time.Duration) {
-	m.idleMu.Lock()
-	defer m.idleMu.Unlock()
-
-	if m.timer == nil {
-		// Only close sets timer to nil. We are done.
+func (m *Manager) resetIdleTimerLocked(d time.Duration) {
+	if m.isClosed() || m.timeout == 0 {
```
easwars, line 95: Should we also return early with a check for `m.timeout == 0` in `OnCallXxx` and save cycles when idleness is disabled.

And in `m.ExitIdleMode`, we probably need a check like this at the top?
```
if m.timeout == 0 {
  m.actuallyIdle = false
  m.activeCallsCount = 0
  return
}
```

easwars:
@dfawley : I've only looked through the idleness code so far. But assigning back to you to make the requested changes while I will continue to review other changes.

dfawley (commented):
All comments address (I think!)

balancer_wrapper.go
```diff
@@ -57,84 +48,89 @@ const (
...
-
-	// We get here only if the above call to Schedule succeeds, in which case it
-	// is guaranteed that the scheduled function will run. Therefore it is safe
-	// to block on this channel.
-	err := <-errCh
-	if logger.V(2) && err != nil {
-		logger.Infof("error from balancer.UpdateClientConnState: %v", err)
-	}
-	return err
+	return <-errCh
```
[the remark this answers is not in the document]
zasweq: Oh yeah my bad I see it now.

internal/idle/idle.go
zasweq: Sounds good.

balancer_wrapper.go
```diff
@@ -57,84 +49,87 @@ const (
...
-	// mu guards access to the below fields. Access to the serializer and its
-	// cancel function needs to be mutex protected because they are overwritten
-	// when the wrapper exits idle mode.
-	mu               sync.Mutex
-	serializer       *grpcsync.CallbackSerializer // To serialize all outoing calls.
-	serializerCancel context.CancelFunc           // To close the seralizer at close/enterIdle time.
-	mode             ccbMode                      // Tracks the current mode of the wrapper.
+	cc               *ClientConn
+	opts             balancer.BuildOptions
+	balancer         *gracefulswitch.Balancer
```
zasweq, line 54: Perhaps mention calls into this are protected by serializer?

balancer_wrapper.go
zasweq: Right, but the graceful switch and this component hold a field called cc. Which cc is it referring to? I don't really feel too strongly though

internal/idle/idle.go
easwars: > We always need to call `Stop` to prevent a timer leak

I know for a fact that we need to `Stop` a `Ticker` to prevent a leak, but i thought that wasn't the case with a `Timer`.

clientconn.go
```diff
@@ -188,20 +185,9 @@ func newClient(target string, opts ...DialOption) (conn *ClientConn, err error)
...
-		ChannelzParentID: cc.channelzID,
-		Target:           cc.parsedTarget,
-	})
 
-	// Configure idleness support with configured idle timeout or default idle
-	// timeout duration. Idleness can be explicitly disabled by the user, by
-	// setting the dial option to 0.
-	cc.idlenessMgr = idle.NewManager(idle.ManagerOptions{Enforcer: (*idler)(cc), Timeout: cc.dopts.idleTimeout, Logger: logger})
+	cc.idlenessMgr = idle.NewManager((*idler)(cc), cc.dopts.idleTimeout)
+	cc.initIdleStateLocked()
```
zasweq, line 190: This method is suffixed with Locked but I see no mutex being held within this method.

clientconn.go
```diff
@@ -321,102 +307,80 @@ func (cc *ClientConn) addTraceEvent(msg string) {
...
 	}
 
 	cc.addTraceEvent("exiting idle mode")
 	return nil
 }
 
+// initIdleStateLocked initializes common state to how it should be while idle.
+func (cc *ClientConn) initIdleStateLocked() {
+	cc.resolverWrapper = newCCResolverWrapper(cc)
+	cc.balancerWrapper = newCCBalancerWrapper(cc)
```
zasweq, line 343: It feels weird to recreate whole components, but I guess idle transitions aren't happening a lot in practice and also it simplifies code.

internal/idle/idle.go
easwars: > we still need to exit idle mode in `OnCallBegin` even when disabled

True

> We can short circuit the accounting logic

Do you think adding that short circuit would be better done as a follow-up PR since it doesn't affect correctness. If so, could we add a TODO. Or do you think it's not worth adding the short circuit?

clientconn.go
```diff
@@ -637,55 +600,23 @@ type ClientConn struct {
...
+	resolverWrapper *ccResolverWrapper         // Always recreated whenever entering idle to simplify Close.
+	balancerWrapper *ccBalancerWrapper         // Always recreated whenever entering idle to simplify Close.
 	sc              *ServiceConfig             // Latest service config received from the resolver.
 	conns           map[*addrConn]struct{}     // Set to nil on close.
 	mkp             keepalive.ClientParameters // May be updated upon receipt of a GoAway.
-	idlenessState   ccIdlenessState            // Tracks idleness state of the channel.
-	exitIdleCond    *sync.Cond                 // Signalled when channel exits idle.
+	// firstResolveEvent is used to track whether the name resolver sent us at
+	// least one update. RPCs block on this event.  May be accessed without mu
+	// if we know we cannot enter idle mode while accessing it.
```
zasweq, line 613: Can you elaborate what "cannot enter idle mode" more specifically technically means here?

clientconn.go
```diff
@@ -1174,12 +1108,12 @@ func (cc *ClientConn) applyServiceConfigAndBalancer(sc *ServiceConfig, configSel
...
 	cc.mu.RUnlock()
-	if r == nil {
-		return
-	}
-	go r.resolveNow(o)
+}
+
+func (cc *ClientConn) resolveNowLocked(o resolver.ResolveNowOptions) {
+	cc.resolverWrapper.resolveNow(o)
 }
```
zasweq, line 1117: Can you make this one and grab lock at callsite? Any reason for two?

resolver_balancer_ext_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"google.golang.org/grpc/resolver"
+	"google.golang.org/grpc/resolver/manual"
+)
+
+// TestResolverBalancerInteraction tests:
+// 1. resolver.Builder.Build()
+// 2. resolver.ClientConn.UpdateState()
+// 3. balancer.Balancer.UpdateClientConnState()
+// 4. balancer.ClientConn.ResolveNow()
+// 5. resolver.Resolver.ResolveNow()
```
zasweq, line 42: Put some arrows so I know these operations are sync. Right now looks like independent operations, not within each other.

zasweq (approved):
LGTM module some nits

internal/idle/idle.go
```diff
@@ -238,42 +235,44 @@ func (m *manager) OnCallBegin() error {
...
+	if m.isClosed() || !m.actuallyIdle {
+		// This can happen in three scenarios:
 		// - handleIdleTimeout() set the calls count to -math.MaxInt32 and called
 		//   tryEnterIdleMode(). But before the latter could grab the lock, an RPC
 		//   came in and OnCallBegin() noticed that the calls count is negative.
 		// - Channel is in idle mode, and multiple new RPCs come in at the same
 		//   time, all of them notice a negative calls count in OnCallBegin and get
 		//   here. The first one to get the lock would got the channel to exit idle.
+		// - Channel is in idle mode, and the user calls Connect which calls
+		//   m.ExitIdleMode.
```
easwars, line 254: Shouldn't this be: `Channel is not in idle mode`?

internal/idle/idle.go
```diff
@@ -203,22 +176,46 @@ func (m *manager) tryEnterIdleMode() bool {
...
+func (m *Manager) OnCallBegin() error {
 	if m.isClosed() {
 		return nil
 	}
-
+	if m.timeout == 0 {
+		// When the manager is disabled (timeout==0), we just exit idle mode if
+		// needed and return.
+		return m.ExitIdleMode()
+	}
```
easwars, line 218: If idleness is disabled, this means that for every RPC, we would be taking the lock (in `m.ExitIdleMode()`), which seems like a lot of wasted effort and a potential for lock contention issues in environments with a high QPS.

internal/idle/idle.go
dfawley: I believe this PR does this now.  Maybe I've missed something, in which case let me know.

internal/idle/idle.go
```diff
@@ -203,22 +176,46 @@ func (m *manager) tryEnterIdleMode() bool {
...
-		return false
-	}
+	m.enforcer.EnterIdleMode()
 
 	// Successfully entered idle mode.
 	m.actuallyIdle = true
 	return true
 }
 
+func (m *Manager) EnterIdleModeForTesting() {
```
easwars, line 186: The only reason for having this function defined in the internal package was to be able to test this functionality from a test (wherein, the actual functionality is not exported). Now, with this change, we have a different code path for testing and a different one for non-testing workloads. This makes me feel nervous because our tests are now using a different code and also, we need to ensure that we maintain this code to be in sync with the code that gets run to enter idle mode in real workloads.

internal/idle/idle.go
dfawley: Hmm, good point.  So should we just add back in all the accounting for the disabled idleness manager?

And this means we take this lock on every RPC except when there is a concurrent call already running.  I wonder if we can improve this in the base case, too?

internal/idle/idle.go
dfawley: Fixed.

resolver_balancer_ext_test.go
dfawley: Done

internal/idle/idle.go
dfawley: Fixed.

resolver_wrapper.go
```diff
@@ -23,127 +23,98 @@ import (
...
 	"google.golang.org/grpc/internal/pretty"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/serviceconfig"
 )
 
-// resolverStateUpdater wraps the single method used by ccResolverWrapper to
-// report a state update from the actual resolver implementation.
-type resolverStateUpdater interface {
-	updateResolverState(s resolver.State, err error) error
-}
```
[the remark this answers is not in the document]
dfawley: Do you have suggestions for how else to handle this?  I think these things are very closely tied together, and without this, we can have deadlocks due to lock ordering.

resolver_wrapper.go
```diff
@@ -23,127 +23,98 @@ import (
...
-// ccResolverWrapper.
-type ccResolverWrapperOpts struct {
-	target     resolver.Target       // User specified dial target to resolve.
-	builder    resolver.Builder      // Resolver builder to use.
-	bOpts      resolver.BuildOptions // Resolver build options to use.
-	channelzID *channelz.Identifier  // Channelz identifier for the channel.
 }
 
 // newCCResolverWrapper uses the resolver.Builder to build a Resolver and
 // returns a ccResolverWrapper object which wraps the newly built resolver.
```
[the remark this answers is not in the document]
dfawley: Updated

resolver_wrapper.go
```diff
@@ -23,127 +23,98 @@ import (
...
-	// Fields accessed *only* in these serializer callbacks, can therefore be
-	// accessed without a mutex.
-	curState resolver.State
+	serializer          *grpcsync.CallbackSerializer
+	serializerCancel    context.CancelFunc
+
+	resolver resolver.Resolver // only accessed within the serializer
 
-	// mu guards access to the below fields.
+	// The following fields are protected by mu.  Caller must take cc.mu before
```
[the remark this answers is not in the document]
dfawley: Yeah I think we'll need to sort things out at that time.  I believe the AI of splitting up `cc.mu` should make things simpler, though, as we'd probably still have one big top-level lock and all the others would be for very short-term accessing of specific fields.

easwars:
Thinking about the original deadlock:

Goroutine1 is entering IDLE:
- idleManager calls into the `ClientConn` and asks it to enter IDLE mode
- here in `cc.enterIdleMode`, we hold `cc.mu` for the whole duration and the comment says:
```
	// TODO: Currently, we close the resolver wrapper upon entering idle mode
	// and create a new one upon exiting idle mode. This means that the
	// `cc.resolverWrapper` field would be overwritten everytime we exit idle
	// mode. While this means that we need to hold `cc.mu` when accessing
	// `cc.resolverWrapper`, it makes the code simpler in the wrapper. We should
	// try to do the same for the balancer and picker wrappers too.
```
- and `cc.resolverWrapper.close` waits for all serializer callbacks to complete before returning

Goroutine2 is handling an update from the resolver in `ccResolverWrapper`:
- this schedules a callback to the serializer which calls `cc.updateResolverState` which tries to grab `cc.mu`
- and hence the deadlock

I feel that the approach of recreating a wrapper when entering and exiting IDLE was a bad idea now. And worse idea was to call one of the wrapper's blocking method with `cc.mu` being held.

And in line with the TODO, we have also made the balancer wrapper be recreated for every trip out of IDLE in this PR. I feel that it would be better to create the wrappers once and have them handle the trip into and out of IDLE (by destroying and creating the thing that they wrap respectively), and ensure that any of their blocking methods are only ever called from the `ClientConn` without holding `cc.mu`.

dfawley:
> I feel that the approach of recreating a wrapper when entering and exiting IDLE was a bad idea now.

I don't think that is a problem.  We need to recreate _some kind_ of channel wrapper when the channel goes idle, because we need to prevent the old resolver instance's calls from being honored if it calls back into the channel any time later, and there is no way to distinguish what's calling us besides using the wrapper itself.  Maybe there's more room for changes, e.g. not using the same "cc wrapper" in order to call _into_ the resolver/balancer wrapper, but that would be room for future investigation; I think this model is okay enough as-is.

> ensure that any of their blocking methods are only ever called from the ClientConn without holding cc.mu.

This is the key, for sure.  For the resolver, the only thing that blocks (AKA needs to be synchronous) is `start`, and for the balancer, the only thing that blocks is `updateClientConnState`, and in both of these cases we release cc.mu before we call them.

internal/idle/idle.go
dfawley: I just removed all short circuiting of the accounting for m.timeout==0.  We can save any other improvements for later.

internal/idle/idle_test.go
```diff
@@ -306,51 +309,59 @@ const (
...
+	state   *racyState // Accessed atomically.
+	started bool
 }
 
 // ExitIdleMode sets the internal state to stateExitedIdle. We should only ever
 // exit idle when we are currently in idle.
 func (ri *racyEnforcer) ExitIdleMode() error {
+	// Set only on the initial ExitIdleMode
+	if ri.started == false {
+		if *ri.state != stateInitial {
```
easwars, line 322: Don't we have to do a `atomic.Load(ri.State)` here?

easwars (commented):
Happy with the changes in the `idle` package. Continuing to review other changes.

resolver_wrapper.go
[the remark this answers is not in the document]
easwars: I tried spending a little bit of time to make the resolver wrapper handle enter and exit idle mode. The wrapper in this change also handles creation of the new resolver and skipping updates from the old resolver. It needs another wrapper, I called it `genWrapper` to indicate that it wraps one generation of the underlying resolver. The changes are not very pretty, but I guess, with a little more time, it can be made to look a little better. But it does seem to work:
https://github.com/grpc/grpc-go/compare/master...easwars:resolver_wrapper_handle_enter_exit_idle?expand=1

Taking the approach of making all the wrappers handle enter and exit idle modes does make the wrappers a bit more verbose, but I feel it might be less complicated in the sense that the wrappers won't be reaching into and grabbing `cc.mu`, and won't have any imposed lock ordering that spans types. Let me know what you think.

internal/idle/idle_test.go
dfawley: I believe if there _is_ a bug, then this would show up as a data race.  But yes, it'd be better to do it atomically, and best to do as a cmpswp, but expecting the previous state to be `Initial`.  Done.

resolver_wrapper.go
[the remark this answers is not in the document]
dfawley: I have some concerns about your diffs:

- I am pretty sure the same deadlock is still there:

1. Resolver calls `UpdateState()`, grabs `ccr.mu`, enqueues work into serializer.
1. Channel does `EnterIdleMode` and takes `cc.mu`.
1. `UpdateState`'s work item gives up `ccr.mu`.
1. Channel calls `ccr.enterIdleMode` which grabs `ccr.mu` and then blocks on `serializer.Done()`.
1. Serializer calls `UpdateState`, which needs `cc.mu`.

- The `genWrapper` still reaches into its parent and grabs its mutex, which is potentially just as problematic readability/correctness-wise as the `ccr` reaching into its parent for `cc.mu`.

- How are you synchronizing calls _into_ the resolver?  It seems like `resolver.ResolveNow` and `resolver.Close` could happen concurrently or `ResolveNow`-after`Close`, which is not desired / illegal.

- The use of the serializer is odd.  You always hold a lock and release it once you're in the serializer, meaning only one thing can be in it at a time.  Why not just remove it and get the lock?  That would remove the deadlock that happens above, but.....

- I think you can also end up with the resolver able to call `cc.updateResolverState` _while_ entering idle mode, and without other changes, could cause problems.

This change went through a lot of iteration and has a number of improvements vs. master, and while I'm fine with continuing discussions about other ways of doing things, this fixes a deadlock that is blocking our release, so I would appreciate it if we can push forward with this approach (unless you spot any non-theoretical problems with it), and consider refining or revising it as follow-ups.

resolver_balancer_ext_test.go
```diff
@@ -0,0 +1,122 @@
...
+	enterIdle(cc)
+	const errStr = "test error from resolver builder"
+	t.Log("pushing res err")
+	resErrCh <- errors.New(errStr)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	if err := cc.Invoke(ctx, "/a/b", nil, nil); err == nil || !strings.Contains(err.Error(), errStr) {
+		t.Fatalf("Invoke = %v; want %v", err, errStr)
+	}
+}
```
easwars, line 122: Could we also please add a test case for the deadlock scenario?

balancer_wrapper.go
```diff
@@ -186,113 +184,41 @@ func (ccb *ccBalancerWrapper) buildLoadBalancingPolicy(name string) {
...
-		ccb.mode = ccbModeActive
-		channelz.Info(logger, ccb.cc.channelzID, "ccBalancerWrapper: exiting idle mode")
-
+		ccb.balancer.Close()
+		ccb.balancer = nil
 	})
-	ccb.mu.Unlock()
-
-	<-done
+	ccb.serializerCancel()
```
easwars, line 194: I see that the wait for `<-ccb.serializer.Done()` has moved to `cc.Close()` and `cc.enterIdleMode()`. The convention that we generally try to use is that any type's close/stop method should be blocking and should ensure that when it returns, all associated resources are released. 

What is the reasoning behind this change? Can this be documented explicitly.

balancer_wrapper.go
```diff
@@ -57,84 +49,89 @@ const (
...
-	// cancel function needs to be mutex protected because they are overwritten
-	// when the wrapper exits idle mode.
-	mu               sync.Mutex
-	serializer       *grpcsync.CallbackSerializer // To serialize all outoing calls.
-	serializerCancel context.CancelFunc           // To close the seralizer at close/enterIdle time.
-	mode             ccbMode                      // Tracks the current mode of the wrapper.
+	// The following field is protected by mu.  Caller must take cc.mu before
+	// taking mu.
+	mu     sync.Mutex
+	closed bool
```
easwars, line 65: I don't see this being set to `true` anywhere? Am I missing something?

balancer_wrapper.go
```diff
@@ -186,113 +184,41 @@ func (ccb *ccBalancerWrapper) buildLoadBalancingPolicy(name string) {
...
 func (ccb *ccBalancerWrapper) NewSubConn(addrs []resolver.Address, opts balancer.NewSubConnOptions) (balancer.SubConn, error) {
-	if ccb.isIdleOrClosed() {
-		return nil, fmt.Errorf("grpc: cannot create SubConn when balancer is closed or idle")
+	ccb.cc.mu.Lock()
+	defer ccb.cc.mu.Unlock()
+
+	ccb.mu.Lock()
+	if ccb.closed {
+		ccb.mu.Unlock()
+		return nil, errConnIdling
```
easwars, line 214: Why is this `errConnIdling`? Should be something else saying the balancer is closed, right?

clientconn.go
```diff
@@ -637,55 +599,25 @@ type ClientConn struct {
...
 	conns           map[*addrConn]struct{}     // Set to nil on close.
 	mkp             keepalive.ClientParameters // May be updated upon receipt of a GoAway.
-	idlenessState   ccIdlenessState            // Tracks idleness state of the channel.
-	exitIdleCond    *sync.Cond                 // Signalled when channel exits idle.
+	// firstResolveEvent is used to track whether the name resolver sent us at
+	// least one update. RPCs block on this event.  May be accessed without mu
+	// if we know we cannot be asked to enter idle mode while accessing it (e.g.
+	// when the idle manager has already been closed, or if we are already
+	// entering idle mode).
+	firstResolveEvent *grpcsync.Event
```
easwars, line 615: But this is also accessed on the fast path in `cc.waitForResolvedAddrs` without `cc.mu`.

clientconn.go
```diff
@@ -725,10 +657,15 @@ func (cc *ClientConn) GetState() connectivity.State {
 // Notice: This API is EXPERIMENTAL and may be changed or removed in a later
 // release.
 func (cc *ClientConn) Connect() {
-	cc.exitIdleMode()
+	if err := cc.idlenessMgr.ExitIdleMode(); err != nil {
+		cc.addTraceEvent(fmt.Sprintf("error exiting idle mode: %v", err))
```
easwars, line 661: This error message will contain repeated text since the error returned from the idleness manager already is:
```
return fmt.Errorf("channel failed to exit idle mode: %w", err)
```

resolver_wrapper.go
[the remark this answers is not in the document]
easwars: My diff was not put together with the rigor required when actually sending for review. So, apologies for the many things that I missed (actually I didn't realize I missed so many things).

> so I would appreciate it if we can push forward with this approach (unless you spot any non-theoretical problems with it), and consider refining or revising it as follow-ups

The main problem I have is that now `cc.mu` is held when handling every update from the balancer and resolver. It is also held every time `cc` has to call into the resolver or the balancer through the appropriate wrapper. This makes me very nervous. Although I cannot point out any actual possible races or deadlocks (and clearly all our tests are passing), I feel that maybe we could have gone with the minimum change required to fix the deadlock and made the other changes subsequently. Also, not being around when the many iterations happened clearly doesn't help me.

resolver_balancer_ext_test.go
dfawley: I honestly had very low hopes for this test, which is why I didn't write it alreayd, but it does seem to be quite effective at hitting the deadlock.

resolver_wrapper.go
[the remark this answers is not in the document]
dfawley: > The main problem I have is that now cc.mu is held when handling every update from the balancer and resolver

This is a good thing.  We block the channel from doing anything big and scary (e.g. entering idle mode) while the plugin is updating our state.

> It is also held every time cc has to call into the resolver or the balancer through the appropriate wrapper

I don't think this is true.  They need to grab it to read the wrapper, since the wrappers can come and go while entering/exiting idle, but these are all non-RPC events and grabbing the mutex during these events should be reasonable.

> Although I cannot point out any actual possible races or deadlocks (and clearly all our tests are passing), I feel that maybe we could have gone with the minimum change required to fix the deadlock

I'm not sure what that minimum change would be.  Like I said, this change took a _lot_ of iterations, and I definitely tried to start small, but kept running into other deadlocks or races.  I think the sync model we have now has some fundamental flaws that need structural changes to address.  It's not a simple problem, either, because we have a synchronous path from the channel to the resolver, back to the channel, to the LB policy, and back to the resolver through the channel, and during the call to the LB policy, we need to be able to handle other calls coming from it.

balancer_wrapper.go
dfawley: Added tests for these things as well.

easwars (approved):
Minor comments.

balancer_wrapper.go
```diff
@@ -32,21 +32,13 @@ import (
...
-)
-
 // ccBalancerWrapper sits between the ClientConn and the Balancer.
 //
 // ccBalancerWrapper implements methods corresponding to the ones on the
 // balancer.Balancer interface. The ClientConn is free to call these methods
 // concurrently and the ccBalancerWrapper ensures that calls from the ClientConn
-// to the Balancer happen synchronously and in order.
+// to the Balancer happen in order by performing them in the serializer, without
+// any mutexes held.
```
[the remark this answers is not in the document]
dfawley: This comment is about how the calls into the plugin, themselves, are made without cc.mu held.  Not that calls into ccb must not hold cc.mu.

Probably we should just document the functions.  (Done.)

balancer_wrapper.go
```diff
@@ -184,115 +168,49 @@ func (ccb *ccBalancerWrapper) buildLoadBalancingPolicy(name string) {
 	ccb.curBalancerName = builder.Name()
 }
 
+// close initiates async shutdown of the wrapper.  To determine the wrapper has
+// finished shutting down, the channel should block on ccb.serializer.Done()
+// without cc.mu held.
```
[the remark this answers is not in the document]
dfawley: Done.

resolver_balancer_ext_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+// TestEnterIdleDuringResolverUpdateState tests a scenario that used to deadlock
+// while calling UpdateState at the same time as the resolver being closed while
+// the channel enters idle mode.
+func (s) TestEnterIdleDuringResolverUpdateState(t *testing.T) {
+	enterIdle := internal.EnterIdleModeForTesting.(func(*grpc.ClientConn))
+	const name = "testeidrus"
```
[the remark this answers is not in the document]
dfawley: I tried that but it didn't work because of case/etc.  But I can fix it.  Done.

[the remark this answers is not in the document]
dfawley: I tried that originally, but it doesn't work because of case/underscores.  I can fix it, though.  Done.

resolver_balancer_ext_test.go
```diff
@@ -0,0 +1,264 @@
+/*
+ *
+ * Copyright 2014 gRPC authors.
```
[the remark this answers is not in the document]
dfawley: O_O

balancer_wrapper.go
```diff
@@ -313,10 +231,6 @@ func (ccb *ccBalancerWrapper) RemoveSubConn(sc balancer.SubConn) {
 }
 
 func (ccb *ccBalancerWrapper) UpdateAddresses(sc balancer.SubConn, addrs []resolver.Address) {
-	if ccb.isIdleOrClosed() {
```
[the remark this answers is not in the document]
dfawley: It's more relevant whether the SubConn/addrConn was closed and not whether the balancer was closed.  And we do that check here:

https://github.com/grpc/grpc-go/blob/93389b7f02e9309ff7a586ca7fc33592809969d7/clientconn.go#L1031

PR #6809, opened 2023-11-17
interop: Custom creds for stress test client

temawi:
Allows custom, Google specific credentials to be specified from the command line.

Also adds a final printout to stdout with the total amount of calls made. Stress tests can use that to determine the average QPS of a run.

RELEASE NOTES: none

### DIFF
--- a/interop/stress/client/main.go
+++ b/interop/stress/client/main.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+  "os"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


temawi:
@dfawley Could you please help get this merged?

interop/stress/client/main.go
```diff
@@ -76,7 +87,8 @@ func parseTestCases(testCaseString string) []testCaseWithWeight {
 			panic(fmt.Sprintf("invalid test case with weight: %s", str))
 		}
 		// Check if test case is supported.
-		switch testCase[0] {
+    testCaseName := strings.ToLower(testCase[0])
+		switch strings.ToLower(testCaseName) {
```
dfawley, line 91: Repeated `ToLower`.

Consider renaming `testCase` to something else like `testCaseNameAndWeight` since this new variable name plus the old one were confusing to me.

interop/stress/client/main.go
```diff
@@ -276,7 +289,15 @@ func logParameterInfo(addresses []string, tests []testCaseWithWeight) {
...
 	var opts []grpc.DialOption
-	if useTLS {
+	if *customCredentialsType != "" {
+		if *customCredentialsType == googleDefaultCredsName {
+			opts = append(opts, grpc.WithCredentialsBundle(google.NewDefaultCredentials()))
+		} else if *customCredentialsType == computeEngineCredsName {
+			opts = append(opts, grpc.WithCredentialsBundle(google.NewComputeEngineCredentials()))
+		} else {
+			logger.Fatalf("Unknown custom credentials: %v", *customCredentialsType)
+		}
```
dfawley, line 299: Consider:

```go
switch *customCredentialsType {
	case googleDefaultCredsName:
	case computeEngineCredsName:
	case "":
		if useTLS {
		}
	default:
		// unknown custom creds error
}
```

Also if we can change the spec here, consider `custom_creds_type=="tls"` means to use tls instead of having a separate flag for it.  Having two flags for one setting is problematic.

temawi: That's how the flags work in the interop test client across languages as well as the java stress test one. Makes sense to revisit, but I think that would be a broad change outside the scope of this PR.

interop/stress/client/main.go
temawi: Renamed the variable and removed the extra call to `ToLower`

PR #6810, opened 2023-11-20
Mark old CRL APIs as deprecated

gtcooke94:
With the addition of the CRL Provider interface and implementations, these old CRL APIs should not be used anymore.

RELEASE NOTES: n/a


zasweq (approved):
LGTM.

PR #6827, opened 2023-12-01
interop: Replace context.Background() with passed ctx

### LINKED ISSUE #6763: interop: interop functions should not call context.Background() for RPCs
E.g. https://github.com/grpc/grpc-go/blob/70f1a4045da95b93f73b6dbdd7049f3f053c0680/interop/test_utils.go#L82-L83

This function and all others in this file should accept a context, which would allow passing RPC timeouts, metadata, etc.  `context.Background` is really only supposed to be called by `main`/`TestXyz` functions wherever possible.

Aditya-Sood:
Fixes #6763 

RELEASE NOTES: none

### DIFF
--- a/interop/client/client.go
+++ b/interop/client/client.go
@@ -169,6 +169,8 @@ func main() {
 		logger.Fatalf("only one of TLS, ALTS, google default creds, or compute engine creds can be used")
 	}
 
+	ctx := context.Background()
+
 	var credsChosen credsMode
 	switch {
 	case *useTLS:
@@ -242,7 +244,7 @@ func main() {
 			}
 			opts = append(opts, grpc.WithPerRPCCredentials(jwtCreds))
 		} else if *testCase == "oauth2_auth_token" {
-			opts = append(opts, grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: oauth2.StaticTokenSource(interop.GetToken(*serviceAccountKeyFile, *oauthScope))}))
+			opts = append(opts, grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: oauth2.StaticTokenSource(interop.GetToken(ctx, *serviceAccountKeyFile, *oauthScope))}))
 		}
 	}
 	if len(*serviceConfigJSON) > 0 {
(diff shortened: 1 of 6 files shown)


interop/client/client.go
```diff
@@ -267,103 +269,103 @@ func main() {
 	tc := testgrpc.NewTestServiceClient(conn)
 	switch *testCase {
 	case "empty_unary":
-		interop.DoEmptyUnaryCall(tc)
+		interop.DoEmptyUnaryCall(tc, ctx)
```
dfawley, line 272: Please make `ctx` the first parameter according to common Go style.

interop/test_utils.go
```diff
@@ -724,9 +724,9 @@ func doOneSoakIteration(ctx context.Context, tc testgrpc.TestServiceClient, rese
 // If resetChannel is false, then each RPC will be performed on tc. Otherwise, each RPC will be performed on a new
 // stub that is created with the provided server address and dial options.
 // TODO(mohanli-ml): Create SoakTestOptions as a parameter for this method.
-func DoSoakTest(tc testgrpc.TestServiceClient, serverAddr string, dopts []grpc.DialOption, resetChannel bool, soakIterations int, maxFailures int, soakRequestSize int, soakResponseSize int, perIterationMaxAcceptableLatency time.Duration, minTimeBetweenRPCs time.Duration, overallDeadline time.Time) {
+func DoSoakTest(tc testgrpc.TestServiceClient, ctx context.Context, serverAddr string, dopts []grpc.DialOption, resetChannel bool, soakIterations int, maxFailures int, soakRequestSize int, soakResponseSize int, perIterationMaxAcceptableLatency time.Duration, minTimeBetweenRPCs time.Duration, overallDeadline time.Time) {
 	start := time.Now()
-	ctx, cancel := context.WithDeadline(context.Background(), overallDeadline)
+	ctx, cancel := context.WithDeadline(ctx, overallDeadline)
```
dfawley, line 729: Ideally the caller would do this and not pass `overallDeadline`, and check `ctx.Err()` instead of `time.Now.After(overallDeadline)`.  If you don't mind making that change, too.  Or we can do it separately if it turns out to be too much.

Thanks!

Aditya-Sood:
hi @dfawley, could you please reopen the PR?
I have pushed the code review changes now

interop/test_utils.go
```diff
@@ -724,9 +724,9 @@ func doOneSoakIteration(ctx context.Context, tc testgrpc.TestServiceClient, rese
 // If resetChannel is false, then each RPC will be performed on tc. Otherwise, each RPC will be performed on a new
 // stub that is created with the provided server address and dial options.
 // TODO(mohanli-ml): Create SoakTestOptions as a parameter for this method.
-func DoSoakTest(tc testgrpc.TestServiceClient, serverAddr string, dopts []grpc.DialOption, resetChannel bool, soakIterations int, maxFailures int, soakRequestSize int, soakResponseSize int, perIterationMaxAcceptableLatency time.Duration, minTimeBetweenRPCs time.Duration, overallDeadline time.Time) {
+func DoSoakTest(ctx context.Context, cancel context.CancelFunc, tc testgrpc.TestServiceClient, serverAddr string, dopts []grpc.DialOption, resetChannel bool, soakIterations int, maxFailures int, soakRequestSize int, soakResponseSize int, perIterationMaxAcceptableLatency time.Duration, minTimeBetweenRPCs time.Duration) {
```
dfawley, line 727: You should not pass a `CancelFunc`.  If this function needs it, it should do `ctx, cancel := context.WithCancel(ctx); defer cancel();` to avoid cancelling the parent's context.

But in this case, I don't think it needs it anyway.  The `defer cancel()` should be done in the caller, not by this function.

dfawley (commented):
Sorry for the delay.  Mostly LGTM; just one thing (inline).

Aditya-Sood:
thanks @dfawley, I've pushed the change in the latest commit

one query - is the use of `CancelFunc` unidirectional (parent to child ctx)? or can a child be allowed to cancel the ctx for itself and its siblings?

dfawley:
> one query - is the use of CancelFunc unidirectional (parent to child ctx)? or can a child be allowed to cancel the ctx for itself and its siblings?

Derived contexts cannot cancel their parents/siblings and cannot be cancelled via a method on the context itself (by design).

The thing that creates a context should be the only thing able to cancel it.

interop/client/client.go
```diff
@@ -358,10 +358,12 @@ func main() {
 		interop.DoPickFirstUnary(ctx, tc)
 		logger.Infoln("PickFirstUnary done")
 	case "rpc_soak":
-		interop.DoSoakTest(ctxWithDeadline, cancel, tc, serverAddr, opts, false /* resetChannel */, *soakIterations, *soakMaxFailures, *soakRequestSize, *soakResponseSize, time.Duration(*soakPerIterationMaxAcceptableLatencyMs)*time.Millisecond, time.Duration(*soakMinTimeMsBetweenRPCs)*time.Millisecond)
+		interop.DoSoakTest(ctxWithDeadline, tc, serverAddr, opts, false /* resetChannel */, *soakIterations, *soakMaxFailures, *soakRequestSize, *soakResponseSize, time.Duration(*soakPerIterationMaxAcceptableLatencyMs)*time.Millisecond, time.Duration(*soakMinTimeMsBetweenRPCs)*time.Millisecond)
+		cancel()
```
dfawley, line 362: `cancel` should be called by a `defer` right after the context is created (line 271).

interop/xds_federation/client.go
```diff
@@ -122,7 +122,8 @@ func main() {
 		go func(c clientConfig) {
 			overallDeadline := time.Now().Add(time.Duration(*soakOverallTimeoutSeconds) * time.Second)
 			ctxWithDeadline, cancel := context.WithDeadline(ctx, overallDeadline)
```
dfawley, line 124: Similar: it's best to `defer cancel()` here instead (nearly always).

The main time you wouldn't `defer` is if the context is created inside a loop, because you don't want to stack up a very large number of `defer`s.

Aditya-Sood:
thank you for the clarifications @dfawley! I've pushed the changes to the branch

interop/client/client.go
```diff
@@ -265,105 +267,108 @@ func main() {
 	}
 	defer conn.Close()
 	tc := testgrpc.NewTestServiceClient(conn)
+	overallDeadline := time.Now().Add(time.Duration(*soakOverallTimeoutSeconds) * time.Second)
+	ctxWithDeadline, cancel := context.WithDeadline(ctx, overallDeadline)
```
dfawley, line 271: Sorry, I missed this before: please use the more ergonomic `WithTimeout` instead:

```go
ctxWithDeadline, cancel := context.WithTimeout(ctx, *soakOverallTimeoutSeconds * time.Second)
```

interop/test_utils.go
```diff
@@ -387,10 +387,10 @@ func DoPerRPCCreds(tc testgrpc.TestServiceClient, serviceAccountKeyFile, oauthSc
 		FillUsername:   true,
 		FillOauthScope: true,
 	}
-	token := GetToken(serviceAccountKeyFile, oauthScope)
+	token := GetToken(ctx, serviceAccountKeyFile, oauthScope)
 	kv := map[string]string{"authorization": token.Type() + " " + token.AccessToken}
-	ctx := metadata.NewOutgoingContext(context.Background(), metadata.MD{"authorization": []string{kv["authorization"]}})
-	reply, err := tc.UnaryCall(ctx, req)
+	outCtx := metadata.NewOutgoingContext(ctx, metadata.MD{"authorization": []string{kv["authorization"]}})
+	reply, err := tc.UnaryCall(outCtx, req)
```
dfawley, line 393: This can just shadow or replace `ctx` instead of making a new thing called `outCtx`.

interop/test_utils.go
```diff
@@ -516,19 +516,19 @@ func validateMetadata(header, trailer metadata.MD) {
...
+func DoCustomMetadata(ctx context.Context, tc testgrpc.TestServiceClient, args ...grpc.CallOption) {
 	// Testing with UnaryCall.
 	pl := ClientNewPayload(testpb.PayloadType_COMPRESSABLE, 1)
 	req := &testpb.SimpleRequest{
 		ResponseType: testpb.PayloadType_COMPRESSABLE,
 		ResponseSize: int32(1),
 		Payload:      pl,
 	}
-	ctx := metadata.NewOutgoingContext(context.Background(), customMetadata)
+	outCtx := metadata.NewOutgoingContext(ctx, customMetadata)
```
dfawley, line 527: As above; just shadow (`:=`) or overwrite (`=`) `ctx`.

interop/xds_federation/client.go
```diff
@@ -115,10 +116,14 @@ func main() {
 	// run soak tests with the different clients
 	logger.Infof("Clients running with test case %q", *testCase)
 	var wg sync.WaitGroup
+	ctx := context.Background()
 	for i := range clients {
 		wg.Add(1)
 		go func(c clientConfig) {
-			interop.DoSoakTest(c.tc, c.uri, c.opts, resetChannel, *soakIterations, *soakMaxFailures, *soakRequestSize, *soakResponseSize, time.Duration(*soakPerIterationMaxAcceptableLatencyMs)*time.Millisecond, time.Duration(*soakMinTimeMsBetweenRPCs)*time.Millisecond, time.Now().Add(time.Duration(*soakOverallTimeoutSeconds)*time.Second))
+			overallDeadline := time.Now().Add(time.Duration(*soakOverallTimeoutSeconds) * time.Second)
+			ctxWithDeadline, cancel := context.WithDeadline(ctx, overallDeadline)
```
dfawley, line 124: As above, please use `WithTimeout` instead of `WithDeadline`.

dfawley (commented):
LGTM; just a couple minor style things remain.

Aditya-Sood:
hi @dfawley, thanks for the feedback - I'd missed the obvious use of `WithTimeout` instead of `WithDeadline`
have pushed the review changes in new commit

dfawley (approved):
Thank you for the PR!

arvindbr8 (approved):
LGTM! Thanks @Aditya-Sood

PR #6832, opened 2023-12-03
Don't compress empty or small messages

### LINKED ISSUE #6831: Compression is applied to empty messages, which makes them bigger, not smaller
From my reading of the source code, it appears that compression is applied to empty messages. This makes no sense, an empty message is 0 bytes long, while a gzipped compressed empty message is at least 21 bytes long. Empty messages should never be compressed. It makes no sense to compress any message that is less than 21 bytes, since the compressed message will always end up being longer than the original.

Example of where compression is done without checking the message length first:

https://github.com/grpc/grpc-go/blob/1b05500d8019885575a37efa2e961e278d680ee9/server.go#L1138C19-L1138C28

jroper:
Fixes #6831.

gzip compressing messages less than 21 bytes is always unfruitful, it always results in "compressed" messages that are longer than the original message length, since the minimum length of a gzipped message is 21 bytes, due to headers and CRC trailer.

### DIFF
--- a/rpc_util.go
+++ b/rpc_util.go
@@ -640,12 +640,21 @@ func encode(c baseCodec, msg any) ([]byte, error) {
 	return b, nil
 }
 
-// compress returns the input bytes compressed by compressor or cp.  If both
-// compressors are nil, returns nil.
+const (
+	// minCompressionLength is the number of bytes that the gzip algorithm "compresses" an empty message to
+	minCompressionLength = 21
+)
+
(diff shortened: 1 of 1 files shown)


jroper:
I can see I've broken the tests, because the tests expect empty messages to be compressed. Before I fix, I'd appreciate feedback on whether this PR is likely to be accepted.

dfawley:
In light of the comments in #6831, I'll close this.  Feel free to reopen or send a new PR when the changes are ready for the alternate approach.

PR #6842, opened 2023-12-09
grpc: skip compression of empty messages

### LINKED ISSUE #6831: Compression is applied to empty messages, which makes them bigger, not smaller

jroper:
Fixes #6831.

This avoids compressing messages that are empty, since you can't compress zero bytes to anything smaller than zero bytes, and most compression algorithms output headers and trailers which means the resulting message will be non-zero bytes.

RELEASE NOTES:
- grpc: skip compression of empty messages (as an optimization)

### DIFF
--- a/rpc_util.go
+++ b/rpc_util.go
@@ -640,12 +640,16 @@ func encode(c baseCodec, msg any) ([]byte, error) {
 	return b, nil
 }
 
-// compress returns the input bytes compressed by compressor or cp.  If both
-// compressors are nil, returns nil.
+// compress returns the input bytes compressed by compressor or cp.
+// If both compressors are nil, or if the message has zero length, returns nil,
+// indicating no compression was done.
 //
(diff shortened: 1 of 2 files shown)


easwars (approved):

arvindbr8:
Although the patch set looks good to me. I would still wait for @dfawley to approve the changes since he has the most context on the issue #6831

dfawley (approved):

PR #6852, opened 2023-12-12
 Forbid dependabot from performing major version bumps

pnacht:
As requested in https://github.com/grpc/grpc-go/pull/6835#issuecomment-1846253818, this PR configures dependabot to ignore major version bumps, reducing the risk of breaking changes.

If you wish, I can instead configure dependabot to send separate PRs with major version bumps. This way, you can easily identify and merge the "easy" minor/patch bumps, but also be notified of major version bumps that may require a bit more attention.

RELEASE NOTES: none

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,6 @@ updates:
       github-actions:
         patterns:
           - "*"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]


pnacht:
See the equivalent of #6835 on my fork: https://github.com/pnacht/grpc-go/pull/1, which only performs the minor version bumps, ignoring `actions/setup-go`.

dfawley:
> If you wish, I can instead configure dependabot to send separate PRs with major version bumps. This way, you can easily identify and merge the "easy" minor/patch bumps, but also be notified of major version bumps that may require a bit more attention.

Sure, that sounds even better, thanks!

pnacht:
Sorry, I just tried it and it's unfortunately not currently possible, which surprised me...

The issue is tracked in https://github.com/dependabot/dependabot-core/issues/1778.

So I think we'll just have to keep this PR as-is for now.

easwars (approved):

dannyskoog:
@pnacht I think it's achievable if you look at https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-3 for example.

Basically specifying the update-type under the group instead. And either:

1. Specify minor and patch update-type for the existing group. Which will result in 1 PR per major update
2. Do the same as number 1 but also create an additional group with major update-type. This will result in 1 PR for ALL major updates

And in both alternatives above, minor and patch updates will continue to be grouped in 1 PR as today (just without major updates)

PR #6884, opened 2023-12-19
alts: Queue ALTS handshakes once limit is reached rather than dropping.

matthewstevenson88:
See b/312467484.

RELEASE NOTES: none

### DIFF
--- a/credentials/alts/internal/handshaker/handshaker.go
+++ b/credentials/alts/internal/handshaker/handshaker.go
@@ -61,8 +61,6 @@ var (
 	// control number of concurrent created (but not closed) handshakes.
 	clientHandshakes = semaphore.NewWeighted(int64(envconfig.ALTSMaxConcurrentHandshakes))
 	serverHandshakes = semaphore.NewWeighted(int64(envconfig.ALTSMaxConcurrentHandshakes))
-	// errDropped occurs when maxPendingHandshakes is reached.
-	errDropped = errors.New("maximum number of concurrent ALTS handshakes is reached")
 	// errOutOfBound occurs when the handshake service returns a consumed
 	// bytes value larger than the buffer that was passed to it originally.
 	errOutOfBound = errors.New("handshaker service consumed bytes value is out-of-bound")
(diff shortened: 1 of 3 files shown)


credentials/alts/internal/handshaker/handshaker.go
```diff
@@ -156,8 +154,8 @@ func NewServerHandshaker(ctx context.Context, conn *grpc.ClientConn, c net.Conn,
 // ClientHandshake starts and completes a client ALTS handshake for GCP. Once
 // done, ClientHandshake returns a secure connection.
 func (h *altsHandshaker) ClientHandshake(ctx context.Context) (net.Conn, credentials.AuthInfo, error) {
-	if !clientHandshakes.TryAcquire(1) {
-		return nil, nil, errDropped
+	if err := clientHandshakes.Acquire(ctx, 1); err != nil {
+		return nil, nil, err
```
cesarghali, line 158: Can you please clarify how returning the actual error help queuing instead of dropping?

matthewstevenson88: In the current code, we call `TryAcquire` on the semaphore. If the semaphore is capped out, then this immediately returns false, so we immediately return `errDropped` to the user.

When we replace the `TryAcquire` call with `Acquire` instead, we will block on trying to acquire the semaphore until `ctx` times out. Internally to the semaphore, there is a queue of "acquire attempts", so calling `Acquire` effectively adds the current goroutine to this queue.

As explained in b/312467484, we want to do this so that we have uniform behavior among the 3 languages.

Does that make sense / help clarify?

cesarghali: Thanks for the clarification. This makes sense. Can you please add a comment that `Acquire` blocks until it can acquire? Thank you!

cesarghali (approved):

matthewstevenson88:
@easwars If you have time, would you also be able to take a quick pass?

easwars (commented):
Is there a test for this queueing behavior? Something like an e2e style test would be nice.

Issue #6782, opened 2023-11-10
Remove use of deprecated functions


dfawley:
https://pkg.go.dev/net/http#CloseNotifier
```sh
$ grep -r CloseNotifier
internal/transport/handler_server_test.go:				type onlyCloseNotifier interface {
internal/transport/handler_server_test.go:					http.CloseNotifier
internal/transport/handler_server_test.go:				return struct{ onlyCloseNotifier }{w.(onlyCloseNotifier)}
```

https://pkg.go.dev/crypto/x509#CertPool.Subjects
```sh
$ grep -r Roots.Subjects
credentials/tls/certprovider/store_test.go:	if gotR, wantR := got.Roots.Subjects(), want.Roots.Subjects(); !cmp.Equal(gotR, wantR, cmpopts.EquateEmpty()) {
credentials/tls/certprovider/pemfile/watcher_test.go:	if gotR, wantR := got.Roots.Subjects(), want.Roots.Subjects(); !cmp.Equal(gotR, wantR, cmpopts.EquateEmpty()) {
```

And corresponding ignores of those usages in `vet.sh` [pending PR](https://github.com/grpc/grpc-go/pull/6780/files#diff-cc1d6f4ed975bfca89b648d8c9078fa94362af099b7b449d0fbe1094b83e912bR188-R189); should hopefully be submitted relatively soon.

milencium:
Heey there, I would love to help on this, can I maybe be assigned ? Thanks in advance

dfawley:
Sounds great, please send a PR; thanks!

zedGGs:
Heey @dfawley , could you assign me on this acc ?  Im just using other acc, thank you

PR #6886, opened 2023-12-20
removing deprecated http closenotifier function

zedGGs:
Regarding this little issue with deprecated function http.CloseNotify, 
https://github.com/grpc/grpc-go/issues/6782

I have decided to add just small pr for http.CloseNotify deprecated 

In this pull request, I have just removed the implementation, tests are going well, 

my other approach was this: 

to add and update `onlyCloseNotifier` interface, by doing this: 

```
type onlyContext {
  http.ResponseWriter
  context.Context
}
```
then to update whole `testHandlerResponseWriter` with corresponding context.Context methods, like these: 
```
type testHandlerResponseWriter struct {
    *httptest.ResponseRecorder
    done chan struct{}
    err error
    value any
    deadline time.Time
    ok bool
}

func (w testHandlerResponseWriter) Deadline() (deadline time.Time, ok bool) {return time.Time{}, false}
func (w testHandlerResponseWriter) Done() <-chan struct{}                              { return w.Result().Request.Context().Done() }
func (w testHandlerResponseWriter) Err() error                                                    {return nil}
func (w testHandlerResponseWriter) Value(key any) any                                     { return nil}

testHandlerResponseWriter{
        ResponseRecorder: httptest.NewRecorder(),
        done: make(chan struct{}, 1),
        err: nil,
        value: nil,
        deadline: time.Time{},
        ok: true,
}
```

But I suppose non of this is required ? 

RELEASE NOTES: none

### DIFF
--- a/internal/transport/handler_server_test.go
+++ b/internal/transport/handler_server_test.go
@@ -227,7 +227,7 @@ type testHandlerResponseWriter struct {
 	*httptest.ResponseRecorder
 }
 
-func (w testHandlerResponseWriter) Flush()                   {}
+func (w testHandlerResponseWriter) Flush() {}
 
 func newTestHandlerResponseWriter() http.ResponseWriter {
 	return testHandlerResponseWriter{
--- a/internal/transport/handler_server_test.go
+++ b/internal/transport/handler_server_test.go
@@ -93,7 +93,6 @@ func (s) TestHandlerTransport_NewServerHandlerTransport(t *testing.T) {
 				// Return w without its Flush method
 				type onlyCloseNotifier interface {
 					http.ResponseWriter
-					http.CloseNotifier
 				}
 				return struct{ onlyCloseNotifier }{w.(onlyCloseNotifier)}
 			},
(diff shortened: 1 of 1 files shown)


zasweq:
Hello, as I mentioned in the issue, we currently cannot take this change. If this change was taken, we would need to do it cross language. Hence, can you please file an issue at the link I listed in the issue you brought up. Thank you!

dfawley:
I think this was the wrong PR?  @zasweq ?

dfawley:
> But I suppose non of this is required ?

Yeah...it looks like nothing ever read from the notifier channel?  SGTM as long as all the tests pass!

In the meantime, there is a lint error (just need to run `gofmt`):

https://github.com/grpc/grpc-go/actions/runs/7282068214/job/19843764206?pr=6886#step:7:259

zedGGs:
@dfawley , yes, thats exactly what I meant, notifier channel seems not to be required, 
I have added update, 

thank you for time

zasweq:
Oh whoops sorry wrong PR

dfawley (approved):

dfawley:
+@easwars for a second pass

easwars (approved):

Issue #6887, opened 2023-12-20
Need support HTTPS proxy server for proxy


zhiweiyin318:
Please see the FAQ in our main README.md before submitting your issue.

### Use case(s) - what problem will this feature solve?
When the grcp client is behind a https forward proxy server (like squid), the client cannot access the server after set env HTTP_PROXY or HTTPS_PROXY.

### Proposed Solution
Need to setup a TLS Conn to do CONNECT handshake with the proxy server if the proxy server address is https in this file https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go

I have a simple solution that define an env var named ROOT_CA_CERT which is to specify the root ca file. 
And setup TLS Conn using the root ca when the address of HTTP_PROXY or HTTPS_PROXY is https://. 
Otherwise to skip secure verify.

### Alternatives Considered
Pass the root ca to the https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go from transparent path. 

### Additional Context
releated issue: 
https://github.com/grpc/grpc-go/issues/6846
solution PR: https://github.com/grpc/grpc-go/pull/6888

dfawley:
Are you sure you actually need https on the connection to the proxy?  That's usually very rare, and doesn't provide much value since you are doing end-to-end encryption of the connection via the tunnel.  But please file an issue in the main (github.com/grpc/grpc) repo if you would like to request this feature.  Thanks!

PR #6888, opened 2023-12-20
another solution to support https proxy server

zhiweiyin318:
issue: https://github.com/grpc/grpc-go/issues/6846
feature request: https://github.com/grpc/grpc-go/issues/6887


zasweq:
Hello, as recommended by Doug, if you do want this feature, can you please file a cross language issue at https://github.com/grpc/grpc/issues/new/choose. If we do choose to support this feature, we need to do it cross language.

PR #6906, opened 2024-01-02
alts: Forward-fix of ALTS queuing of handshake requests.

matthewstevenson88:
This PR builds on #6884, which was reverted in #6903 because of flakiness in `alts_test.go`. It turned out that this flakiness  was just due to the test occasionally bumping up on the timeout, so we've fixed this in this PR by increasing the timeout.

RELEASE NOTES: none

### DIFF
--- a/credentials/alts/alts_test.go
+++ b/credentials/alts/alts_test.go
@@ -44,7 +44,7 @@ import (
 )
 
 const (
-	defaultTestLongTimeout  = 10 * time.Second
+	defaultTestLongTimeout  = 60 * time.Second
 	defaultTestShortTimeout = 10 * time.Millisecond
 )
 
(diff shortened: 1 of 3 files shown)


ginayeh:
@erm-g can you review this PR first?

matthewstevenson88:
Note that the real changes were already reviewed in #6884, the only delta here is bumping up the test timeout in alts_test.go.

matthewstevenson88:
@erm-g Friendly ping.

Also cc @cesarghali.

credentials/alts/alts_test.go
```diff
@@ -44,7 +44,7 @@ import (
 )
 
 const (
-	defaultTestLongTimeout  = 10 * time.Second
+	defaultTestLongTimeout  = 60 * time.Second
```
erm-g, line 47: I'm not sure I fully understand this long timeout. It's used only for `establishAltsConnection' and has the comment `			// The server is not ready yet. Try again.` However the server should be up and running because of synchronous `startServer` call preceding the  `establishAltsConnection'. What am I missing here?

matthewstevenson88: That comment was just a bit out-of-date. I've updated it to better reflect the current state of affairs. Thanks for flagging this!

erm-g (approved):

PR #6907, opened 2024-01-04
removing Roots deprecated Subjects field in tests

zedGGs:
Regarding this issue, https://github.com/grpc/grpc-go/issues/6782

one of the goals was to remove usage of deprecated field `Subjects`, which is not being returned by `SystemCertPool` anymore, more details here https://pkg.go.dev/crypto/x509#CertPool.Subjects
also I have updated these tests so we compare got and wanted CertPool structs, by using https://pkg.go.dev/crypto/x509#CertPool.Equal

RELEASE NOTES: none

### DIFF
--- a/credentials/tls/certprovider/pemfile/watcher_test.go
+++ b/credentials/tls/certprovider/pemfile/watcher_test.go
@@ -61,12 +61,6 @@ func compareKeyMaterial(got, want *certprovider.KeyMaterial) error {
 		}
 	}
 
-	// x509.CertPool contains only unexported fields some of which contain other
-	// unexported fields. So usage of cmp.AllowUnexported() or
-	// cmpopts.IgnoreUnexported() does not help us much here. Also, the standard
-	// library does not provide a way to compare CertPool values. Comparing the
-	// subjects field of the certs in the CertPool seems like a reasonable
-	// approach.
 	if gotR, wantR := got.Roots, want.Roots; !gotR.Equal(wantR) {
 		return fmt.Errorf("keyMaterial roots = %v, want %v", gotR, wantR)
 	}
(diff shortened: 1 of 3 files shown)


arvindbr8 (approved):
LGTM. Thanks for the PR. We need another maintainer's review before we can check this in. cc: @ginayeh could you please take a look at this?

zedGGs:
Heey @arvindbr8 , thanks for reply, as I see everyone is on vacation and you are in charge now hehe, cheers ! 
btw I can also update comments in tests

zedGGs:
@arvindbr8 , I had just removed unneeded comments, where `Subjects` are mentioned, that are not used anymore
and `cmp` that is not used anymore,

I suppose there's no need for comments since its obvious from `Equal` what are we doing, similar to code above in same files, where `Equal` is only used without additional explanation, 

I can still update and change this if you want

arvindbr8 (changes requested):
Thanks! The updates made to the comments LGTM

However, I think we remove corresponding ignores of those usages in [vet.sh]( github.com/grpc/grpc-go/blob/master/vet.sh#L190). Could you please append your PR with that?

arvindbr8:
Marking this PR as fixing #6782 since the other half of the issue was already fixed by you @zedGGs

zedGGs:
@arvindbr8 I have updated pr, but not sure is this still okey ? since vet-proto is not good

binarylog/grpc_binarylog_v1/binarylog.pb.go
arvindbr8, line 1: Please roll this back

arvindbr8 (changes requested):
Overall the changes looks good to me.  Please address that one comment about the `pb.go` file

> btw I added small update on vet-proto

We should not be doing that as part of this change. I have actually merged a change to update the `pb.go` file. Please rollback that file and rebase from master -- and vet-proto should be happy :)

binarylog/grpc_binarylog_v1/binarylog.pb.go
zedGGs: thank you for the reivew, I have updated pr but tests seems not to pass Im not sure is that because of flakines with xds tests 

I apologise, I have managed somehow to have push from wrong account, I 
if this matters I can add separated pr for this

arvindbr8 (approved):

arvindbr8:
Thanks, this looks good to me. cc'ing @ginayeh for a second pass

ginayeh (approved):

PR #6917, opened 2024-01-11
DNS resolving with timeout

and1truong:
Currently, the DNS resolver does not have a timeout mechanism for DNS resolution. In some edge cases, where the DNS server is slow to respond, this can lead to stale records being used, resulting in failed connections or other issues.

This pull request adds a configurable timeout for DNS resolution, which can be set by the user to a value that suits their needs. The default value is set to 5 seconds, which should be sufficient for most cases. If the DNS server does not respond within the specified time, the resolver will cancel the request and move on to the next available server.

Please review and let me know if you have any questions or concerns.

Thanks!

RELEASE NOTES:
* resolver/dns: add SetResolvingTimeout to allow configuring the DNS resolver's timeout globally.

### DIFF
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -107,9 +107,7 @@ type dnsBuilder struct{}
 
 // Build creates and starts a DNS resolver that watches the name resolution of
 // the target.
-func (b *dnsBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (
-	resolver.Resolver, error,
-) {
+func (b *dnsBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
 	host, port, err := parseTarget(target.Endpoint(), defaultPort)
 	if err != nil {
 		return nil, err
@@ -325,12 +323,10 @@ func (d *dnsResolver) lookupHost(ctx context.Context) ([]resolver.Address, error
 }
 
 func (d *dnsResolver) lookup() (*resolver.State, error) {
-	ctxSRV, cancelSRV := context.WithTimeout(d.ctx, ResolvingTimeout)
-	defer cancelSRV()
-	srv, srvErr := d.lookupSRV(ctxSRV)
-	ctxHost, cancelHost := context.WithTimeout(d.ctx, ResolvingTimeout)
-	defer cancelHost()
-	addrs, hostErr := d.lookupHost(ctxHost)
+	ctx, cancel := context.WithTimeout(d.ctx, ResolvingTimeout)
+	defer cancel()
+	srv, srvErr := d.lookupSRV(ctx)
+	addrs, hostErr := d.lookupHost(ctx)
 	if hostErr != nil && (srvErr != nil || len(srv) == 0) {
 		return nil, hostErr
 	}
@@ -340,9 +336,7 @@ func (d *dnsResolver) lookup() (*resolver.State, error) {
 		state = grpclbstate.Set(state, &grpclbstate.State{BalancerAddresses: srv})
 	}
 	if !d.disableServiceConfig {
-		ctxTXT, cancelTXT := context.WithTimeout(d.ctx, ResolvingTimeout)
-		defer cancelTXT()
-		state.ServiceConfig = d.lookupTXT(ctxTXT)
+		state.ServiceConfig = d.lookupTXT(ctx)
 	}
 	return &state, nil
 }
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -45,6 +45,13 @@ import (
 // addresses from SRV records.  Must not be changed after init time.
 var EnableSRVLookups = false
 
+// ResolvingTimeout specifies the maximum duration for a DNS resolution request.
+// If the timeout expires before a response is received, the request will be canceled.
+//
+// It is recommended to set this value at application startup. Avoid modifying this variable
+// after initialization as it's not thread-safe for concurrent modification.
+var ResolvingTimeout = 10 * time.Second
+
 var logger = grpclog.Component("dns")
 
 func init() {
@@ -100,7 +107,9 @@ type dnsBuilder struct{}
 
 // Build creates and starts a DNS resolver that watches the name resolution of
 // the target.
-func (b *dnsBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
+func (b *dnsBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (
+	resolver.Resolver, error,
+) {
 	host, port, err := parseTarget(target.Endpoint(), defaultPort)
 	if err != nil {
 		return nil, err
@@ -221,18 +230,18 @@ func (d *dnsResolver) watcher() {
 	}
 }
 
-func (d *dnsResolver) lookupSRV() ([]resolver.Address, error) {
+func (d *dnsResolver) lookupSRV(ctx context.Context) ([]resolver.Address, error) {
 	if !EnableSRVLookups {
 		return nil, nil
 	}
 	var newAddrs []resolver.Address
-	_, srvs, err := d.resolver.LookupSRV(d.ctx, "grpclb", "tcp", d.host)
+	_, srvs, err := d.resolver.LookupSRV(ctx, "grpclb", "tcp", d.host)
 	if err != nil {
 		err = handleDNSError(err, "SRV") // may become nil
 		return nil, err
 	}
 	for _, s := range srvs {
-		lbAddrs, err := d.resolver.LookupHost(d.ctx, s.Target)
+		lbAddrs, err := d.resolver.LookupHost(ctx, s.Target)
 		if err != nil {
 			err = handleDNSError(err, "A") // may become nil
 			if err == nil {
@@ -269,8 +278,8 @@ func handleDNSError(err error, lookupType string) error {
 	return err
 }
 
-func (d *dnsResolver) lookupTXT() *serviceconfig.ParseResult {
-	ss, err := d.resolver.LookupTXT(d.ctx, txtPrefix+d.host)
+func (d *dnsResolver) lookupTXT(ctx context.Context) *serviceconfig.ParseResult {
+	ss, err := d.resolver.LookupTXT(ctx, txtPrefix+d.host)
 	if err != nil {
 		if envconfig.TXTErrIgnore {
 			return nil
@@ -297,8 +306,8 @@ func (d *dnsResolver) lookupTXT() *serviceconfig.ParseResult {
 	return d.cc.ParseServiceConfig(sc)
 }
 
-func (d *dnsResolver) lookupHost() ([]resolver.Address, error) {
-	addrs, err := d.resolver.LookupHost(d.ctx, d.host)
+func (d *dnsResolver) lookupHost(ctx context.Context) ([]resolver.Address, error) {
+	addrs, err := d.resolver.LookupHost(ctx, d.host)
 	if err != nil {
 		err = handleDNSError(err, "A")
 		return nil, err
@@ -316,8 +325,12 @@ func (d *dnsResolver) lookupHost() ([]resolver.Address, error) {
 }
 
 func (d *dnsResolver) lookup() (*resolver.State, error) {
-	srv, srvErr := d.lookupSRV()
-	addrs, hostErr := d.lookupHost()
+	ctxSRV, cancelSRV := context.WithTimeout(d.ctx, ResolvingTimeout)
+	defer cancelSRV()
+	srv, srvErr := d.lookupSRV(ctxSRV)
+	ctxHost, cancelHost := context.WithTimeout(d.ctx, ResolvingTimeout)
+	defer cancelHost()
+	addrs, hostErr := d.lookupHost(ctxHost)
 	if hostErr != nil && (srvErr != nil || len(srv) == 0) {
 		return nil, hostErr
 	}
@@ -327,7 +340,9 @@ func (d *dnsResolver) lookup() (*resolver.State, error) {
 		state = grpclbstate.Set(state, &grpclbstate.State{BalancerAddresses: srv})
 	}
 	if !d.disableServiceConfig {
-		state.ServiceConfig = d.lookupTXT()
+		ctxTXT, cancelTXT := context.WithTimeout(d.ctx, ResolvingTimeout)
+		defer cancelTXT()
+		state.ServiceConfig = d.lookupTXT(ctxTXT)
 	}
 	return &state, nil
 }
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -50,7 +50,7 @@ var EnableSRVLookups = false
 //
 // It is recommended to set this value at application startup. Avoid modifying this variable
 // after initialization as it's not thread-safe for concurrent modification.
-var ResolvingTimeout = 10 * time.Second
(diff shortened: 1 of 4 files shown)


dfawley:
Hi @and1truong,

Such a wide-sweeping change would require cross-language agreement for adding this.

Instead, maybe you can add a global in the `resolver/dns` package to control this (which would be set globally at initialization time)?  This is equivalent to what gRPC-Java has for this kind of thing.

and1truong:
Thanks for comment. I will start working on this next week.

internal/resolver/dns/dns_resolver.go
```diff
@@ -100,7 +107,9 @@ type dnsBuilder struct{}
 
 // Build creates and starts a DNS resolver that watches the name resolution of
 // the target.
-func (b *dnsBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
+func (b *dnsBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (
+	resolver.Resolver, error,
+) {
```
dfawley, line 112: Please revert these changes.

internal/resolver/dns/dns_resolver.go
```diff
@@ -316,8 +325,12 @@ func (d *dnsResolver) lookupHost() ([]resolver.Address, error) {
 }
 
 func (d *dnsResolver) lookup() (*resolver.State, error) {
-	srv, srvErr := d.lookupSRV()
-	addrs, hostErr := d.lookupHost()
+	ctxSRV, cancelSRV := context.WithTimeout(d.ctx, ResolvingTimeout)
+	defer cancelSRV()
```
dfawley, line 329: Could you just use (and overwrite) `ctx` for all of these?

```go
ctx, cancel := context.WithTimeout(d.ctx, ResolvingTimeout)
defer cancel()
d.lookupSRV(ctx)
ctx, cancel = context.WithTimeout(d.ctx, ResolvingTimeout)
defer cancel
d.lookupHost(ctx)
// same with TXT
```

(It would probably be ideal to do these calls in parallel so we only need to make the context once, but that would be a bigger behavior change that there's no need to introduce here.)

resolver/dns/dns_resolver.go
```diff
@@ -24,10 +24,29 @@
...
 )
 
+// SetResolvingTimeout sets the maximum duration for DNS resolution requests.
+//
+// This function updates the global `ResolvingTimeout` variable, which specifies
+// how long a DNS resolution attempt will wait before it is canceled.
+//
+// It is recommended to call this function at application startup, before any
+// gRPC calls are made. Modifying this value after initialization is not
+// thread-safe.
```
dfawley, line 40: Please replace "It is recommended to call this function..." with "This function must be called only...".

resolver/dns/dns_resolver.go
```diff
@@ -24,10 +24,29 @@
...
+	"time"
+
 	"google.golang.org/grpc/internal/resolver/dns"
 	"google.golang.org/grpc/resolver"
 )
 
+// SetResolvingTimeout sets the maximum duration for DNS resolution requests.
+//
+// This function updates the global `ResolvingTimeout` variable, which specifies
+// how long a DNS resolution attempt will wait before it is canceled.
```
dfawley, line 36: Instead of mentioning "the global `ResolvingTimeout`", maybe something like "This function affects the global timeout used by all channels using the DNS name resolver scheme"?

resolver/dns/dns_resolver.go
```diff
@@ -24,10 +24,29 @@
...
+// It is recommended to call this function at application startup, before any
+// gRPC calls are made. Modifying this value after initialization is not
+// thread-safe.
+//
+// The default value is 10 seconds. Setting the timeout too low may result
+// in premature timeouts during resolution, while setting it too high may
+// lead to unnecessary delays in service discovery. Choose a value appropriate
+// for your specific needs and network environment.
+func SetResolvingTimeout(timeout time.Duration) {
+	dns.ResolvingTimeout = timeout
```
dfawley, line 47: Ideally we'd have an end-to-end style tests that sets this and confirms it works as expected.

I think that would include overriding `internal.NewNetResolver` to produce a resolver that does nothing but block on the context, and then just doing `grpc.Dial("dns:///doesntmatter.com:123")` and waiting for the `State()` of the channel to change to `TransientFailure`.  It probably wouldn't be that hard to make - would you mind adding a test like that in this directory?

internal/resolver/dns/fake_net_resolver_test.go
```diff
@@ -47,9 +49,28 @@ func (tr *testNetResolver) LookupHost(ctx context.Context, host string) ([]strin
 	tr.mu.Lock()
 	defer tr.mu.Unlock()
 
+	// trigger a timeout if the host name contains "infinity"
+	if strings.Contains(host, "infinity") {
+		<-time.After(88 * time.Millisecond)
+	}
```
dfawley, line 55: Timing based tests like this are problematic, because if the machine hiccups for >88 ms, you could end up with a race where this case wins and "infinity" actually resolves.

Please instead use a `testNetResolver.lookupHostCh` that is unbuffered, and don't read from it until after the timeout occurs.

resolver/dns/dns_resolver.go
```diff
@@ -32,12 +32,10 @@ import (
...
 // SetResolvingTimeout sets the maximum duration for DNS resolution requests.
 //
-// This function updates the global `ResolvingTimeout` variable, which specifies
-// how long a DNS resolution attempt will wait before it is canceled.
+// This function affects the global timeout used by all channels using the DNS name resolver scheme.
 //
-// It is recommended to call this function at application startup, before any
-// gRPC calls are made. Modifying this value after initialization is not
-// thread-safe.
+// It must be called only at application startup, before any  gRPC calls are made.
```
dfawley, line 37: Nit: fix the double-space after "any"

internal/resolver/dns/dns_resolver.go
```diff
@@ -45,6 +45,13 @@ import (
 // addresses from SRV records.  Must not be changed after init time.
 var EnableSRVLookups = false
 
+// ResolvingTimeout specifies the maximum duration for a DNS resolution request.
+// If the timeout expires before a response is received, the request will be canceled.
+//
+// It is recommended to set this value at application startup. Avoid modifying this variable
+// after initialization as it's not thread-safe for concurrent modification.
+var ResolvingTimeout = 10 * time.Second
```
dfawley, line 53: It looks like there was no timeout at all before.  Could you set this to 30 seconds, then, to be safer?

resolver/dns/dns_resolver.go
```diff
@@ -24,10 +24,27 @@
...
 )
 
+// SetResolvingTimeout sets the maximum duration for DNS resolution requests.
+//
+// This function affects the global timeout used by all channels using the DNS name resolver scheme.
+//
+// It must be called only at application startup, before any gRPC calls are made.
+// Modifying this value after initialization is not thread-safe.
+//
+// The default value is 10 seconds. Setting the timeout too low may result
```
dfawley, line 40: Please update to "30 seconds".

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -1215,3 +1216,56 @@ func (s) TestReportError(t *testing.T) {
...
+func overrideResolveTimeoutDuration(t *testing.T, dur time.Duration) {
+	origDur := dns.ResolvingTimeout
+	dnspublic.SetResolvingTimeout(dur)
+
+	t.Cleanup(func() { dnspublic.SetResolvingTimeout(origDur) })
+}
+
+// Test verifies that when the DNS resolver gets timeout error when net.Resolver
+// takes too long to resolve a target.
+func (s) TestResolveTimeout(t *testing.T) {
```
dfawley, line 1230: I'm not understanding this test now.

I think the test was perfect the way you had it before, but instead of using a `Sleep`, you'd use a blocked channel send instead to cause the error.

I.e. in `func (tr *testNetResolver) LookupHost()`, use `SendContext` instead of `Send` so the function can return an error when the context it receives expires:

```go
if tr.lookupHostCh != nil {
	if err := tr.lookupHostCh.SendContext(ctx, nil); err != nil {
		return err
	}
}
```

Then in this test; mostly like how you had it before:

```go
	// Set DNS resolving timeout to 7ms
	// ...

	defer cancel()

	// ...
	const target = "infinity"
	overrideNetResolver(t, &testNetResolver{
		lookupHostCh:    testutils.NewChannelWithSize(0),
		hostLookupTable: map[string][]string{},
	})

	// Building will produce a timeout error on errCh since hostLookupCh is blocked.
	r, _, errCh := buildResolverWithTestClientConn(t, target)

	select {
	case <-ctx.Done():
		t.Fatal("Timeout when waiting for an error from the resolver")
	case err := <-errCh:
		if err == nil || !strings.Contains(err.Error(), "hostLookup timeout") {
			t.Fatalf(`we expect to see timed out error`)
		}
	}
}
```

resolver/dns/dns_resolver.go
and1truong: Change pushed. Thanks.

arvindbr8:
@and1truong -- let us know if this is ready for another review

arvindbr8:
okay, I'll request @dfawley to take another look at this

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -1215,3 +1216,46 @@ func (s) TestReportError(t *testing.T) {
...
+// Override the default dns.ResolvingTimeout with a test duration.
+func overrideResolveTimeoutDuration(t *testing.T, dur time.Duration) {
+	origDur := dns.ResolvingTimeout
+	dnspublic.SetResolvingTimeout(dur)
+
+	t.Cleanup(func() { dnspublic.SetResolvingTimeout(origDur) })
+}
+
+// Test verifies that when the DNS resolver gets timeout error when net.Resolver
+// takes too long to resolve a target.
```
dfawley, line 1229: Nit: remove the first "when".

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -1215,3 +1216,46 @@ func (s) TestReportError(t *testing.T) {
...
+	// Define a testNetResolver with lookupHostCh, an unbuffered channel,
+	// so we can block the resolver until reaching timeout.
+	tr := &testNetResolver{
+		lookupHostCh:    testutils.NewChannel(),
+		hostLookupTable: map[string][]string{target: {"1.2.3.4"}},
+	}
+	overrideNetResolver(t, tr)
+
+	// block testNetResolver.testNetResolver until timeout
+	_ = tr.lookupHostCh.SendContext(ctx, nil)
```
dfawley, line 1251: This should not be needed.

I think the problem is that you have a buffered channel.  Try `testutils.NewChannelWithSize(0)` instead?

internal/resolver/dns/fake_net_resolver_test.go
```diff
@@ -41,7 +41,9 @@ type testNetResolver struct {
 
 func (tr *testNetResolver) LookupHost(ctx context.Context, host string) ([]string, error) {
 	if tr.lookupHostCh != nil {
-		tr.lookupHostCh.Send(nil)
+		if err := tr.lookupHostCh.SendContext(ctx, nil); nil != err {
```
dfawley, line 44: Nit: `err != nil` instead, please.

dfawley (approved):
LGTM except for the last few small things.  Thank you for the PR!

internal/resolver/dns/dns_resolver_test.go
and1truong: Great point. Change pushed. Thanks.

internal/resolver/dns/dns_resolver.go
```diff
@@ -45,6 +45,13 @@ import (
 // addresses from SRV records.  Must not be changed after init time.
 var EnableSRVLookups = false
 
+// ResolvingTimeout specifies the maximum duration for a DNS resolution request.
+// If the timeout expires before a response is received, the request will be canceled.
+//
+// It is recommended to set this value at application startup. Avoid modifying this variable
+// after initialization as it's not thread-safe for concurrent modification.
```
arvindbr8, line 52:
```suggestion
// ResolvingTimeout specifies the maximum duration for a DNS resolution request.
// If the timeout expires before a response is received, the request will be
// canceled.
//
// It is recommended to set this value at application startup. Avoid modifying
// this variable after initialization as it's not thread-safe for concurrent
// modification.
```

To keep the comment line length consistent to 80-column wide

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -1215,3 +1216,44 @@ func (s) TestReportError(t *testing.T) {
 		}
 	}
 }
+
+// Override the default dns.ResolvingTimeout with a test duration.
+func overrideResolveTimeoutDuration(t *testing.T, dur time.Duration) {
+	origDur := dns.ResolvingTimeout
```
arvindbr8, line 1222: optional: 

call t.Helper() here?

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -1215,3 +1216,44 @@ func (s) TestReportError(t *testing.T) {
...
+}
+
+// Test verifies that the DNS resolver gets timeout error when net.Resolver
+// takes too long to resolve a target.
+func (s) TestResolveTimeout(t *testing.T) {
+	// Set DNS resolving timeout duration to 7ms
+	timeoutDur := 7 * time.Millisecond
+	overrideResolveTimeoutDuration(t, timeoutDur)
+
+	// context with a bit larger
```
arvindbr8, line 1235: mm, bit larger... ?

I think this line is self explanatory and you could prolly nix this comment.  
```suggestion
```

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -1215,3 +1216,44 @@ func (s) TestReportError(t *testing.T) {
...
+	}
+	overrideNetResolver(t, tr)
+
+	_, _, errCh := buildResolverWithTestClientConn(t, target)
+	select {
+	case <-ctx.Done():
+		t.Fatal("Timeout when waiting for the DNS resolver to timeout")
+	case err := <-errCh:
+		if err == nil || !strings.Contains(err.Error(), "context deadline exceeded") {
+			t.Fatalf(`We expect to see timed out error`)
```
arvindbr8, line 1256: nit:
```suggestion
			t.Fatalf(`Expected to see Timeout error; got: %v`, err)
```

resolver/dns/dns_resolver.go
```diff
@@ -24,10 +24,27 @@
...
+//
+// This function affects the global timeout used by all channels using the DNS name resolver scheme.
+//
+// It must be called only at application startup, before any gRPC calls are made.
+// Modifying this value after initialization is not thread-safe.
+//
+// The default value is 30 seconds. Setting the timeout too low may result
+// in premature timeouts during resolution, while setting it too high may
+// lead to unnecessary delays in service discovery. Choose a value appropriate
+// for your specific needs and network environment.
```
arvindbr8, line 43: Similar nit formatting suggestion

```suggestion
// SetResolvingTimeout sets the maximum duration for DNS resolution requests.
//
// This function affects the global timeout used by all channels using the DNS
// name resolver scheme.
//
// It must be called only at application startup, before any gRPC calls are
// made. Modifying this value after initialization is not thread-safe.
//
// The default value is 30 seconds. Setting the timeout too low may result in
// premature timeouts during resolution, while setting it too high may lead to
// unnecessary delays in service discovery. Choose a value appropriate for your
// specific needs and network environment.
```

arvindbr8 (changes requested):
Few comments, mostly formatting.

arvindbr8 (approved):

arvindbr8:
Thanks for the PR @and1truong. We look forward to seeing your next PR.

PR #6924, opened 2024-01-17
protoc: Change protoc to include generated call option

zasweq:
This PR adds a call option in the generated code that can be used by stats handlers to determine whether a call comes from a registered method or not.

RELEASE NOTES:
* protoc: Change protoc to include generated call option


zasweq:
Discussed offline; closing this as we decided to add an exported call option from gRPC, release it, and use that symbol in generated code.

PR #6931, opened 2024-01-19
alts: Extract AuthInfo after handshake in ALTS e2e test.

matthewstevenson88:
RELEASE NOTES: none

### DIFF
--- a/credentials/alts/alts_test.go
+++ b/credentials/alts/alts_test.go
@@ -40,6 +40,7 @@ import (
 	"google.golang.org/grpc/internal/testutils"
 	testgrpc "google.golang.org/grpc/interop/grpc_testing"
 	testpb "google.golang.org/grpc/interop/grpc_testing"
+	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 )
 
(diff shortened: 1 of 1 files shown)


credentials/alts/alts_test.go
```diff
@@ -409,6 +411,16 @@ func establishAltsConnection(t *testing.T, handshakerAddress, serverAddress stri
 	if !success {
 		t.Fatalf("c.UnaryCall() timed out after %v", defaultTestShortTimeout)
 	}
+	if got, want := peer.AuthInfo.AuthType(), "alts"; got != want {
```
erm-g, line 414: I'm not sure I fully understand these checks. IIUC, alts here is a result of calling new creds [function](https://github.com/grpc/grpc-go/blob/987df1309236a950d3f618eb5287279d81a7396e/credentials/alts/alts.go#L143)
Shouldn't we check the result of it instead?

matthewstevenson88: That's the top-level ALTS credentials API, and there are other tests in this file that check the result of that API.

This PR is modifying a helper function that is used during the e2e test for the ALTS credentials API. Specifically, in addition to checking that the ALTS handshake succeeds, we are also checking that the `AuthInfo` object produced from the ALTS handshake has the expected "alts" type.

matthewstevenson88: Added a clarifying comment, per offline request.

erm-g (approved):

PR #6940, opened 2024-01-24
test: add test for invalid streamID

AnandInguva:
Fixes: https://github.com/grpc/grpc-go/issues/104

RELEASE NOTES: none


test/end2end_test.go
```diff
@@ -3901,6 +3901,29 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+		_, err := stream.Recv()
+		return err
+	}}
+	te.startServer(ts)
+	defer te.tearDown()
+	te.withServerTester(func(st *serverTester) {
+		st.writeHeadersGRPC(2, "/grpc.testing.TestService/StreamingInputCall", true)
+		_, err := st.fr.ReadFrame()
+		if err != nil {
+			fmt.Println(err)
```
AnandInguva, line 3916: How to handle the error here?

AnandInguva:
@dfawley @arvindbr8 can you suggest how could i assert/verify the error?

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+		if err == nil {
+			t.Fatalf("Error expected when Client StreamID is even %v", err)
+		}
+	})
+}
+
+// Client must always send a streamID in odd numbers according to
+// https://httpwg.org/specs/rfc7540.html#StreamIdentifiers. This test
+// makes sure that the transport throws an error when the client streamID is
+// even.
```
arvindbr8, line 3924: It would be nicer for readability if the docstring follows this structure - 
1. What/Why about the test? - A general overview of the test  
2. How is it tested? - How the test is set up, mentioning if there are any special setup done for the test and what is the assertion

In the case of the test failing (due to a regression or flake) - This helps us quickly understand why did it fail. It's always nice to be as descriptive as possible 

I would slightly modify the comment here. How about? 
```suggestion
// Tests gRPC server's behavior when a gRPC client sends a frame with an invalid
// streamID. Per [HTTP/2 spec]: Streams initiated by a client MUST use
// odd-numbered stream identifiers. This test sets up a test server and send a
// header frame with stream ID of 2. The test asserts that a subsequent read on
// the transport throws an error. 
//
// [HTTP/2 spec]: https://httpwg.org/specs/rfc7540.html#StreamIdentifiers
```

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
 	}
 }
 
+func testClientInvalidStreamID(t *testing.T, e env) {
```
arvindbr8, line 3904: Looking at the tests, we are missing a test for the case where the server sends an invalid stream ID as well. Might be worth adding that add that as part of this diff.

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+		}
+	})
+}
+
+// Client must always send a streamID in odd numbers according to
+// https://httpwg.org/specs/rfc7540.html#StreamIdentifiers. This test
+// makes sure that the transport throws an error when the client streamID is
+// even.
+func (s) TestClientInvalidStreamID(t *testing.T) {
+	for _, e := range listTestEnv() {
```
arvindbr8, line 3926: What's the benefit of running for every `TestEnv` here?

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+		if e.httpHandler {
+			continue
+		}
+		testClientInvalidStreamID(t, e)
```
arvindbr8, line 3930: I prefer having all the testing logic inline. Also if you want to run sub tests for each env I would suggest using [t.Run()](https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks)

arvindbr8 (changes requested):
I've taken a first pass at your PR. Please let me know if you have any questions

test/end2end_test.go
AnandInguva: Looking at the behavior of servertester.go, we send requests to server to test the server and server sends appropriate response to the client conn? 

If I want to test if the client is throwing error if it receives odd stream IDs from the server, servertester is the not the right one to use IIUC?

test/end2end_test.go
AnandInguva: No benefit really. Following pattern with other tests.

arvindbr8: I believe we can do this test without that.

test/end2end_test.go
arvindbr8: This comment would not apply if you remove the testing for every environment.

test/end2end_test.go
arvindbr8: That's correct. Your best bet must be a similar [clienttester](https://github.com/AnandInguva/grpc-go/blob/master/test/clienttester.go) in the same test package.

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
 	}
 }
 
+func testClientInvalidStreamID(t *testing.T, e env) {
+	te := newTest(t, e)
```
arvindbr8, line 3905: Do we need to create a `test` object here? The test struct does a lot of other things which are not significant for this test. 

Can we just create a gRPC server (calling `grpc.NewServer()`) and use that instead?

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+	ts := &funcServer{streamingInputCall: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+		_, err := stream.Recv()
+		return err
+	}}
+	te.startServer(ts)
```
arvindbr8, line 3910: `funcServer` is not required for this case. You can pass in `nil` to `startServer`.  (this can be fully avoided if you use a gRPC server instead of a test server)

test/end2end_test.go
```diff
@@ -3901,6 +3901,36 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+	defer te.tearDown()
+	te.withServerTester(func(st *serverTester) {
+		st.writeHeadersGRPC(2, "/grpc.testing.TestService/StreamingInputCall", true)
+		_, err := st.fr.ReadFrame()
+		if err == nil {
```
arvindbr8, line 3915: We should assert for the exact error here. would be hard to catch the situation where ReadFrame could have error'd for an unrelated issue

arvindbr8 (changes requested):
I've made a few more comments. PTAL

AnandInguva:
Yeah, let me make some changes soon

test/end2end_test.go
AnandInguva: I didn't use a gRPC server as the servertester takes care of setting up both client and server for you.

test/end2end_test.go
AnandInguva: As discussed offline, since the stream is always initiated by the client, we don't need to test the stream id on the client side.

test/end2end_test.go
AnandInguva: I was looking at how to assert the error here. 

```
--- FAIL: Test (0.00s)
    --- FAIL: Test/ClientInvalidStreamID (0.00s)
        end2end_test.go:3917: Running test in tcp-clear-v1-balancer environment...
        tlogger.go:116: INFO server.go:667 [core] [Server #1] Server created  (t=+1.368417ms)
        tlogger.go:116: INFO server.go:867 [core] [Server #1 ListenSocket #2] ListenSocket created  (t=+1.448209ms)
        tlogger.go:116: INFO syscall_nonlinux.go:39 [core] CPU time info is unavailable on non-linux environments.  (t=+2.208959ms)
        tlogger.go:116: INFO http2_server.go:1210 [transport] [server-transport 0x140002a89c0] Closing: received an illegal stream id: 2. headers frame: [FrameHeader HEADERS flags=END_STREAM|END_HEADERS stream=2 len=60]  (t=+2.488042ms)
        tlogger.go:116: INFO controlbuf.go:543 [transport] [server-transport 0x140002a89c0] loopyWriter exiting with error: transport closed by client  (t=+2.506959ms)
        end2end_test.go:3928: Error : EOF
        tlogger.go:116: INFO server.go:806 [core] [Server #1 ListenSocket #2] ListenSocket deleted  (t=+2.577042ms)
FAIL
```

This is the error log for me.  I was printing what error I was getting and it is EOF. It makes sense because there are no frames sent from the server. Can you suggest how to verify the expected error? string matching :p?

AnandInguva: One more thought, Instead of the test here, can we do a test in the `transport.go` and test http2_server's operateHeaders func. let me see

test/end2end_test.go
```diff
@@ -3901,6 +3901,27 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+// streamID. Per [HTTP/2 spec]: Streams initiated by a client MUST use
+// odd-numbered stream identifiers. This test sets up a test server and send a
+// header frame with stream ID of 2. The test asserts that a subsequent read on
+// the transport throws an error.
+//
+// [HTTP/2 spec]: https://httpwg.org/specs/rfc7540.html#StreamIdentifiers
+func (s) TestClientInvalidStreamID(t *testing.T) {
+	te := newTest(t, tcpClearEnv)
+	te.startServer(nil)
+	defer te.tearDown()
```
arvindbr8, line 3914: I didnt like using `newTest()` because it requires you to pass in a `env` parameter. The `test` struct is already overloaded, and let's not reuse that again. At some time in the future, we would love to split this end2end_test package. 

Also we dont need a server and a client here. The setup should essentially be server and serverTester.

I'm thinking about something like this
```suggestion
	lis, err := net.Listen("tcp", "localhost:0")
	if err != nil {
		t.Fatalf("Failed to listen: %v", err)
	}
	defer lis.Close()
	s := grpc.NewServer()
	defer s.Stop()
	go s.Serve(lis)
	conn, err := net.DialTimeout("tcp", lis.Addr().String(), defaultTestTimeout)
	if err != nil {
		t.Fatalf("Failed to dial: %v", err)
	}
	st := newServerTesterFromConn(t, conn)
	st.greet()
```

arvindbr8: Also, would be nice to have st.greet() really be part of newServerTesterFromConn.. Not imp for this PR

internal/transport/transport_test.go
```diff
@@ -2592,3 +2592,20 @@ func TestConnectionError_Unwrap(t *testing.T) {
 		t.Error("ConnectionError does not unwrap")
 	}
 }
+
+func TestServerOperateHeaderOnEvenStreamID(t *testing.T) {
```
arvindbr8, line 2596: This is a nice approach!

internal/transport/transport_test.go
```diff
@@ -2592,3 +2592,20 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	err := s.operateHeaders(ctx, metaHeaderFrame, nil)
+	if err == nil {
+		t.Fatalf("Error expected on the even stream ID. got %v", err)
+	}
+	possibleErrMsg := "received an illegal stream id: 2"
+	if !strings.Contains(err.Error(), possibleErrMsg) {
+		t.Fatalf("Expected partial error message: %v, got %v", possibleErrMsg, err.Error())
+	}
+	defer cancel()
```
arvindbr8, line 2609: could you move the defer call to right after creating ctx pls?

internal/transport/transport_test.go
```diff
@@ -2592,3 +2592,20 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	metaHeaderFrame := &http2.MetaHeadersFrame{HeadersFrame: hf}
+	s := http2Server{}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	err := s.operateHeaders(ctx, metaHeaderFrame, nil)
+	if err == nil {
+		t.Fatalf("Error expected on the even stream ID. got %v", err)
+	}
+	possibleErrMsg := "received an illegal stream id: 2"
+	if !strings.Contains(err.Error(), possibleErrMsg) {
+		t.Fatalf("Expected partial error message: %v, got %v", possibleErrMsg, err.Error())
```
arvindbr8, line 2607: got before want. See: go/got-before-want
```suggestion
		t.Fatalf("http2Server.operateHeaders() returned err: %v; want: %v, got, want)
```

internal/transport/transport_test.go
```diff
@@ -2592,3 +2592,20 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	}
+	defer cancel()
+
```
arvindbr8, line 2610: nix newline pls

internal/transport/transport_test.go
arvindbr8: But maybe the test name should mention fails to operate somewhere? Also pls add a comment on the test for readability?

internal/transport/transport_test.go
```diff
@@ -2592,3 +2592,20 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	hf := &http2.HeadersFrame{FrameHeader: http2.FrameHeader{StreamID: 2}}
+	metaHeaderFrame := &http2.MetaHeadersFrame{HeadersFrame: hf}
+	s := http2Server{}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	err := s.operateHeaders(ctx, metaHeaderFrame, nil)
+	if err == nil {
+		t.Fatalf("Error expected on the even stream ID. got %v", err)
+	}
+	possibleErrMsg := "received an illegal stream id: 2"
+	if !strings.Contains(err.Error(), possibleErrMsg) {
```
arvindbr8, line 2606: how about an exact match? Also the nil check can be combined with the assertion. 


```suggestion
	got := s.operateHeaders(ctx, metaHeaderFrame, nil)
	want := "received an illegal stream id: 2"
	if got := err.Error(); got != want {
```

arvindbr8 (changes requested):

arvindbr8 (dismissed):
LGTM! Nice work on the PR @AnandInguva. Congrats on the first one to the repo!

arvindbr8 (changes requested):
I think we missed a small excerpt from the HTTP2 spec -- 

> An endpoint that receives an unexpected stream identifier MUST respond with a connection error ([Section 5.4.1](https://httpwg.org/specs/rfc7540.html#ConnectionErrorHandler)) of type [PROTOCOL_ERROR](https://httpwg.org/specs/rfc7540.html#PROTOCOL_ERROR).

Right now when operateHeaders returns with an error the transport calls Close on itself (also note that the only error returned by operateHeaders is in the same invalid stream ID scenario). We should instead put a `GOAWAY` into the controlbuf and just continue the loop in HandleStreams. The `GOAWAY` should have the error code PROTOCOL_ERROR and also should close the connection. Since the connection is now closed, the next iteration in HandleStreams would then error when reading the frame and close the transport. 

By doing this, we should be able to use just the end2end test and assert that the server returns a GOAWAY frame with the error message. 

Let's please add that as part of this fix.

AnandInguva:
> I think we missed a small excerpt from the HTTP2 spec --
> 
> > An endpoint that receives an unexpected stream identifier MUST respond with a connection error ([Section 5.4.1](https://httpwg.org/specs/rfc7540.html#ConnectionErrorHandler)) of type [PROTOCOL_ERROR](https://httpwg.org/specs/rfc7540.html#PROTOCOL_ERROR).
> 
> Right now when operateHeaders returns with an error the transport calls Close on itself (also note that the only error returned by operateHeaders is in the same invalid stream ID scenario). We should instead put a `GOAWAY` into the controlbuf and just continue the loop in HandleStreams. The `GOAWAY` should have the error code PROTOCOL_ERROR and also should close the connection. Since the connection is now closed, the next iteration in HandleStreams would then error when reading the frame and close the transport.
> 
> By doing this, we should be able to use just the end2end test and assert that the server returns a GOAWAY frame with the error message.
> 
> Let's please add that as part of this fix.

I made changes according to this. PTAL.

internal/transport/http2_server.go
```diff
@@ -349,10 +349,15 @@ func (t *http2Server) operateHeaders(ctx context.Context, frame *http2.MetaHeade
 		})
 		return nil
 	}
-
+	
 	if streamID%2 != 1 || streamID <= t.maxStreamID {
 		// illegal gRPC stream id.
-		return fmt.Errorf("received an illegal stream id: %v. headers frame: %+v", streamID, frame)
+		t.controlBuf.put(&goAway{
```
arvindbr8, line 355: This is not where I would put the goAway to the controlBuf. This would result in operateHeaders never returning an error and the err handling portion in HandleStreams never getting executed. How about we do exactly the same where t.Close(err) is called in HandleStreams and the goAway frame is created using the error returned from here?

test/end2end_test.go
```diff
@@ -3901,6 +3901,41 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+	go s.Serve(lis)
+
+	conn, err := net.DialTimeout("tcp", lis.Addr().String(), defaultTestTimeout)
+	if err != nil {
+		t.Fatalf("Failed to dial: %v", err)
+	}
+	st := newServerTesterFromConn(t, conn)
+	st.greet()
+	st.writeHeadersGRPC(2, "/grpc.testing.TestService/StreamingInputCall", true)
+	frame, err := st.fr.ReadFrame()
```
arvindbr8, line 3928: servertester already has a `wantGoAway` API  which already does the assertion. We should still assert on `DebugData()`

arvindbr8 (changes requested):
LGTM modulo 2 comments

internal/transport/http2_server.go
AnandInguva: I am getting EOF when I read st.fr.ReadFrame(). This might how HandleStream operates on the `t.Close()`. 

When we put goAwayFrame and don't return error, the transport will close later? but if I put the goAwayFrame in the HandleStream and close the transport `t.Close()`, seems like the client is not getting GoAwayFrame, instead EOF is sent.

Looking at the operateHeader class, this is the only place where we return an error. For everything else, it was queue into `controlBuf`.  This https://github.com/grpc/grpc-go/blob/f135e982e6ca1949979c4e944942ccf0b20acc96/internal/transport/http2_server.go#L436 is also similar to what we are doing. I am in favor of going with this change. lmk

test/end2end_test.go
AnandInguva: Done

internal/transport/http2_server.go
arvindbr8: > I am getting EOF when I read st.fr.ReadFrame() if I put GoAwayFrame in the HandleStream and then call t.Close().

Once you put the goaway Frame in HandleStreams, you should just `continue` and the next iteration of the loop would close the transport here.

 https://github.com/grpc/grpc-go/blob/f135e982e6ca1949979c4e944942ccf0b20acc96/internal/transport/http2_server.go#L639

when the next read here fails 

https://github.com/grpc/grpc-go/blob/f135e982e6ca1949979c4e944942ccf0b20acc96/internal/transport/http2_server.go#L639

test/end2end_test.go
```diff
@@ -3901,6 +3901,38 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+
+	conn, err := net.DialTimeout("tcp", lis.Addr().String(), defaultTestTimeout)
+	if err != nil {
+		t.Fatalf("Failed to dial: %v", err)
+	}
+	st := newServerTesterFromConn(t, conn)
+	st.greet()
+	st.writeHeadersGRPC(2, "/grpc.testing.TestService/StreamingInputCall", true)
+	goAwayFrame := st.wantGoAway(http2.ErrCodeProtocol)
+	got := string(goAwayFrame.DebugData()[:])
```
arvindbr8, line 3929: nit: 
```suggestion
	got := string(goAwayFrame.DebugData())
```

arvindbr8 (approved):
LGTM, modulo one optional comment. 

I would also change the PR title to say there is a behavior change here. + PR description.

@dfawley - Could you please take a look at the patch set and see if it looks good to you?

internal/transport/http2_server.go
```diff
@@ -642,8 +642,12 @@ func (t *http2Server) HandleStreams(ctx context.Context, handle func(*Stream)) {
 		switch frame := frame.(type) {
 		case *http2.MetaHeadersFrame:
 			if err := t.operateHeaders(ctx, frame, handle); err != nil {
-				t.Close(err)
-				break
+				t.controlBuf.put(&goAway{
```
arvindbr8, line 645: Also, please add a comment in here to mention that an error from operateHeaders is when this invalid stream ID situation happens.

dfawley: I'd phrase it as: "Any error processing client headers, e.g. invalid stream ID, is considered a protocol violation."

dfawley (approved):

dfawley (approved):

internal/transport/http2_server.go
```diff
@@ -642,6 +642,8 @@ func (t *http2Server) HandleStreams(ctx context.Context, handle func(*Stream)) {
 		switch frame := frame.(type) {
 		case *http2.MetaHeadersFrame:
 			if err := t.operateHeaders(ctx, frame, handle); err != nil {
+				// Any error processing client headers,
+				// e.g. invalid stream ID, is considered a protocol violation.
```
dfawley, line 646: Nit: can you wrap this at 80 columns (consistently) please?

internal/transport/http2_server.go
```diff
@@ -642,8 +642,14 @@ func (t *http2Server) HandleStreams(ctx context.Context, handle func(*Stream)) {
 		switch frame := frame.(type) {
 		case *http2.MetaHeadersFrame:
 			if err := t.operateHeaders(ctx, frame, handle); err != nil {
-				t.Close(err)
-				break
+				// Any error processing client headers,e.g. invalid stream ID,
```
dfawley, line 645: Nit: space after comma please.

test/end2end_test.go
```diff
@@ -3901,6 +3901,37 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
 	}
 }
 
+// Tests gRPC server's behavior when a gRPC client sends a frame with an invalid
+// streamID. Per [HTTP/2 spec]: Streams initiated by a client MUST use
+// odd-numbered stream identifiers. This test sets up a test server and send a
```
dfawley, line 3906: Nit: s/send/sends/

test/end2end_test.go
```diff
@@ -3901,6 +3901,37 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+// header frame with stream ID of 2. The test asserts that a subsequent read on
+// the transport throws an error.
+//
+// [HTTP/2 spec]: https://httpwg.org/specs/rfc7540.html#StreamIdentifiers
+func (s) TestClientInvalidStreamID(t *testing.T) {
```
dfawley, line 3911: Since we have this test now, and if we don't already have tests for other invalid stream ID conditions (e.g. a lower ID than the previous, or a stream ID of zero for the first stream), then now might be a good time to add related test cases.

test/end2end_test.go
```diff
@@ -3901,6 +3901,37 @@ func (s) TestClientRequestBodyErrorCloseAfterLength(t *testing.T) {
...
+
+	conn, err := net.DialTimeout("tcp", lis.Addr().String(), defaultTestTimeout)
+	if err != nil {
+		t.Fatalf("Failed to dial: %v", err)
+	}
+	st := newServerTesterFromConn(t, conn)
+	st.greet()
+	st.writeHeadersGRPC(2, "/grpc.testing.TestService/StreamingInputCall", true)
+	goAwayFrame := st.wantGoAway(http2.ErrCodeProtocol)
+	want := "received an illegal stream id: 2. headers frame: [FrameHeader HEADERS flags=END_STREAM|END_HEADERS stream=2 len=60]"
```
dfawley, line 3929: I'm a bit worried about matching the string generated by another library (the headers frame debug data).

Can we make this exclude that and do `strings.Contains()` in the `if` below, instead?

test/end2end_test.go
AnandInguva: I will take a look today and add if tests are not present.

AnandInguva: >> stream ID of zero for the first stream

For this case, the http2/frame itself throws an error when we write frames of streamID 0 using serverTester. So I am guessing we can't do this test end to end and ask for question if we need this test at all?

dfawley: It would still be nice to have a test for this, but I'm fine with punting for now if it's difficult to write.  Presumably this is a different code path on the server (the framer may error reading the frame), so a test would still be a good idea to have one day.

AnandInguva: Added one more test. PTAL

test/end2end_test.go
```diff
@@ -3926,8 +3926,37 @@ func (s) TestClientInvalidStreamID(t *testing.T) {
 	st.greet()
 	st.writeHeadersGRPC(2, "/grpc.testing.TestService/StreamingInputCall", true)
 	goAwayFrame := st.wantGoAway(http2.ErrCodeProtocol)
-	want := "received an illegal stream id: 2. headers frame: [FrameHeader HEADERS flags=END_STREAM|END_HEADERS stream=2 len=60]"
-	if got := string(goAwayFrame.DebugData()); got != want {
+	want := "received an illegal stream id: 2."
+	if got := string(goAwayFrame.DebugData()); !strings.Contains(got, want) {
+		t.Fatalf("Error received: %v, want: %v", got, want)
```
dfawley, line 3931: Nit: "`want: Contains(%v)`" or something to indicate what's happening.

dfawley (approved):

test/end2end_test.go
AnandInguva: Done

dfawley: I see it changed below but not here.

AnandInguva: ah right. changed in the previous test as well. thanks

arvindbr8 (approved):

PR #6945, opened 2024-01-26
Propagate resolver attributes from cdsbalancer via the addresses. 

anicr7:
To make the propagation easier add a helper method in attributes library which merges two attributes together.

This feature allows any resolver attributes to be propagated further along the xds layers. This is especially useful for resolvers that embed the xds resolvers for example and want to propagated custom attributes.

@dfawley

### DIFF
--- a/attributes/attributes.go
+++ b/attributes/attributes.go
@@ -70,6 +70,33 @@ func (a *Attributes) Value(key any) any {
 	return a.m[key]
 }
 
+// Merge returns a new Attributes containing the union of keys and values 
+// associated with two Attributes. If the same key appears multiple times, the 
+// last value overwrites all previous values for that key. To remove an
+// existing key, use a nil value.  Value should not be modified later.
+func (a *Attributes) Merge(b *Attributes) *Attributes {
+	totalLen := 0
+	if a != nil {
+		totalLen += len(a.m)
+	}
+	if b != nil {
+		totalLen += len(b.m)
+	}
+	c := &Attributes{m: make(map[any]any, totalLen)}
+	if a != nil {
+		for k, v := range a.m {
+			c.m[k] = v
+		}
+	}
+	if b != nil {
+		for k, v := range b.m {
+			c.m[k] = v
+		}
+	}
+
+	return c
+}
+
 // Equal returns whether a and o are equivalent.  If 'Equal(o any) bool' is
 // implemented for a value in the attributes, it is called to determine if the
 // value matches the one stored in the other attributes.  If Equal is not
--- a/attributes/attributes.go
+++ b/attributes/attributes.go
@@ -45,7 +45,7 @@ func New(key, value any) *Attributes {
 	return &Attributes{m: map[any]any{key: value}}
 }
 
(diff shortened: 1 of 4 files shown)


attributes/attributes.go
```diff
@@ -45,7 +45,7 @@ func New(key, value any) *Attributes {
 	return &Attributes{m: map[any]any{key: value}}
 }
 
-// WithValue returns a new Attributes containing the previous keys and values
+// WithValue returns a new Attribute containing the previous keys and values
```
dfawley, line 48: This was technically correct before.

xds/internal/balancer/cdsbalancer/cdsbalancer.go
```diff
@@ -666,6 +670,11 @@ type ccWrapper struct {
 func (ccw *ccWrapper) NewSubConn(addrs []resolver.Address, opts balancer.NewSubConnOptions) (balancer.SubConn, error) {
 	newAddrs := make([]resolver.Address, len(addrs))
 	for i, addr := range addrs {
+		// First merge the resolver attributes if any and the add the handshake
+		// attributes.
+		if ccw.resolverAttr != nil {
+			addr.Attributes = addr.Attributes.Merge(ccw.resolverAttr)
+		}
```
dfawley, line 677: This mutates the input parameter `addrs` -- is that desirable?

xds/internal/balancer/cdsbalancer/cdsbalancer.go
```diff
@@ -677,6 +686,11 @@ func (ccw *ccWrapper) NewSubConn(addrs []resolver.Address, opts balancer.NewSubC
 func (ccw *ccWrapper) UpdateAddresses(sc balancer.SubConn, addrs []resolver.Address) {
 	newAddrs := make([]resolver.Address, len(addrs))
 	for i, addr := range addrs {
+		// First merge the resolver attributes if any and the add the handshake
+		// attributes.
+		if ccw.resolverAttr != nil {
+			addr.Attributes = addr.Attributes.Merge(ccw.resolverAttr)
+		}
```
dfawley, line 693: So we add it to both the addresses we send down to the child policies AND the addresses we get from the children when subchannels are created for them?  Is that necessary?

xds/internal/balancer/cdsbalancer/cdsbalancer.go
```diff
@@ -287,6 +288,7 @@ func (b *cdsBalancer) UpdateClientConnState(state balancer.ClientConnState) erro
 		}
 		b.xdsClient = c
 	}
+	b.ccw.resolverAttr = state.ResolverState.Attributes
```
dfawley, line 291: This is reading global attributes for the resolver itself.

But then you're setting it on the individual addresses.

I don't think that's correct.  I think we should propagate it down to the children in the same place, `ResolverState.Attributes`.  It's not necessarily okay to turn global or endpoint attributes into address-level attributes, as different attributes serve different purposes.

dfawley:
I'm pretty sure it's decided that we will not be pursuing this approach anymore.

PR #6946, opened 2024-01-29
internal/transport: Remove redundant if conditional in http2_server

chengfang:
In internal/transport/http2_server.go, around line 640, the if-else branches contain the identical code, and should be combined:

```go
if err == io.EOF || err == io.ErrUnexpectedEOF {
  t.Close(err)
  return
}
t.Close(err)
return
```

RELEASE NOTES: none


arvindbr8 (approved):
LGTM. @zasweq -- do you mind taking a second pass at this?

zasweq (approved):

PR #6952, opened 2024-01-31
fix enabling compression by trimming whitespaces in accept encoding header

### LINKED ISSUE #6951: Response is not compressed when grpc-accept-encoding header contains spaces between values
1.61.0

1.21.2

macOS

### What did you do?
Call grpc-go server from https://github.com/grpc/grpc client with compression enabled.

[grpc-core](https://github.com/grpc/grpc/blob/58300b9d1fcce62069f793050e6d65da874a98d9/src/core/lib/compression/compression_internal.cc#L72) adds a whitespace character between values:
```
grpc-accept-encoding=identity, deflate, gzip
```
go server should accept this header.

### What did you see instead?
There is no compression happened between server and client.
The reason for that the [go code](https://github.com/grpc/grpc-go/blob/02858ee5064040458f7e3e04e10765a62814c50b/server.go#L2171) splits the value with `,` and doesn't trim the results.
The only valid header for go is the following:
```
grpc-accept-encoding=identity,deflate,gzip
```

[grpc-dotnet](https://github.com/grpc/grpc-dotnet/pull/1882) repo is has done something similar before.

sercand:
fixes #6951 

RELEASE NOTES: none

### DIFF
--- a/server.go
+++ b/server.go
@@ -2119,7 +2119,7 @@ func ClientSupportedCompressors(ctx context.Context) ([]string, error) {
 		return nil, fmt.Errorf("failed to fetch the stream from the given context %v", ctx)
 	}
 
-	return strings.Split(stream.ClientAdvertisedCompressors(), ","), nil
+	return stream.ClientAdvertisedCompressors(), nil
 }
 
 // SetTrailer sets the trailer metadata that will be sent when an RPC returns.
@@ -2159,7 +2159,7 @@ func (c *channelzServer) ChannelzMetric() *channelz.ServerInternalMetric {
 
 // validateSendCompressor returns an error when given compressor name cannot be
 // handled by the server or the client based on the advertised compressors.
-func validateSendCompressor(name, clientCompressors string) error {
+func validateSendCompressor(name string, clientCompressors []string) error {
 	if name == encoding.Identity {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


arvindbr8 (approved):
LGTM. Thanks for the PR @sercand

server.go
```diff
@@ -2119,7 +2119,11 @@ func ClientSupportedCompressors(ctx context.Context) ([]string, error) {
 		return nil, fmt.Errorf("failed to fetch the stream from the given context %v", ctx)
 	}
 
-	return strings.Split(stream.ClientAdvertisedCompressors(), ","), nil
+	values := strings.Split(stream.ClientAdvertisedCompressors(), ",")
```
dfawley, line 2122: It seems `ClientAdvertisedCompressors()` is always used with a `strings.Split(",")`.  Can we make that return `[]string` instead and do the split/trim there, please?

dfawley (changes requested):

server.go
sercand: @dfawley I commited the changes you requested. Could you review again 🙏

dfawley (approved):
Thanks!

arvindbr8 (approved):

arvindbr8:
LGTM. Thanks for the PR @sercand

PR #6919, opened 2024-01-12
deps: move from github.com/golang/protobuf to google.golang.org/protobuf/proto

Clement-Jean:
This is the second part of #6736 which is moving away from github.com/golang/protobuf in favor of google.golang.org/protobuf/proto

RELEASE NOTES:
* Use google.golang.org/protobuf/proto instead of github.com/golang/protobuf.

PR #6953, opened 2024-01-31
status: modify TestStatus_ErrorDetails_Fail to replace protoimpl package

arvindbr8:
The protoimpl package is being currently used in a test (which got introduced in #6919). But this package cannot be accessed directly in google3, hence we need a different way of testing the case. 

RELEASE NOTES: none

### DIFF
--- a/status/status_test.go
+++ b/status/status_test.go
@@ -24,14 +24,12 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	cpb "google.golang.org/genproto/googleapis/rpc/code"
 	epb "google.golang.org/genproto/googleapis/rpc/errdetails"
 	spb "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/protoadapt"
 	"google.golang.org/protobuf/reflect/protoreflect"
-	"google.golang.org/protobuf/runtime/protoimpl"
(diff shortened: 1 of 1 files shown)


status/status_test.go
```diff
@@ -419,17 +418,24 @@ func (s) TestStatus_ErrorDetails_Fail(t *testing.T) {
 		},
 	}
 	for _, tc := range tests {
-		got := tc.s.Details()
-		if !cmp.Equal(got, tc.i, cmp.Comparer(proto.Equal), cmp.Comparer(equalError)) {
-			t.Errorf("(%v).Details() = %+v, want %+v", str(tc.s), got, tc.i)
+		details := tc.s.Details()
```
dfawley, line 421: You should first ensure `len(details) == len(tc.want)` here or it could panic.

status/status_test.go
```diff
@@ -419,17 +418,24 @@ func (s) TestStatus_ErrorDetails_Fail(t *testing.T) {
...
+		for i, d := range details {
+			// s.Deatils can either contain an error or a proto message.  We
```
dfawley, line 423: Spelling

dfawley (approved):

status/status_test.go
arvindbr8: that's right! thanks for catching that

status/status_test.go
arvindbr8: done

PR #6954, opened 2024-02-01
grpc: add ability to compile with or without tracing

hugelgupf:
golang.org/x/net/trace is the only dependency of gRPC Go that depends -- through html/template and text/template -- on reflect's MethodByName. Binaries with MethodByName are not eligible for Go dead code elimination.

As of protobuf v1.32.0 combined with Go 1.22, Go protobufs support compilation with dead code elimination. This commit allows users of gRPC Go to also make use of dead code elimination for smaller binary sizes.

---

What's the best place to insert a build test to make sure that `go build -tags=grpcnotrace ./...` still builds?

RELEASE NOTES:
* *: Allow building without `x/net/trace` by using `grpcnotrace` to enable dead code elimination

### DIFF
--- a/trace_notrace.go
+++ b/trace_notrace.go
@@ -1,0 +1,52 @@
+//go:build grpcnotrace
+
+/*
+ *
+ * Copyright 2024 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


hugelgupf:
I added a release notes string, the test failure looks like a flake to me. Seems ready for review :)

trace_notrace.go
```diff
@@ -0,0 +1,52 @@
+//go:build grpcnotrace
+
+/*
+ *
+ * Copyright 2015 gRPC authors.
```
dfawley, line 5: Please update the date here & below

dfawley (commented):
LGTM; just one administrivia item.

trace_notrace.go
hugelgupf: Done.

hugelgupf:
Thanks @dfawley!

dfawley (approved):
Thanks!

PR #6956, opened 2024-02-01
Set the security level of Windows named pipes to NoSecurity

irsl:
Communication over named pipes on Windows is local IPC.

RELEASE NOTES:
- Set the security level of Windows named pipes to NoSecurity

### DIFF
--- a/credentials/local/local.go
+++ b/credentials/local/local.go
@@ -65,6 +65,9 @@ func getSecurityLevel(network, addr string) (credentials.SecurityLevel, error) {
 	// Local TCP connection
 	case strings.HasPrefix(addr, "127."), strings.HasPrefix(addr, "[::1]:"):
 		return credentials.NoSecurity, nil
+	// Windows named pipe connection
+	case network == "pipe" && strings.HasPrefix(addr, `\\.\pipe\`):
+		return credentials.NoSecurity, nil
 	// UDS connection
 	case network == "unix":
 		return credentials.PrivacyAndIntegrity, nil


credentials/local/local.go
```diff
@@ -65,6 +65,9 @@ func getSecurityLevel(network, addr string) (credentials.SecurityLevel, error) {
 	// Local TCP connection
 	case strings.HasPrefix(addr, "127."), strings.HasPrefix(addr, "[::1]:"):
 		return credentials.NoSecurity, nil
+	// Windows named pipe connection
+	case network == "pipe" && strings.HasPrefix(addr, `\\.\pipe\`):
+		return credentials.NoSecurity, nil
```
dfawley, line 70: Can you explain why `NoSecurity` was chosen over `PrivacyAndIntegrity`?  I'm forgetting the technical details why `localhost` was considered `NoSecurity`, but UDS was P&I, but I believe the fact that unix permissions could be applied to sockets to prevent tampering had a lot to do with it.

cc @matthewstevenson88 @gtcooke94 for their security perspective on this.

irsl: The project I'm working on connects to containerd and they use NoSecurity:

https://github.com/containerd/containerd/blob/main/client/client.go#L126

https://github.com/grpc/grpc-go/blob/v1.61.0/credentials/insecure/insecure.go#L43

The company I work for doesn't allow using the insecure package, only the local package, so porting the behavior for the unsupported platform seemed to be the way to go.

matthewstevenson88: My understanding of why UDS connections are considered `PrivacyAndIntegrity` matches what @dfawley describes.

Having `NoSecurity` here sounds reasonable to me, FWIW.

dfawley: @irsl `Insecure` is what they're using because that works for them.  `Insecure`->`NoSecurity` which means we shouldn't send credentials on the connection.  This might work for your project, but it will be more limiting for others if they need per-call credentials.

@matthewstevenson88 why isn't `PrivacyAndIntegrity` okay in this scenario?  Do we not believe windows pipes are as secure as UDS, or do we not know / have time to investigate?

dfawley: I think it's okay to use `NoSecurity` for now since it's probably not worth prioritizing, but then we'd want a TODO and an issue to investigate whether P&I would be even more appropriate.

irsl: I just tested the connection to containerd with `PrivacyAndIntegrity` set here and it works just fine. This makes me quite confused about what this security flag does on the client side - I either misunderstand something fundamental or there is a potential for MitM attacks here. This was already suspicious as the code has been working fine on unix platforms - where PrivacyAndIntegrity is set and containerd server clearly doesn't encrypt/authenticate the gRPC requests.
Regardless, I'm adjusting the PR soon.

irsl: Thanks for the explanation. I did some more tests to be able to answer the question from this aspect. I created a simple users with default settings, no special permissions or group memberships (SeChangeNotifyPrivilege is the only privilege).

- I called ListenPipe as this low privileged user without a [SecurityDescriptor](https://github.com/microsoft/go-winio/blob/main/pipe.go#L491). (An empty SD means no additional access restrictions are enforced by the kernel wrt client connections.)
- It succeeded and did show up for other users (confirmed via `[System.IO.Directory]::GetFiles("\\.\\pipe\\")`)
- **This confirms it is possible to start listening on any pipe names not yet in use**
- Other processes running as the same user were able to connect seamlessly
- Other low privileged users couldn't connect at all (access denied)
- **Processes running with elevated privileges (e.g. the admin user after UAC or NT_AUTHORITY\SYSTEM) could connect to the pipe of the low-privileged user**
- In the case of containerd, both the server and client is typically executed with elevated privileges, so hijacking connections is definitely possible in the case of race condition

According to your previous comment, `NoSecurity` is the right choice here. Re-adjusting the PR.

irsl: Friendly ping.

dfawley: SGTM, and thanks for looking into this.

dfawley (approved):

arvindbr8 (approved):
LGTM. Thanks for the PR

PR #6957, opened 2024-02-01
Delay closing connection on the server after all streams are completed

s-matyukevich:
We hit exactly the same bug that was previously reported in grpc-java https://github.com/grpc/grpc-java/issues/9566

It was also fixed in http2 go standard lib https://github.com/golang/net/commit/cd69bc3fc700721b709c3a59e16e24c67b58f6ff#diff-75c1d03aa4d9dfad1fe759ba9fce0462d6e88f2e33242425c4563d4ffb454951

TL;DR is that when server sends large responses and configures MaxConnectionAge, sometimes it closes the connection too early after finishing processing all active streams (the client is not done reading all data from the connection yet, the data is still stored inside TCP kernel buffer on the server) In such cases the client receives RST and gets the following error:

```
error while calling server error="rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: read tcp 10.131.194.232:41856->10.132.142.4:3000: read: connection reset by peer\", received prior goaway: code: NO_ERROR, debug data: \"max_age\""
```

The fix is very simple: sleep for 1 sec after finishing processing all streams and before closing the connection.

I can reliably reproduce this error in our staging environment, and I verified that the proposed fix actually works. 

I am still working on a unit test that reproduces the error locally. My plan is to modify [this](https://github.com/grpc/grpc-go/blob/master/test/goaway_test.go#L55) test and make the server send large responses, but I'll appreciate any guidance on how we should test this.

I can also make the sleep duration configurable if you prefer.

### DIFF
--- a/server.go
+++ b/server.go
@@ -31,6 +31,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"testing"
 	"time"
 
 	"golang.org/x/net/trace"
@@ -66,6 +67,8 @@ const (
 	listenerAddressForServeHTTP = "listenerAddressForServeHTTP"
 )
 
+var goAwayTimeout = 1 * time.Second
+
 func init() {
 	internal.GetServerCredentials = func(srv *Server) credentials.TransportCredentials {
 		return srv.opts.creds
@@ -83,6 +86,10 @@ func init() {
 	internal.BinaryLogger = binaryLogger
 	internal.JoinServerOptions = newJoinServerOption
 	internal.RecvBufferPool = recvBufferPool
+
+	if testing.Testing() {
+		goAwayTimeout = 0
+	}
 }
 
 var statusOK = status.New(codes.OK, "")
@@ -1013,7 +1020,7 @@ func (s *Server) serveStreams(ctx context.Context, st transport.ServerTransport,
 	}
 
 	defer func() {
-		time.Sleep(2 * time.Second)
+		time.Sleep(goAwayTimeout)
 		st.Close(errors.New("finished serving streams for the server transport"))
 		for _, sh := range s.opts.statsHandlers {
 			sh.HandleConn(ctx, &stats.ConnEnd{})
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -83,6 +86,10 @@ func init() {
 	internal.BinaryLogger = binaryLogger
 	internal.JoinServerOptions = newJoinServerOption
 	internal.RecvBufferPool = recvBufferPool
+
+	if testing.Testing() {
+		goAwayTimeout = 0
+	}
```
s-matyukevich, line 92: The tests that use `grpc.NewServer` and define very short timeouts should not be affected.

server.go
```diff
@@ -83,6 +86,10 @@ func init() {
 	internal.BinaryLogger = binaryLogger
 	internal.JoinServerOptions = newJoinServerOption
 	internal.RecvBufferPool = recvBufferPool
+
+	if testing.Testing() {
```
arvindbr8, line 90: seems like `testing.Testing()` was only made available since go.1.21.0 https://pkg.go.dev/testing#Testing

We maintain a backward compatibility to the last 3 versions of go. (i.e., v1.19+)

s-matyukevich: Any suggestions what should I do about it? Options I can think of:
* add a hack to determine if this is a testing build, something like this https://stackoverflow.com/questions/14249217/how-do-i-know-im-running-within-go-test
* Don't enable the delay by default, add a server option to activate it
* Enable delay by default and ignore the fact that it could break some tests that expect the server to be shut-down faster then 1 sec. Add a server option so that users can disable (or maybe configure?) the delay in their tests.

dfawley:
Oh this is especially nasty, and is likely the root cause behind #5358, which has confounded me for some time.

So the root cause of the problem is that if the server's receive buffer is not empty, then closing the connection results in RST instead of FIN.

https://www.freesoft.org/CIE/RFC/1122/99.htm
> If such a host issues a CLOSE call while received data is still pending in TCP, or if new data is received after CLOSE is called, its TCP SHOULD send a RST to show that data was lost.

I have a slightly different idea here, for a way to fix this without incurring a fixed delay on everything.  Let me play around with it a bit.

dfawley:
Thank you for the PR.  Please see https://github.com/grpc/grpc-go/pull/6977 which has a preferred fix.

PR #6926, opened 2024-01-18
grpc: Add StaticMethod CallOption

zasweq:
This PR adds a call option that specifies whether a method is static. This is intended to be used by generated code and users to get a method attribute within OpenTelemetry while protecting the cardinality of OpenTelemetry timeseries.

RELEASE NOTES:
* grpc: Add StaticMethod CallOption as a signal to stats handler that a method is safe to use as an instrument key.

PR #6960, opened 2024-02-02
protoc: Change protoc to include generated call option

zasweq:
This PR changes protoc to include generated call option. This will be used in stats handlers. This cannot be merged until gRPC is released with #6926, and we will have to update the minimum version of gRPC required for new versions of protoc-gen-go, and features which use this will need to regenerate their protos (such as Envoy's protos). Per offline discussion, OpenTelemetry e2e testing will be sufficient enough for testing this feature, as protoc-gen-go is implicitly tested in our tests which all use generated code from protoc-gen-go.

RELEASE NOTES:
* protoc: Change protoc to include generated call option


cmd/protoc-gen-go-grpc/grpc.go
```diff
@@ -170,7 +170,7 @@ func generateFileContent(gen *protogen.Plugin, file *protogen.File, g *protogen.
 
 	g.P("// This is a compile-time assertion to ensure that this generated file")
 	g.P("// is compatible with the grpc package it is being compiled against.")
-	g.P("// Requires gRPC-Go v1.32.0 or later.")
+	g.P("// Requires gRPC-Go v1.32.0 or later.")                     // TODO: Update to latest released gRPC version
 	g.P("const _ = ", grpcPackage.Ident("SupportPackageIsVersion7")) // When changing, update version number above.
```
dfawley, line 174: This needs to be updated to `Version8` now.  The comment just lets users discover more easily when `Version8` became available.

cmd/protoc-gen-go-grpc/grpc.go
```diff
@@ -309,9 +309,15 @@ func genClientMethod(gen *protogen.Plugin, file *protogen.File, g *protogen.Gene
 		g.P(deprecationComment)
 	}
 	g.P("func (c *", unexport(service.GoName), "Client) ", clientSignature(g, method), "{")
+	g.P("cOpts := make([]", grpcPackage.Ident("CallOption"), ", len(opts))")
+	g.P("for i, opt := range opts {")
+	g.P("	cOpts[i] = opt")
+	g.P("}")
+	g.P("cOpts = append(cOpts, ", grpcPackage.Ident("RegisteredMethod()"), "\"")
```
dfawley, line 316: Can be simplified to:

```go
opts := append([]CallOption{grpc.RegisteredMethod()}, opts...) // shadows `opts` so you don't need the subsequent diffs even
``` 

Or use `copy` instead of a loop (and also pre-allocate `len(opts)+1`).

dfawley (commented):
LGTM with one nit.  But also blocked on the next gRPC release.

cmd/protoc-gen-go-grpc/grpc.go
zasweq: Done.

cmd/protoc-gen-go-grpc/grpc.go
zasweq: Changed to former. Chose that since less lines of code and get rid of subsequent diffs.

zasweq:
Finished comments, as discussed offline, let's run this by Eric to see if we even want to go down this route, or if we should preregister on ClientConn.

zasweq:
Also cannot merge until 1.62 release.

zasweq:
Reopened this since I think this is ready to be merged as the release is out.

dfawley (approved):
Please regenerate all the protos and make sure vet-proto is green before merging.

PR #6961, opened 2024-02-02
 deps: move from github.com/golang/protobuf to google.golang.org/protobuf/proto

Clement-Jean:
@dfawley this is the continuation of #6919

I'm still unsure about the reflection package. I read the [README](https://github.com/grpc/grpc-go/blob/master/reflection/grpc_testing_not_regenerate/README.md) there and it says that we shouldn't regenerate the testv3.go file. I see two solutions, either we manually edit the generated file (this was previously done) or we regenerate with an older version of protoc. What do you think?

RELEASE NOTES: none


internal/pretty/pretty.go
```diff
@@ -36,17 +35,17 @@ const jsonIndent = "  "
 // If marshal fails, it falls back to fmt.Sprintf("%+v").
 func ToJSON(e any) string {
 	switch ee := e.(type) {
-	case protov1.Message:
+	case protoadapt.MessageV1:
```
arvindbr8, line 38: nit: can we combine some of this now from below `case: protoadapt.MessageV2`?

security/advancedtls/examples/go.mod
```diff
@@ -9,7 +9,6 @@ require (
 )
 
 require (
-	github.com/golang/protobuf v1.5.3 // indirect
```
arvindbr8, line 12: How are these deps changing without any changes under this sub module?

credentials/credentials.go
```diff
@@ -287,5 +287,5 @@ type ChannelzSecurityValue interface {
 type OtherChannelzSecurityValue struct {
 	ChannelzSecurityValue
 	Name  string
-	Value proto.Message
+	Value protoadapt.MessageV1
```
arvindbr8, line 290: this could just be 
```suggestion
	Value proto.message
```
of `google.golang.org/protobuf/proto` if you also fix the usage in https://github.com/grpc/grpc-go/blob/d41b01db97ca2e3627b2c9949fffe8f152a4255d/channelz/service/service.go#L193 to use Value as is

internal/pretty/pretty.go
```diff
@@ -36,17 +35,17 @@ const jsonIndent = "  "
 // If marshal fails, it falls back to fmt.Sprintf("%+v").
 func ToJSON(e any) string {
```
arvindbr8, line 36: @dfawley, I think we should still support both versions to keep backwards compatibility to older proto messages?

arvindbr8 (changes requested):

dfawley:
> 
I'm still unsure about the reflection package. I read the README there and it says that we shouldn't regenerate the testv3.go file. I see two solutions, either we manually edit the generated file (this was previously done) or we regenerate with an older version of protoc. What do you think?

I don't believe this should be changed.  It tests that we will work with old generated code; we'd like to still confirm that fact in our tests.

internal/pretty/pretty.go
Clement-Jean: Do you mean removing the switch and receive a `proto.message` as ToJSON argument?

arvindbr8:
@Clement-Jean -- friendly ping here, didnt want the bot to mark this stale

Clement-Jean:
@arvindbr8 could you take a look at the comment on your suggestions? Not entirely sure about how to proceed

arvindbr8:
Apologies for the delay -- @Clement-Jean. I have been busy with some internal priorities. Taking a look now.

arvindbr8:
Also it is weird that I'm not able to see your reply on my comment. however there was a email notification for it

> Do you mean removing the switch and receive a proto.message as ToJSON argument?

you can ignore this comment. That was only a nit to combine the logic in the both the switch cases. Im not too inclined for it

Please let me know if you have other questions.

Clement-Jean:
@arvindbr8 so basically right now we are blocked at this point. We need to wait for @dfawley rewrite of channelz. Once this is done, we can go further and do the change in `credentials/credentials.go` that you suggested.

arvindbr8:
Ah I see.. I would prefer if we could decouple those changes. You can ignore my comment there. I will track that suggestion in Doug's PR.

arvindbr8 (approved):

arvindbr8:
Approved. Thanks for the PR. We need 1 more person from the team to take a look at it.

security/advancedtls/examples/go.mod
dfawley: This is an indirect dep, so it changes because the grpc deps change.

internal/pretty/pretty.go
dfawley: +1; it would be nice to unify these two codepaths completely, but feel free to disregard if you don't want to do it.

I.e.

```go
if ee, ok := e.(protoadapt.MessageV1); ok {
	e = protoadapt.MessageV2Of(e)
}
switch ee := e.(type) {
	// no case for MessageV1 now
...
```

I don't know why the existing code sets different `MarshalOptions` but I highly suspect it's an oversight caused by having two codepaths that do essentially the same thing.

dfawley (approved):
LGTM.  Thanks for the change.  Please see the optional requested change and let us know if you don't want to do it and we can merge this as-is.

dfawley:
Thanks for the PR!  I went ahead and made some minor cleanup edits to speed things up; hope that's okay.  :)

dfawley (approved):

PR #6962, opened 2024-02-04
resolver/dns: Add SetMinResolutionInterval Option

HomayoonAlimohammadi:
In this merge request I tried to make `SetMinResolutionInterval` which is currently a constant duration variable by providing options. SetMinResolutionInterval sets the default minimum interval at which DNS re-resolutions are allowed. This helps to prevent excessive re-resolution.

RELEASE NOTES:
- resolver/dns: Add SetMinResolutionInterval sets the default minimum interval at which DNS re-resolutions are allowed.

### DIFF
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -49,16 +49,16 @@ var (
 	// MinResolutionRate is the minimum rate at which re-resolutions are
 	// allowed. This helps to prevent excessive re-resolution.
 	MinResolutionRate = 30 * time.Second
+
+	// ResolvingTimeout specifies the maximum duration for a DNS resolution request.
+	// If the timeout expires before a response is received, the request will be canceled.
+	//
+	// It is recommended to set this value at application startup. Avoid modifying this variable
+	// after initialization as it's not thread-safe for concurrent modification.
+	ResolvingTimeout = 30 * time.Second
+
+	logger = grpclog.Component("dns")
 )
-
-// ResolvingTimeout specifies the maximum duration for a DNS resolution request.
-// If the timeout expires before a response is received, the request will be canceled.
-//
-// It is recommended to set this value at application startup. Avoid modifying this variable
-// after initialization as it's not thread-safe for concurrent modification.
-var ResolvingTimeout = 30 * time.Second
-
-var logger = grpclog.Component("dns")
 
 func init() {
 	resolver.Register(NewBuilder())
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -41,9 +41,15 @@ import (
 	"google.golang.org/grpc/serviceconfig"
 )
 
-// EnableSRVLookups controls whether the DNS resolver attempts to fetch gRPCLB
-// addresses from SRV records.  Must not be changed after init time.
-var EnableSRVLookups = false
+var (
+	// EnableSRVLookups controls whether the DNS resolver attempts to fetch gRPCLB
+	// addresses from SRV records.  Must not be changed after init time.
+	EnableSRVLookups = false
+
+	// MinResolutionRate is the minimum rate at which re-resolutions are
+	// allowed. This helps to prevent excessive re-resolution.
+	MinResolutionRate = 30 * time.Second
+)
 
 // ResolvingTimeout specifies the maximum duration for a DNS resolution request.
 // If the timeout expires before a response is received, the request will be canceled.
@@ -208,7 +214,7 @@ func (d *dnsResolver) watcher() {
 			// Success resolving, wait for the next ResolveNow. However, also wait 30
 			// seconds at the very least to prevent constantly re-resolving.
 			backoffIndex = 1
-			waitTime = internal.MinResolutionRate
+			waitTime = MinResolutionRate
 			select {
 			case <-d.ctx.Done():
 				return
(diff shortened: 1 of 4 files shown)


HomayoonAlimohammadi:
@dfawley I would really appreciate it if you would take a look at this merge request.

HomayoonAlimohammadi:
I'm not able to add any labels or milestones. Would you please help me around?

dfawley:
Hi, I think a change like this is OK but we'd want it to be a global configuration knob like we are doing with DNS resolution timeouts in #6917.  Please let us know if you have any questions about that approach.

HomayoonAlimohammadi:
> Hi, I think a change like this is OK but we'd want it to be a global configuration knob like we are doing with DNS resolution timeouts in #6917. Please let us know if you have any questions about that approach.

Hello again, sorry for the later reply. I've changed the structure to the best of my knowledge according to the pull request you've mentioned. Please let me know if there's any problem. @dfawley

HomayoonAlimohammadi:
I resolved the conflicts and it's ready to go! Let me know if I need to change anything.

dfawley:
Sorry for the delay.. I will review it this week for sure.

internal/resolver/dns/dns_resolver.go
```diff
@@ -41,9 +41,15 @@ import (
 	"google.golang.org/grpc/serviceconfig"
 )
 
-// EnableSRVLookups controls whether the DNS resolver attempts to fetch gRPCLB
-// addresses from SRV records.  Must not be changed after init time.
-var EnableSRVLookups = false
+var (
```
dfawley, line 44: Could you please also pull in `ResolvingTimeout` here?

dfawley (commented):
LGTM, just a few minor comments.  Thank you for the PR!

internal/resolver/dns/dns_resolver.go
HomayoonAlimohammadi: Thanks for the comments! I also pulled the `logger` variable. Let me know if it was not a good idea.

resolver/dns/dns_resolver.go
```diff
@@ -52,3 +52,12 @@ func SetResolvingTimeout(timeout time.Duration) {
 func NewBuilder() resolver.Builder {
 	return dns.NewBuilder()
 }
+
+// SetMinResolutionRate sets the default minimum rate at which DNS re-resolutions are
+// allowed. This helps to prevent excessive re-resolution.
+//
+// Using this option overwrites the default [MinResolutionRate] specified
+// in the internal dns resolver package.
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: Got it! Done.

internal/resolver/dns/dns_resolver_test.go
```diff
@@ -70,9 +70,9 @@ func overrideNetResolver(t *testing.T, r *testNetResolver) {
 
 // Override the DNS Min Res Rate used by the resolver.
 func overrideResolutionRate(t *testing.T, d time.Duration) {
-	origMinResRate := dnsinternal.MinResolutionRate
-	dnsinternal.MinResolutionRate = d
-	t.Cleanup(func() { dnsinternal.MinResolutionRate = origMinResRate })
+	origMinResRate := dns.MinResolutionRate
+	dns.MinResolutionRate = d
+	t.Cleanup(func() { dns.MinResolutionRate = origMinResRate })
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: That's right, thanks for pointing it out! Done.

HomayoonAlimohammadi:
I struggle to understand why the `vet-proto` fails. Would you please help me out?

dfawley:
> I struggle to understand why the `vet-proto` fails. Would you please help me out?

This check is optional, and the failure is caused by an upstream protobuf definition change.  Please ignore it.

HomayoonAlimohammadi:
> > I struggle to understand why the `vet-proto` fails. Would you please help me out?
> 
> This check is optional, and the failure is caused by an upstream protobuf definition change. Please ignore it.

Thanks a lot. I think all the comments are resolved. Let me know if I need to change anything.

resolver/dns/dns_resolver.go
```diff
@@ -52,3 +52,9 @@ func SetResolvingTimeout(timeout time.Duration) {
 func NewBuilder() resolver.Builder {
 	return dns.NewBuilder()
 }
+
+// SetMinResolutionRate sets the default minimum rate at which DNS re-resolutions are
+// allowed. This helps to prevent excessive re-resolution.
+func SetMinResolutionRate(d time.Duration) {
```
dfawley, line 58: I swear I wrote this comment in the last pass, but it's gone now.  :)  Sorry!

Please rename this to `SetMinResolutionInterval` (and change the global variable name to match).  A "rate" is the opposite of what this is.  (This actually determines the _maximum_ rate, or the _minimum_ interval.)

HomayoonAlimohammadi: You're totally right, thanks for pointing it out!
I think I managed to change all of them. I triple checked to make sure but let me know if there's still anything wrong.

dfawley (approved):
Thanks for the PR!

arvindbr8 (approved):
Looks good to me. Thanks for the PR @HomayoonAlimohammadi.

Slightly updated the PR title/description to match the latest diff.

PR #6965, opened 2024-02-05
deps: fix backwards compatibility with encoding

arvindbr8:
In `encoding/proto.go` when Marshal() and Unmarshal() we should be able to handle both v1 and v2 of proto messages. But in #6919 we broke backwards compatibility of able to handle v1 proto messages. This change fixes that. 

cc: @Clement-Jean 

RELEASE NOTES: none

### DIFF
--- a/encoding/proto/proto.go
+++ b/encoding/proto/proto.go
@@ -39,13 +39,8 @@ func init() {
 type codec struct{}
 
 func (codec) Marshal(v any) ([]byte, error) {
-	var vv proto.Message
-	switch v := v.(type) {
-	case protoadapt.MessageV1:
-		vv = protoadapt.MessageV2Of(v)
-	case protoadapt.MessageV2:
-		vv = v
-	default:
+	vv := messageV2Of(v)
+	if vv == nil {
 		return nil, fmt.Errorf("failed to marshal, message is %T, want proto.Message", v)
 	}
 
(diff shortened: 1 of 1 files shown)


encoding/proto/proto.go
```diff
@@ -38,18 +39,30 @@ func init() {
...
 func (codec) Unmarshal(data []byte, v any) error {
-	vv, ok := v.(proto.Message)
-	if !ok {
+	var vv proto.Message
+	switch v := v.(type) {
+	case protoadapt.MessageV1:
+		vv = protoadapt.MessageV2Of(v)
+	case protoadapt.MessageV2:
+		vv = v
+	default:
```
dfawley, line 62: Optional: make this a helper function and share with above.

dfawley (approved):

encoding/proto/proto.go
```diff
@@ -39,31 +39,32 @@ func init() {
...
 	}
 
 	return proto.Marshal(vv)
 }
 
 func (codec) Unmarshal(data []byte, v any) error {
-	var vv proto.Message
+	vv := messageV2Of(v)
+	if vv == nil {
+		return fmt.Errorf("failed to marshal, message is %T, want proto.Message", v)
```
dfawley, line 53: *unmarshal

arvindbr8: err... https://github.com/grpc/grpc-go/pull/6966 pls

PR #6968, opened 2024-02-07
security/advancedtls: CRL checks improvement

erm-g:
Part of CRL Provider effort. The functionality in this PR includes enhancements for verifyCRL function to get it aligned with RFC 5280

RELEASE NOTES: N/A


security/advancedtls/advancedtls_test.go
```diff
@@ -381,14 +381,14 @@ func (s) TestClientServerHandshake(t *testing.T) {
 		return &GetRootCAsResults{TrustCerts: cs.ServerTrust3}, nil
 	}
 
-	makeStaticCRLProvider := func(crlPath string) *RevocationConfig {
+	makeStaticCRLProvider := func(crlPath string, allowUndetermined bool) *RevocationConfig {
```
gtcooke94, line 384: Can we rename this to something like `MakeRevocationConfigWithStaticCrlProvider`?

security/advancedtls/advancedtls_test.go
```diff
@@ -750,7 +750,24 @@ func (s) TestClientServerHandshake(t *testing.T) {
...
+			serverVType:            CertVerification,
+			serverExpectError:      true,
+		},
+		// Client: set valid credentials with the revocation config
+		// Server: set valid credentials with the revocation config
+		// Expected Behavior: fail, because CRL is issued by the malicious CA. It
+		// can't be properly processed, and we don't allow RevocationUndetermined.
+		{
+			desc:                   "Client sets peer cert, reload root function with verifyFuncGood; Server sets peer cert, reload root function; Client uses CRL; mutualTLS",
+			clientCert:             []tls.Certificate{cs.ClientCert3},
```
gtcooke94, line 766: Please rename `ClientCert3` to something more meaningful, it's hard to review that this is the right cert for the test

security/advancedtls/crl.go
```diff
@@ -303,24 +303,30 @@ func fetchCRL(c *x509.Certificate, crlVerifyCrt []*x509.Certificate, cfg Revocat
 		if crl == nil {
 			return nil, fmt.Errorf("no CRL found for certificate's issuer")
 		}
+		if err := verifyCRL(crl, nil, crlVerifyCrt); err != nil {
```
gtcooke94, line 306: is the `nil` `rawIssuer` arg actually used at all in this fn?

security/advancedtls/crl.go
```diff
@@ -303,24 +303,30 @@ func fetchCRL(c *x509.Certificate, crlVerifyCrt []*x509.Certificate, cfg Revocat
...
 	crl, err := fetchCRL(c, crlVerifyCrt, cfg)
 	if err != nil {
-		// We couldn't load any CRL files for the certificate, so we don't know
-		// if it's RevocationUnrevoked or not.  This is not necessarily a
+		// We couldn't load any valid CRL files for the certificate, so we don't
+		// know if it's RevocationUnrevoked or not. This is not necessarily a
 		// problem - it's not invalid to have no CRLs if you don't have any
-		// revocations for an issuer. We just return RevocationUndetermined and
-		// there is a setting for the user to control the handling of that.
+		// revocations for an issuer. It also might be an indication of CRL file to
```
gtcooke94, line 326: "It also might be an indication of CRL file to be invalid" -> "It also might be an indication that the CRL file in invalid."

security/advancedtls/crl.go
```diff
@@ -547,11 +553,15 @@ func verifyCRL(crl *CRL, rawIssuer []byte, chain []*x509.Certificate) error {
 		// include this extension in all CRLs issued."
 		// So, this is much simpler than RFC4158 and should be compatible.
 		if bytes.Equal(c.SubjectKeyId, crl.authorityKeyID) && bytes.Equal(c.RawSubject, crl.rawIssuer) {
+			// RFC5280, 6.3.3 (f) Key usage and cRLSign bit.
+			if c.KeyUsage != 0 && c.KeyUsage&x509.KeyUsageCRLSign == 0 {
+				return fmt.Errorf("verifyCRL: The trust anchor can't be used for issuing CRLs")
```
gtcooke94, line 558: replace trust anchor with "certificate" or "ca"

security/advancedtls/crl_test.go
```diff
@@ -471,22 +474,43 @@ func TestVerifyCrl(t *testing.T) {
...
+		{
+			desc:    "Fail CRL by malicious CA",
+			crl:     loadCRL(t, testdata.Path("crl/provider_malicious_crl_empty.pem")),
+			certs:   makeChain(t, testdata.Path("crl/provider_client_trust_cert.pem")),
+			cert:    makeChain(t, testdata.Path("crl/provider_client_trust_cert.pem"))[0],
+			errWant: "verification error",
+		},
+		{
+			desc:    "Fail KeyUsage without cRLSign bit",
+			crl:     loadCRL(t, testdata.Path("crl/provider_malicious_crl_empty.pem")),
```
gtcooke94, line 509: Is the "malicious" CA just a CA that doesn't have the cRLSign bit set? Otherwise are we sure that we are failing this because of the non-set bit and not because of some signature mismatch or something?

security/advancedtls/testdata/crl/README.md
```diff
@@ -116,4 +119,14 @@ cert are below.
 $ openssl ca -gencrl -keyfile provider_client_trust_key.pem -cert provider_client_trust_cert.pem -out provider_crl_empty.pem -config provider_crl.cnf
 $ openssl ca -revoke provider_server_cert.pem -keyfile provider_client_trust_key.pem -cert provider_client_trust_cert.pem -config provider_crl.cnf
 $ openssl ca -gencrl -keyfile provider_client_trust_key.pem -cert provider_client_trust_cert.pem -out provider_crl_server_revoked.pem -config provider_crl.cnf
-```
\ No newline at end of file
+```
+
+The commands to generate CRL file by 'malicious' CA are below.
+```
+$ openssl genrsa -out provider_malicious_client_trust_key.pem 4096
```
gtcooke94, line 126: I think it's best to have this entirely in a script and just link the script, rather than having terminal commands in the README

gtcooke94: It looks like you do have it in a script already, so by having it here we're just creating a future documentation problem for ourselves where we might not update both things at the same time, let's have one authoritative source of truth for the creation, and that should be a script. WDYT?

security/advancedtls/testdata/crl/provider_create.sh
```diff
@@ -72,5 +72,35 @@ openssl ca -gencrl                       \
   -out provider_crl_server_revoked.pem   \
   -config provider_crl.cnf
 
+openssl genrsa                                      \
+  -out provider_malicious_client_trust_key.pem 4096
+
+SKI=$(openssl x509 -in provider_client_trust_cert.pem \
```
gtcooke94, line 78: SKI -> SubjectKeyIdentifier

security/advancedtls/testdata/crl/provider_extensions.conf
```diff
@@ -1,4 +1,5 @@
-subjectKeyIdentifier = hash
+[extensions]
+subjectKeyIdentifier = 54:1B:29:00:6E:3C:B1:54:68:75:C2:F8:26:90:BB:46:D5:70:45:E9
```
gtcooke94, line 2: What does setting this do exactly? It just immediately strikes me as a little odd to hard-code a value like this in what is meant as a configuration file?

security/advancedtls/advancedtls_test.go
erm-g: Since it's private for the test function, I changed it to a shorter `makeStaticCRLRevocationConfig`

security/advancedtls/advancedtls_test.go
erm-g: Done for client and server certs. However the only way to check if the cert is right is still checking what we write there in `LoadCerts`

security/advancedtls/crl.go
erm-g: Not for this functionality, let's discuss one previous usage offline

security/advancedtls/crl.go
erm-g: Done

security/advancedtls/crl.go
erm-g: My bad, forgot about this discussion. Done

security/advancedtls/crl_test.go
erm-g: For this particular test (we validate a crl by the correct cert that was used to issue it), it's just a CA without the cRLSign bit

security/advancedtls/testdata/crl/README.md
erm-g: I think atm it's intermediate state. We want to have a script just to 'do all the setup' but it's tough to understand the intent. That's why I created a README with some explanations there.
I think the end goal would be to drop the README in favor of actual documentation once it's ready, so I'd prefer to keep the README for now

security/advancedtls/testdata/crl/provider_create.sh
erm-g: Since it's internal var for bash script, I'd prefer to keep it short.

security/advancedtls/testdata/crl/provider_extensions.conf
erm-g: That's a bug, I forgot to include a 'revert' part of the script. The updated flow is
Create a cert
Grab it's subjectKeyIdentifier
Update the value of subjectKeyIdentifier in .cnf
Generate malicious certs using that .cnf
Revert .cnf back to subjectKeyIdentifier = hash

security/advancedtls/crl.go
```diff
@@ -323,8 +324,8 @@ func checkCert(c *x509.Certificate, crlVerifyCrt []*x509.Certificate, cfg Revoca
 		// We couldn't load any valid CRL files for the certificate, so we don't
 		// know if it's RevocationUnrevoked or not. This is not necessarily a
 		// problem - it's not invalid to have no CRLs if you don't have any
-		// revocations for an issuer. It also might be an indication of CRL file to
-		// be invalid.
+		// revocations for an issuer. It also might be an indication that the CRL
+		// file to is invalid.
```
gtcooke94, line 328: Remove "to"

security/advancedtls/testdata/crl/README.md
gtcooke94: But right now it's an exact duplicate of the script, right? That's a maintenance headache - there's no reason to copy that to the README. I think we should only have bash code in the README that aren't covered in the scripts

security/advancedtls/testdata/crl/provider_create.sh
gtcooke94: I'd prefer to go with the full name - longer names are free, and this way I don't have to take time to search what `SKI` could stand for if I forget

gtcooke94 (approved):
LGTM

security/advancedtls/crl.go
erm-g: removed

security/advancedtls/testdata/crl/README.md
erm-g: refactored

security/advancedtls/testdata/crl/provider_create.sh
erm-g: Renamed

security/advancedtls/crl.go
erm-g: Done

dfawley (approved):

PR #6989, opened 2024-02-16
transport: Make error-handling for bad HTTP method consistent between HTTP/2 server transport and handler server transport

jhump:
This tweaks `transport.NewServerHandlerTransport` so that it behaves more like the normal HTTP/2 transport returned by `transport.NewServerTransport` with regard to requests with the wrong HTTP method. In particular, it now sends back a _405 Method Not Allowed_ status code and correctly sets an `Allow` response header.

I've also moved the HTTP version check to after we've confirmed that this is a gRPC request (via content-type check) and changed that to also use a more appropriate HTTP status code: _505 HTTP Version Not Supported_. (This is not really related to consistency since the other transport never has to worry about HTTP 1.1 requests. But it makes the checks a little more coherent since IMO it seems more appropriate to reject the HTTP version only once we've confirmed that it's a gRPC request.)

RELEASE NOTES:
* server: Fix some error cases when using a `grpc.Server` as an `http.Handler` with the Go stdlib HTTP server.

### DIFF
--- a/internal/transport/handler_server.go
+++ b/internal/transport/handler_server.go
@@ -51,14 +51,10 @@ import (
 // inside an http.Handler, or writes an HTTP error to w and returns an error.
 // It requires that the http Server supports HTTP/2.
 func NewServerHandlerTransport(w http.ResponseWriter, r *http.Request, stats []stats.Handler) (ServerTransport, error) {
-	if r.ProtoMajor != 2 {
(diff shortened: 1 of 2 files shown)


dfawley (approved):

zasweq (approved):
LGTM

PR #6969, opened 2024-02-07
channelz: major cleanup / reorganization

dfawley:
Several major changes are included:

- Stop using an "ID" to reference the channelz object from the channel/etc that contains it.  Instead get a pointer to the object directly.
- Stop storing the channelz metrics in the channel/etc and instead store them in the channelz struct representation of that thing.
- Reorganize files so they are categorized by type instead of "func" vs. "types"
- Lots of small things

To be done:
- Represent tree of channelz objects using a tree rooted on the top-level types (channel, server) instead of a bunch of flat buffers storing each type with references between them.  We still need the flat buffers to look things up through the channelz service, but this might remove the need for manual reference counting, e.g.
- Cleanup the trace logs to be simpler.  There's a pretty complex interaction between the trace logs and the things they reference right now.

RELEASE NOTES: none

PR #7004, opened 2024-02-27
Use 'google.golang.org/protobuf' instead of 'github.com/golang/protobuf/' as much as possible

mmorel-35:
RELEASE NOTES: none

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/channelz/service/func_linux.go
+++ b/channelz/service/func_linux.go
@@ -21,15 +21,14 @@ package service
 import (
 	"time"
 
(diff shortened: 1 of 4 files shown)


dfawley:
The changes to channelz will conflict with #6969, so I'd rather do them later (as that is a huge PR), and the other changes you've made are already included in #6961.

Thank you for the PR, though!

Issue #1786, opened 2018-01-08
Consider deprecating Dial and DialContext and replacing with a NewClient


dfawley:
`Dial` and `DialContext` have several differences from how gRPC works in other languages.  A `NewClient` function that works more like other languages would differ in at least two ways:

1. Default to DNS resolver instead of passthrough.
1. Make target compliant with RFC 3986 (#1911).
1. Start IDLE instead of CONNECTING.

If we create a `NewClient`, we'd probably also want to make matching `ClientOptions` instead of `DialOptions`, as some `DialOptions` (e.g. `WithBlock` or `WithTimeout`) do not make sense when not dialing synchronously.

nhooyr:
> Default to DNS resolver instead of passthrough.

Could you elaborate on what passthrough would be? Wouldn't all Dial's use a DNS resolver?

dfawley:
Passthrough sends the entire target string (after "passthrough:///") to the dialer as the address, verbatim.

A DNS resolver would resolve the target (after "dns:///"), turn it into IP addresses & ports, and _those_ would be handed to the dialer.

It just so happens that the default dialer (`net.Dial`) is able to do DNS name resolution, and so passing "google.com:22" will work just fine with it.

However, this passthrough method means we can't do things like load-balance across all the DNS results for that name.

You can see more about the resolver/balancer design in the gRFC here: https://github.com/grpc/proposal/blob/master/L9-go-resolver-balancer-API.md

raphoester:
Hello guys, I notice that the `DialContext` method is now deprecated 
I was using it a lot for my tests that included the presentation layer, eg :
```go 
conn, err := grpc.DialContext(
		context.Background(), "bufnet",
		grpc.WithContextDialer(
			func(context.Context, string) (net.Conn, error) {
				return s.lis.Dial()
			},
		),
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
	// then create the client using the new conn
	client := someservice.NewSomeServiceClient(conn)
```

I tried replacing `DialContext` with `NewClient`, e.g : 
```go 
conn, err := grpc.NewClient(
		"passthrough", // ?
		grpc.WithContextDialer(
			func(context.Context, string) (net.Conn, error) {
				return s.lis.Dial()
			},
		),
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
```

But now i get a DNS error on RPC call from the client : `rpc error: code = Unavailable desc = name resolver error: produced zero addresses`

[...]

PR #7010, opened 2024-02-28
chore: expose `NewClient` method to end users

bruuuuuuuce:
This will allow users to create a connection, leaving it in idle state, rather than having it automatically try and connect in the background, which is the current behavior of `Dial`. Leaving the connection in idle state is more consistent with how other grpc implementations deal with connections

Looks like this change was already planned, but just has not been commited yet https://github.com/grpc/grpc-go/blob/99ded5c4905feb5d6ef305114c6f8a7a942011c7/clientconn.go#L222-L225

RELEASE NOTES:

- grpc: allow users to create new clients with connections remaining in idle state


dfawley (commented):
One more thing needs to be figured out before we can do this: we want the default name resolver used when calling `NewClient` directly by users to be `dns` instead of the awkward and non-standard `passthru` used by `Dial`.

However, the default is set globally, and can be overridden, so we need to come up with a way to have a different default for the two different paths, override them both when the user calls `resolver.SetDefaultScheme`, and have the proper code path use the proper one.  (This is mentioned in #1786.)

Let me know if you're interested in taking that work on as well.

bruuuuuuuce:
> Let me know if you're interested in taking that work on as well.

Sure I can take a stab at this

balancer/rls/config.go
```diff
@@ -195,13 +195,13 @@ func parseRLSProto(rlsProto *rlspb.RouteLookupConfig) (*lbConfig, error) {
 	parsedTarget, err := url.Parse(lookupService)
 	if err != nil {
 		// url.Parse() fails if scheme is missing. Retry with default scheme.
-		parsedTarget, err = url.Parse(resolver.GetDefaultScheme() + ":///" + lookupService)
+		parsedTarget, err = url.Parse(resolver.GetDefaultSchemeOrPassthrough() + ":///" + lookupService)
```
bruuuuuuuce, line 198: I don't really love these names, am open to changing them. Maybe something like `GetOverridenSchemeOrPassthrough`?

dialoptions.go
```diff
@@ -79,6 +79,7 @@ type dialOptions struct {
 	resolvers                   []resolver.Builder
 	idleTimeout                 time.Duration
 	recvBufferPool              SharedBufferPool
+	eagerConnect                bool
```
bruuuuuuuce, line 82: Not totally sold on this name either

resolver/resolver.go
```diff
@@ -36,7 +36,9 @@ var (
 	// m is a map from scheme to resolver builder.
 	m = make(map[string]Builder)
 	// defaultScheme is the default scheme to use.
-	defaultScheme = "passthrough"
+	defaultScheme     = ""
```
bruuuuuuuce, line 39: Possibly change it to `overriddenScheme`?

bruuuuuuuce:
@dfawley lmk if this was what you were thinking, also any naming suggestions would be appreciated 😅

balancer/rls/config.go
dfawley: Hmm, can we avoid changing the existing things entirely, please, and keep the behavior the same?

I think that could work if we have a way to determine whether `SetDefaultScheme` was ever called by the user (ideally only exposed via `internal`), and a way to distinguish whether the user was directly calling `NewClient` or `Dial`.

clientconn.go
```diff
@@ -208,7 +208,11 @@ func newClient(target string, opts ...DialOption) (conn *ClientConn, err error)
 // https://github.com/grpc/grpc/blob/master/doc/naming.md.
 // e.g. to use dns resolver, a "dns:///" prefix should be applied to the target.
 func DialContext(ctx context.Context, target string, opts ...DialOption) (conn *ClientConn, err error) {
-	cc, err := newClient(target, opts...)
+	// At the end of this method, we kick the channel out of idle, rather than waiting for the first rpc.
+	// The eagerConnect option is used to tell the ClientChannel that DialContext was called to make
+	// this channel, and hence tries to connect immediately.
+	opts = append(opts, WithEagerConnect())
+	cc, err := NewClient(target, opts...)
```
dfawley, line 215: Something like this could work, though I'd want to not export `WithEagerConnect`.

Another option would be:

```go
func newClient(target, defaultScheme string, opts ...DialOption) {}

func NewClient(..) (..) {
	return newClient(target, "dns", opts...)
}

func DialContext(..) (..) {
	cc, err := newClient(target, "passthrough", opts...)
	...
}
```

We could also combine the two ideas and make a `DialOption` that specifies the default scheme, and if it isn't specified, then we assume "dns" (and `DialContext` sets it with `"passthrough"`).

clientconn.go
```diff
@@ -1740,8 +1744,13 @@ func (cc *ClientConn) parseTargetAndFindResolver() error {
...
 	// specified an unregistered scheme. We should fallback to the default
 	// scheme, except when a custom dialer is specified in which case, we should
-	// always use passthrough scheme.
-	defScheme := resolver.GetDefaultScheme()
+	// always use passthrough scheme. For either case, we need to respect any overridden
+	// global defaults set by the user.
+	defScheme := resolver.GetDefaultSchemeOrDNS()
+	if cc.dopts.eagerConnect {
+		defScheme = resolver.GetDefaultSchemeOrPassthrough()
+	}
```
dfawley, line 1752: Based on the above, this could be:

```go
defScheme := cc.dopts.defScheme
if defScheme == "" || internal.UserSetDefaultScheme {
	defScheme = resolver.GetDefaultScheme()
}
```

dfawley (commented):
Thanks for continuing with this!  I think it's probably pretty close, though I would like to massage it a bit to avoid making any user visible changes besides the `NewClient` function being added.

clientconn.go
bruuuuuuuce: Yep I like this, looks clean

clientconn_parsed_target_test.go
```diff
@@ -34,7 +34,7 @@ import (
 )
 
 func (s) TestParsedTarget_Success_WithoutCustomDialer(t *testing.T) {
-	defScheme := resolver.GetDefaultScheme()
+	defScheme := "passthrough"
```
bruuuuuuuce, line 37: these tests use the `Dial` method, which we overrode to use the passthrough scheme, since the default scheme is now `dns`. I can instead have these tests use the `NewClient` method? Or I can also add additional tests for `Dial` vs `NewClient` methods

clientconn.go
```diff
@@ -1740,8 +1748,13 @@ func (cc *ClientConn) parseTargetAndFindResolver() error {
 	// We are here because the user's dial target did not contain a scheme or
 	// specified an unregistered scheme. We should fallback to the default
 	// scheme, except when a custom dialer is specified in which case, we should
-	// always use passthrough scheme.
-	defScheme := resolver.GetDefaultScheme()
+	// always use passthrough scheme. For either case, we need to respect any overridden
+	// global defaults set by the user.
+	defScheme := cc.dopts.defScheme
+	if defScheme == "" || internal.UserSetDefaultScheme {
```
dfawley, line 1754: `defscheme` will never be `""` with these changes, since you always set it.

I think it's OK to just delete the `defScheme == ""` part.  (We should use our default default unless the user set a different default.)

clientconn_parsed_target_test.go
dfawley: Yes, I like the sound of making this test check both `Dial` and `NewClient`.

It would also be nice to have some tests for `NewClient` generally (at least creating a client and performing RPCs).

resolver/resolver.go
```diff
@@ -36,7 +36,7 @@ var (
 	// m is a map from scheme to resolver builder.
 	m = make(map[string]Builder)
 	// defaultScheme is the default scheme to use.
-	defaultScheme = "passthrough"
+	defaultScheme = "dns"
```
dfawley, line 39: Let's just leave this as `"passthrough"` and then the tests don't have to change.  AFAICT this value isn't ever used by gRPC anymore anyway, since you always force the default default to either `"dns"` or `"passthrough"` depending on the code path in `clientconn.go`.

resolver/resolver.go
```diff
@@ -72,7 +72,8 @@ func SetDefaultScheme(scheme string) {
 	defaultScheme = scheme
 }
 
-// GetDefaultScheme gets the default scheme that will be used.
+// GetDefaultScheme gets the default scheme that will be used, if no scheme was
+// overridden, defaults to the "dns" scheme.
```
dfawley, line 76: Maybe this can say `...the default scheme that will be used by grpc.Dial.  If SetDefaultScheme is never called, the default scheme used by grpc.NewClient is "dns" instead.`

resolver/resolver.go
```diff
@@ -72,7 +72,8 @@ func SetDefaultScheme(scheme string) {
 	defaultScheme = scheme
```
dfawley, line 72: This also needs to set `internal.UserSetDefaultScheme=true`.

clientconn.go
```diff
@@ -705,6 +712,7 @@ func init() {
 	internal.ExitIdleModeForTesting = func(cc *ClientConn) error {
 		return cc.idlenessMgr.ExitIdleMode()
 	}
+	internal.UserSetDefaultScheme = resolver.GetDefaultScheme() != "dns"
```
dfawley, line 715: Let's do this via `resolver.SetDefaultScheme` instead of trying to infer it.

resolver/resolver.go
bruuuuuuuce: I had initially tried that, but was getting an `import cycle ` error when building. I saw this note on top of the `SetDefaultScheme` method

```go
// NOTE: this function must only be called during initialization time (i.e. in
// an init() function)
func SetDefaultScheme(scheme string) {
```

which was why I put it in the `init` function inside of `clientconn.go`

bruuuuuuuce: This is the cycle its detecting

```bash
go build google.golang.org/grpc/...
package google.golang.org/grpc
	imports google.golang.org/grpc/balancer
	imports google.golang.org/grpc/channelz
	imports google.golang.org/grpc/internal/channelz
	imports google.golang.org/grpc/internal
	imports google.golang.org/grpc/resolver
	imports google.golang.org/grpc/internal: import cycle not allowed
```

dfawley: `internal` should not be importing `resolver`.  Actually it shouldn't import anything else in the grpc tree, as part of its intent, but we don't need to fix all the offenses right away.

Can you move `Get/SetXDSHandshakeClusterName` into another package?  Just `internal/xds/xds.go` seems fine to me.

clientconn.go
```diff
@@ -191,6 +190,11 @@ func newClient(target string, opts ...DialOption) (conn *ClientConn, err error)
 	return cc, nil
 }
 
+// NewClient returns a new client in idle mode.
```
dfawley, line 193: Just FYI / for the next reviewer's reference:

After this PR, I plan to rewrite things a little bit.  I'd like this to become the primary API for users, and to call `Dial` and `DialContext` both "deprecated" in preference of this.  As such, this will contain a bit more documentation, and the others will state they call it and then initiate and wait for a connection.

clientconn.go
```diff
@@ -208,7 +212,10 @@ func newClient(target string, opts ...DialOption) (conn *ClientConn, err error)
 // https://github.com/grpc/grpc/blob/master/doc/naming.md.
 // e.g. to use dns resolver, a "dns:///" prefix should be applied to the target.
 func DialContext(ctx context.Context, target string, opts ...DialOption) (conn *ClientConn, err error) {
-	cc, err := newClient(target, opts...)
+	// At the end of this method, we kick the channel out of idle, rather than waiting for the first rpc.
+	// The eagerConnect option is used to tell the ClientChannel that DialContext was called to make
+	// this channel, and hence tries to connect immediately.
```
dfawley, line 217: Please delete the bit about the "eagerConnect" option.

clientconn_parsed_target_test.go
```diff
@@ -33,29 +33,75 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
+func generateTarget(scheme string, target string) resolver.Target {
+	return resolver.Target{URL: *testutils.MustParseURL(fmt.Sprintf("%s:///%s", scheme, target))}
+}
+
 func (s) TestParsedTarget_Success_WithoutCustomDialer(t *testing.T) {
-	defScheme := resolver.GetDefaultScheme()
+	dialScheme := resolver.GetDefaultScheme()
```
dfawley, line 41: I believe this should probably try to re-create the initial state in case another test calls `SetDefaultScheme`.   I.e. do `resolver.SetDefaultScheme("passthrough")` and `internal.UserSetDefaultScheme = false`?

resolver/resolver.go
```diff
@@ -63,16 +64,18 @@ func Get(scheme string) Builder {
 }
 
 // SetDefaultScheme sets the default scheme that will be used. The default
-// default scheme is "passthrough".
+// scheme is "passthrough".
```
dfawley, line 67: "default default" was not actually a typo.  But maybe would be clearer if it said "the default scheme is initially set to ..." or something like that.

dfawley (approved):
Thank you for the contribution!  I have a few minor things, but I think we can also ask for a 2nd pass review now (2 approvals are required by policy).

arvindbr8 (approved):
Thanks for the PR. Looks good to me 👍

clientconn.go
zbindenren: @dfawley ok now with the newest release both are deprecated. But how do I add a timeout?

When I try the follwing:

```go
	conn, err := grpc.NewClient(addr,
		grpc.WithTransportCredentials(creds),
		grpc.WithBlock(),
		grpc.WithChainUnaryInterceptor(interceptors...),
		grpc.WithTimeout(dialTimeout),
	)
```

the linter tells me that `grpc.WithTimeout` is deprecated I should use `grpc.DialContext` (which is deprecated too). So how do I use `grpc.NewClient` with `grpc.WithBlock` that doesn't block forever?

dfawley: Setting a timeout when creating a client is an anti-pattern, and is not possible in any other gRPC implementations.

See https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#dialing-in-grpc for more info.

stijnderyckere: I agree that connection failures should always be checked, not only on start-up.

But if you're working with setups such as Kubernetes, you don't want your new pod to pass liveness/readiness checks in case it has an invalid connection string. It's way safer to have this "ping" check to confirm there are no misconfigurations in your new deployment, before you terminate your previous pods.

dfawley: There are other, better, more robust ways of doing that.  For example, health checking: https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md

You can also use the connectivity state API if you just want to see if the client was able to connect: https://pkg.go.dev/google.golang.org/grpc#ClientConn.GetState and https://pkg.go.dev/google.golang.org/grpc#ClientConn.WaitForStateChange

PR #7015, opened 2024-03-04
Send GOAWAY to server on Client Transport Shutdown

### LINKED ISSUE #460: Send GOAWAY on client shutdown
The HTTP/2 spec encourages sending GOAWAY before dropping the connection. GOAWAY is necessary for "graceful" shutdowns.

When error handling:

> Endpoints SHOULD send a GOAWAY frame when ending a connection, providing that circumstances permit it.

In the GOAWAY frame description section:

> Endpoints SHOULD always send a GOAWAY frame before closing a connection so that the remote peer can know whether a stream has been partially processed or not.

.

> A GOAWAY frame might not immediately precede closing of the connection; a receiver of a GOAWAY that has no more use for the connection SHOULD still send a GOAWAY frame before terminating the connection.

When discussing connection management:

> When either endpoint chooses to close the transport-layer TCP connection, the terminating endpoint SHOULD first send a GOAWAY (Section 6.8) frame so that both endpoints can reliably determine whether previously sent frames have been processed and gracefully complete or terminate any necessary remaining tasks.

aranjans:
Fixes #460 

RELEASE NOTES:

* client: client sends GOAWAY to server on graceful shutdown

### DIFF
--- a/internal/transport/controlbuf.go
+++ b/internal/transport/controlbuf.go
@@ -193,7 +193,7 @@ type goAway struct {
 	code         http2.ErrCode
 	debugData    []byte
 	headsUp      bool
-	closeConnErr error // if set, loopyWriter will exit, resulting in conn closure
+	closeConnErr error // if set, loopyWriter will exit with this error
 }
 
 func (*goAway) isTransportResponseFrame() bool { return false }
--- a/internal/transport/controlbuf.go
+++ b/internal/transport/controlbuf.go
@@ -340,11 +340,9 @@ func (c *controlBuffer) executeAndPut(f func(it any) bool, it cbItem) (bool, err
 	var wakeUp bool
 	c.mu.Lock()
 	if c.err != nil {
-		fmt.Printf("ControlBUf already have error %v: ", c.err)
 		c.mu.Unlock()
 		return false, c.err
 	}
-	fmt.Println("Success putting the goaway")
 	if f != nil {
 		if !f(it) { // f wasn't successful
 			c.mu.Unlock()
@@ -544,7 +542,6 @@ func (l *loopyWriter) run() (err error) {
 		if l.logger.V(logLevel) {
 			l.logger.Infof("loopyWriter exiting with error: %v", err)
 		}
-		l.logger.Infof("loopyWriter exiting with error: %v", err)
 		if !isIOError(err) {
 			l.framer.writer.Flush()
 		}
--- a/internal/transport/controlbuf.go
+++ b/internal/transport/controlbuf.go
@@ -190,10 +190,10 @@ type incomingGoAway struct {
 func (*incomingGoAway) isTransportResponseFrame() bool { return false }
 
 type goAway struct {
-	code      http2.ErrCode
-	debugData []byte
-	headsUp   bool
-	closeConn error // if set, loopyWriter will exit, resulting in conn closure
+	code         http2.ErrCode
+	debugData    []byte
+	headsUp      bool
+	closeConnErr error // if set, loopyWriter will exit, resulting in conn closure
 }
 
 func (*goAway) isTransportResponseFrame() bool { return false }
@@ -340,9 +340,11 @@ func (c *controlBuffer) executeAndPut(f func(it any) bool, it cbItem) (bool, err
 	var wakeUp bool
 	c.mu.Lock()
 	if c.err != nil {
+		fmt.Printf("ControlBUf already have error %v: ", c.err)
 		c.mu.Unlock()
 		return false, c.err
 	}
+	fmt.Println("Success putting the goaway")
 	if f != nil {
 		if !f(it) { // f wasn't successful
 			c.mu.Unlock()
@@ -542,6 +544,7 @@ func (l *loopyWriter) run() (err error) {
 		if l.logger.V(logLevel) {
 			l.logger.Infof("loopyWriter exiting with error: %v", err)
 		}
+		l.logger.Infof("loopyWriter exiting with error: %v", err)
 		if !isIOError(err) {
 			l.framer.writer.Flush()
 		}
--- a/internal/transport/controlbuf.go
+++ b/internal/transport/controlbuf.go
@@ -190,10 +190,10 @@ type incomingGoAway struct {
 func (*incomingGoAway) isTransportResponseFrame() bool { return false }
 
 type goAway struct {
-	code         http2.ErrCode
-	debugData    []byte
-	headsUp      bool
-	closeConnErr error // if set, loopyWriter will exit with this error
+	code      http2.ErrCode
+	debugData []byte
+	headsUp   bool
+	closeConn error // if set, loopyWriter will exit with this error
 }
 
 func (*goAway) isTransportResponseFrame() bool { return false }
@@ -495,21 +495,22 @@ type loopyWriter struct {
 	ssGoAwayHandler func(*goAway) (bool, error)
 }
 
-func newLoopyWriter(s side, fr *framer, cbuf *controlBuffer, bdpEst *bdpEstimator, conn net.Conn, logger *grpclog.PrefixLogger) *loopyWriter {
+func newLoopyWriter(s side, fr *framer, cbuf *controlBuffer, bdpEst *bdpEstimator, conn net.Conn, logger *grpclog.PrefixLogger, goAwayHandler func(*goAway) (bool, error)) *loopyWriter {
 	var buf bytes.Buffer
 	l := &loopyWriter{
-		side:          s,
-		cbuf:          cbuf,
-		sendQuota:     defaultWindowSize,
-		oiws:          defaultWindowSize,
-		estdStreams:   make(map[uint32]*outStream),
-		activeStreams: newOutStreamList(),
-		framer:        fr,
-		hBuf:          &buf,
-		hEnc:          hpack.NewEncoder(&buf),
-		bdpEst:        bdpEst,
-		conn:          conn,
-		logger:        logger,
+		side:            s,
+		cbuf:            cbuf,
+		sendQuota:       defaultWindowSize,
+		oiws:            defaultWindowSize,
+		estdStreams:     make(map[uint32]*outStream),
+		activeStreams:   newOutStreamList(),
+		framer:          fr,
+		hBuf:            &buf,
+		hEnc:            hpack.NewEncoder(&buf),
+		bdpEst:          bdpEst,
+		conn:            conn,
+		logger:          logger,
+		ssGoAwayHandler: goAwayHandler,
 	}
 	return l
 }
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -20,6 +20,7 @@ package transport
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -116,6 +117,10 @@ type http2Client struct {
 	waitingStreams        uint32
 	nextID                uint32
 	registeredCompressors string
+
+	// goAwaySent is initialised with http2Client and fired when client
+	// transport is shutdown
+	goAwaySent *grpcsync.Event
 
 	// Do not access controlBuf with mu held.
 	mu            sync.Mutex // guard the following variables
@@ -326,6 +331,7 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 		registeredCompressors: grpcutil.RegisteredCompressors(),
 		address:               addr,
 		conn:                  conn,
+		goAwaySent:            grpcsync.NewEvent(),
 		remoteAddr:            conn.RemoteAddr(),
 		localAddr:             conn.LocalAddr(),
 		authInfo:              authInfo,
@@ -411,6 +417,7 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 		if err == nil {
 			err = <-readerErrCh
 		}
+		t.logger.Infof("Defer: Got error as %v", err)
 		if err != nil {
 			t.Close(err)
 		}
@@ -460,6 +467,7 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 	}
 	go func() {
 		t.loopy = newLoopyWriter(clientSide, t.framer, t.controlBuf, t.bdpEst, t.conn, t.logger)
+		t.loopy.ssGoAwayHandler = t.outgoingGoAwayHandler
 		if err := t.loopy.run(); !isIOError(err) {
 			// Immediately close the connection, as the loopy writer returns
 			// when there are no more active streams and we were draining (the
@@ -515,6 +523,17 @@ func (t *http2Client) getPeer() *peer.Peer {
 		AuthInfo:  t.authInfo, // Can be nil
 		LocalAddr: t.localAddr,
 	}
+}
+
+// Handles outgoing GoAway and returns true if loopy needs to put itself
+// in draining mode.
+func (t *http2Client) outgoingGoAwayHandler(g *goAway) (bool, error) {
+	// Send out a GOAWAY frame so server is aware of client transport shutdown
+	t.logger.Infof("Sending goaway")
+	if err := t.framer.fr.WriteGoAway(math.MaxUint32, http2.ErrCodeNo, g.debugData); err != nil {
+		return false, err
+	}
+	return false, g.closeConnErr
 }
 
 func (t *http2Client) createHeaderFields(ctx context.Context, callHdr *CallHdr) ([]hpack.HeaderField, error) {
@@ -991,7 +1010,9 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.finish()
+	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte("GOAWAY from client"), closeConnErr: errors.New("graceful_shutdown")})
+	<-t.writerDone
+	t.logger.Infof("Writer done")
 	t.cancel()
 	t.conn.Close()
 	channelz.RemoveEntry(t.channelz.ID)
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -118,10 +118,6 @@ type http2Client struct {
 	nextID                uint32
 	registeredCompressors string
 
-	// goAwaySent is initialised with http2Client and fired when client
-	// transport is shutdown
-	goAwaySent *grpcsync.Event
-
 	// Do not access controlBuf with mu held.
 	mu            sync.Mutex // guard the following variables
 	state         transportState
@@ -331,7 +327,6 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 		registeredCompressors: grpcutil.RegisteredCompressors(),
 		address:               addr,
 		conn:                  conn,
-		goAwaySent:            grpcsync.NewEvent(),
 		remoteAddr:            conn.RemoteAddr(),
 		localAddr:             conn.LocalAddr(),
 		authInfo:              authInfo,
@@ -417,7 +412,6 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 		if err == nil {
 			err = <-readerErrCh
 		}
-		t.logger.Infof("Defer: Got error as %v", err)
 		if err != nil {
 			t.Close(err)
 		}
@@ -529,7 +523,6 @@ func (t *http2Client) getPeer() *peer.Peer {
 // in draining mode.
 func (t *http2Client) outgoingGoAwayHandler(g *goAway) (bool, error) {
 	// Send out a GOAWAY frame so server is aware of client transport shutdown
-	t.logger.Infof("Sending goaway")
 	if err := t.framer.fr.WriteGoAway(math.MaxUint32, http2.ErrCodeNo, g.debugData); err != nil {
 		return false, err
 	}
@@ -1010,9 +1003,10 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte("GOAWAY from client"), closeConnErr: errors.New("graceful_shutdown")})
-	<-t.writerDone
-	t.logger.Infof("Writer done")
+	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte("graceful_shutdown"), closeConnErr: errors.New("graceful_shutdown")})
+	if t.loopy != nil {
+		<-t.writerDone
+	}
 	t.cancel()
 	t.conn.Close()
 	channelz.RemoveEntry(t.channelz.ID)
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -20,7 +20,6 @@ package transport
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -408,14 +407,6 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 	// returning from this function.
 	readerErrCh := make(chan error, 1)
 	go t.reader(readerErrCh)
-	defer func() {
-		if err == nil {
-			err = <-readerErrCh
-		}
-		if err != nil {
-			t.Close(err)
-		}
-	}()
 
 	// Send connection preface to server.
 	n, err := t.conn.Write(clientPreface)
@@ -471,6 +462,14 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 		}
 		close(t.writerDone)
 	}()
+	defer func() {
+		if err == nil {
+			err = <-readerErrCh
+		}
+		if err != nil {
+			t.Close(err)
+		}
+	}()
 	return t, nil
 }
 
@@ -519,10 +518,9 @@ func (t *http2Client) getPeer() *peer.Peer {
 	}
 }
 
-// Handles outgoing GoAway and returns true if loopy needs to put itself
-// in draining mode.
+// OutgoingGoAwayHandler writes GOAWAY to the connection.  Always returns (false, err) as we want the GoAway
+// to be the last frame loopy writes to the transport.
 func (t *http2Client) outgoingGoAwayHandler(g *goAway) (bool, error) {
-	// Send out a GOAWAY frame so server is aware of client transport shutdown
 	if err := t.framer.fr.WriteGoAway(math.MaxUint32, http2.ErrCodeNo, g.debugData); err != nil {
 		return false, err
 	}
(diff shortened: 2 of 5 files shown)


ginayeh:
@aranjans can you check test failures?

internal/transport/http2_client.go
```diff
@@ -983,6 +1027,18 @@ func (t *http2Client) Close(err error) {
 	}
 	t.mu.Unlock()
 	t.controlBuf.finish()
+	t.controlBuf.setError(nil)
+	err1 := t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte("GOAWAY from client"), headsUp: true})
```
arvindbr8, line 1031: After calling `controlBuf.finish()`, a `put` would just be a no-op. I think there are a couple of options here. 

One other way I can think about is, if we can make sure the loopywriter,run() has returned, it should be safe to access the framer directly and write the goAway frame directly. 

Or the other option is to somehow make sure we `put()` the GoAway and also assure that this is the last thing that is processed in the controlbuf.

arvindbr8: setError(nil) is only required since finish() sets the error which causes the put to be a no-op. I would prefer taking another approach.

internal/transport/transport_test.go
```diff
@@ -1188,6 +1188,88 @@ func (s) TestServerConnDecoupledFromApplicationRead(t *testing.T) {
 
 }
 
+// Test that a transport level client shutdown successfully sends a GOAWAY frame
+// to underlying connection.
+func (s) TestClientSendsAGoAwayFrame(t *testing.T) {
```
arvindbr8, line 1193: We might have discussed this offline about writing a test in the end2end test package using the clienttester to test this behavior. Might be worth to explore that.

arvindbr8 (changes requested):
I've taken a first pass at the PR. Let me know if you have any questions

arvindbr8:
Also the tests are failing due to a panic on a nil pointer reference. Seems like its the `goAwaySent` event that was added.

internal/transport/http2_client.go
aranjans: Sure, I have updated the PR

shashkin:
@aranjans vet-proto test should be fixed after #7028 so you need to pull updated master

test/goaway_test.go
```diff
@@ -761,3 +761,50 @@ func (s) TestTwoGoAwayPingFrames(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to dial: %v", err)
+	}
+
+	st := newServerTesterFromConn(t, conn)
+	st.greet()
+	goAwayFrameReceived := testutils.NewChannel()
+	go func() {
+		for {
+			f, err := st.readFrame()
```
arvindbr8, line 785: Wouldnt this race with the server created above?

arvindbr8 (changes requested):
The changes to transport looks right to me. However I think we are missing something in the test. Could you take another look?

arvindbr8 (changes requested):
The changes to transport looks right to me. However I think we are missing something in the test. Could you take another look?

test/goaway_test.go
```diff
@@ -761,3 +761,50 @@ func (s) TestTwoGoAwayPingFrames(t *testing.T) {
 		t.Fatalf("Error waiting for graceful shutdown of the server: %v", err)
 	}
 }
+
+func (s) TestClientSendsAGoAwayFrame(t *testing.T) {
```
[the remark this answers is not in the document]
arvindbr8: clienttester might be missing a few methods, which you might need for this.

test/goaway_test.go
```diff
@@ -761,3 +761,60 @@ func (s) TestTwoGoAwayPingFrames(t *testing.T) {
...
+				t.Logf("server tester received unexpected frame type %T", f)
+			}
+		}
+	}()
+
+	// Wait for the GOAWAY frame or timeout
+	select {
+	case <-goAwayFrameReceived:
+		t.Logf("GOAWAY frame received by server")
+	case <-time.After(5 * time.Second): // Adjust timeout if needed
```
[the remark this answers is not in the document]
arvindbr8: using timer like this might leak resources. You should either create a timer and defer a cancel on it. Or simply use context with timeout.

test/goaway_test.go
```diff
@@ -761,3 +761,60 @@ func (s) TestTwoGoAwayPingFrames(t *testing.T) {
...
+	conn, err := net.DialTimeout("tcp", lis.Addr().String(), defaultTestTimeout)
+	if err != nil {
+		t.Fatalf("Error dialing: %v", err)
+	}
+	defer conn.Close()
+
+	// Create a helper to read frames (similar to your serverTester)
+	st := newServerTesterFromConn(t, conn)
+
+	// Channel to signal if the GOAWAY frame was received
```
arvindbr8, line 794: Prefer using complete sentences. i.e., punctuate the comment like an actual sentence (like terminating with a period).

arvindbr8: See. https://google.github.io/styleguide/go/decisions#doc-comments

arvindbr8 (changes requested):
I think we need to iterate over the test once again. Left a few comments.

internal/transport/http2_client.go
```diff
@@ -454,13 +454,21 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 	}
 
 	t.connectionID = atomic.AddUint64(&clientConnectionCounter, 1)
+	t.loopy = newLoopyWriter(clientSide, t.framer, t.controlBuf, t.bdpEst, t.conn, t.logger)
+	t.loopy.lWriterDone = grpcsync.NewEvent()
```
arvindbr8, line 458: 2 comments, 

1. Why not init `lWriterDone` inside of `newLoopyWriter`
2. I think we can get away without the lWriterDone
  - Seems like this channel is required for blocking the client from calling conn.Close before we return from the loopywriter. We already have writeDone channel which is used for the same purpose.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,104 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+		}
+	}()
+
+	dialer := func(ctx context.Context, addr string) (net.Conn, error) {
+		return (&net.Dialer{}).DialContext(ctx, "tcp", addr)
+	}
+	copts := ConnectOptions{
+		Dialer:         dialer,
+		ChannelzParent: channelz.RegisterSubChannel(-1, "test subchannel"),
+	}
```
arvindbr8, line 2743: copts is not necessary for this test.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,104 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	ct, err := NewClientTransport(connectCtx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, copts, func(GoAwayReason) {})
+	if err != nil {
+		t.Fatalf("Error while creating client transport: %v", err)
+	}
+	callHdr := &CallHdr{
+		Host:   "localhost",
+		Method: "foo.Small",
+	}
```
arvindbr8, line 2751: callHdr is also not required to be set for this test.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,104 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+		t.Fatalf("wrong stream id: %d", s1.id)
+	}
+	ct.Close(errors.New("manually closed by client"))
+	t.Logf("Closed the client connection")
+	e, err := success.Receive(connectCtx)
+	if e != nil || err != nil {
+		t.Fatalf("Error in frame received: %v. Error receiving from channel: %v", e, err)
+	} else {
+		t.Logf("Server received the GOAWAY from client")
+	}
```
arvindbr8, line 2766: we should prolly have two channels? one for success and the other for error?

arvindbr8 (changes requested):
Nice work on getting it this far. I think we should be able to wrap it up in a couple of iterations. I've commented an alternative. We can also discuss in detail offline.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,91 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+			if goAwayFrame.ErrCode == http2.ErrCodeNo && string(goAwayFrame.DebugData()) == "graceful_shutdown" {
+				t.Logf("Received goAway frame from client")
+				close(successCh)
+			} else {
+				close(errorCh)
+			}
+			return
+		default:
+			// The client should have sent any frame other than GOAWAY
+			t.Logf("The server received a frame other than GOAWAY")
```
arvindbr8, line 2728: Seems like this is also a failure path? We should `close(errorCh)` here as well?

internal/transport/http2_client.go
```diff
@@ -991,7 +1003,10 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.finish()
+	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte("graceful_shutdown"), closeConnErr: errors.New("graceful_shutdown")})
```
arvindbr8, line 1006: It would be nice if we could send more info in the goAway about the graceful shutdown


```suggestion
	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte(fmt.Sprintf("client shutdown with: %v", err)), closeConnErr: err})
```

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,91 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+			return
+		}
+	}()
+
+	ct, err := NewClientTransport(connectCtx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{}, func(GoAwayReason) {})
+	if err != nil {
+		t.Fatalf("Error while creating client transport: %v", err)
+	}
+	s1, err1 := ct.NewStream(connectCtx, &CallHdr{})
```
arvindbr8, line 2737: nit:
```suggestion
	s1, err := ct.NewStream(connectCtx, &CallHdr{})
```

internal/transport/controlbuf.go
```diff
@@ -190,10 +190,10 @@ type incomingGoAway struct {
...
 
 type goAway struct {
-	code      http2.ErrCode
-	debugData []byte
-	headsUp   bool
-	closeConn error // if set, loopyWriter will exit, resulting in conn closure
+	code         http2.ErrCode
+	debugData    []byte
+	headsUp      bool
+	closeConnErr error // if set, loopyWriter will exit, resulting in conn closure
```
arvindbr8, line 196: The comment here isn't accurate. wdyt?

arvindbr8 (changes requested):

internal/transport/http2_client.go
```diff
@@ -991,7 +1003,10 @@ func (t *http2Client) Close(err error) {
...
+	if t.loopy != nil {
```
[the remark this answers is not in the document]
arvindbr8: Also seems like this read here causes a data race

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,87 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+		}
+		sfr := http2.NewFramer(sconn, sconn)
+		if err := sfr.WriteSettings(); err != nil {
+			t.Errorf("Error while writing settings %v", err)
+			return
+		}
+
+		sfr.ReadFrame()
+		sfr.ReadFrame()
+		sfr.ReadFrame()
```
arvindbr8, line 2705:
```suggestion
		fr, _ := sfr.ReadFrame()
		if _, ok := fr.(*http2.SettingsFrame); !ok {
			t.Errorf("Expected settings frame, got %v", fr)
		}
		fr, _ = sfr.ReadFrame()
		if fr, ok := fr.(*http2.SettingsFrame); !ok && fr.IsAck() {
			t.Errorf("Expected settings ACK frame, got %v", fr)
		}
		fr, _ = sfr.ReadFrame()
		if fr, ok := fr.(*http2.HeadersFrame); !ok && fr.Flags.Has(http2.FlagHeadersEndStream) {
			t.Errorf("Expected Headers frame with END_HEADERS frame, got %v", fr)
		}
```

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,87 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+// clientTransport.Close(), client sends a goaway to the server with the correct
+// error code and debug data.
+func (s) TestClientSendsAGoAwayFrame(t *testing.T) {
+	// Create a server.
+	lis, err := net.Listen("tcp", "localhost:0")
+	if err != nil {
+		t.Fatalf("Error while listening: %v", err)
+	}
+	defer lis.Close()
+	// greetDone channel verifies that greet is done with the connection
```
arvindbr8, line 2678: nit: the channel is "used" to verify. So i might rephrase like
```suggestion
	// greetDone is used to notify when server is done greeting the client.
```

and for the following ones.

arvindbr8 (changes requested):

internal/transport/http2_client.go
```diff
@@ -468,6 +472,7 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 			t.conn.Close()
 		}
 		close(t.writerDone)
+		t.writerMu.Unlock()
```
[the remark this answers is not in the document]
aranjans: Wouldn't that cause deadlock? Let's say we acquire lock to initialise it and we release it. Now it acquires lock for [this](https://github.com/grpc/grpc-go/pull/7015/files#diff-5334018444bcb34d15bcb3c8a3050cdf9c8a40fe41305c3a602ea4bfeb0cd178R1013) and since the lock is already acquired, it will never execute [this](https://github.com/grpc/grpc-go/pull/7015/files#diff-5334018444bcb34d15bcb3c8a3050cdf9c8a40fe41305c3a602ea4bfeb0cd178R474)

arvindbr8 (changes requested):
Looked at the changes more closely. 

The reason why we had to do this nil check [here](https://github.com/aranjans/grpc-go/blob/71a70f7d05e613a4c18f47d4fa1c0c1b82c50ece/internal/transport/http2_client.go#L1013) is to handle the case when t.Close() is called before we create the goroutine which creates the loopy. And it is because of this check we are having to use a mutex to handle synchronization of the channel. 

Instead we could just ensure that we dont block read on writerDone channel before calling Close. And according to me I think [this](https://github.com/aranjans/grpc-go/blob/71a70f7d05e613a4c18f47d4fa1c0c1b82c50ece/internal/transport/http2_client.go#L415) defer is the only place where Close is called before the loopy is created. 

I think it might help making `writerDone` a `grpcsync.Event`, which should make it easier to close the channel. It might be tricky otherwise with dealing with panics of closing a closed channel.

Also per our discussion from easier, try to see if its easier to do the same test in a more end2end style.

arvindbr8:
Discussed with @aranjans offline. We figured a way to do it without the grpcsync.Event. @aranjans let me know when the PR is ready for another review.

internal/transport/http2_client.go
```diff
@@ -991,7 +1003,14 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.finish()
+	// The HTTP/2 spec mentions that a GOAWAY frame should be sent before a connection close. If this close() function
+	// ever starts to take in an HTTP/2 error code the peer will be able to get more information about the reason
+	// behind the connection close.
+	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte(fmt.Sprintf("client shutdown with: %v", err)), closeConnErr: err})
+
```
arvindbr8, line 1010: please nix this \n

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,95 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+		if _, err := io.ReadFull(sconn, make([]byte, len(clientPreface))); err != nil {
+			t.Errorf("Error while writing settings ack: %v", err)
+			return
+		}
+		sfr := http2.NewFramer(sconn, sconn)
+		if err := sfr.WriteSettings(); err != nil {
+			t.Errorf("Error while writing settings %v", err)
+			return
+		}
+
```
arvindbr8, line 2702: please nix this newline. I think the general philosophy should be a newline separator for logical blocks.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,95 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+		}
+		switch fr := frame.(type) {
+		case *http2.GoAwayFrame:
+			// Records that the server successfully received a GOAWAY frame.
+			goAwayFrame := fr
+			if goAwayFrame.ErrCode == http2.ErrCodeNo {
+				t.Logf("Received goAway frame from client")
+				close(successCh)
+			} else {
+				close(errorCh)
```
arvindbr8, line 2729: We should log the unexpected situation maybe by signalling the error back to the read. 

The philosophy here should be that when this test fails later in a regression, it should be clear for the user from the logs to understand what failed.

arvindbr8 (approved):
LGTM modulo the few comments. I'm ccing @dfawley here for a second look at this patchset.

internal/transport/http2_client.go
```diff
@@ -469,6 +471,7 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 		}
 		close(t.writerDone)
 	}()
+
```
arvindbr8, line 474: nit: i would remove this newline 

```suggestion
```

internal/transport/http2_client.go
```diff
@@ -991,7 +1002,13 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.finish()
+	// The HTTP/2 spec mentions that a GOAWAY frame should be sent before a connection close. If this close() function
+	// ever starts to take in an HTTP/2 error code the peer will be able to get more information about the reason
+	// behind the connection close.
```
arvindbr8, line 1007: nit: most of the comment is actually obvious. Calling out an excerpt from the spec might be more useful here. how about? 
```suggestion
	// Per HTTP/2 spec, a GOAWAY frame must be sent before closing the
	// connection. See https://httpwg.org/specs/rfc7540.html#GOAWAY.
```

internal/transport/controlbuf.go
```diff
@@ -190,10 +190,10 @@ type incomingGoAway struct {
...
 
 type goAway struct {
-	code      http2.ErrCode
-	debugData []byte
-	headsUp   bool
-	closeConn error // if set, loopyWriter will exit, resulting in conn closure
+	code         http2.ErrCode
+	debugData    []byte
+	headsUp      bool
+	closeConnErr error // if set, loopyWriter will exit with this error
```
arvindbr8, line 196: Still this is only true on the server side. Is this true on the client side as well?

aranjans: Yeah this is valid for client side as well. We are looping out from writer when we receive this error.

internal/transport/http2_client.go
```diff
@@ -991,7 +1002,12 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.finish()
+	// Per HTTP/2 spec, a GOAWAY frame must be sent before closing the
+	// connection. See https://httpwg.org/specs/rfc7540.html#GOAWAY.
+	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte(fmt.Sprintf("client shutdown with: %v", err)), closeConnErr: err})
+	if t.writerDone != nil {
```
dfawley, line 1008: When is `writerDone == nil`?

internal/transport/controlbuf.go
dfawley, line 196: Maybe just revert this change since it results in churn and isn't a substantial improvement.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,95 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Error while listening: %v", err)
+	}
+	defer lis.Close()
+	// greetDone is used to notify when server is done greeting the client.
+	greetDone := make(chan struct{})
+	// successCh verifies that GOAWAY is received at server side
+	successCh := make(chan struct{})
+	// errorCh verifies that desired GOAWAY not received by server
+	errorCh := make(chan struct{})
```
dfawley, line 2683: These can probably be combined..  The pattern is usually:

```go
errCh := make(chan error)
...
// on error:
errCh<-err
close(errCh)

// on success:
close(errCh)
```

test/goaway_test.go
```diff
@@ -761,3 +761,66 @@ func (s) TestTwoGoAwayPingFrames(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	val, err := ctCh.Receive(ctx)
+	if err != nil {
+		t.Fatalf("timeout waiting for client transport (should be given after http2 creation)")
+	}
+	ct := val.(*clientTester)
+	goAwayReceived := make(chan struct{})
+	errCh := make(chan struct{})
```
dfawley, line 797: As above, you can just use 1 channel for both purposes.

internal/transport/http2_client.go
```diff
@@ -517,6 +519,15 @@ func (t *http2Client) getPeer() *peer.Peer {
 	}
 }
 
+// OutgoingGoAwayHandler writes GOAWAY to the connection.  Always returns (false, err) as we want the GoAway
+// to be the last frame loopy writes to the transport.
+func (t *http2Client) outgoingGoAwayHandler(g *goAway) (bool, error) {
+	if err := t.framer.fr.WriteGoAway(math.MaxUint32, http2.ErrCodeNo, g.debugData); err != nil {
```
dfawley, line 525: This should write the max stream ID and not maxint.

dfawley (commented):
Looks good overall.  Thanks for fixing this ancient issue!

internal/transport/http2_client.go
```diff
@@ -408,10 +408,9 @@ func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts
 	readerErrCh := make(chan error, 1)
 	go t.reader(readerErrCh)
 	defer func() {
-		if err == nil {
-			err = <-readerErrCh
-		}
 		if err != nil {
+			// writerDone should be closed since the loopy goroutine wouldn't have started in the case this function returns an error.
```
dfawley, line 412: Please wrap all comments at 80 columns (there's more below, I believe).

internal/transport/http2_client.go
```diff
@@ -991,7 +1003,10 @@ func (t *http2Client) Close(err error) {
 		t.kpDormancyCond.Signal()
 	}
 	t.mu.Unlock()
-	t.controlBuf.finish()
+	// Per HTTP/2 spec, a GOAWAY frame must be sent before closing the
+	// connection. See https://httpwg.org/specs/rfc7540.html#GOAWAY.
+	t.controlBuf.put(&goAway{code: http2.ErrCodeNo, debugData: []byte(fmt.Sprintf("client shutdown with: %v", err)), closeConn: err})
```
dfawley, line 1008: Let's omit the client shutdown error from the data we send to the server.  It doesn't seem like a good idea to send that.  It's probably just `ErrConnClosing`, but better safe than sorry.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,98 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+		switch fr := frame.(type) {
+		case *http2.GoAwayFrame:
+			// Records that the server successfully received a GOAWAY frame.
+			goAwayFrame := fr
+			if goAwayFrame.ErrCode == http2.ErrCodeNo {
+				t.Logf("Received goAway frame from client")
+				close(errorCh)
+			} else {
+				t.Logf("Received unexpected goAway frame from client")
+				errorCh <- errors.New("received unexpected goAway frame from client")
```
dfawley, line 2727: To be more helpful these need to include the ErrCode received (and also the one expected)

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,98 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+				close(errorCh)
+			}
+			return
+		default:
+			t.Logf("The server received a frame other than GOAWAY")
+			errorCh <- errors.New("server received a frame other than GOAWAY")
+			close(errorCh)
```
dfawley, line 2734: Similar as above.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,98 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	t.Logf("Closed the client connection")
+	select {
+	case err = <-errorCh:
+	case <-ctx.Done():
+		t.Errorf("Timed out")
+	}
+	if err != nil {
+		t.Errorf("Error receiving the GOAWAY frame: %v", err)
+	} else {
+		t.Logf("Received a GOAWAY frame from client")
```
dfawley, line 2759: You log this above, so no need to repeat it here.

internal/transport/transport_test.go
```diff
@@ -2664,3 +2664,98 @@ func TestConnectionError_Unwrap(t *testing.T) {
...
+	_, err = ct.NewStream(ctx, &CallHdr{})
+	if err != nil {
+		t.Fatalf("failed to open stream: %v", err)
+	}
+	// Wait until server receives the headers and settings frame as part of greet.
+	<-greetDone
+	ct.Close(errors.New("manually closed by client"))
+	t.Logf("Closed the client connection")
+	select {
+	case err = <-errorCh:
```
dfawley, line 2752: Simplify to:

```go
select {
case err := <-errorCh:
	t.Errorf(...)
}
```

please.

test/goaway_test.go
```diff
@@ -761,3 +761,66 @@ func (s) TestTwoGoAwayPingFrames(t *testing.T) {
...
+				close(errCh)
+			}
+		}
+	}()
+	cc.Close()
+	defer ct.conn.Close()
+	select {
+	case <-goAwayReceived:
+	case <-errCh:
+		t.Errorf("Error receiving the goAway: %v", err)
```
dfawley, line 822: You never send the error on the channel or read it here, so this `err` will be `nil` (from the last `ctCh.Receive`).

internal/transport/http2_client.go
dfawley: This still includes the error.  I would prefer if there was no dynamic data in here, just a static string that says the client is shutting down and closing the connection.

internal/transport/transport_test.go
dfawley: `:=` please, to limit scope

test/goaway_test.go
```diff
@@ -818,7 +820,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 	defer ct.conn.Close()
 	select {
 	case <-goAwayReceived:
-	case <-errCh:
+	case err = <-errCh:
```
dfawley, line 823: `:=` please, to limit scope

internal/transport/http2_client.go
aranjans: I think this is already fixed, but i am not sure why github doesn't shows the change made post the comment.

dfawley: I don't know, but please fix it.

dfawley (approved):
LGTM modulo the requested change to the last stream ID field in the GOAWAY.

Thanks and congrats on your first PR!

arvindbr8 (approved):
LGTM. Congrats on the first PR!

PR #7025, opened 2024-03-06
Update go.mod

SpaceSuffer:
updated to the latest version


arvindbr8:
This diff didnt update all the modules which ref protobuf library(s). And as part of updating that we need to re-generate pb.go's as well. Seems like https://github.com/grpc/grpc-go/pull/7028 has already made the change. Thanks for the PR.

PR #7029, opened 2024-03-07
grpc: clean up doc strings and some code around Dial vs NewClient

dfawley:
Also added a test and removed an unused parameter from a function and call sites.  (The reason it was needed in the past is to auto-detect the gRPCLB LB policy, which is functionality we no longer support.)

RELEASE NOTES: none


arvindbr8 (approved):
LGTM, just a minor nit

PR #1902, opened 2018-03-07
Call options provide methods so interceptors in other packages can recover settings

jhump:
This fixes #1495.

If I understood correctly, @dfawley, this is the approach you were suggesting. Can you take a look and let me know?

PR #7033, opened 2024-03-11
grpc: Export header list sizes in DialOption and ServerOption

ZhouyihaiDing:
Similar to https://github.com/grpc/grpc-go/pull/1902, it is good to export the header list sizes for users to know what is the existing metadata size limit.

RELEASE NOTES: none

### DIFF
--- a/server.go
+++ b/server.go
@@ -528,12 +528,22 @@ func ConnectionTimeout(d time.Duration) ServerOption {
 	})
 }
 
+// MaxHeaderListSizeServerOption is a ServerOption that sets the max
+// (uncompressed) size of header list that the server is prepared to accept.
+type MaxHeaderListSizeServerOption struct {
+	MaxHeaderListSize uint32
+}
+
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -528,12 +528,22 @@ func ConnectionTimeout(d time.Duration) ServerOption {
...
+	so.maxHeaderListSize = &o.MaxHeaderListSize
+}
+
 // MaxHeaderListSize returns a ServerOption that sets the max (uncompressed) size
 // of header list that the server is prepared to accept.
 func MaxHeaderListSize(s uint32) ServerOption {
-	return newFuncServerOption(func(o *serverOptions) {
-		o.maxHeaderListSize = &s
-	})
+	return &MaxHeaderListSizeServerOption{
```
dfawley, line 544: Can we keep the two types consistent please?  I.e. return this as a value here instead of a pointer to be consistent with the DialOption.

dfawley (approved):

PR #7035, opened 2024-03-12
gracefulswitch: add ParseConfig and make UpdateClientConnState call SwitchTo if needed

dfawley:
cc @anicr7

This change does not migrate some users of gracefulswitch; they can be done as a follow-up: clusterimpl, outlierdetection, balancergroup.

RELEASE NOTES: none


internal/balancer/gracefulswitch/gracefulswitch.go
```diff
@@ -152,12 +161,32 @@ func (gsb *Balancer) latestBalancer() *balancerWrapper {
 }
 
 // UpdateClientConnState forwards the update to the latest balancer created.
+//
+// If the state's BalancerConfig is the config returned by a call to
+// gracefulswitch.ParseConfig, then this function will automatically SwitchTo
```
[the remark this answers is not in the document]
zasweq: maybe after by the config...", if applicable," or something along those lines.

service_config.go
```diff
@@ -164,7 +156,7 @@ type jsonMC struct {
 // TODO(lyuxuan): delete this struct after cleaning up old service config implementation.
 type jsonSC struct {
 	LoadBalancingPolicy *string
-	LoadBalancingConfig *internalserviceconfig.BalancerConfig
+	LoadBalancingConfig *json.RawMessage
```
zasweq, line 159: Why not continue to use the old logic of looping through list on the internalserviceconfig.BalancerConfig type? Now you need to prepare a graceful switch config alongside looping through this json list to find first config for balancer that is registered?

internal/balancer/gracefulswitch/gracefulswitch.go
```diff
@@ -94,14 +94,23 @@ func (gsb *Balancer) balancerCurrentOrPending(bw *balancerWrapper) bool {
 // process is not complete when this method returns. This method must be called
 // synchronously alongside the rest of the balancer.Balancer methods this
 // Graceful Switch Balancer implements.
+//
+// Deprecated: use ParseConfig and pass a parsed config to UpdateClientConnState
+// to cause the Balancer to automatically change to the new child.
```
zasweq, line 99: same here, add "if applicable" or something

internal/balancer/gracefulswitch/config.go
```diff
@@ -0,0 +1,83 @@
...
+// ServiceConfig.  It returns a type that should be passed to
+// UpdateClientConnState in the BalancerConfig field.
+func ParseConfig(cfg json.RawMessage) (serviceconfig.LoadBalancingConfig, error) {
+	var lbCfg []map[string]json.RawMessage
+	if err := json.Unmarshal(cfg, &lbCfg); err != nil {
+		return nil, err
+	}
+	for _, e := range lbCfg {
+		if len(e) != 1 {
+			return nil, fmt.Errorf("expected a JSON struct with one entry; received: %v", e)
```
zasweq, line 54: Optional nit: add info of index too, since this just gives the struct that doesn't meet this requirement, but no info about what part of overall JSON array. (I.e. https://github.com/grpc/grpc-go/blob/master/internal/serviceconfig/serviceconfig.go#L84)

dfawley: Done

internal/balancer/gracefulswitch/gracefulswitch.go
dfawley: Done.

service_config.go
dfawley: > internalserviceconfig.BalancerConfig

I want to delete this type.  It's not a good idea to combine parsing and additional business logic in the same step.

zasweq: Yeah I remember our effort to do it. I had planned to do it as part of custom lb plumbing in xDS, but never got around to it. I think the configs are being validated twice IIRC as a result.

internal/balancer/gracefulswitch/gracefulswitch.go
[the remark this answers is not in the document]
zasweq: Ah, automatically from callers perspective. I meant too strong as I thought it meant "unconditionally", now with "if necessary" this works.

zasweq:
This change LGTM, what was the technical reason this came up (other than the fact this is now how Java and C does it)? Will let Arvind gate the merge though since he's official reviewer.

dfawley (commented):
Comments addressed

> This change LGTM, what was the technical reason this came up (other than the fact this is now how Java and C does it)? Will let Arvind gate the merge though since he's official reviewer.

The reason we did this is that we had a user that wanted to delegate to a child using GSB but didn't want to have to do all the parsing logic.

arvindbr8 (approved):

PR #7038, opened 2024-03-14
Fix: error message using correct keepalive config value

BatmanAoD:
`kpTimeoutLeft` is calculated from `kp.Timeout`, but the error message implies it's based on `kp.Time`.

RELEASE NOTES:
- transport/server: use the proper timeout value when keepalive pings are not ack'd in time

### DIFF
--- a/internal/transport/http2_server.go
+++ b/internal/transport/http2_server.go
@@ -1198,7 +1198,7 @@ func (t *http2Server) keepalive() {
 				continue
 			}
 			if outstandingPing && kpTimeoutLeft <= 0 {
-				t.Close(fmt.Errorf("keepalive ping not acked within timeout %s", t.kp.Time))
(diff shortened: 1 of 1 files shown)


dfawley (approved):
LGTM, but you will need to sign off on the CLAs (by clicking the links from the comment from the bot).  Thanks!

BatmanAoD:
@dfawley Talking to my company about it now; it will probably be a couple of weeks because this fix isn't actually particularly high-priority. (I should have just opened a ticket, since the fix is so trivial...)

arvindbr8:
@BatmanAoD -- Did your meeting happen? Hopefully it went well.

BatmanAoD:
@arvindbr8 It sounds like the CLA shouldn't be a problem for us; I've started the signature process.

arvindbr8 (approved):
Nice, now we can see more PR's from you @BatmanAoD  :) . Thanks so much for the PR!

PR #7040, opened 2024-03-16
googlec2p: use the bootstrap parsing code to generate parsed bootstrap config instead of handcrafting it

proxfly:
Fixes: https://github.com/grpc/grpc-go/issues/6901

It looks like the test cases are failing because the bootstrap parsing code is populating fallback values that are not anticipated by the failed test cases. I’m unfamiliar with these fallback values and would appreciate everyone’s insight.

RELEASE NOTES: none


arvindbr8 (changes requested):
I've made a pass at the PR. I think the `TestBuildXDS` test also needs to be updated. But I need some clarity on what the expected behavior of the ClientListenerResourceNameTemplate is for c2p. I will get back to that in a bit.

cc: @apolcyn

arvindbr8 (approved):
@proxfly Heads up - I'm merging a change that would cause a  conflict with your branch. 

There was a behavior regression in the resolver. I wanted to fix that as a different PR. So that this change looks cleaner. Please rebase your changes with master when you get a chance.

xds/googledirectpath/googlec2p.go
```diff
@@ -106,25 +102,21 @@ func (c2pResolverBuilder) Build(t resolver.Target, cc resolver.ClientConn, opts
 	if balancerName == "" {
 		balancerName = tdURL
 	}
-	serverConfig, err := bootstrap.ServerConfigFromJSON([]byte(fmt.Sprintf(`
+
+	node := newNode(<-zoneCh, <-ipv6CapableCh)
+	xdsServer := newXdsServer(balancerName)
+	authorities := newAuthorities(xdsServer)
```
arvindbr8, line 108: optional: how about `new(Whatever)Config`? `newXdsServer` sounds to me like we are creating something new here. But essentially its only the config

xds/googledirectpath/googlec2p.go
```diff
@@ -106,25 +102,21 @@ func (c2pResolverBuilder) Build(t resolver.Target, cc resolver.ClientConn, opts
...
+
+	config, err := bootstrap.NewConfigFromContents([]byte(fmt.Sprintf(`
 	{
-		"server_uri": "%s",
-		"channel_creds": [{"type": "google_default"}],
-		"server_features": ["xds_v3", "ignore_resource_deletion"]
-	}`, balancerName)))
+		"xds_servers": [%s],
+	    "authorities": %s,
+		"node": %s
```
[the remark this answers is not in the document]
arvindbr8: `\t` instead of `\s`

arvindbr8 (changes requested):

arvindbr8:
The changes in #7048 should resolve the diff for `ClientListenerResourceNameTemplate` in the test. For the diff in  `server_features` -- `xds.config.resource-in-sotw` is expected to be part of the config - So I would augment the test golden config to have that as well.

xds/googledirectpath/googlec2p.go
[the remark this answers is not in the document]
proxfly: Done

xds/googledirectpath/googlec2p.go
```diff
@@ -106,25 +102,21 @@ func (c2pResolverBuilder) Build(t resolver.Target, cc resolver.ClientConn, opts
...
+	}`, xdsServer, authorities, node)))
+
 	if err != nil {
 		return nil, fmt.Errorf("failed to build bootstrap configuration: %v", err)
 	}
-	config := &bootstrap.Config{
-		XDSServer: serverConfig,
-		ClientDefaultListenerResourceNameTemplate: "%s",
```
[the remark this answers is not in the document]
proxfly: Done

xds/googledirectpath/googlec2p.go
proxfly: Done. Good insight!

xds/googledirectpath/googlec2p.go
```diff
@@ -166,30 +160,45 @@ func (r *c2pResolver) Close() {
...
 	}
-	return ret
+	return fmt.Sprintf(`
+	{
+		"id": "%s",
+		"user_agent_name": "%s",
+		"UserAgentVersionType": {
+			"userAgentVersion": "%s"
+		},
+		"client_features": ["%s", "%s"],
```
arvindbr8, line 177: `UserAgentVersionType` is in camel case while the rest seem to be in snake case. 


But this got me thinking `newConfigFromContents` overrides these values here https://github.com/grpc/grpc-go/blob/cce163274b6cb9de2b2bf0bc742384e5755fee42/xds/internal/xdsclient/bootstrap/bootstrap.go#L568-L570


I would prefer if we didnt set it here. Sorry about the back and forth on this.

```suggestion
```

arvindbr8 (approved):
Thanks for the PR. Looks really good to me modulo one comment. 

Per policy we need another pass from one of the maintainer. cc: @zasweq

xds/internal/xdsclient/bootstrap/bootstrap.go
```diff
@@ -567,7 +565,7 @@ func newConfigFromContents(data []byte) (*Config, error) {
 	}
 	node.UserAgentName = gRPCUserAgentName
 	node.UserAgentVersionType = &v3corepb.Node_UserAgentVersion{UserAgentVersion: grpc.Version}
-	node.ClientFeatures = append(node.ClientFeatures, clientFeatureNoOverprovisioning, clientFeatureResourceWrapper)
+	node.ClientFeatures = AppendIfNotPresent(node.ClientFeatures, clientFeatureNoOverprovisioning, clientFeatureResourceWrapper)
```
arvindbr8, line 568: Thanks for making the change. I think with the last commit, this can also be reverted

zasweq (approved):
LGTM.

PR #7046, opened 2024-03-19
channelz: re-add state for subchannels

dfawley:
RELEASE NOTES: none


sergiitk:
For posterity: the regression introduced in #6969, internal ref b/330220025.

PR #7048, opened 2024-03-19
googlec2p: use xdstp style template for client LDS resource name

arvindbr8:
The google c2p resolver should be using the new xdstp style names for LDS resource names. Also adds "xds.config.resource-in-sotw" to `server_features` which is also required by the google c2p resolver. 

RELEASE NOTES: none

### DIFF
--- a/xds/googledirectpath/googlec2p.go
+++ b/xds/googledirectpath/googlec2p.go
@@ -110,7 +110,7 @@ func (c2pResolverBuilder) Build(t resolver.Target, cc resolver.ClientConn, opts 
 	{
 		"server_uri": "%s",
 		"channel_creds": [{"type": "google_default"}],
-		"server_features": ["xds_v3", "ignore_resource_deletion"]
+		"server_features": ["xds_v3", "ignore_resource_deletion", "xds.config.resource-in-sotw"]
 	}`, balancerName)))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


xds/googledirectpath/googlec2p.go
```diff
@@ -120,7 +120,8 @@ func (c2pResolverBuilder) Build(t resolver.Target, cc resolver.ClientConn, opts
 		ClientDefaultListenerResourceNameTemplate: "%s",
 		Authorities: map[string]*bootstrap.Authority{
 			c2pAuthority: {
-				XDSServer: serverConfig,
```
apolcyn, line 123: nit: can we add a TODO comment to explain how this only needed b/c it doesn't use the parsing logic in bootstrap.go which would normally set this default?

(otherwise it's a little confusing how this looks different from java/c++)

arvindbr8: This is all changing in https://github.com/grpc/grpc-go/pull/7040. I wanted to fix the behavior as part of this PR and try to make it more obvious in the other PR

apolcyn (approved):

apolcyn (approved):

apolcyn (approved):

dfawley (approved):

Issue #7030, opened 2024-03-08
codegen: Use Go generics for stream types


aarongable:
### Use case(s)

Suppose you have the following service definitions:

```proto
message HelloRequest {
	string name = 1;
}

message HelloReply {
	string greeting = 1;
}

service Greeter {
	rpc StoreHello (HelloRequest) returns (google.protobuf.Empty) {}
	rpc SayHello (google.protobuf.Empty) returns (stream HelloReply) {}
}

service GreeterReadOnly {
	rpc SayHello (google.protobuf.Empty) returns (stream HelloReply) {}
}
```

The idea here being that someone with access to a Greeter client could call StoreHello multiple times to store many names that should be greeted, and then either a Greeter client or a GreeterReadOnly client can call SayHello to get a stream of all the stored replies. Obviously this is a toy example, but you can see a real example of this same pattern [here](https://github.com/letsencrypt/boulder/blob/main/sa/proto/sa.proto). The salient feature here is this: we have two different services which define the _exact same streaming RPC_.

Now because both Greeter and GreeterReadOnly share functionality, we want to use a single implementation type to implement both of them. So we write something like this:

```go
type GreeterImpl struct {
	names []string
}

func (g *GreeterImpl) StoreHello(ctx context.Context, req *pb.HelloRequest) (*emptypb.Empty, error) {
	...
}

func (g *GreeterImpl) SayHello(req *emptypb.Empty, stream pb.Greeter_SayHelloServer) error {
	...
}
```

But of course when we go to Register this implementation, it doesn't actually work. It satisfies the `Greeter` interface, but _not_ the `GreeterReadOnly` interface.

```go
func main() {
	grpcServer := grpc.NewServer()
	pb.RegisterGreeterServer(grpcServer, &GreeterImpl{}) // This succeeds
	pb.RegisterGreeterReadOnlyServer(grpcServer, &GreeterImpl{}) // This fails at compile time
}
```

This is because grpc-go has generated the following interfaces for us:

```go
type GreeterServer interface {
	StoreHello(*HelloRequest) (*emptypb.Empty, error)
	SayHello(*HelloRequest, Greeter_SayHelloServer) error
}

type GreeterReadOnlyServer interface {
	SayHello(*HelloRequest, GreeterReadOnly_SayHelloServer) error
}
```

The `SayHello` methods in these two interfaces do not have the same signature, despite having the exact same definition in the original .proto file. This means that they can never both be implemented by the same Go struct.

[...]

PR #7057, opened 2024-03-20
Add an optional implementation of streams using generics

aarongable:
Add six new interfaces, and two new structs implementing those interfaces, to the grpc package. These interfaces and implementations represent the client and server sides of client-streaming, server-streaming, and bidi-streaming RPCs.

Add a new "use_generic_streams" flag to protoc-gen-go-grpc, which causes the code generation to use these new experimental types instead of generating individual implementations for each and every streaming method. This simplifies both the code generation process, and the resulting generated code.

Release notes for protoc-gen-go-grpc:
* the new "use_generic_stream_experimental=true" option will cause the gRPC codegen to use prebuilt generic types to implement client and server stream objects, rather than generating new types and implementations for every RPC method.

Fixes https://github.com/grpc/grpc-go/issues/7030

RELEASE NOTES: none


dfawley:
Thanks for the PR and sorry I haven't gotten to it yet.  I like the way it looks, but I need to spend more time thinking about it than I currently have available.  Hopefully I will have more time by early-mid next week.

cmd/protoc-gen-go-grpc/grpc.go
```diff
@@ -318,11 +319,25 @@ func genClientMethod(gen *protogen.Plugin, file *protogen.File, g *protogen.Gene
...
+		} else { // i.e. if method.Desc.IsStreamingServer()
+			streamInterface = g.QualifiedGoIdent(experimentalPackage.Ident("ServerStreamClient")) + "[" + g.QualifiedGoIdent(method.Output.GoIdent) + "]"
+		}
+	}
+
 	serviceDescVar := service.GoName + "_ServiceDesc"
 	g.P("stream, err := c.cc.NewStream(ctx, &", serviceDescVar, ".Streams[", index, `], `, fmSymbol, `, opts...)`)
 	g.P("if err != nil { return nil, err }")
-	g.P("x := &", streamType, "{stream}")
+	g.P("x := &", streamType, "{ClientStream: stream}")
```
aarongable, line 340: Note to reviewers: this change is not necessary for my main change to work. I made this change only so that the "[tests (vet)](https://github.com/grpc/grpc-go/actions/runs/8365438407/job/22903238268)" task would pass -- it appears to me that the existing codegen fails vet. But then of course making this change caused all of the existing example generated code to have diffs and fail "[vet-proto](https://github.com/grpc/grpc-go/actions/runs/8366309053/job/22906239552?pr=7057)", so I'm happy to revert this portion of the change.

examples/route_guide/routeguide/route_guide.pb.go
```diff
@@ -14,8 +14,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.32.0
-// 	protoc        v4.25.2
+// 	protoc-gen-go v1.33.0
+// 	protoc        v3.21.12
```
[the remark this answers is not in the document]
aarongable: Yeah, I'd chosen not to care about this because I wasn't sure if we were going to be landing the updated example generated code at all (see my note at the bottom of the PR description). I've now regenerated all of the examples using protoc v4.25.2.

vet.sh
```diff
@@ -41,7 +41,7 @@ if [[ "$1" = "-install" ]]; then
   popd
   if [[ -z "${VET_SKIP_PROTO}" ]]; then
     if [[ "${GITHUB_ACTIONS}" = "true" ]]; then
-      PROTOBUF_VERSION=25.2 # a.k.a. v4.22.0 in pb.go files.
```
dfawley, line 44: Oof.  Maybe just change this to `v4.x.y` so that it doesn't have to be kept in sync, if you don't mind?

dfawley (commented):
So even with the generics being generated, the `route_guide` server and client do not need any changes?

I thought we were saying things would not be backward compatible...can you remind me where the trouble comes in please?

vet.sh
arvindbr8: This should not be changed. The release version 25.2 uses v4.22.0 as the version number in the generated pb.go files

aarongable: Ah, I did not understand! I assumed that someone had updated the PROTOBUF_VERSION without updating the comment next to it. I'll revert this, thanks.

aarongable:
> So even with the generics being generated, the `route_guide` server and client do not need any changes?
> 
> I thought we were saying things would not be backward compatible...can you remind me where the trouble comes in please?

Correct, as this change is written now, everything is backwards compatible. (This is why I asked about maybe putting the new type definitions directly into the grpc package, rather than into experimental, since technically this change could be made completely invisibly to everyone as long as they're on a version of go with generics.)

In my _ideal_ world, the type aliases wouldn't exist. Rather than the generated code saying
```go
type RouteGuideServer interface {
	RouteChat(RouteGuide_RouteChatServer) error
}
...
type RouteGuide_RouteChatServer = grpc.BidiStreamServer[RouteNote, RouteNote]
```
it would just say
```go
type RouteGuideServer interface {
	RouteChat(grpc.BidiStreamServer[RouteNote, RouteNote]) error
}
```
But _that_ change would be backwards incompatible, since everyone would need to update the function signatures of the types that they use to implement such servers.

---

One way to split the difference between how this PR currently looks and my imagined ideal would be to use the new types everywhere (including in the interface definitions, and in the client implementation function signatures) _and_ still provide the type aliases purely for backwards compatibility:

```go
type RouteGuideServer interface {
	RouteChat(grpc.BidiStreamServer[RouteNote, RouteNote]) error
}
...
type RouteGuide_RouteChatServer = grpc.BidiStreamServer[RouteNote, RouteNote]
```

 This would be a slightly more visible change, as anyone who regenerates their protos will see the interface definitions change, but all of their existing code will still continue to work thanks to the type alias.

dfawley:
Both of the two versions (this and the "split the difference") seem to be the same in practice.  Is there anything you're missing from this current version that could be done with the other version (besides maybe looking nicer / being more understandable)?

aarongable:
Nope, they're roughly equivalent to me. From a "will it compile" perspective, they're identical to me.

I do slightly prefer the "split the difference" version for one reason. The original motivation behind this change was allowing a single struct to implement two different gRPC Services that share the same streaming method, and both the current version and the "split the difference" version allow that. However, the current version makes it _look_ a little awkward, because the function signature on the type you implement won't match at least one of the two interfaces you're implementing:

```go
// Generated code
type GreeterServer interface {
	SayHello(*HelloRequest, Greeter_SayHelloServer) error
}

type Greeter_SayHelloServer = grpc.ServerStreamServer[HelloReply]

type GreeterReadOnlyServer interface {
	SayHello(*HelloRequest, GreeterReadOnly_SayHelloServer) error
}

type GreeterReadOnly_SayHelloServer = grpc.ServerStreamServer[HelloReply]

// Implementation
type GreeterImpl struct { ... }

// Any of the three lines below will compile, but none of them are identical to both interface lines above.
func (g *GreeterImpl) SayHello(req *emptypb.Empty, stream pb.Greeter_SayHelloServer) error { ... }
func (g *GreeterImpl) SayHello(req *emptypb.Empty, stream pb.GreeterReadOnly_SayHelloServer) error { ... }
func (g *GreeterImpl) SayHello(req *emptypb.Empty, stream grpc.ServerStreamServer[pb.HelloReply]) error { ... }
```

The "spllit the difference" version would cause the third of the three options above to exactly match both generated interfaces.

But as you said, that's just "looking nicer / being more understandable", and the current version completely works.

dfawley:
Thanks for the insights.  I would agree with your assessment, then, and also prefer the alternate version.  I think we can generate those type aliases and comment them as "these are only here for backward compatibility".

Let's go ahead and move the interfaces into `grpc` (in a new file, please, maybe `services.go` or `service_interfaces.go` or something like that) and bump the version [here](https://github.com/grpc/grpc-go/blob/master/rpc_util.go#L978).  But we'll have to make the codegen only _optionally_ emit these and, when enabled, also require ([here](https://github.com/grpc/grpc-go/blob/308dbc446644b2f783c29a7abc03f9bec60e8512/cmd/protoc-gen-go-grpc/grpc.go#L178)) the new version until the next release of gRPC (next month).  At that time, we can turn it on by default unless we discover any problems.

Also, I would like to use that option to regenerate all the protos that we build in this repo via `regenerate.sh`.

Does that sound good?

aarongable:
Sounds great! I'll probably get to updating this next week. Thanks for the pointers to the version locks.

dfawley:
Also, if `extras` is failing in CI, #7120 will make it much easier to see what's happening by eliminating a ton of spam.

aarongable:
As I work on this today: is it okay for me to force-push to this PR branch to keep things maximally tidy? Or would you prefer that I merge `master` and continue forward from here?

aarongable:
Okay, I believe I've addressed all comments so far. The latest version puts the new generic types in the top-level grpc package (in stream_interfaces.go, right next to stream.go, which they're closely tied to), includes code to generate those types as part of the generated interfaces if the `use_generic_streams=true` opt is set, uses the version flag to ensure that the right version of the grpc package is being used alongside that new generated code, and leaves comments on the generated type aliases indicating that they exist for backwards compatibility.

All of the examples in the repo have been regenerated with the `use_generic_streams=true` flag set, so you can see what the final result looks like, and all of the examples still work with no changes necessary to the hand-written parts of their code.

I'm not sure why the "extras" test is failing, the log looks fully successful to me until it exits 1 at the very end. I'd appreciate a pointer on how best to debug that, if you have one. Thanks!

dfawley (commented):
Everything looks great to me, but I have a few comments about naming / documentation.

dfawley:
> Conflicting files
> cmd/protoc-gen-go-grpc/testdata/golden_grpc.pb.go

FYI this file was deleted, so you can just delete it in your branch as well.

stream_interfaces.go
```diff
@@ -18,80 +18,84 @@
...
 package grpc
 
-// ServerStreamClient represents the client side of a server-streaming (one
+// ServerStreamingClient represents the client side of a server-streaming (one
 // request, many responses) RPC. It is generic over the type of the response
-// message.
-type ServerStreamClient[T any] interface {
-	Recv() (*T, error)
+// message. It is used in generated code.
+type ServerStreamingClient[Req any] interface {
```
dfawley, line 24: This should be `Res`, not `Req`.

dfawley (approved):

dfawley:
I love it.  Thank you for the contribution.  We just need a second pass review (and also please make the q->s change).

cmd/protoc-gen-go-grpc/main.go
```diff
@@ -55,6 +56,7 @@ func main() {
 
 	var flags flag.FlagSet
 	requireUnimplemented = flags.Bool("require_unimplemented_servers", true, "set to false to match legacy behavior")
+	useGenericStreams = flags.Bool("use_generic_streams", false, "set to true to use generic types for streaming client and server objects")
```
arvindbr8, line 59: Should the new flag be marked experimental in the next release of protoc-gen-go-grpc? By experimental, I mean flag string be `use_generic_streams_experimental` and the usage says something like `This flag is EXPERIMENTAL and may be changed or removed in a later release.`

Later when we stabilize the API we can switch the flag's default to true and remove the experimental tags. 

@dfawley, wdyt?

arvindbr8 (commented):
Looks great. Modulo one comment on the new feature flag.

cmd/protoc-gen-go-grpc/main.go
dfawley: That makes sense - the idea would be to delete the flag entirely one release after it is on by default.  So:

1. Add flag, disabled, release.
2. Switch default of flag, release.
3. Remove flag, release.

(All separated by some reasonable amount of time to find bugs/etc.)

aarongable: Makes sense to me! I've updated the name and description of the flag, updated the two places it is referenced in scripts in this repo, and updated the release notes in the PR description.

dfawley (approved):

arvindbr8 (approved):
Thanks for taking care of it. Looks great!

PR #7059, opened 2024-03-21
deps: move from github.com/golang/protobuf to google.golang.org/protobuf/proto

Clement-Jean:
@dfawley @arvindbr8 this is the next step after #6919 since #6969 is closed.

I'm still unsure about the remaining deprecated dependency in channelz/service/service_test.go. First, this line:

```go
proto.RegisterType((*OtherSecurityValue)(nil), "grpc.credentials.OtherChannelzSecurityValue")
```

it seems to be possible to use `protoregistry.GlobalTypes.RegisterMessage` instead of `proto.RegisterType`. I just don't know how yet.

Second, it seems that `proto.UnmarshalText(want[i], w)` and `prototext.Unmarshall(string(want[i]), w)` are not doing the same thing. The first one accepts '\x00' in the string and the second does not. Any idea why?

RELEASE NOTES: none


dfawley:
> it seems to be possible to use protoregistry.GlobalTypes.RegisterMessage instead of proto.RegisterType. I just don't know how yet.

This seems reasonable.  But I think it's also possible, if you want, to create a proto file and import it instead of doing it dynamically like the test currently does.

> Second, it seems that proto.UnmarshalText(want[i], w) and prototext.Unmarshall(string(want[i]), w) are not doing the same thing. The first one accepts '\x00' in the string and the second does not. Any idea why?

I'm not sure.  Can you do `prototext.Marshal()` of the output of the specific `proto.UnmarshalText`?  Maybe that will show you the format `prototext.Unmarshal` wants?

Clement-Jean:
> This seems reasonable. But I think it's also possible, if you want, to create a proto file and import it instead of doing it dynamically like the test currently does.

For now, I found a simpler solution that does pretty much the same as before:

```go
func init() {
	// Ad-hoc registering the proto type here to facilitate Unmarshal of OtherSecurityValue.
	m := (*OtherSecurityValue)(nil)
	s := "grpc.credentials.OtherChannelzSecurityValue"
	mt := protoimpl.X.LegacyMessageTypeOf(m, protoreflect.FullName(s))
	if err := protoregistry.GlobalTypes.RegisterMessage(mt); err != nil {
		panic(err)
	}
}
```

We are still using legacy way of registering messages but not the deprecated dependency. The main problem I have with having a proto file or registering a descriptor is that it is not linked to `OtherSecurityValue`.

> I'm not sure. Can you do prototext.Marshal() of the output of the specific proto.UnmarshalText? Maybe that will show you the format prototext.Unmarshal wants?

Doing:

```go
err := proto.UnmarshalText(want[i], w)
if err != nil {
  t.Fatal(err)
}
_, err = prototext.Marshal(w)
if err != nil {
  t.Fatal(err)
}
```

there are not errors returned...

dfawley:
> there are not errors returned...

We wouldn't expect errors.  I was curious what the output would look like as that should really be something that would be accepted by `prototext.Unmarshal`.

Clement-Jean:
The issue seems to be coming from the function called `addr` which doesn't return valid UTF8 strings. And `prototext.Unmarshal` seems to be expecting UTF8 strings.

Clement-Jean:
I believe we could also remove this line in vet.sh:

```bash
# - Ensure all ptypes proto packages are renamed when importing.
not git grep "\(import \|^\s*\)\"github.com/golang/protobuf/ptypes/" -- "*.go"
```

Other than that, if there is no way to remove the dependency in `reflection/grpc_testing_not_regenerate/testv3.go`, I think we are pretty much done.

dfawley:
> The issue seems to be coming from the function called addr which doesn't return valid UTF8 strings. And prototext.Unmarshal seems to be expecting UTF8 strings.

Good find!  Thanks for the fix.

> I believe we could also remove this line in vet.sh:

I'm fine with that, but if we do that we should replace it with something that looks for `"github.com/golang/protobuf"` and forbids it anywhere but in `reflection/grpc_testing_not_regenerate`.

channelz/service/service_test.go
```diff
@@ -27,14 +27,18 @@ import (
...
 	"github.com/google/go-cmp/cmp"
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/internal/channelz"
 	"google.golang.org/grpc/internal/grpctest"
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/protoadapt"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+	"google.golang.org/protobuf/runtime/protoimpl"
```
dfawley, line 39: > WARNING: This package should only ever be imported by generated messages. The compatibility agreement covers nothing except for functionality needed to keep existing generated messages operational. Breakages that occur due to unauthorized usages of this package are not the author's responsibility.

I'd rather not use this package if it's at all possible to avoid it, unless you are extremely sure backward compatibility will be guaranteed for all our uses of it.

Clement-Jean: yes, I agree that using protoimpl is half a solution. I'm figuring out how to imrpove this.

Clement-Jean:
@dfawley I found other references to github.com/golang/protobuf and I changed the vet.sh. Could you check?

The GetSocket test seems to fail. Not sure why yet.

Clement-Jean:
@dfawley It was because protobuf does not ensure encoding order of fields. Going with a generated proto message is easier.

channelz/service/service_test.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package grpc.credentials;
+
+option go_package = "google.golang.org/grpc/channelz/service";
```
dfawley, line 5: Can this go into `service_test`?  If not, please put it in a subdirectory so it doesn't get pulled into the production build.  (Even if it is eventually pruned out (?), it still feels wrong.)

credentials/credentials.go
```diff
@@ -287,5 +287,5 @@ type ChannelzSecurityValue interface {
 type OtherChannelzSecurityValue struct {
 	ChannelzSecurityValue
 	Name  string
-	Value protoadapt.MessageV1
```
dfawley, line 290: This may be a breaking change.  It is probably fine since it's experimental, and it's unlikely anyone is using it....so I think I'm OK with it.  Just wanted to call it out.

dfawley (changes requested):
This looks great to me.  Just one change to request.  Thanks!

channelz/service/service_test.proto
Clement-Jean: Yeah, that's why I wanted to keep it as a descriptor in the test file at first... I think I will move the info into the test file like before.

dfawley (approved):
Thanks!

arvindbr8 (approved):
LGTM. Thanks @Clement-Jean for making huge strides in fixing https://github.com/grpc/grpc-go/issues/5316. 

---

[gRPConf 2024 Registration](https://twitter.com/grpcio/status/1775676303288783254) is now open! Please find more details at grpc.io. Hope to see you there.

PR #7085, opened 2024-04-03
xds/internal/xdsclient: Process string metadata in CDS for com.google.csm.telemetry_labels

zasweq:
This PR adds processing in CDS for string telemetry labels with filter_metadata string "com.google.csm.telemetry_labels". Other value types and filter_metadata identifiers are ignored. This also does not introduce NACK's. This is based off the C-Core implementation, which I disagree with technically but understand there were good reasons for the decisions made. See discussion in xDS Chat room for context as to why C-Core did it this way and why I followed it.

The relevant language in the CSM Observability Design is "The above key_value pairs from `com.google.csm.telemetry_labels` are going to be parsed by xDS Client from the CDS Resource and propagated to the watcher on this resource in the “cds” load balancing policy."

RELEASE NOTES: N/A

### DIFF
--- a/xds/internal/xdsclient/xdsresource/type_cds.go
+++ b/xds/internal/xdsclient/xdsresource/type_cds.go
@@ -85,7 +85,8 @@ type ClusterUpdate struct {
 
 	// Raw is the resource from the xds response.
 	Raw *anypb.Any
-	// All the string valued metadata of filter_metadata type
-	// "com.google.csm.telemetry_labels".
-	StringMD map[string]string
+	// TelemetryLabels are the string valued metadata of filter_metadata type
+	// "com.google.csm.telemetry_labels" with keys "service_name" or
+	// "service_namespace".
+	TelemetryLabels map[string]string
 }
--- a/xds/internal/xdsclient/xdsresource/type_cds.go
+++ b/xds/internal/xdsclient/xdsresource/type_cds.go
@@ -85,4 +85,7 @@ type ClusterUpdate struct {
 
 	// Raw is the resource from the xds response.
 	Raw *anypb.Any
+	// All the string valued metadata of filter_metadata type
+	// "com.google.csm.telemetry_labels".
+	StringMD map[string]string
 }
--- a/xds/internal/xdsclient/xdsresource/unmarshal_cds.go
+++ b/xds/internal/xdsclient/xdsresource/unmarshal_cds.go
@@ -85,9 +85,20 @@ func validateClusterAndConstructClusterUpdate(cluster *v3clusterpb.Cluster) (Clu
 	stringMD := make(map[string]string)
 	if fmd := cluster.GetMetadata().GetFilterMetadata(); fmd != nil {
 		if val, ok := fmd["com.google.csm.telemetry_labels"]; ok {
-			for key, val := range val.GetFields() {
-				if _, ok := val.GetKind().(*structpb.Value_StringValue); ok {
-					stringMD[key] = val.GetStringValue()
+			/*
+				"service_name" = "",
+				"service_namespace" = ""
+			*/
+			if fields := val.GetFields(); fields != nil {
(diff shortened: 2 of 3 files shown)


xds/internal/xdsclient/xdsresource/type_cds.go
```diff
@@ -85,4 +85,7 @@ type ClusterUpdate struct {
 
 	// Raw is the resource from the xds response.
 	Raw *anypb.Any
+	// All the string valued metadata of filter_metadata type
+	// "com.google.csm.telemetry_labels".
+	StringMD map[string]string
```
dfawley, line 90: Bikeshedding: How about `TelemetryLabels`?

xds/internal/xdsclient/xdsresource/unmarshal_cds.go
```diff
@@ -81,6 +82,28 @@ const (
 )
 
 func validateClusterAndConstructClusterUpdate(cluster *v3clusterpb.Cluster) (ClusterUpdate, error) {
+	stringMD := make(map[string]string)
+	if fmd := cluster.GetMetadata().GetFilterMetadata(); fmd != nil {
+		if val, ok := fmd["com.google.csm.telemetry_labels"]; ok {
+			/*
+				"service_name" = "",
+				"service_namespace" = ""
+			*/
```
dfawley, line 91: This comment looks out of place.  Are you documenting JSON schema or.......?

xds/internal/xdsclient/xdsresource/unmarshal_cds.go
```diff
@@ -81,6 +82,28 @@ const (
...
+			if fields := val.GetFields(); fields != nil {
+				if val, ok := fields["service_name"]; ok {
+					if _, isStringVal := val.GetKind().(*structpb.Value_StringValue); isStringVal {
```
dfawley, line 94: s/isStringVal/ok/ .. `ok` is the de-facto standard for type assertion checks unless you're doing something especially complex.  Yes, it's nested, but that's fine since you never need the outer `ok` again.

xds/internal/xdsclient/xdsresource/unmarshal_cds.go
```diff
@@ -81,6 +82,28 @@ const (
...
+						stringMD["service_name"] = val.GetStringValue()
```
dfawley, line 95: This feels weird.  We check the type of a thing, then call `GetStringValue` on some _other_ thing.  Does `GetStringValue` do something different if the `Kind()` is not `StringValue`?

xds/internal/xdsclient/xdsresource/type_cds.go
zasweq: Sure. That's what the scaled up LB Policy configs use as the field name, so that makes sense.

xds/internal/xdsclient/xdsresource/unmarshal_cds.go
zasweq: Deleted.

xds/internal/xdsclient/xdsresource/unmarshal_cds.go
zasweq: Yeah, this is the best solution that I could come up with. GetStringValue() returns a string, not a *string, so there is no way to distinguish between an empty string (a valid label value, which will be recorded) and an unset string (should eventually record method labels as "unknown" if unset). Thus, I don't set if not a string value (not set = "unknown" in OpenTelemetry component) and set it even if it's empty if it's present.

dfawley: I _believe_ you can do:

```go
if s, ok := val.AsInterface().(string); ok {
	stringMD["service_name"] = s
}
```

But I'm not sure that that's much better.

dfawley (approved):

Issue #1234, opened 2017-05-12
Get an unexpected "context canneled" when server stream.Recv()


deafwolf:
I want my grpc server to know when client disconnected. So I use stream.Recv to solve it.
But I get a "rpc error: code = Canceled desc = context canceled".
I found #1061 but I'm not sure if they are the same problem.

Here is the demo to reproduce this issue
https://github.com/deafwolf/grpc-stream

My enviroment
go version go1.7.1 darwin/amd64
grpc and golang.org/x/net/context is up-to-date

menghanl:
Server indicates the RPC is done by returning from the service handler. When the service handler returns, gRPC cancels this stream because the RPC is done.

So in your handler for RPC `Start`, you need to either call `Recv()` in the same goroutine, or add some synchronization in the service handler to make sure the handler doesn't return until you are done with this RPC.

deafwolf:
@menghanl Thank you very much.

I remove the goroutine in server.Start, it works fine. 

I just need an error when client disconnected, I don't care the real reason such as EOF/canceled. I don't know the Start RPC is done, so I think the client cancel the RPC and server receive an "unexpected" error.

PR #7094, opened 2024-04-04
channelz: ensure `SubChannel` is always registered with parent

daniel-weisse:
`channelz.SubChannel` may be missing parent information if created using `channelz.RegisterSubChannel`.
This is because `channelz.RegisterSubChannel` does not take a `channelz.Channel` as argument, but just the ID of this channel.
The function itself then tries to load the Channel from a package scoped `channelMap`.
This map does not contain any entries if channelz data collection is turned off, resulting in the SubChannel being created without a parent.
Which in turn results in the SubChannel's string representation being logged as `<nil> SubChannel #1234` (1234 being some ID of the SubChannel).

This PR fixes this by restoring the behavior of `v1.62.2`, which instead of passing just the ID of a SubChannel's parent, passed a reference to the parent Channel, returning an error if it is `nil`: https://github.com/grpc/grpc-go/blob/10baa6bd42af1c0fc4d65e58bddf1b3f3c6a5aa8/internal/channelz/funcs.go#L199-L206

RELEASE NOTES: none

### DIFF
--- a/internal/channelz/funcs.go
+++ b/internal/channelz/funcs.go
@@ -22,6 +22,7 @@
 package channelz
 
 import (
+	"errors"
 	"sync/atomic"
 	"time"
 
@@ -143,21 +144,29 @@ func RegisterChannel(parent *Channel, target string) *Channel {
 // Returns a unique channelz identifier assigned to this subChannel.
 //
 // If channelz is not turned ON, the channelz database is not mutated.
-func RegisterSubChannel(pid int64, ref string) *SubChannel {
+func RegisterSubChannel(pid *Channel, ref string) (*SubChannel, error) {
+	if pid == nil {
+		return nil, errors.New("a SubChannel's parent id cannot be nil")
+	}
+
 	id := IDGen.genID()
 	if !IsOn() {
-		return &SubChannel{ID: id}
+		return &SubChannel{
+			RefName: ref,
+			parent:  pid,
+			ID:      id,
+		}, nil
 	}
 
 	sc := &SubChannel{
 		RefName: ref,
 		ID:      id,
 		sockets: make(map[int64]string),
-		parent:  db.getChannel(pid),
+		parent:  pid,
 		trace:   &ChannelTrace{CreationTime: time.Now(), Events: make([]*traceEvent, 0, getMaxTraceEntry())},
 	}
-	db.addSubChannel(id, sc, pid)
-	return sc
+	db.addSubChannel(id, sc, pid.ID)
+	return sc, nil
 }
 
 // RegisterServer registers the given server s in channelz database. It returns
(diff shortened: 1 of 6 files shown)


dfawley:
@daniel-weisse do you know if this is a regression with the latest release?  If so we may prefer to restore the previous behavior instead.

daniel-weisse:
The Channel and SubChannel types were introduced in `v1.63`.
v1.62 used the [`Identifier`](https://github.com/grpc/grpc-go/blob/10baa6bd42af1c0fc4d65e58bddf1b3f3c6a5aa8/internal/channelz/id.go#L25) struct, which served the same purpose.

From a brief look at the v1.62.2 code, it looks like parent information was always omitted from its string representation if there was no parent: https://github.com/grpc/grpc-go/blob/10baa6bd42af1c0fc4d65e58bddf1b3f3c6a5aa8/internal/channelz/id.go#L69-L75

https://github.com/grpc/grpc-go/blob/10baa6bd42af1c0fc4d65e58bddf1b3f3c6a5aa8/internal/channelz/id.go#L48-L50

dfawley:
> One fix would be adjusting the signature of the RegisterSubChannel function to take the SubChannel's parent *Channel as an argument, instead of just its pid.

This sounds good.  Subchannels should always have parents, so I was skeptical about the initial change proposal -- we should ideally be logging both the channel+subchannel IDs in all the subchannel events.

daniel-weisse:
Updated the PR to restore the `RegisterSubChannel` function to the behavior of `v1.62.2`.
This should ensure SubChannels are always created with a valid parent, so this information is never logged as `nil`

internal/channelz/funcs.go
```diff
@@ -143,21 +144,29 @@ func RegisterChannel(parent *Channel, target string) *Channel {
 // Returns a unique channelz identifier assigned to this subChannel.
 //
 // If channelz is not turned ON, the channelz database is not mutated.
-func RegisterSubChannel(pid int64, ref string) *SubChannel {
+func RegisterSubChannel(pid *Channel, ref string) (*SubChannel, error) {
+	if pid == nil {
+		return nil, errors.New("a SubChannel's parent id cannot be nil")
+	}
```
dfawley, line 150: Let's not return an error here since it complicates the code so much, and it should just be an invariant of the code like any other dereference.  Just panic and we'll make sure `RegisterSubChannel` is always called properly.

Actually, this change is pretty important to get done quickly, so I created #7101 and we will be doing a patch release with it ASAP.  Thank you for the fix!

dfawley:
As mentioned in the review comment, we'll be continuing this in #7101.  Sorry for that, but it needs to get done very quickly.  Thank you!

Issue #7095, opened 2024-04-04
Deprecate WithBlock and WithReturnConnectionError


pellared:
### Use case(s) - what problem will this feature solve?

https://github.com/grpc/grpc-go/pull/7029 has deprecated `Dial` and `DialContext`. 
`NewClient` ignores `WithBlock`, `WithTimeout`, `WithReturnConnectionError`.

`grpc.WithTimeout` is already deprecated.

`WithBlock` and `WithReturnConnectionError` have a GoDoc comment which discourages its usage.

### Proposed Solution

Deprecate `WithBlock` and `WithReturnConnectionError` so to better notify that these options should not be used and additionally that they are ignored by `NewClient`.

Deprecation is more visible in https://pkg.go.dev/google.golang.org/grpc and more usable for static analysis tools such as `golangci-lint`.

dfawley:
Thank you for creating this issue.  I had not realized these functions were not already tagged as "deprecated" properly.  Let's do that.

pellared:
@dfawley I think that `FailOnNonTempDialError` can also be deprecated. Can you please double-check that?

dfawley:
> @dfawley I think that FailOnNonTempDialError can also be deprecated. Can you please double-check that?

Yes, that sounds right to me, thanks!

PR #7097, opened 2024-04-04
grpc: Deprecate WithBlock, WithReturnConnectionError, FailOnNonTempDialError

pellared:
Fixes https://github.com/grpc/grpc-go/issues/7095

Notice in `NewClient` GoDoc comment that `FailOnNonTempDialError` dial option is ignored.

Deprecate options that are ignored by `NewClient` function.
These options are only supported by deprecated `Dial` and `DialContext` functions.

RELEASE NOTES:
- grpc: Deprecate WithBlock, WithReturnConnectionError, FailOnNonTempDialError which are ignored by NewClient

### DIFF
--- a/dialoptions.go
+++ b/dialoptions.go
@@ -317,12 +317,8 @@ func WithBlock() DialOption {
 // Use of this feature is not recommended.  For more information, please see:
 // https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md
 //
-// # Experimental
-//
-// Notice: This API is EXPERIMENTAL and may be changed or removed in a
-// later release.
-//
 // Deprecated: this DialOption is not supported by NewClient.
+// Will be supported throughout 1.x.
 func WithReturnConnectionError() DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
 		o.block = true
--- a/dialoptions.go
+++ b/dialoptions.go
@@ -471,12 +471,9 @@ func withBinaryLogger(bl binarylog.Logger) DialOption {
 // Use of this feature is not recommended.  For more information, please see:
 // https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md
 //
-// # Experimental
-//
-// Notice: This API is EXPERIMENTAL and may be changed or removed in a
-// later release.
-//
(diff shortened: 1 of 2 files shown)


pellared:
I guess [Type: Documentation](https://github.com/grpc/grpc-go/labels/Type%3A%20Documentation) label can be added.

dialoptions.go
```diff
@@ -318,6 +321,8 @@ func WithBlock() DialOption {
 //
 // Notice: This API is EXPERIMENTAL and may be changed or removed in a
 // later release.
```
dfawley, line 323: Could you please remove this experimental warning and we'll just call it supported now ("throughout 1.x")?

dialoptions.go
```diff
@@ -474,6 +479,8 @@ func withBinaryLogger(bl binarylog.Logger) DialOption {
...
```
dfawley, line 481: For this one, I'd say we should delete "experimental", but we probably want to reserve the right to remove it.  IIRC this makes its way through multiple layers of our library and might be more painful to maintain long-term.

dfawley (approved):
Thank you for the PR.  Two small requests, otherwise LGTM!

dialoptions.go
pellared: PTAL https://github.com/grpc/grpc-go/pull/7097/commits/cce30008634fa201fb6d8500b9c12f6b3c806d04

dialoptions.go
pellared: PTAL https://github.com/grpc/grpc-go/pull/7097/commits/cbdf3519fb67fb293b164642a2068b1502ca0451

dfawley (approved):
Looks great, thank you!

pellared:
Someone can fix vet failure for me: https://github.com/grpc/grpc-go/actions/runs/8573454634/job/23498177256?pr=7097

Otherwise I will fix it on Monday

dfawley:
> Someone can fix vet failure for me

This is an optional checker, and the current breakage is caused by an upstream change.  No need to do anything for this PR.

arvindbr8 (approved):
Looks great! Thanks for the PR @pellared

PR #7099, opened 2024-04-05
vet: split vet-proto from vet.sh

arvindbr8:
Move all the vet-proto logic to its own file

RELEASE NOTES: none


arvindbr8:
Discussed offline with @dfawley and decided to get rid of the env-variables fully, and just split to the logic to another bash file.

scripts/util.sh
dfawley, line 1: As mentioned offline: add pre-check and cleanup and name ~`vet-common.sh`

scripts/vet-proto.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash 
+
+set -ex  # Exit on error; debugging enabled.
+set -o pipefail  # Fail a pipe if any sub-command fails.
+
+source "$(dirname $0)/util.sh" || exit 1
```
dfawley, line 6: As discussed `|| exit` is not needed because of `-e`

dfawley (approved):

PR #7118, opened 2024-04-10
advancedtls: unexport parts of API not meant to be public

gtcooke94:
ServerNameOverride is test-only
RevocationStatus, CheckRevocation, and CheckChainRevocation are not used externally and should not be exported

RELEASE NOTES: none

### DIFF
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -172,10 +172,6 @@ type ClientOptions struct {
 	// If this is set, we will perform this customized check after doing the
 	// normal check(s) indicated by setting VType.
 	VerifyPeer CustomVerificationFunc
-	// ServerNameOverride is for testing only. If set to a non-empty string,
-	// it will override the virtual host name of authority (e.g. :authority
-	// header field) in requests.
-	ServerNameOverride string
 	// RootOptions is OPTIONAL on client side. If not set, we will try to use the
 	// default trust certificates in users' OS system.
 	RootOptions RootCertificateOptions
@@ -193,6 +189,10 @@ type ClientOptions struct {
 	// By default, the maximum version supported by this package is used,
 	// which is currently TLS 1.3.
 	MaxVersion uint16
+	// serverNameOverride is for testing only. If set to a non-empty string,
+	// it will override the virtual host name of authority (e.g. :authority
+	// header field) in requests.
+	serverNameOverride string
 }
 
 // ServerOptions contains the fields needed to be filled by the server.
@@ -244,7 +244,7 @@ func (o *ClientOptions) config() (*tls.Config, error) {
 		return nil, fmt.Errorf("the minimum TLS version is larger than the maximum TLS version")
 	}
 	config := &tls.Config{
-		ServerName: o.ServerNameOverride,
+		ServerName: o.serverNameOverride,
 		// We have to set InsecureSkipVerify to true to skip the default checks and
 		// use the verification function we built from buildVerifyFunc.
 		InsecureSkipVerify: true,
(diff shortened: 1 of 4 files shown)


security/advancedtls/advancedtls.go
```diff
@@ -193,6 +189,10 @@ type ClientOptions struct {
 	// By default, the maximum version supported by this package is used,
 	// which is currently TLS 1.3.
 	MaxVersion uint16
+	// serverNameOverride is for testing only. If set to a non-empty string,
+	// it will override the virtual host name of authority (e.g. :authority
```
matthewstevenson88, line 193: Also the SNI and the target hostname that we used during server cert verification. It might be worth commenting on these as well. :)

matthewstevenson88 (approved):

security/advancedtls/crl.go
```diff
@@ -192,25 +188,25 @@ func x509NameHash(r pkix.RDNSequence) string {
 	return fmt.Sprintf("%08x", fileHash)
 }
 
-// CheckRevocation checks the connection for revoked certificates based on RFC5280.
+// checkRevocation checks the connection for revoked certificates based on RFC5280.
```
erm-g, line 191: I don't see any usages of it inside the codebase, but it looks like it can be used as an embedded part of tls.Config. In case of OSS users use it that way, would it make sense just to mark it as deprecated (in favor of new CRL solution) and remove later?

security/advancedtls/crl.go
```diff
@@ -192,25 +188,25 @@ func x509NameHash(r pkix.RDNSequence) string {
...
-	return CheckChainRevocation(conn.VerifiedChains, cfg)
+func checkRevocation(conn tls.ConnectionState, cfg RevocationConfig) error {
+	return checkChainRevocation(conn.VerifiedChains, cfg)
 }
 
-// CheckChainRevocation checks the verified certificate chain
+// checkChainRevocation checks the verified certificate chain
 // for revoked certificates based on RFC5280.
-func CheckChainRevocation(verifiedChains [][]*x509.Certificate, cfg RevocationConfig) error {
+func checkChainRevocation(verifiedChains [][]*x509.Certificate, cfg RevocationConfig) error {
```
erm-g, line 205: Thinking about the target state - aren't we going to provide some interface that will allow users to define such functions? If yes, I think it's better to keep this one public, or maybe provide a builder (in this case private is fine).

security/advancedtls/crl.go
gtcooke94: Can you link what you mean? I didn't see any usage in grpc-go or in google

security/advancedtls/crl.go
gtcooke94: I don't think so - the interface is the CRLProvider, so the user is just responsible for giving us CRLs. We don't provide any overrides for the actual revocation check

security/advancedtls/crl.go
erm-g: https://github.com/grpc/grpc-go/blob/0baa668e3dcc443261c266d8914d3321d20d2eaa/security/advancedtls/crl_test.go#L739

gtcooke94: For me, I think we should unexport - we don't want anyone depending on this. If someone is using it in this way currently, they can stay on a version of the advancedTLS package that has it until they move off.
It's not really something to deprecate, rather it's an internal API that was written with a capital letter

@dfawley what do you think here?

security/advancedtls/advancedtls.go
gtcooke94: To make sure I'm clear, you're saying to comment something like "it will override the virtual host name of authority... and the target hostname used during server cert verification"

matthewstevenson88: Correct.

gtcooke94: Added the comment

dfawley (approved):

PR #7121, opened 2024-04-11
stats: mark InPayload.Data and OutPayload.Data for deletion (experimental)

dfawley:
cc @papacharlie

RELEASE NOTES:
* stats: deprecate `InPayload.Data` and `OutPayload.Data`; they were experimental and will be deleted in the next release


dfawley:
> I think it should also probably be explicitly documented that the Payload field can also be modified after HandleRPC returns since it will be handed over to the user code. If the handler needs access to it after the fact, it needs to be copied.

Done!

zasweq (approved):

PR #7127, opened 2024-04-15
Move revocation to subpackage

gtcooke94:
This PR moves revocation related code to a subpackage in advancedtls.
This required a large shift of code to a new internal revocation package as well. The internal package now exports `CheckChainRevocation` which is called in advancedtls.go, but is not intended to be exposed publicly.

I apologize in advanced if the diff is difficult - I had to refactor everything that is called in the `CheckChainRevocation` call tree to `internal` which didn't result in a super clean diff.

RELEASE NOTES: none


matthewstevenson88:
Deferring this review to @erm-g.

PR #7130, opened 2024-04-16
xds/internal/xdsclient/xdslbregistry: remove unused call to type url

marcoferrer:
I was reading through the implementation for custom lb polices and noticed a call to the policy type URL wasn't being used

RELEASE NOTES: n/a


aranjans:
@marcoferrer can you please add something like below as part of PR description?
Screenshot 2024-04-17 at 1 41 34 PM

dfawley:
> can you please add something like below as part of PR description?

FYI for non-user-visible changes like this, we would use `RELEASE NOTES: n/a` (or `none` also works).

zasweq (approved):

PR #7132, opened 2024-04-16
client: Drop two calls to `pretty.ToJSON` and move code outside of lock

townba:
RELEASE NOTES:
* client: improve RPC performance by reducing work while holding a lock

### DIFF
--- a/clientconn.go
+++ b/clientconn.go
@@ -37,7 +37,6 @@ import (
 	"google.golang.org/grpc/internal/channelz"
 	"google.golang.org/grpc/internal/grpcsync"
 	"google.golang.org/grpc/internal/idle"
-	"google.golang.org/grpc/internal/pretty"
 	iresolver "google.golang.org/grpc/internal/resolver"
 	"google.golang.org/grpc/internal/transport"
 	"google.golang.org/grpc/keepalive"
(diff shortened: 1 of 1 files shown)


clientconn.go
```diff
@@ -934,10 +933,14 @@ func equalAddresses(a, b []resolver.Address) bool {
...
 func (ac *addrConn) updateAddrs(addrs []resolver.Address) {
-	ac.mu.Lock()
-	channelz.Infof(logger, ac.channelz, "addrConn: updateAddrs curAddr: %v, addrs: %v", pretty.ToJSON(ac.curAddr), pretty.ToJSON(addrs))
-
 	addrs = copyAddressesWithoutBalancerAttributes(addrs)
+	limit := len(addrs)
+	if limit > 5 {
+		limit = 5
+	}
+	channelz.Infof(logger, ac.channelz, "addrConn: updateAddrs addrs (%d of %d): %v", limit, len(addrs), addrs[:limit])
```
dfawley, line 941: Can you show what this looks like, please?  (Mainly I'm not sure: does `%v` on a slice call `String()` for each member?)

dfawley (commented):
Could you please add a comment to `addrConn`'s `mu` field that says something like "`This is taken on the RPC path, so its usage should be minimized as much as possible.  TODO: find a lock-free way to retrieve the transport and state from the addrConn without a lock`"

clientconn.go
townba: Looks something like this:
```
addrConn: updateAddrs addrs (3 of 3): [{Addr: "127.0.0.1:35771", ServerName: "", } {Addr: "127.0.0.1:41729", ServerName: "", } {Addr: "127.0.0.1:36587", ServerName: "", }]
```

townba:
> Could you please add a comment to `addrConn`'s `mu` field that says something like …

Done.

townba:
PTAL

clientconn.go
dfawley: Great, thanks!

clientconn.go
```diff
@@ -1172,6 +1175,10 @@ type addrConn struct {
 	// is received, transport is closed, ac has been torn down).
 	transport transport.ClientTransport // The current transport.
 
+	// This mutex is used on the RPC path, so its usage should be minimized as
+	// much as possible.
+	// TODO: Find a lock-free way to retrieve the transport and state from the
+	// addrConn without a lock.
```
dfawley, line 1181: Oops, "lock-free ... without a look" - please remove one or the other.  Thanks!

dfawley (approved):

clientconn.go
townba: Done. PTAL

dfawley (approved):

PR #7139, opened 2024-04-17
*: fix regenerate.sh

arvindbr8:
`vet.sh` was failing in macOS. The logger defined in channelz/internal/protoconv package was not used when build on non_linux machines, which caused staticcheck to output this: 

```
channelz/internal/protoconv/util.go:27:5: var logger is unused (U1000)
```

also fixes the go_package for grpc_testing_not_regenerate proto

RELEASE NOTES: none

### DIFF
--- a/channelz/internal/protoconv/sockopt_nonlinux.go
+++ b/channelz/internal/protoconv/sockopt_nonlinux.go
@@ -26,6 +26,7 @@ import (
 	"google.golang.org/grpc/internal/channelz"
 )
 
-func sockoptToProto(skopts *channelz.SocketOptionData) []*channelzpb.SocketOption {
+func sockoptToProto(_ *channelz.SocketOptionData) []*channelzpb.SocketOption {
+	logger.Info("sockoptToProto is not implemented on non-linux platforms")
 	return nil
 }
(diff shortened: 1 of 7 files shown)


scripts/vet.sh
```diff
@@ -119,7 +119,6 @@ grep "(SA1019)" "${SC_OUT}" | not grep -Fv 'XXXXX PleaseIgnoreUnused
 XXXXX Protobuf related deprecation errors:
 "github.com/golang/protobuf
 .pb.go:
-grpc_testing_not_regenerate
```
dfawley, line 122: Staticcheck is staying within the module?

Ugh, should we be running it for every `go.mod` then?

arvindbr8: That's right. Let's track this here https://github.com/grpc/grpc-go/issues/7141

channelz/internal/protoconv/sockopt_nonlinux.go
```diff
@@ -26,6 +26,6 @@ import (
 	"google.golang.org/grpc/internal/channelz"
 )
 
-func sockoptToProto(skopts *channelz.SocketOptionData) []*channelzpb.SocketOption {
+func sockoptToProto(_ *channelz.SocketOptionData) []*channelzpb.SocketOption {
```
dfawley, line 29: No need to change this here but there is even an even more simplified version:
```go
func sockoptToProto(*channelz.SocketOptionData) []*channelzpb.SocketOption {
```

dfawley (approved):

PR #7140, opened 2024-04-17
advancedtls: Rename custom verification function APIs

gtcooke94:
This change renames the APIs for injecting a custom verification function such that the names are clearer that it is occurs post-handshake with already built chains and does not replace the base chain building and chain verification that the underlying security library does.

I've left type aliases to the old names and marked them as deprecated.

(To add to release notes? -  In advancedtls, change `CustomVerificationFunc` to `PostHandshakeVerificationFunc`, `VerificationFuncParams` to `HandshakeVerificationInfo`, and `VerificationResults` to `PostHandshakeVerificationResults`.)

RELEASE NOTES: none


security/advancedtls/advancedtls.go
```diff
@@ -35,10 +35,10 @@ import (
 	credinternal "google.golang.org/grpc/internal/credentials"
 )
 
-// VerificationFuncParams contains parameters available to users when
-// implementing CustomVerificationFunc.
+// HandshakeVerificationInfo contains parameters available to users when
+// implementing PostHandshakeVerificationFunc.
 // The fields in this struct are read-only.
-type VerificationFuncParams struct {
+type HandshakeVerificationInfo struct {
```
erm-g, line 41: I'm a little bit confused by naming - there are entries of 'xInfo' and 'xOptions'. It looks like that 'xInfo' is more about 'general purpose container' and 'xOptions' is more about 'container for func params' but I'd suggest double checking with Doug.

security/advancedtls/examples/credential_reloading_from_files/client/main.go
```diff
@@ -76,8 +76,8 @@ func main() {
 		IdentityOptions: advancedtls.IdentityCertificateOptions{
 			IdentityProvider: identityProvider,
 		},
-		VerifyPeer: func(params *advancedtls.VerificationFuncParams) (*advancedtls.VerificationResults, error) {
-			return &advancedtls.VerificationResults{}, nil
+		VerifyPeer: func(params *advancedtls.HandshakeVerificationInfo) (*advancedtls.PostHandshakeVerificationResults, error) {
```
erm-g, line 79: Looking at both 'main.go' examples it looks like 'xOptions' is used more here, so I'm leaning towards 'HandshakeVerificationOptions'

erm-g (changes requested):

security/advancedtls/advancedtls.go
gtcooke94: I disagree with `xOptions` for this case - `xOptions` is for users to create configuration in our APIs.
This struct represents the input to a function that users can override - it's not `Options` or anything that is configured. Rather, the struct contains and represents information about the handshake verification steps that have already been taken, such as the raw certs, the built chains, the server name

security/advancedtls/examples/credential_reloading_from_files/client/main.go
gtcooke94: See other comment

security/advancedtls/advancedtls.go
```diff
@@ -35,10 +35,10 @@ import (
 	credinternal "google.golang.org/grpc/internal/credentials"
 )
 
-// VerificationFuncParams contains parameters available to users when
-// implementing CustomVerificationFunc.
+// HandshakeVerificationInfo contains parameters available to users when
+// implementing PostHandshakeVerificationFunc.
```
dfawley, line 39: Similarly, should this say "contains information about a handshake needed for verification" or something?

security/advancedtls/advancedtls.go
```diff
@@ -54,17 +54,34 @@ type VerificationFuncParams struct {
 	Leaf *x509.Certificate
 }
 
-// VerificationResults contains the information about results of
-// CustomVerificationFunc.
-// VerificationResults is an empty struct for now. It may be extended in the
+// DEPRECATED: Renamed to `HandshakeVerificationInfo`
```
dfawley, line 57: See other PR re: how Go writes "deprecated" docstrings.  This ties in with tooling in godoc and static checkers, so isn't just opinion.

security/advancedtls/advancedtls.go
```diff
@@ -54,17 +54,34 @@ type VerificationFuncParams struct {
...
+type PostHandshakeVerificationResults struct{}
+
+// DEPRECATED: Renamed to `PostHandshakeVerificationResults`
+type VerificationResults = PostHandshakeVerificationResults
+
+// PostHandshakeVerificationFunc is the function defined by users to perform
+// custom verification checks after chain building and regular handshake
+// verification has been completed.
+// PostHandshakeVerificationFunc returns nil
+// if the authorization fails; otherwise returns an empty struct.
```
dfawley, line 76: I don't understand this.

The convention is that functions reutrn `<zero value AKA nil here>, error` or error and `_, nil` on success.  If `nil, nil` is illegal, then that makes the API confusing and hard to use -- in which case you should have it return a value instead of a pointer.

security/advancedtls/advancedtls.go
```diff
@@ -54,17 +54,34 @@ type VerificationFuncParams struct {
...
+type PostHandshakeVerificationFunc func(params *HandshakeVerificationInfo) (*PostHandshakeVerificationResults, error)
```
dfawley, line 77: Suggestion: `PeerVerificationFunc` (and all associated types)?  It's a shorter name and (maybe?) is named after the intent of the thing vs. when it happens.

security/advancedtls/advancedtls.go
gtcooke94: Changed

security/advancedtls/advancedtls.go
gtcooke94: I think `Info` here is the best in this case

security/advancedtls/advancedtls.go
gtcooke94: Fixed

security/advancedtls/advancedtls.go
gtcooke94: We want to make sure that we are differentiating between verification happening after the normal chain building and verification that happens by default (which this is), and overriding the base chain building and verification itself.

security/advancedtls/advancedtls.go
gtcooke94: It returns `nil, error` in the case the authorization fails, `<currently empty struct>, nil` on success

We discussed and don't think there is harm in keeping the struct even though it's currently unused - there's a good chance it could be in the future and it's already implemented as such

security/advancedtls/advancedtls.go
dfawley: I'm not sure I follow all this.

What is "post handshake verification" exactly?  And you're saying this can be used to _replace_ the default behavior?  What is the default behavior?  Do you have any examples of what users would want to do with this?

security/advancedtls/advancedtls.go
dfawley: OK - it's a little unusual but should be OK.  If that's what you want you should make sure you're panicking if you call it and the function ever returns `nil, nil`.

security/advancedtls/advancedtls.go
dfawley: What's the API for (2)?  `VerifyPeer`?

So is it the case that if you specify this callback then `VerifyPeer` isn't called?

security/advancedtls/advancedtls.go
gtcooke94: I don't think there's a compelling reason to return a pointer vs. a struct if that is what is making this unclear, but that would represent a breaking behavior change. Looking at it, I don't think `nil, nil` is illegal, it just means there's no information to pass back out.

I added some text to the comment to make it a little more clear.

That being said, looking at it again today with fresh eyes, it's called [here](https://github.com/grpc/grpc-go/blob/34de5cf4832ffa43dec384ddb20c91ca0805954f/security/advancedtls/advancedtls.go#L581-L582) in our code. And the struct is always thrown away. So I'm rethinking the value of keeping this in the API. Will discuss with my team and add another comment soon.

dfawley: It says "otherwise returns an empty struct" which implies that `nil, nil` is invalid.  If `nil, nil` is valid, then you can just delete this part of the documentation.  Returning `<zero>, error` or `_, nil` is just how regular Go conventions always work.

gtcooke94: Ah I see, I'll remove that part and then this is resolved? I think that it's more pain that its worth to remove the struct return

gtcooke94: PTAL

security/advancedtls/advancedtls.go
gtcooke94: Renamed `VerifyPeer` in the settings struct, PTAL

dfawley:
Can you please resolve the conflicts as well?  Thanks!

gtcooke94:
Fixed merge conflicts, should be all good

security/advancedtls/advancedtls.go
```diff
@@ -186,10 +186,17 @@ type ClientOptions struct {
 	// IdentityOptions is OPTIONAL on client side. This field only needs to be
 	// set if mutual authentication is required on server side.
 	IdentityOptions IdentityCertificateOptions
+	// AdditionalPeerVerification is a custom verification check after certificate signature
+	// check.
+	// If this is set, we will perform this customized check after doing the
+	// normal check(s) indicated by setting VType.
```
dfawley, line 192: VType->VerificationType?

security/advancedtls/advancedtls.go
```diff
@@ -225,10 +232,17 @@ type ClientOptions struct {
 type ServerOptions struct {
 	// IdentityOptions is REQUIRED on server side.
 	IdentityOptions IdentityCertificateOptions
+	// AdditionalPeerVerification is a custom verification check after certificate signature
+	// check.
+	// If this is set, we will perform this customized check after doing the
+	// normal check(s) indicated by setting VType.
```
dfawley, line 238: VerificationType

security/advancedtls/advancedtls.go
```diff
@@ -258,13 +272,18 @@ type ServerOptions struct {
 }
 
 func (o *ClientOptions) config() (*tls.Config, error) {
+	// TODO(gtcooke94) Remove this block when remove o.VerifyPeer
+	// Set AdditionalPeerVerification if the user is still using VerifyPeer.
+	if o.VerifyPeer != nil {
+		o.AdditionalPeerVerification = o.VerifyPeer
```
dfawley, line 278: `if o.AdditionalPeerVerification != nil { return <err> }`?

Or ignore the old deprecated field `VerifyPeer` instead by reversing it:

```go
if o.AdditionalPeerVerification == nil {
	o.AdditionalPeerVerification = o.VerifyPeer
}
```

security/advancedtls/advancedtls.go
gtcooke94: Ah I think I lost some in the merge, good catch

security/advancedtls/advancedtls.go
gtcooke94: Good catch

security/advancedtls/advancedtls.go
gtcooke94: It's not required, so the error return I think doesn't make the most sense.

The reverse works too, I guess it just matters for the precedence - I was going with "if the old field is set they probably haven't migrated", but the bottom is good too as a "if the new field isn't set take whatever is in the old field"

dfawley: The error would be "you set two fields that mean the same thing, one of which is deprecated: what were you thinking?" :laughing:

dfawley (approved):

Issue #7135, opened 2024-04-17
In picker logic(picker_wrapper.go), clearly understand the real reason for failures after a context.Error(timeout)


holdno:
https://github.com/grpc/grpc-go/blob/a2f8e612d700df49e6b88ec47cbcf8bbd6076149/picker_wrapper.go#L153  

Should we consider setting the error content for lastPickErr so that clients can clearly understand the real reason for failures after a timeout? This might improve troubleshooting and user experience.

``` go
if err == balancer.ErrNoSubConnAvailable {
    lastPickErr = err
    continue
}
```

aranjans:
Hi @holdno, can you please share your findings on running your code?

PR #7143, opened 2024-04-18
add ErrNoSubConnAvailable reason to pick ctx timeout

holdno:
If the server-side balancer picks balancer.ErrNoSubConnAvailable, the gRPC picking logic will cause the context (ctx) to block until it times out. As a result, from the client's perspective, it appears as if there is a server response timeout, without knowing that it is due to NoSubConnAvailable.  

https://github.com/grpc/grpc-go/issues/7135

### DIFF
--- a/picker_wrapper.go
+++ b/picker_wrapper.go
@@ -151,6 +151,7 @@ func (pw *pickerWrapper) pick(ctx context.Context, failfast bool, info balancer.
 		pickResult, err := p.Pick(info)
 		if err != nil {
 			if err == balancer.ErrNoSubConnAvailable {
+				lastPickErr = err
 				continue
 			}
 			if st, ok := status.FromError(err); ok {
--- a/picker_wrapper_test.go
+++ b/picker_wrapper_test.go
@@ -70,8 +70,7 @@ func (s) TestBlockingPickTimeout(t *testing.T) {
 	bp := newPickerWrapper(nil)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
 	defer cancel()
-	var err error
-	if _, _, err = bp.pick(ctx, true, balancer.PickInfo{}); status.Code(err) != codes.DeadlineExceeded {
+	if _, _, err := bp.pick(ctx, true, balancer.PickInfo{}); status.Code(err) != codes.DeadlineExceeded {
 		t.Errorf("bp.pick returned error %v, want DeadlineExceeded", err)
 	}
 }
--- a/picker_wrapper_test.go
+++ b/picker_wrapper_test.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc/balancer"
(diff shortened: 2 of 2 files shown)


aranjans:
@holdno There could be other errors due to which picker would be blocked, like `no connections available` or context deadline error [here](https://github.com/grpc/grpc-go/blob/9cf408ec48fe2b20d020b2430981a148fd93d187/picker_wrapper.go#L120). Probably consider those cases as well.

picker_wrapper.go
```diff
@@ -151,6 +151,7 @@ func (pw *pickerWrapper) pick(ctx context.Context, failfast bool, info balancer.
 		pickResult, err := p.Pick(info)
 		if err != nil {
 			if err == balancer.ErrNoSubConnAvailable {
+				lastPickErr = err
```
dfawley, line 154: This is not the right way to do this.  `ErrNoSubConnAvailable` is not a meaningful error to a user.

Instead, we should probably change line 120's error to something like `fmt.Sprintf("received context error while waiting for new LB policy update: %v", ctx.Err())`

holdno: ctx.Err() could either indicate an issue with the implementation of a certain load balancer, or it might be caused by the absence of a corresponding server. I believe that it would be more user-friendly for the client if the specific reason for the timeout were included along with the ctx.Err().

holdno: Thank you for your response, I hope we can have further discussions.

In my current scenario, during the gray release process, the methods corresponding to the new and old versions of the service are different. In the load balancer, I retrieve the service with the new method by using the fullMethodName. When there are no new version services online, I return an ErrNoSubConnAvailable error, which clearly indicates that there is indeed no effective connection available for the method the client is trying to access. However, the client only receives a timeout error and is not aware of the actual reason for the timeout.

I could indeed use a custom error to convey this scenario, but I feel that the semantics of the ErrNoSubConnAvailable error match my current situation well, and since it is a built-in error of the balancer, the client could also use this error to identify the type of error.

dfawley: > I could indeed use a custom error to convey this scenario, but I feel that the semantics of the ErrNoSubConnAvailable error match my current situation well

I don't think so.  You really _should_ be returning an error in this case.  Also, things generally don't work well if you have multiple versions of a service deployed.  All clients should use the old version of the service until you know all servers have been upgraded, and only then should you update your clients to use the new version.  Otherwise you will run into trouble with things like this.

`ErrNoSubConnAvailable` is not intended for an end user to see.  It is just part of the API between the LB policy and the ClientConn.

holdno: Thank you, I understand this design. However, I think it is still necessary to clarify to the client the reason for the ctx.Error, such as a timeout in updating the client-side load balancer. This is user-friendly for those who are new to gRPC.

new PR for this. https://github.com/grpc/grpc-go/pull/7206

PR #7144, opened 2024-04-18
fix: race detected on accessing http2Client.nextID

aranjans:
RELEASE NOTES: 

- fixes race condition detected on accessing t.nextID while sending GOAWAY frame from client.

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -526,7 +526,7 @@ func (t *http2Client) getPeer() *peer.Peer {
 func (t *http2Client) outgoingGoAwayHandler(g *goAway) (bool, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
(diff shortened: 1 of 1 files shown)


internal/transport/http2_client.go
```diff
@@ -526,7 +526,7 @@ func (t *http2Client) getPeer() *peer.Peer {
 func (t *http2Client) outgoingGoAwayHandler(g *goAway) (bool, error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
-	if err := t.framer.fr.WriteGoAway(t.nextID-2, http2.ErrCodeNo, g.debugData); err != nil {
+	if err := t.framer.fr.WriteGoAway(math.MaxInt32*3/4, http2.ErrCodeNo, g.debugData); err != nil {
```
dfawley, line 529: This is not right.  We want to send the GOAWAY with the ID of the last stream started.  We need to find another way to fix the race.

PR #7150, opened 2024-04-18
add: information about endpoint connection when auth handshake fails

holgerson97:
Wanted in Kubernetes: https://github.com/kubernetes/kubernetes/issues/122639

It adds more information to the log when the authentication handshake fails in gRPC connections.


holgerson97:
@arvindbr8 

There are some tests failing which don't correlate with my change. Do I need to do something?

arvindbr8:
hi @holgerson97 -- I thought I had replied to this PR earlier. Thanks for sending this to us, we think it's definitely useful add some server related info in the log. But we feel like the change could be more wholesome such that all logs in the transport gets it. I am working on a branch to get this functionality. Will get a PR out soon.

Closing this PR in that regard. Thanks again for the PR.

PR #7152, opened 2024-04-19
*: Reduce use of `pretty.ToJSON` due to performance concerns

townba:
RELEASE NOTES: None


dfawley:
For all of these, can you please show before and after for what the message will look like?  Thanks!

PR #7155, opened 2024-04-19
vet: run staticcheck for all sub modules

### LINKED ISSUE #7141: run staticcheck for all modules
In a recent PR, we discovered that `staticcheck` is aware of go modules, and is only currently run for the main module. We need to make sure that it runs for every sub module. This task might involve resolving check errors in other modules which slip through.

arvindbr8:
1. Run staticcheck for all submodules
2. Make changes inorder for vet to pass
3. Add `|| :` to grep.  This is because grep returns error when there are no lines found. 

Basically, fixes #7141 

RELEASE NOTES: none

### DIFF
--- a/gcp/observability/observability_test.go
+++ b/gcp/observability/observability_test.go
@@ -39,7 +39,6 @@ import (
 	"google.golang.org/grpc/internal/leakcheck"
 	"google.golang.org/grpc/internal/stubserver"
 	"google.golang.org/grpc/internal/testutils"
-	"google.golang.org/grpc/metadata"
 
 	testgrpc "google.golang.org/grpc/interop/grpc_testing"
 	testpb "google.golang.org/grpc/interop/grpc_testing"
@@ -64,21 +63,16 @@ func init() {
 }
 
 var (
-	defaultTestTimeout        = 10 * time.Second
-	testHeaderMetadata        = metadata.MD{"header": []string{"HeADer"}}
-	testTrailerMetadata       = metadata.MD{"trailer": []string{"TrAileR"}}
-	testOkPayload             = []byte{72, 101, 108, 108, 111, 32, 87, 111, 114, 108, 100}
-	testErrorPayload          = []byte{77, 97, 114, 116, 104, 97}
-	testErrorMessage          = "test case injected error"
-	infinitySizeBytes   int32 = 1024 * 1024 * 1024
-	defaultRequestCount       = 24
+	defaultTestTimeout  = 10 * time.Second
+	testOkPayload       = []byte{72, 101, 108, 108, 111, 32, 87, 111, 114, 108, 100}
+	defaultRequestCount = 24
 )
 
 const (
-	TypeOpenCensusViewDistribution string = "distribution"
-	TypeOpenCensusViewCount               = "count"
-	TypeOpenCensusViewSum                 = "sum"
-	TypeOpenCensusViewLastValue           = "last_value"
+	TypeOpenCensusViewDistribution = "distribution"
+	TypeOpenCensusViewCount        = "count"
+	TypeOpenCensusViewSum          = "sum"
+	TypeOpenCensusViewLastValue    = "last_value"
 )
 
 type fakeOpenCensusExporter struct {
@@ -503,12 +497,12 @@ func (s) TestCustomTagsTracingMetrics(t *testing.T) {
 		"labels":{"customtag1":"wow","customtag2":"nice"}
 	}`
 
-	cleanup, err := createTmpConfigInFileSystem(configJSON)
+	cleanup, _ := createTmpConfigInFileSystem(configJSON)
 	defer cleanup()
 
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
-	err = Start(ctx)
+	err := Start(ctx)
 	defer End()
 	if err != nil {
 		t.Fatalf("Start() failed with err: %v", err)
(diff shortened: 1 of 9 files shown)


arvindbr8:
Verified the behavior of `vet.sh` for Darwin. Letting the Github Action test for behavior in Linux

gcp/observability/observability_test.go
```diff
@@ -503,12 +497,12 @@ func (s) TestCustomTagsTracingMetrics(t *testing.T) {
 		"labels":{"customtag1":"wow","customtag2":"nice"}
 	}`
 
-	cleanup, err := createTmpConfigInFileSystem(configJSON)
+	cleanup, _ := createTmpConfigInFileSystem(configJSON)
```
dfawley, line 500: `t.Fatal` on `err` instead?

scripts/vet.sh
```diff
@@ -40,34 +40,34 @@ not grep 'func Test[^(]' *_test.go
...
 
 # - Check for typos in test function names
-git grep 'func (s) ' -- "*_test.go" | not grep -v 'func (s) Test'
-git grep 'func [A-Z]' -- "*_test.go" | not grep -v 'func Test\|Benchmark\|Example'
+git grep --quiet 'func (s) ' -- "*_test.go" | not grep -v 'func (s) Test'
+git grep --quiet 'func [A-Z]' -- "*_test.go" | not grep -v 'func Test\|Benchmark\|Example'
 
 # - Do not import x/net/context.
-not git grep -l 'x/net/context' -- "*.go"
+not git grep --quiet -l 'x/net/context' -- "*.go"
```
dfawley, line 47: Let's just remove this?  It's a symlink now, and there's no reason we would be using it anymore, anyway.

scripts/vet.sh
```diff
@@ -86,18 +86,21 @@ for MOD_FILE in $(find . -name 'go.mod'); do
...
 
-# - Collection of static analysis checks
-SC_OUT="$(mktemp)"
-staticcheck -go 1.19 -checks 'all' ./... > "${SC_OUT}" || true
+for MOD_FILE in $(find . -name 'go.mod'); do
+  MOD_DIR=$(dirname ${MOD_FILE})
+  pushd ${MOD_DIR}
+  # - Collection of static analysis checks
+  SC_OUT="$(mktemp)"
+  staticcheck -go 1.19 -checks 'all' ./... >"${SC_OUT}" || true
```
dfawley, line 94: Maybe just move this into the previous for loop instead of adding another one?

scripts/vet.sh
```diff
@@ -86,18 +86,21 @@ for MOD_FILE in $(find . -name 'go.mod'); do
...
 
-# Error for anything other than checks that need exclusions.
-grep -v "(ST1000)" "${SC_OUT}" | grep -v "(SA1019)" | grep -v "(ST1003)" | not grep -v "(ST1019)\|\(other import of\)"
+  # Error for anything other than checks that need exclusions.
+  (grep -v "(ST1000)" "${SC_OUT}" || : )| (grep -v "(SA1019)" || :) | (grep -v "(ST1003)" || :) | (grep -v "(ST1019)\|\(other import of\)" || :) | not grep -v "(SA4000)"
```
dfawley, line 97: Can we define a function for this `grep || :` pattern instead?

security/advancedtls/crl.go
```diff
@@ -583,6 +583,7 @@ func extractCRLIssuer(crlBytes []byte) ([]byte, error) {
 	}
 	der := cryptobyte.String(crlBytes)
 	var issuer cryptobyte.String
+
```
dfawley, line 586: ? was this a linter?

dfawley:
+@gtcooke94 to review the advancedtls changes.

gcp/observability/observability_test.go
arvindbr8: Done

scripts/vet.sh
arvindbr8: Sure. I'm not so sure about it. Removing it from vet.

gtcooke94 (approved):
LGTM, very small changes in advancedtls

scripts/vet.sh
arvindbr8: Done

scripts/vet.sh
arvindbr8: Done

security/advancedtls/crl.go
arvindbr8: Done

scripts/vet.sh
dfawley: Did I say "symlink"?  I meant "type alias" O_O

scripts/vet.sh
```diff
@@ -1,13 +1,18 @@
...
+set -o pipefail # Fail a pipe if any sub-command fails.
 
 source "$(dirname $0)/vet-common.sh"
 
 # Check to make sure it's safe to modify the user's git repo.
 git status --porcelain | fail_on_output
 
+# Suppresses the return code of grep since grep would fail vet when there are no matches (for eg, empty file).
+noret_grep() {
+  grep "$@" || :
```
dfawley, line 13: Like we did with diff, should this only suppress if the exit code is 1 ("no lines selected")?  i.e. `|| [[ $? == 1 ]]` instead of `|| :`?

dfawley (approved):

scripts/vet.sh
arvindbr8: Nice catch. Thanks! Fixed

dfawley (approved):

PR #7158, opened 2024-04-22
Add lazyDebugf method to process the arguments lazily

### LINKED ISSUE #7154: Delete grpclog's Debugf function
https://github.com/grpc/grpc-go/blob/9cf408ec48fe2b20d020b2430981a148fd93d187/internal/grpclog/prefixLogger.go#L65

If a parameter passed to these functions is expensive to evaluate, that expense is borne regardless of whether verbosity would output the message .

All usages should be audited, and we should write another function that allows the arguments to be processed lazily (i.e. pass in closures).

aranjans:
Fixes #7154 

RELEASE NOTES:

- Add LazyDebugf function.

### DIFF
--- a/internal/grpclog/prefixLogger.go
+++ b/internal/grpclog/prefixLogger.go
@@ -79,6 +79,34 @@ func (pl *PrefixLogger) Debugf(format string, args ...any) {
 
 }
 
+// LazyDebugf does info logging at verbose level 2, only
+// evaluating arguments if logging is enabled.
+func (pl *PrefixLogger) LazyDebugf(format string, args ...func() any) {
+	if !Logger.V(2) {
+		return
+	}
+	if pl != nil {
+		format = pl.prefix + format
+
+		// Build the argument list by evaluating the closures
+		var evaluatedArgs []any
+		for _, argFunc := range args {
+			evaluatedArgs = append(evaluatedArgs, argFunc())
+		}
+
+		pl.logger.InfoDepth(1, fmt.Sprintf(format, evaluatedArgs...))
+		return
+	}
+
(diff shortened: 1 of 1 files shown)


internal/grpclog/prefixLogger.go
```diff
@@ -79,6 +79,34 @@ func (pl *PrefixLogger) Debugf(format string, args ...any) {
 
 }
 
+// LazyDebugf does info logging at verbose level 2, only
+// evaluating arguments if logging is enabled.
+func (pl *PrefixLogger) LazyDebugf(format string, args ...func() any) {
```
dfawley, line 84: Is this worth it?

This makes the call site:

```go
logger.LazyDebugf("blah %s %s %v", func() { blah1() }, func() { blah2() }, func() { blah3() })
```

vs.

```go
if logger.V(2) {
	logger.Infof("blah %s %s %v", blah1(), blah2(), blah3())
}
```

Yes it's "3 lines instead of 1 line" but:
1. it's not obvious that Debugf is checking V(2), and
2. the usage with Lazy is really hard to read and very long because of all the extraneous closures.

I think I'd rather just delete `Debugf` instead.

purnesh42H: How many occurrences of Debugf do we have?

aranjans: 17 i believe

aranjans:
Taking pause on this sometime to prioritize other issues. Will update this PR based on [this](https://github.com/grpc/grpc-go/issues/7154#issuecomment-2074222144) comment

arvindbr8:
Let's close this PR for now and re-open when we are actively working on it

PR #7151, opened 2024-04-18
advancedtls: Rename RevocationConfig

gtcooke94:
This PR renames `RevocationConfig` to `RevocationOptions` to match the naming of all of our other `Options` structs. A type alias remains as to not break existing code.

RELEASE NOTES: none

PR #7168, opened 2024-04-24
Bring back the doc string for VerificationResults to fix static check

arjan-bal:
The static checks enabled in https://github.com/grpc/grpc-go/pull/7155 are failing due https://github.com/grpc/grpc-go/pull/7140 being merged around the same time. #7140 removed the doc string for VerificationResults and replaced it with a deprecation note. As a result the static check failed with the following error:
```
Error: advancedtls.go:70:1: comment on exported type VerificationResults should be of the form "VerificationResults ..." (with optional leading article) (ST1021)
````

This change brings back the first line of the doc string to satisfy the test.

### CODE REVIEW

File: security/advancedtls/advancedtls.go [hunk 1]
```diff
@@ -67,6 +67,9 @@ type VerificationFuncParams = HandshakeVerificationInfo
 // future to include more information.
 type PostHandshakeVerificationResults struct{}
 
+// VerificationResults contains the information about results of
+// CustomVerificationFunc.
```
purnesh42H, line 71: is this moved to new line by auto formatter?

purnesh42H (approved):
lgtm. Thanks for quick fix!

security/advancedtls/advancedtls.go
arjan-bal: I think so, it was the same earlier: https://github.com/grpc/grpc-go/blob/5fe2e74bf4f092addfb0a36f53964474914357a3/security/advancedtls/advancedtls.go#L57-L59

aranjans: So what's the fix here, how come tests are passed now?

aranjans:
I think this fix is also part of this PR https://github.com/grpc/grpc-go/pull/7151/files

arjan-bal:
> I think this fix is also part of this PR https://github.com/grpc/grpc-go/pull/7151/files

I see. @gtcooke94, will #7151 be merged soon? If not, should we fix only the lint error to make the tests healthy again?

gtcooke94 (approved):
No worries just merging this as-is, I'll deal with any git weirdness on #7151 on my own

arjan-bal:
@dfawley PTAL and merge if it looks good. I can't merge since I don't have write access.

PR #7169, opened 2024-04-25
*: Limit calls to `pretty.ToJSON` to more verbose debugging, warnings, and errors to improve performance

townba:
RELEASE NOTES: None

### DIFF
--- a/resolver_wrapper.go
+++ b/resolver_wrapper.go
@@ -194,5 +194,7 @@ func (ccr *ccResolverWrapper) addChannelzTraceEvent(s resolver.State) {
 	} else if len(ccr.curState.Addresses) == 0 && len(s.Addresses) > 0 {
 		updates = append(updates, "resolver returned new addresses")
 	}
-	channelz.Infof(logger, ccr.cc.channelz, "Resolver state updated: %s (%v)", pretty.ToJSON(s), strings.Join(updates, "; "))
+	if channelz.IsOn() || logger.V(2) {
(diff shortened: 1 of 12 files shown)


zasweq:
How did this make these logging calls look before and after? How much did this improve performance?

resolver_wrapper.go
```diff
@@ -194,5 +194,7 @@ func (ccr *ccResolverWrapper) addChannelzTraceEvent(s resolver.State) {
 	} else if len(ccr.curState.Addresses) == 0 && len(s.Addresses) > 0 {
 		updates = append(updates, "resolver returned new addresses")
 	}
-	channelz.Infof(logger, ccr.cc.channelz, "Resolver state updated: %s (%v)", pretty.ToJSON(s), strings.Join(updates, "; "))
+	if channelz.IsOn() || logger.V(2) {
```
zasweq, line 197: What is the reason for the additional channelz.IsOn() bool conditional?

zasweq (approved):
LGTM.

aranjans (approved):

purnesh42H (approved):
lgtm

resolver_wrapper.go
townba: It's similar to the reason for the `logger.V(2)` check. We only want to call `pretty.ToJSON` when we know we're going to use the result. `channelz.Infof` can add a trace event, log, or both, so we want to call it when channelz data collection is on, we're logging at depth 2, or both.

townba:
**Please don't merge this yet.** Doug Fawley and I are still discussing this approach.

dfawley:
I'm going to close this since I think we got the same win just from the Debugf calls.  We can reopen or recreate in the future if needed.

PR #7173, opened 2024-04-30
advancedTLS: Rename {Min/Max}Version to {Min/Max}TLSVersion

gtcooke94:
This PR renames the `MinVersion` and `MaxVersion` on `ClientOptions` and `ServerOptions` to `MinTLSVersion` and `MaxTLSVersion` to make it clear that this field represents the TLS version.

It also adds logic to set the default min and max to TLS12 and TLS13 respectively, and a tests for that logic.
Previously our code comments indicated this was the default, but we really just delegated to the defaults of the crypto/tls package (which are currently the same). Now we explicitly set to the defaults that we say we do, rather than depending on a dependency's defaults to match ours.

RELEASE NOTES: none

### DIFF
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -221,7 +221,7 @@ type ClientOptions struct {
 	MinVersion uint16
 	// MaxVersion contains the maximum TLS version that is acceptable.
 	//
-	// Deprecated: use MinTLSVersion instead.
+	// Deprecated: use MaxTLSVersion instead.
 	MaxVersion uint16
 	// MinTLSVersion contains the minimum TLS version that is acceptable.
 	// By default, TLS 1.2 is currently used as the minimum when acting as a
@@ -279,7 +279,7 @@ type ServerOptions struct {
 	MinVersion uint16
 	// MaxVersion contains the maximum TLS version that is acceptable.
 	//
-	// Deprecated: use MinTLSVersion instead.
+	// Deprecated: use MaxTLSVersion instead.
 	MaxVersion uint16
 	// MinTLSVersion contains the minimum TLS version that is acceptable.
 	// By default, TLS 1.2 is currently used as the minimum when acting as a
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -224,12 +224,14 @@ type ClientOptions struct {
 	// Deprecated: use MaxTLSVersion instead.
 	MaxVersion uint16
 	// MinTLSVersion contains the minimum TLS version that is acceptable.
+	// The value should be set using tls.VersionTLSxx from https://pkg.go.dev/crypto/tls
 	// By default, TLS 1.2 is currently used as the minimum when acting as a
 	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
 	// supported by this package, both as a client and as a server.  This
 	// default may be changed over time affecting backwards compatibility.
 	MinTLSVersion uint16
 	// MaxTLSVersion contains the maximum TLS version that is acceptable.
+	// The value should be set using tls.VersionTLSxx from https://pkg.go.dev/crypto/tls
 	// By default, the maximum version supported by this package is used,
 	// which is currently TLS 1.3.  This default may be changed over time
 	// affecting backwards compatibility.
@@ -282,12 +284,14 @@ type ServerOptions struct {
 	// Deprecated: use MaxTLSVersion instead.
 	MaxVersion uint16
 	// MinTLSVersion contains the minimum TLS version that is acceptable.
+	// The value should be set using tls.VersionTLSxx from https://pkg.go.dev/crypto/tls
 	// By default, TLS 1.2 is currently used as the minimum when acting as a
 	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
 	// supported by this package, both as a client and as a server.  This
 	// default may be changed over time affecting backwards compatibility.
 	MinTLSVersion uint16
 	// MaxTLSVersion contains the maximum TLS version that is acceptable.
+	// The value should be set using tls.VersionTLSxx from https://pkg.go.dev/crypto/tls
 	// By default, the maximum version supported by this package is used,
 	// which is currently TLS 1.3.  This default may be changed over time
 	// affecting backwards compatibility.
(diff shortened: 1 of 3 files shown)


security/advancedtls/advancedtls.go
```diff
@@ -216,14 +216,24 @@ type ClientOptions struct {
...
 	MaxVersion uint16
+	// MinTLSVersion contains the minimum TLS version that is acceptable.
+	// By default, TLS 1.2 is currently used as the minimum when acting as a
+	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
+	// supported by this package, both as a client and as a server.  This
+	// default may be changed over time affecting backwards compatibility.
+	MinTLSVersion uint16
+	// MaxTLSVersion contains the maximum TLS version that is acceptable.
+	// By default, the maximum version supported by this package is used,
+	// which is currently TLS 1.3.  This default may be changed over time
```
erm-g, line 234: Mb 'By default, the maximum version supported by this package is currently TLS 1.3' ?

erm-g (approved):

security/advancedtls/advancedtls.go
```diff
@@ -216,14 +216,24 @@ type ClientOptions struct {
...
-	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
-	// supported by this package, both as a client and as a server.
+	//
+	// Deprecated: use MinTLSVersion instead.
 	MinVersion uint16
 	// MaxVersion contains the maximum TLS version that is acceptable.
-	// By default, the maximum version supported by this package is used,
-	// which is currently TLS 1.3.
+	//
+	// Deprecated: use MinTLSVersion instead.
```
dfawley, line 224: MaxTLSVersion?

security/advancedtls/advancedtls.go
```diff
@@ -216,14 +216,24 @@ type ClientOptions struct {
...
 	MaxVersion uint16
+	// MinTLSVersion contains the minimum TLS version that is acceptable.
+	// By default, TLS 1.2 is currently used as the minimum when acting as a
+	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
+	// supported by this package, both as a client and as a server.  This
+	// default may be changed over time affecting backwards compatibility.
```
dfawley, line 230: Are we reserving the right to override the user's setting?  Like if they set 1.0 and we decide to set the min to 1.2, then we'll automatically upgrade them?  Or no?

security/advancedtls/advancedtls.go
```diff
@@ -264,14 +274,24 @@ type ServerOptions struct {
...
-	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
-	// supported by this package, both as a client and as a server.
+	//
+	// Deprecated: use MinTLSVersion instead.
 	MinVersion uint16
 	// MaxVersion contains the maximum TLS version that is acceptable.
-	// By default, the maximum version supported by this package is used,
-	// which is currently TLS 1.3.
+	//
+	// Deprecated: use MinTLSVersion instead.
```
dfawley, line 282: `Max`

security/advancedtls/advancedtls.go
```diff
@@ -264,14 +274,24 @@ type ServerOptions struct {
...
 	MaxVersion uint16
+	// MinTLSVersion contains the minimum TLS version that is acceptable.
+	// By default, TLS 1.2 is currently used as the minimum when acting as a
+	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
+	// supported by this package, both as a client and as a server.  This
+	// default may be changed over time affecting backwards compatibility.
+	MinTLSVersion uint16
```
dfawley, line 289: It's really unfortunate to have to repeat all this documentation.  It seems like 99% of client and server options are identical and the only field unique to servers is `RequireClientCert`, is that right?  Can that somehow be implemented by some other feature instead, like VerifyPeer or VerificationType or something?

Or maybe we can merge them anyway and just document that field as ignored on clients.

security/advancedtls/advancedtls.go
gtcooke94: Discussed offline

security/advancedtls/advancedtls.go
gtcooke94: I agree with you - I'm very open to doing that in another PR, does that sound good to you? We can discuss specifics offline to not pollute this PR?

security/advancedtls/advancedtls.go
gtcooke94: Good catch

security/advancedtls/advancedtls.go
gtcooke94: I don't think so - if we decide to change the minimum supported version from 1.0 to 1.2, I think we'd add in a check that throws an error rather than override the user setting.

security/advancedtls/advancedtls.go
gtcooke94: Good catch

security/advancedtls/advancedtls.go
dfawley: That's fine.  It seems like when they made a change like this in Go's tls package they added an env var so users could get the old 1.0-default behavior back (just FYI).

Also, I assume this is supposed to be a `tls.VersionTLSxx` value?  (I wish they had made a type name for it.)  That might be worth documenting.

security/advancedtls/advancedtls.go
dfawley: SG; I wasn't trying to block this PR, just making an observation.

security/advancedtls/advancedtls.go
gtcooke94: I think we could explore adding an env var like that down the line if it ever happens that we need it
Added notes in the docs about using tls.VersionTLSxx

dfawley: > Added notes in the docs about using tls.VersionTLSxx


What doc?  The godocs?  If so, I think you forgot to push the change.  It seems worth documenting in these comments, since this isn't inside the `tls` package, but I won't argue too strongly.

dfawley (approved):

security/advancedtls/advancedtls.go
gtcooke94: Whoops I didn't actually push the change to my remote branch, done now

PR #7179, opened 2024-05-02
advancedTLS: Swap to DenyUndetermined from AllowUndetermined in revocation settings

gtcooke94:
Swap `AllowUndetermined` to `DenyUndetermined` in our revocation settings.
We want the default for this setting to be to allow `RevocationUndetermined` to pass revocation checks, so the boolean needs to be `DenyUndetermined` (default false) for this behavior.

RELEASE NOTES: none

### DIFF
--- a/security/advancedtls/crl.go
+++ b/security/advancedtls/crl.go
@@ -61,9 +61,9 @@ type RevocationOptions struct {
 	// Directory format must match OpenSSL X509_LOOKUP_hash_dir(3).
 	// Deprecated: use CRLProvider instead.
 	RootDir string
-	// AllowUndetermined controls if certificate chains with RevocationUndetermined
+	// DenyUndetermined controls if certificate chains with RevocationUndetermined
 	// revocation status are allowed to complete.
-	AllowUndetermined bool
+	DenyUndetermined bool
 	// Cache will store CRL files if not nil, otherwise files are reloaded for every lookup.
 	// Only used for caching CRLs when using the RootDir setting.
 	// Deprecated: use CRLProvider instead.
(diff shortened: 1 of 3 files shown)


security/advancedtls/crl.go
```diff
@@ -61,8 +61,13 @@ type RevocationOptions struct {
 	// Directory format must match OpenSSL X509_LOOKUP_hash_dir(3).
 	// Deprecated: use CRLProvider instead.
 	RootDir string
+	// DenyUndetermined controls if certificate chains with RevocationUndetermined
+	// revocation status are allowed to complete.
+	DenyUndetermined bool
```
erm-g, line 66: I'd prefer to keep the current ['positive boolean' approach](https://testing.googleblog.com/2023/10/improve-readability-with-positive.html), potentially with 'default true' mechanism like 
`
type MyConfig struct {
    AllowChanges bool
}

func NewMyConfig() MyConfig {
    return MyConfig{
        AllowChanges: true,
    }
}`
Leaving it to you&Doug to decide.

erm-g (approved):

security/advancedtls/crl.go
gtcooke94: I don't like this because it requires adding a constructor layer for a relatively simple struct - in this case IMO the positive boolean approach isn't worth the complexity

@dfawley WDYT?

dfawley: Yeah there's unfortunately come conflict between having positive booleans and having default behavior = zero value.

Maybe `DenyUndetermined` is not too negative?  It states what you _will_ be doing, not what you will _not_ be doing.  Or we could go with something more verbose that removes the double-negative like `RequireDetermined[Revocation]Status`?

dfawley (approved):

security/advancedtls/crl.go
dfawley: BTW I'm fine either way.  Another option could also be to replace with an "enum" (dumb Go doesn't actually have them, but):

```go
type RevocationStatusRequirement int

const (
    DenyUndetermined RevocationStatusRequirement = 0
    AllowUndetermined
)
```

(Names are just for illustrative purposes; probably pick better names.)

PR #7180, opened 2024-05-02
xds/internal: Replace calls to `Debugf` with `V(2)` checks and `Infof`

townba:
… to limit calls to `pretty.ToJSON`

RELEASE NOTES: None


dfawley (approved):
Any idea of the impact of this more isolated change yet?

townba:
> Any idea of the impact of this more isolated change yet?

From what I saw, it was about as good as the larger-scale change, though I only had it in two VMs over the weekend.

Issue #7176, opened 2024-05-01
advancedtls package does not set VerifiedChains in TLSInfo.State


mudhireddy:
NOTE: if you are reporting is a potential security vulnerability or a crash,
please follow our CVE process at
https://github.com/grpc/proposal/blob/master/P4-grpc-cve-process.md instead of
filing an issue here.

Please see the FAQ in our main README.md, then answer the questions below
before submitting your issue.

### What version of gRPC are you using?
1.60.0

### What version of Go are you using (`go version`)?
go1.22.1 

### What operating system (Linux, Windows, …) and version?
linux/amd64

### What did you do?
If possible, provide a recipe for reproducing the error.

// The function buildVerifyFunc is used when users want root cert reloading,
// and possibly custom verification check.
// We have to build our own verification function here because current
// tls module:
//  1. does not have a good support on root cert reloading.
//  2. will ignore basic certificate check when setting InsecureSkipVerify
//     to true.

When using buildVerifyFunc() here https://github.com/grpc/grpc-go/blob/e4a6ce3a5489/security/advancedtls/advancedtls.go#L516
it verifies the client certificate chain correctly as expected but it doesn't expose the verifiedChains to the upper layer to make use of them like how tls package does.

### What did you expect to see?
I expect it to set the VerifiedChains in TLSInfo.State structure.

patch for this is at: https://github.com/mudhireddy/grpc-go/commit/9d0785894b24daf021dba8e68c844ac696ba5679#diff-e8cdf2d70ced71ce723591cf60921a380bd6f007888f8883cbb20df4ef406680L444

it sets TLSInfo.State.VerifiedChains so that application can read the necessary information from VerifiedChains.

https://github.com/grpc/grpc-go/blob/e4a6ce3a5489/security/advancedtls/advancedtls.go#L514
514: func (c *advancedTLSCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
         cfg := credinternal.CloneTLSConfig(c.config)
	-  cfg.VerifyPeerCertificate = buildVerifyFunc(c, "", rawConn)
	+  peerVerifiedChains := [][]*x509.Certificate{}
	+  cfg.VerifyPeerCertificate = buildVerifyFunc(c, "", rawConn, &peerVerifiedChains)

https://github.com/grpc/grpc-go/blob/e4a6ce3a5489/security/advancedtls/advancedtls.go#L528
[...]

PR #7181, opened 2024-05-02
advancedtls: populate verified chains when using custom buildVerifyFunc

mudhireddy:
This request is to address the issue opened https://github.com/grpc/grpc-go/issues/7176

Basically this patch will set the verifiedChains correctly when advancedtls package is used with buildVerifyFunc() and applications need to access TLSInfo.State.VerifiedChains in their logic.

RELEASE NOTES: none

### DIFF
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -35,6 +35,8 @@ import (
 	credinternal "google.golang.org/grpc/internal/credentials"
 )
 
+type CertChains [][]*x509.Certificate
+
 // HandshakeVerificationInfo contains information about a handshake needed for
 // verification for use when implementing the `PostHandshakeVerificationFunc`
 // The fields in this struct are read-only.
@@ -47,7 +49,7 @@ type HandshakeVerificationInfo struct {
 	RawCerts [][]byte
 	// The verification chain obtained by checking peer RawCerts against the
 	// trust certificate bundle(s), if applicable.
-	VerifiedChains [][]*x509.Certificate
+	VerifiedChains CertChains
 	// The leaf certificate sent from peer, if choosing to verify the peer
 	// certificate(s) and that verification passed. This field would be nil if
 	// either user chose not to verify or the verification failed.
@@ -484,7 +486,7 @@ func (c *advancedTLSCreds) ClientHandshake(ctx context.Context, authority string
 	if cfg.ServerName == "" {
 		cfg.ServerName = authority
 	}
-	peerVerifiedChains := [][]*x509.Certificate{}
+	peerVerifiedChains := CertChains{}
 	cfg.VerifyPeerCertificate = buildVerifyFunc(c, cfg.ServerName, rawConn, &peerVerifiedChains)
 	conn := tls.Client(rawConn, cfg)
 	errChannel := make(chan error, 1)
@@ -515,7 +517,7 @@ func (c *advancedTLSCreds) ClientHandshake(ctx context.Context, authority string
 
 func (c *advancedTLSCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
 	cfg := credinternal.CloneTLSConfig(c.config)
-	peerVerifiedChains := [][]*x509.Certificate{}
+	peerVerifiedChains := CertChains{}
 	cfg.VerifyPeerCertificate = buildVerifyFunc(c, "", rawConn, &peerVerifiedChains)
 	conn := tls.Server(rawConn, cfg)
 	if err := conn.Handshake(); err != nil {
@@ -562,7 +564,7 @@ func (c *advancedTLSCreds) OverrideServerName(serverNameOverride string) error {
 func buildVerifyFunc(c *advancedTLSCreds,
 	serverName string,
 	rawConn net.Conn,
-	peerVerifiedChains *[][]*x509.Certificate) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	peerVerifiedChains *CertChains) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 		chains := verifiedChains
 		var leafCert *x509.Certificate
@@ -626,7 +628,7 @@ func buildVerifyFunc(c *advancedTLSCreds,
 		if c.revocationOptions != nil {
 			verifiedChains := chains
 			if verifiedChains == nil {
-				verifiedChains = [][]*x509.Certificate{rawCertList}
+				verifiedChains = CertChains{rawCertList}
 			}
 			if err := checkChainRevocation(verifiedChains, *c.revocationOptions); err != nil {
 				return err
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -35,7 +35,7 @@ import (
 	credinternal "google.golang.org/grpc/internal/credentials"
 )
 
-type CertChains [][]*x509.Certificate
+type CertificateChains [][]*x509.Certificate
 
 // HandshakeVerificationInfo contains information about a handshake needed for
 // verification for use when implementing the `PostHandshakeVerificationFunc`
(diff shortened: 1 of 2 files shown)


mudhireddy:
My local testing shows it ok though  

grpc-go/test/xds @ramesh-937457-0.sjc% go test
PASS
ok      google.golang.org/grpc/test/xds 11.131s


Is there a way to trigger the re-test of it?

mudhireddy:
@purnesh42H  any documentation on who and how to add LABEL, MILESTONE so that all checks on this pull request will pass ? right now Mergeable is failing for missing LABEL, MILESTONE

purnesh42H:
@mudhireddy I have assigned it to myself. I will discuss with other maintainers and update you back. Thanks!

security/advancedtls/advancedtls.go
```diff
@@ -611,6 +618,7 @@ func buildVerifyFunc(c *advancedTLSCreds,
 				return err
 			}
 			leafCert = rawCertList[0]
+			*peerVerifiedChains = chains
```
gtcooke94, line 621: We would not want to do this until right before the `return nil` on line 643, otherwise we might set these as verified chains then return an error if we find they are revoked by a CRL. A custom verification func that was supplied could fail as well.

gtcooke94 (commented):
Thanks for the PR! In general looks good, one big comment about where we set the value.
I'm reasoning through a few things on the overall code flow to make sure this wouldn't have any risk of leaking cert chains around if someone used the same credential with multiple connections - I'll update here when I'm done looking through that.

In the meanwhile, could you add tests that validate the expected behaviors here?
Also just curious, what is your use case for this?

security/advancedtls/advancedtls.go
mudhireddy: Thanks for the review. addressed it in the next patch

security/advancedtls/advancedtls.go
```diff
@@ -552,7 +558,8 @@ func (c *advancedTLSCreds) OverrideServerName(serverNameOverride string) error {
 //     to true.
 func buildVerifyFunc(c *advancedTLSCreds,
```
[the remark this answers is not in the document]
mudhireddy: done

mudhireddy:
> Thanks for the PR! In general looks good, one big comment about where we set the value. I'm reasoning through a few things on the overall code flow to make sure this wouldn't have any risk of leaking cert chains around if someone used the same credential with multiple connections - I'll update here when I'm done looking through that.
> 
> In the meanwhile, could you add tests that validate the expected behaviors here? Also just curious, what is your use case for this?

Added tests. Reg use case we use the CommonName details from VerifiedChains

gtcooke94 (dismissed):
This looks good, thanks again for the PR!
We'll get one more reviewer to check it out

gtcooke94:
@dfawley can you review this or assign to another go team member?

security/advancedtls/advancedtls.go
```diff
@@ -508,12 +509,15 @@ func (c *advancedTLSCreds) ClientHandshake(ctx context.Context, authority string
 		},
 	}
 	info.SPIFFEID = credinternal.SPIFFEIDFromState(conn.ConnectionState())
+	// set the VerifiedChains
```
dfawley, line 512: This comment seems to add no value.  Please remove it.

security/advancedtls/advancedtls.go
```diff
@@ -526,6 +530,8 @@ func (c *advancedTLSCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credenti
...
```
dfawley, line 533: As above.

security/advancedtls/advancedtls.go
```diff
@@ -550,9 +556,15 @@ func (c *advancedTLSCreds) OverrideServerName(serverNameOverride string) error {
...
+//
+// peerVerifiedChains(output param): verified chain of certs from leaf to the
+// trust cert that the peer trusts.
+//  1. For server it is, client certs + Root ca that the server trusts
+//  2. For client it is, server certs + Root ca that the client trusts
 func buildVerifyFunc(c *advancedTLSCreds,
 	serverName string,
-	rawConn net.Conn) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	rawConn net.Conn,
+	peerVerifiedChains *[][]*x509.Certificate) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
dfawley, line 567: A pointer to a slice of slices of pointers to x509 certs?

Would it make more sense to return a second value that is `[][]*x509.Certificate` instead?

Also should we define a type / type alias for `[][]*x509.Certificate`?

mudhireddy: to be honest, I couldn't unpack all the syntactic complexity around that to make it work. Any help would be appreciated.

mudhireddy: @dfawley 

   cfg := credinternal.CloneTLSConfig(c.config)
   peerVerifiedChains := [][]*x509.Certificate{}
   cfg.VerifyPeerCertificate = buildVerifyFunc(c, "", rawConn, &peerVerifiedChains)
   conn := tls.Server(rawConn, cfg)
   if err := conn.Handshake(); err != nil {
      conn.Close()
      return nil, nil, err
   }

buildVerifyFunc() is returning a function that gets stored in cfg.VerifyPeerCertificate and later gets called as part of conn.Handshake(). Given this structure returning a second parameter would mean that we have to modify the prevalent use of the type "func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error" and that quickly becomes a big change in itself. 

Reg "type / type alias for [][]*x509.Certificate?" I can define it and replace in advancedtls.go where possible, does this name sound reasonable? (sorry, trying to minimize back and forth and want to close this soon)

type CertChains [][]*x509.Certificate

mudhireddy:
@gtcooke94, can I please get approval for this PR to merge ?

security/advancedtls/advancedtls.go
dfawley: @gtcooke94 - thoughts?

security/advancedtls/advancedtls.go
```diff
@@ -35,6 +35,8 @@ import (
 	credinternal "google.golang.org/grpc/internal/credentials"
 )
 
+type CertChains [][]*x509.Certificate
```
gtcooke94, line 38: Can we call it `CertificateChains`?

gtcooke94 (dismissed):
Giving the type a name makes the code a little more readable, I think it's a good change

gtcooke94 (commented):
Sorry meant to not approve yet because I don't want the shortened name

security/advancedtls/advancedtls.go
gtcooke94: We don't really have too much flexibility in adding return values - we build this verification func then set crypto/tls's `VerifyPeerCertificate` to it https://pkg.go.dev/crypto/tls

security/advancedtls/advancedtls.go
mudhireddy: sure, let me fix that

gtcooke94 (approved):
I think this is looking good - it would be nice to be able to replace and use `CertificateChains` everywhere, but we could potentially just do that in a separate PR

dfawley (approved):

security/advancedtls/advancedtls.go
gtcooke94: As far as I can tell this doesn't involve anything async? Instead of passing the variable directly we could do something with channels? But given the constraints of the callback structure, I think something like this is how we'll have to get it out unfortunately.

mudhireddy:
Thank you guys for the review and approval. 

@gtcooke94, I see that it is still open, is it going to auto merge or do I need to do anything for it to merge ?

security/advancedtls/advancedtls.go
```diff
@@ -35,6 +35,8 @@ import (
 	credinternal "google.golang.org/grpc/internal/credentials"
 )
 
+type CertificateChains [][]*x509.Certificate
```
arjan-bal, line 38: Since this is an exported type, it should have a doc string.

PR #7182, opened 2024-05-03
xds: Move bootstrap config to internal/xds

zasweq:
This PR moves bootstrap config to internal/xds. This allows CSM Plugin Option to call bootstrap config parsing to parse the out the mesh id from the node proto's id.

In the future, we could perhaps make the bootstrap config a singleton (and wrap creation function with something like a sync.OnceFunc).

RELEASE NOTES: N/A


internal/xds/bootstrap/credentials.go
dfawley, line 1: This is fine...

But do you think maybe this logic (and this file) would be even better in `grpc/xds/bootstrap`?  (And then blank import that package from this package's `bootstrap.go`?

I'm OK either way, but would lean toward moving it in with the registry if you can't think of any reason it's better off (or needs to be?) here.  These are essentially "default supported creds", and if it's possible for a user to import the registry but not get these creds in it, that seems weird.

dfawley (approved):

internal/xds/bootstrap/credentials.go
zasweq: I'm fine with either. I'll move it.

zasweq: No need to blank import, the package is already imported :).

dfawley (approved):

internal/xds/bootstrap/credentials.go
dfawley: Oh yeah, it does `Get`s

Issue #434, opened 2015-11-06
TLS does not enforce ALPN protocol 


carl-mastrangelo:
According to https://tools.ietf.org/html/rfc7540#section-3.3 connections over TLS use the "h2" application protocol identifier.  Attempting to use another protocol identifier, such as "h2c", should fail the connection.  Currently, the Grpc go server accepts using this invalid identifier when establishing a TLS connection.  

Here is the test that fails:

https://github.com/grpc/grpc/blob/master/tools/http2_interop/http2interop.go#L235

jroper:
Crazy that this issue hasn't been fixed 8 years on. Here's a work around that can be used:

```golang
import (
	"context"
	"crypto/tls"
	"fmt"
	"google.golang.org/grpc/credentials"
	"net"
)

func NewAlpnEnforcingTLSCredentials(config *tls.Config) credentials.TransportCredentials {
	return &alpnEnforcer{
		delegate: credentials.NewTLS(config),
	}
}

type alpnEnforcer struct {
	delegate credentials.TransportCredentials
}

func (a *alpnEnforcer) ClientHandshake(ctx context.Context, authority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
	conn, authInfo, err := a.delegate.ClientHandshake(ctx, authority, rawConn)
	if err != nil {
		return nil, nil, err
	}

	tlsInfo, ok := authInfo.(*credentials.TLSInfo)
	if !ok {
		_ = conn.Close()
		return nil, nil, fmt.Errorf("TLS credentials did not return a TLS info")
	}
	if tlsInfo.State.NegotiatedProtocol == "" {
		_ = conn.Close()
		return nil, nil, fmt.Errorf("peer does not support ALPN and therefore can't support HTTP/2")
	}
	if tlsInfo.State.NegotiatedProtocol != "h2" {
		_ = conn.Close()
		return nil, nil, fmt.Errorf("negotiated protocol with peer that is not HTTP/2")
	}
	return conn, authInfo, err
}

func (a *alpnEnforcer) ServerHandshake(conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
	return a.delegate.ServerHandshake(conn)
}

func (a *alpnEnforcer) Info() credentials.ProtocolInfo {
	return a.delegate.Info()
}

func (a *alpnEnforcer) Clone() credentials.TransportCredentials {
	return &alpnEnforcer{
		delegate: a.delegate.Clone(),
	}
}

[...]

PR #7184, opened 2024-05-03
credentials/tls: reject connections with ALPN disabled

arjan-bal:
Fixes https://github.com/grpc/grpc-go/issues/434

Since HTTP/2 mandates the use of ALPN during the TLS handshake to establish connections, gRPC clients will reject connections to servers that don't have ALPN enabled. gRPC servers will also reject TLS connections from clients with ALPN disabled.
As this change can break existing clients, this behaviour is guarded by an environment variable flag. The error message returned is similar to the [message returned by the C++ implementation](https://github.com/grpc/grpc/blob/master/src/core/lib/security/security_connector/ssl_utils.cc#L138).

RELEASE NOTES: 
- credentials/tls: clients and servers will reject connections that don't support ALPN when environment variable "GRPC_ENFORCE_ALPN_ENABLED" is set to "true" (case insensitive).  This behavior will become the default in a future release.

### DIFF
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -151,8 +151,20 @@ func (c *tlsCreds) ServerHandshake(rawConn net.Conn) (net.Conn, AuthInfo, error)
 		conn.Close()
 		return nil, nil, err
 	}
+	cs := conn.ConnectionState()
+    // The negotiated application protocol can be empty only if the client doesn't 
+    // support ALPN. In such cases, we can close the connection since ALPN is required
+    // for using HTTP/2 over TLS.
+	if cs.NegotiatedProtocol == "" {
+		if envconfig.EnforceALPNEnabled {
+			conn.Close()
+			return nil, nil, fmt.Errorf("credentials: cannot check peer: missing selected ALPN property")
+		} else if logger.V(2) {
+			logger.Info("Allowing TLS connection from client with ALPN disabled. TLS connections with ALPN disabled will be disallowed in future grpc-go releases")
+		}
+	}
 	tlsInfo := TLSInfo{
-		State: conn.ConnectionState(),
+		State: cs,
 		CommonAuthInfo: CommonAuthInfo{
 			SecurityLevel: PrivacyAndIntegrity,
 		},
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -152,9 +152,9 @@ func (c *tlsCreds) ServerHandshake(rawConn net.Conn) (net.Conn, AuthInfo, error)
 		return nil, nil, err
 	}
 	cs := conn.ConnectionState()
-    // The negotiated application protocol can be empty only if the client doesn't 
-    // support ALPN. In such cases, we can close the connection since ALPN is required
-    // for using HTTP/2 over TLS.
+	// The negotiated application protocol can be empty only if the client doesn't
+	// support ALPN. In such cases, we can close the connection since ALPN is required
+	// for using HTTP/2 over TLS.
 	if cs.NegotiatedProtocol == "" {
 		if envconfig.EnforceALPNEnabled {
 			conn.Close()
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -27,8 +27,12 @@ import (
 	"net/url"
 	"os"
 
+	"google.golang.org/grpc/grpclog"
 	credinternal "google.golang.org/grpc/internal/credentials"
+	"google.golang.org/grpc/internal/envconfig"
 )
+
+var logger = grpclog.Component("credentials")
 
 // TLSInfo contains the auth information for a TLS authenticated connection.
 // It implements the AuthInfo interface.
@@ -113,17 +117,20 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
 		return nil, nil, ctx.Err()
 	}
 
-    // The negotiated protocol can be either of the following:
-    // 1. h2: When the server supports ALPN. Only HTTP/2 can be negotiated since
-    //    it is the only protocol advertised by the client during the handshake.
-    //    The tls library ensures that the server chooses a protocol advertised
-    //    by the client.
-    // 2. "" (empty string): If the server doesn't support ALPN. ALPN is a requirement
-    //    for using HTTP/2 over TLS. We can terminate the connection immediately.
-    np := conn.ConnectionState().NegotiatedProtocol
-    if np == "" {
-		_ = conn.Close()
-		return nil, nil, fmt.Errorf("cannot check peer: missing selected ALPN property")
+	// The negotiated protocol can be either of the following:
+	// 1. h2: When the server supports ALPN. Only HTTP/2 can be negotiated since
+	//    it is the only protocol advertised by the client during the handshake.
+	//    The tls library ensures that the server chooses a protocol advertised
+	//    by the client.
+	// 2. "" (empty string): If the server doesn't support ALPN. ALPN is a requirement
+	//    for using HTTP/2 over TLS. We can terminate the connection immediately.
+	np := conn.ConnectionState().NegotiatedProtocol
+	if np == "" {
+		if envconfig.EnforceALPNEnabled {
+			_ = conn.Close()
+			return nil, nil, fmt.Errorf("cannot check peer: missing selected ALPN property")
+		}
+		logger.Warning("Allowing TLS connection to server %q with ALPN disabled")
 	}
 	tlsInfo := TLSInfo{
 		State: conn.ConnectionState(),
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -112,6 +112,19 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
 		conn.Close()
 		return nil, nil, ctx.Err()
 	}
+
+    // The negotiated protocol can be either of the following:
+    // 1. h2: When the server supports ALPN. Only HTTP/2 can be negotiated since
+    //    it is the only protocol advertised by the client during the handshake.
+    //    The tls library ensures that the server chooses a protocol advertised
+    //    by the client.
+    // 2. "" (empty string): If the server doesn't support ALPN. ALPN is a requirement
+    //    for using HTTP/2 over TLS. We can terminate the connection immediately.
+    np := conn.ConnectionState().NegotiatedProtocol
+    if np == "" {
+		_ = conn.Close()
+		return nil, nil, fmt.Errorf("cannot check peer: missing selected ALPN property")
+	}
 	tlsInfo := TLSInfo{
 		State: conn.ConnectionState(),
 		CommonAuthInfo: CommonAuthInfo{
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -127,10 +127,10 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
 	np := conn.ConnectionState().NegotiatedProtocol
 	if np == "" {
 		if envconfig.EnforceALPNEnabled {
-			_ = conn.Close()
-			return nil, nil, fmt.Errorf("cannot check peer: missing selected ALPN property")
+			conn.Close()
+			return nil, nil, fmt.Errorf("credentials: cannot check peer: missing selected ALPN property")
 		}
-		logger.Warning("Allowing TLS connection to server %q with ALPN disabled")
+		logger.Warningf("Allowing TLS connection to server %q with ALPN disabled. TLS connections to servers with ALPN disabled will be disallowed in future grpc-go releases", cfg.ServerName)
 	}
 	tlsInfo := TLSInfo{
 		State: conn.ConnectionState(),
--- a/credentials/tls_ext_test.go
+++ b/credentials/tls_ext_test.go
@@ -385,11 +385,8 @@ func (s) TestTLS_DisabledALPNServer(t *testing.T) {
 			}
 			defer conn.Close()
 
-			select {
-			case err := <-errCh:
-				if err != nil {
-					t.Fatalf("Unexpected server error: %v", err)
-				}
+			if err := <-errCh; err != nil {
+				t.Fatalf("Unexpected server error: %v", err)
 			}
 		})
 	}
--- a/credentials/tls_ext_test.go
+++ b/credentials/tls_ext_test.go
@@ -236,3 +236,49 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
 		t.Fatalf("EmptyCall err = %v; want <nil>", err)
 	}
 }
+
+// TestTLS_DisabledALPN tests the behaviour of a gRPC client when connecting to
+// a server that doesn't support ALPN.
+func (s) TestTLS_DisabledALPN(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	// Start a non gRPC TLS server.
+	config := &tls.Config{
+		Certificates: []tls.Certificate{serverCert},
+		NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
+	}
+	listner, err := tls.Listen("tcp", ":0", config)
+	if err != nil {
+		t.Fatalf("Error starting TLS server: %v", err)
+	}
+	defer listner.Close()
+
+	// Start listening for server requests in a new go routine.
+	go func() {
+		conn, err := listner.Accept()
+		if err != nil {
+			t.Errorf("tls.Accept failed err = %v", err)
+		} else {
+            _, _ = conn.Write([]byte("Hello, World!"))
+			_ = conn.Close()
+		}
+	}()
+
+	clientCreds := credentials.NewTLS(&tls.Config{
+		ServerName: serverName,
+		RootCAs:    certPool,
+	})
+
+	cc, err := grpc.NewClient("dns:"+listner.Addr().String(), grpc.WithTransportCredentials(clientCreds))
+	if err != nil {
+		t.Fatalf("grpc.NewClient error: %v", err)
+	}
+	defer cc.Close()
+	client := testgrpc.NewTestServiceClient(cc)
+
+	const wantStr = "missing selected ALPN property"
+	if _, err = client.EmptyCall(ctx, &testpb.Empty{}); status.Code(err) != codes.Unavailable || !strings.Contains(status.Convert(err).Message(), wantStr) {
+		t.Fatalf("EmptyCall err = %v; want code=%v, message contains %q", err, codes.Unavailable, wantStr)
+	}
+}
--- a/credentials/tls_ext_test.go
+++ b/credentials/tls_ext_test.go
@@ -239,18 +239,13 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
 	}
 }
 
-// TestTLS_DisabledALPN tests the behaviour of a gRPC client when connecting to
-// a server that doesn't support ALPN.
+// TestTLS_DisabledALPN tests the behaviour of a TransportCredentials when
+// connecting to a server that doesn't support ALPN.
 func (s) TestTLS_DisabledALPN(t *testing.T) {
 	initialVal := envconfig.EnforceALPNEnabled
 	defer func() {
 		envconfig.EnforceALPNEnabled = initialVal
 	}()
-
-	serverCfg := &tls.Config{
-		Certificates: []tls.Certificate{serverCert},
-		NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
-	}
 
 	tests := []struct {
 		name         string
@@ -271,19 +266,25 @@ func (s) TestTLS_DisabledALPN(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			envconfig.EnforceALPNEnabled = tc.alpnEnforced
 
-			listner, err := tls.Listen("tcp", "localhost:0", serverCfg)
+			listner, err := tls.Listen("tcp", "localhost:0", &tls.Config{
+				Certificates: []tls.Certificate{serverCert},
+				NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
+			})
 			if err != nil {
 				t.Fatalf("Error starting TLS server: %v", err)
 			}
 
+			errCh := make(chan error)
+
 			go func() {
 				conn, err := listner.Accept()
 				if err != nil {
-					t.Errorf("tls.Accept failed err = %v", err)
+					errCh <- fmt.Errorf("tls.Accept failed: %v", err)
 				} else {
 					conn.Write([]byte("Hello, World!"))
 					conn.Close()
 				}
+				close(errCh)
 			}()
 
 			serverAddr := listner.Addr().String()
@@ -296,13 +297,18 @@ func (s) TestTLS_DisabledALPN(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 			defer cancel()
 
-			clientCfg := tls.Config{ServerName: serverName,
-				RootCAs: certPool,
+			clientCfg := tls.Config{
+				ServerName: serverName,
+				RootCAs:    certPool,
 			}
 			_, _, err = credentials.NewTLS(&clientCfg).ClientHandshake(ctx, serverName, conn)
 
 			if gotErr := (err != nil); gotErr != tc.wantErr {
-				t.Errorf("ClientHandshake returned unexpected error: got=%v, want=%t", gotErr, tc.wantErr)
+				t.Errorf("ClientHandshake returned unexpected error: got=%v, want=%t", err, tc.wantErr)
+			}
+
+			if err = <-errCh; err != nil {
+				t.Fatalf("Server reported unexpected error: %v", err)
 			}
 		})
 	}
--- a/credentials/tls_ext_test.go
+++ b/credentials/tls_ext_test.go
@@ -23,6 +23,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"net"
 	"os"
 	"strings"
 	"testing"
@@ -241,7 +242,6 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
 // TestTLS_DisabledALPN tests the behaviour of a gRPC client when connecting to
 // a server that doesn't support ALPN.
 func (s) TestTLS_DisabledALPN(t *testing.T) {
-	// ctx := context.Background()
 
 	initialVal := envconfig.EnforceALPNEnabled
 	defer func() {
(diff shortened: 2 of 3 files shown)


credentials/tls.go
```diff
@@ -112,6 +116,22 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
...
+
+	// The negotiated protocol can be either of the following:
+	// 1. h2: When the server supports ALPN. Only HTTP/2 can be negotiated since
+	//    it is the only protocol advertised by the client during the handshake.
+	//    The tls library ensures that the server chooses a protocol advertised
+	//    by the client.
+	// 2. "" (empty string): If the server doesn't support ALPN. ALPN is a requirement
+	//    for using HTTP/2 over TLS. We can terminate the connection immediately.
+	np := conn.ConnectionState().NegotiatedProtocol
+	if np == "" {
```
purnesh42H, line 128: In addition to checking if np is empty (""), shouldn't we also validate the value of np against a list of unsupported protocol like "h2c" (as called out in the [issue](https://github.com/grpc/grpc-go/issues/434))? Alternatively, we can flip the check to np != "h2" if that's the only supported protocol or can maintain a list of supported protocols.

arjan-bal: I checked all the places where a new tls.Config is being created and we're setting the NextProtos field to `{"h2"}` everywhere. However, we're also calling the `AppendH2ToNextProtos` method (https://github.com/grpc/grpc-go/blob/master/internal/credentials/util.go#L28-L37) in all the constructions of the TLS `TransportCredentials`. It seems that gRPC code is written to allow addition of more protocols in the future.

We are using the crypto/tls library to do the TLS handshake on the client side. If we set only "h2" in the `tls.Config.NextProtos` slice, [this function](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_client.go#L828-L845) ensures that the application protocol selected by the server is present in the list supported by the client. If not, the tls library fails the handshake. So the re-validation of protocol selected by the server in gRPC code seems redundant.

purnesh42H: Thanks! This has been discussed in the issue #434 . I think I might have missed it. Appreciate summarizing it here.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+	defer listner.Close()
+
+	tests := []struct {
+		description            string
+		alpnEnforced           bool
+		wantErrMatchPattern    string
+		wantErrNonMatchPattern string
+	}{
+		{
+			description:         "enforced",
```
purnesh42H, line 271: nit: description of tests have more details. For example, its hard to understand which one of these is a success test for handshake. Alternatively, can add comments as well explaining the intent

go/go-test-examples#descriptions

purnesh42H (approved):
LGTM overall! Thanks for picking this up

credentials/tls_ext_test.go
arjan-bal: This field is actually the test name and not a description. Renamed the field accordingly. Based on [the Google style guide](https://google.github.io/styleguide/go/decisions.html#subtest-names), the test names should be similar to identifiers like function names to help filter and identify tests instead of a prose description.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			client := testgrpc.NewTestServiceClient(cc)
+			_, rpcErr := client.EmptyCall(ctx, &testpb.Empty{})
+
+			if gotCode := status.Code(rpcErr); gotCode != codes.Unavailable {
+				t.Errorf("EmptyCall returned unexpected code: got=%v, want=%v", gotCode, codes.Unavailable)
+			}
+
+			matchPat, err := regexp.Compile(tc.wantErrMatchPattern)
+			if err != nil {
+				t.Fatalf("Error message match pattern %q is invalid due to error: %v", tc.wantErrMatchPattern, err)
```
purnesh42H, line 319: it should be changed to `got` before `want`? Similar for line 328

https://google.github.io/styleguide/go/decisions#got-before-want

credentials/tls_ext_test.go
purnesh42H: Thanks for renaming? So, we don't need description here?

credentials/tls_ext_test.go
arjan-bal: The regex pattern is expected to always be valid, the want is implicitly `nil`.  There is not `want` in this message. If this line executes, it means that the test case is invalid.

credentials/tls_ext_test.go
arjan-bal: Having a test description is optional.

credentials/tls.go
```diff
@@ -112,6 +116,22 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
...
+		if envconfig.EnforceALPNEnabled {
+			_ = conn.Close()
+			return nil, nil, fmt.Errorf("cannot check peer: missing selected ALPN property")
+		}
+		logger.Warning("Allowing TLS connection to server %q with ALPN disabled")
```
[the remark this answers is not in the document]
easwars: Also, you are missing the server name in the argument list even though you have the `%q` verb.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+	defer func() {
+		envconfig.EnforceALPNEnabled = initialVal
+	}()
+
+	// Start a non gRPC TLS server.
+	config := &tls.Config{
+		Certificates: []tls.Certificate{serverCert},
+		NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
+	}
+	listner, err := tls.Listen("tcp", ":0", config)
```
easwars, line 258: We have had some issues in the past with starting a listener with address in `:0` (especially on forge). One option might be to use `localhost:0` here, or to use https://github.com/grpc/grpc-go/blob/5d24ee2bd170a0c2c636b7d1ebc7ebdd3ecf81ac/internal/testutils/local_listener.go#L24 and pass it to https://pkg.go.dev/crypto/tls#NewListener instead.

easwars: Also, we have had issues in the past where common stuff like this is started once for the whole test instead of once per entry in the test table below (making the sub-tests not completely hermetic). I would recommend moving this to be inside `t.Run`.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+	tests := []struct {
+		name                   string
+		alpnEnforced           bool
+		wantErrMatchPattern    string
+		wantErrNonMatchPattern string
+	}{
+		{
+			name:                "enforced",
+			alpnEnforced:        true,
+			wantErrMatchPattern: "transport: .*missing selected ALPN property",
```
easwars, line 273: Optional/Nit: `wanErr` or `wantErrStr` would be a more commonly used variable name in our codebase.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+		},
+		{
+			name:                   "not_enforced",
+			wantErrNonMatchPattern: "transport:",
+		},
+		{
+			name:                   "default_value",
+			wantErrNonMatchPattern: "transport:",
+			alpnEnforced:           initialVal,
+		},
```
easwars, line 283: Is this sub-test useful? If so, how? Also, this would need to change when the default value changes, right?

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+		wantErrNonMatchPattern string
+	}{
+		{
+			name:                "enforced",
+			alpnEnforced:        true,
+			wantErrMatchPattern: "transport: .*missing selected ALPN property",
+		},
+		{
+			name:                   "not_enforced",
+			wantErrNonMatchPattern: "transport:",
```
easwars, line 277: Can we have something more than just `transport:` here?

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			envconfig.EnforceALPNEnabled = tc.alpnEnforced
+			// Listen to one TCP connection request.
+			go func() {
+				conn, err := listner.Accept()
+				if err != nil {
+					t.Errorf("tls.Accept failed err = %v", err)
+				} else {
+					_, _ = conn.Write([]byte("Hello, World!"))
```
easwars, line 295: Do something useful with the error value (maybe call t.Error()) or completely ignore it. I don't see the value of assigning it to a blank identifier.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,99 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			}
+
+			nonMatchPat, err := regexp.Compile(tc.wantErrNonMatchPattern)
+			if err != nil {
+				t.Fatalf("Error message non-match pattern %q is invalid due to error: %v", tc.wantErrNonMatchPattern, err)
+			}
+
+			if tc.wantErrNonMatchPattern != "" && nonMatchPat.MatchString(status.Convert(rpcErr).Message()) {
+				t.Errorf("EmptyCall err = %v; want pattern missing %q", rpcErr, nonMatchPat)
+			}
```
easwars, line 333: This whole test logic is quite convoluted. Can you please let me know what exactly you are trying to validate. We should try to simplify this.

arjan-bal: I wanted to validate that an RPC call is rejected when ALPN is enforced by the client and allowed when its not enforced. I was facing the following problems:
1. The gRPC server code automatically adds "h2" to list of supported protocols (named NextProtos), making it hard to disable ALPN. I was able to overcome this by providing a TLS listener to the gRPC server which completes the handshake and not providing any `TranasportCredentials` `ServerOption`.
2. The next problem was that when the client handshake fails, the client was retrying the requests till 10 seconds and getting a `DeadlineExceeded` status. This makes the test waste 10 seconds. I think this would require overriding the timeouts by providing a `Service Config`. This was making things complicated.
3. So I created a non gRPC server that just does a TLS handshake. Now the gRPC client would always see an error when sending it requests, but at different stages depending on if the TLS handshake succeeds. So I was trying to check if the error is a TLS handshake failure or "anything else". Since we're not wrapping any special error types for TLS failures, I was using the error message instead.

I've avoided all this now by testing the `ClientHandshake` method directly without creating a gRPC server or client. Please have another look.

credentials/tls_ext_test.go
arjan-bal: Ignoring the return values completely now. I had added these to make it explicit that an error is being ignored since IDEs don't show any warning for this.

credentials/tls_ext_test.go
arjan-bal: The test has been reworked, please take another look.

credentials/tls_ext_test.go
arjan-bal: Removed. This was added to catch changes in the default behaviour so that we could use this test case to verify the changes to the default in the future.

credentials/tls_ext_test.go
arjan-bal: Changed to wantErr.

credentials/tls_ext_test.go
arjan-bal: Moved inside the closure.

credentials/tls.go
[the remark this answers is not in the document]
arjan-bal: >Please append something like this to this warning log. TLS connections to servers with ALPN disabled will be disallowed in future grpc-go releases. This will increase the possibility of this being picked up by the user.

Changed the log message. 

> Also, you are missing the server name in the argument list even though you have the %q verb.

Nice catch! Changed to `Warningf` and added the server name.

credentials/tls.go
```diff
@@ -112,6 +116,22 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
...
+	// 1. h2: When the server supports ALPN. Only HTTP/2 can be negotiated since
+	//    it is the only protocol advertised by the client during the handshake.
+	//    The tls library ensures that the server chooses a protocol advertised
+	//    by the client.
+	// 2. "" (empty string): If the server doesn't support ALPN. ALPN is a requirement
+	//    for using HTTP/2 over TLS. We can terminate the connection immediately.
+	np := conn.ConnectionState().NegotiatedProtocol
+	if np == "" {
+		if envconfig.EnforceALPNEnabled {
+			_ = conn.Close()
```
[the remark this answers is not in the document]
arjan-bal: I just wanted to make it explicit to readers that an error is being ignored. Removed it now.

credentials/tls.go
```diff
@@ -112,6 +116,22 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
...
+			return nil, nil, fmt.Errorf("cannot check peer: missing selected ALPN property")
```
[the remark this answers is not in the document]
arjan-bal: Done. Noted for the future.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,72 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
 		t.Fatalf("EmptyCall err = %v; want <nil>", err)
 	}
 }
+
+// TestTLS_DisabledALPN tests the behaviour of a gRPC client when connecting to
+// a server that doesn't support ALPN.
```
easwars, line 243: This comment needs updating since we are not creating a grpc client anymore and only testing the ClientHandshake method on the TLS creds.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,72 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+				t.Fatalf("net.Dial(%s) failed: %v", serverAddr, err)
+			}
+			defer conn.Close()
+
+			ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+			defer cancel()
+
+			clientCfg := tls.Config{ServerName: serverName,
+				RootCAs: certPool,
+			}
```
easwars, line 301: Nit: Please use

```golang
clientCfg := tls.Config{
  ServerName: serverName,
  RootCAs: certPool,
}
```

Or

```golang
clientCfg := tls.Config{ServerName: serverName, RootCAs: certPool}
```

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,72 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+func (s) TestTLS_DisabledALPN(t *testing.T) {
+	initialVal := envconfig.EnforceALPNEnabled
+	defer func() {
+		envconfig.EnforceALPNEnabled = initialVal
+	}()
+
+	serverCfg := &tls.Config{
+		Certificates: []tls.Certificate{serverCert},
+		NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
+	}
```
easwars, line 253: Optional nit: This could be inline in the call to `tls.Listen()`.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,72 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Error starting TLS server: %v", err)
+			}
+
+			go func() {
+				conn, err := listner.Accept()
+				if err != nil {
+					t.Errorf("tls.Accept failed err = %v", err)
+				} else {
+					conn.Write([]byte("Hello, World!"))
```
easwars, line 284: What purpose does this call to `Write` serve? Can we get rid of it?

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,72 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			envconfig.EnforceALPNEnabled = tc.alpnEnforced
+
+			listner, err := tls.Listen("tcp", "localhost:0", serverCfg)
+			if err != nil {
+				t.Fatalf("Error starting TLS server: %v", err)
+			}
+
+			go func() {
```
easwars, line 279: Does it make sense to push any errors seen in this goroutine to a channel and have the main test goroutine block on that channel at the end, and ensure that a non-nil error is returned?

easwars (commented):
Changes look mostly good to me. Just some minor nits this time around.

credentials/tls_ext_test.go
arjan-bal: Since this a tls listner, it automatically does a TLS handshake when writing for the first time. An alternate way is to cast the net.Listener into a tls.Listener and call it's handshake method.

credentials/tls_ext_test.go
arjan-bal: Done.

credentials/tls_ext_test.go
arjan-bal: Done.

credentials/tls_ext_test.go
arjan-bal: Fixed.

credentials/tls_ext_test.go
arjan-bal: Missed that, fixed now.

credentials/tls_ext_test.go
easwars: Just a one liner comment to say that this call to `Write` is to ensure that the underlying handshake happens would be helpful for a future reader of this code. Thanks.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,78 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			}
+			_, _, err = credentials.NewTLS(&clientCfg).ClientHandshake(ctx, serverName, conn)
+
+			if gotErr := (err != nil); gotErr != tc.wantErr {
+				t.Errorf("ClientHandshake returned unexpected error: got=%v, want=%t", err, tc.wantErr)
+			}
+
+			if err = <-errCh; err != nil {
+				t.Fatalf("Server reported unexpected error: %v", err)
+			}
```
easwars, line 312: This should also `select` on the test context expiring. That way, when something bad happens and there is no write on `errCh`, we can still guarantee that the test fails within a reasonable amount of time (in the order of seconds) and not wait for the overall test timeout configured on the `go test` command. We use this pattern a lot in our tests, because we were bitten by those kinds of long test failures in the past.

easwars (approved):
Thanks for making this change. Long time coming!!

dfawley:
> Clients will reject connections to servers that don't support ALPN when environment variable "GRPC_ENFORCE_ALPN_ENABLED" is set to "1", "t", "T", "true", "TRUE" or "True".

This doesn't seem right.  IIUC it needs to be "true" (case insensitive).

And do we not need a similar check on the server?

arjan-bal:
> > Clients will reject connections to servers that don't support ALPN when environment variable "GRPC_ENFORCE_ALPN_ENABLED" is set to "1", "t", "T", "true", "TRUE" or "True".
> 
> This doesn't seem right. IIUC it needs to be "true" (case insensitive).

You are correct. I confused environment variables with with command line flags. Updated the PR description.

> And do we not need a similar check on the server?

During the TLS handshake, the server decides the application layer protocol based on the intersection of the client's `NextProto` list and the server's `NextProto` list. The following line in the credentials constructor ensures "h2" (the protocol name for http/2) is always present in the server's protocol list list.
https://github.com/grpc/grpc-go/blob/0020ccf9d9fc448fcfa099173df4ad330e96326c/credentials/tls.go#L172

Looking into the code for the server side handshake, it looks like a gRPC server may accept connections from clients which don't specify the ALPN option.
https://github.com/golang/go/blob/a32e94d43ee34cc6de23ff674c4f5e5b49528e86/src/crypto/tls/handshake_server.go#L290-L296
I think server side changes will also be required. Putting the PR in draft for the time being.

arjan-bal:
@dfawley @easwars added the checks for ALPN in the ServerHandshake method to ensure clients also have ALPN enabled. Added unit tests for the same. Please have another look.

credentials/tls_ext_test.go
arjan-bal: Added a comment.

credentials/tls_ext_test.go
arjan-bal: Done.

credentials/tls.go
```diff
@@ -131,8 +151,20 @@ func (c *tlsCreds) ServerHandshake(rawConn net.Conn) (net.Conn, AuthInfo, error)
...
+	// The negotiated application protocol can be empty only if the client doesn't
+	// support ALPN. In such cases, we can close the connection since ALPN is required
+	// for using HTTP/2 over TLS.
+	if cs.NegotiatedProtocol == "" {
+		if envconfig.EnforceALPNEnabled {
+			conn.Close()
+			return nil, nil, fmt.Errorf("credentials: cannot check peer: missing selected ALPN property")
+		} else if logger.V(2) {
+			logger.Info("Allowing TLS connection from client with ALPN disabled. TLS connections with ALPN disabled will be disallowed in future grpc-go releases")
+		}
```
arjan-bal, line 164: I'm logging at Debug level here as this may add too much noise in the server logs.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,161 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+
+			listener, err := tls.Listen("tcp", "localhost:0", &tls.Config{
+				Certificates: []tls.Certificate{serverCert},
+				NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
+			})
+			if err != nil {
+				t.Fatalf("Error starting TLS server: %v", err)
+			}
+
+			errCh := make(chan error)
```
easwars, line 277: Nit: Make the channel with a buffer of 1, so that the writing goroutine will not get blocked if the main test goroutine does not get around to reading from it.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,161 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			select {
+			case err = <-errCh:
+				break
+			case <-ctx.Done():
+				err = ctx.Err()
+			}
+
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
```
easwars, line 322: This could instead be:

```golang
select {
	case err := <-errCh:
		if err != nil { 
			t.Fatal(err)
		}
	case <-ctx.Done():
		t.Fatal("Timeout when waiting for error from server")
}
```

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,161 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			listener, err := tls.Listen("tcp", "localhost:0", &tls.Config{
+				Certificates: []tls.Certificate{serverCert},
+				NextProtos:   []string{}, // Empty list indicates ALPN is disabled.
+			})
+			if err != nil {
+				t.Fatalf("Error starting TLS server: %v", err)
+			}
+
+			errCh := make(chan error)
+
```
easwars, line 278: Nit: nix this newline.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,161 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			envconfig.EnforceALPNEnabled = tc.alpnEnforced
+
+			listener, err := net.Listen("tcp", "localhost:0")
+			if err != nil {
+				t.Fatalf("Error starting server: %v", err)
+			}
+
+			errCh := make(chan error)
+
```
easwars, line 360: Same comments as in the other test.

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,161 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+			}
+			conn, err := tls.Dial("tcp", serverAddr, clientCfg)
+			if err != nil {
+				t.Fatalf("tls.Dial(%s) failed: %v", serverAddr, err)
+			}
+			defer conn.Close()
+
+			if err := <-errCh; err != nil {
+				t.Fatalf("Unexpected server error: %v", err)
+			}
```
easwars, line 395: We need to do the `select` here as well.

arjan-bal: There was only one channel to wait on since credentials.ServerHandshake doesn't accept a ctx so I used an `if`. Changed to `select` now.

credentials/tls_ext_test.go
arjan-bal: Removed.

credentials/tls_ext_test.go
arjan-bal: Removed.

credentials/tls_ext_test.go
arjan-bal: Changed.

credentials/tls_ext_test.go
arjan-bal: Done

credentials/tls_ext_test.go
```diff
@@ -236,3 +238,159 @@ func (s) TestTLS_CipherSuitesOverridable(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			envconfig.EnforceALPNEnabled = tc.alpnEnforced
+
+			listener, err := net.Listen("tcp", "localhost:0")
+			if err != nil {
+				t.Fatalf("Error starting server: %v", err)
+			}
+
+			errCh := make(chan error)
```
easwars, line 355: This one needs a buffer of size 1 too.

credentials/tls_ext_test.go
arjan-bal: `vet` test is throwing the following error now:
```
Error: credentials/tls_ext_test.go:388:4: should use a simple channel send/receive instead of select with a single case (S1000)
```
Changed back to `if` statement.

easwars: What I meant was something like:

```golang
select {
 case <-time.After(defaultTestTimeout):
   t.Fatal():
 case err := <-errCh:
   if err != nil {
     t.Fatal()
}
```

credentials/tls_ext_test.go
arjan-bal: Changed.

credentials/tls_ext_test.go
arjan-bal: Changed to select with `time.After`

easwars (approved):

easwars:
@dfawley : For second set of eyes.

dfawley (approved):

PR #7190, opened 2024-05-05
update a running server

TalLerner:
Required for ongoing update of certificates


TalLerner:
bot:retest

arjan-bal:
Can you please open an issue describing what your use case and want to achieve? There may be a better way of solving the problem that doesn't require a code change. 

Also, allowing updating of server options could potentially break code which depends on them being immutable.

PR #7202, opened 2024-05-06
advancedTLS: Combine `ClientOptions` and `ServerOptions` to just `Options`

gtcooke94:
`ClientOptions` and `ServerOptions` were basically identical minus a few fields. Duplicating so much code and documentation is messy. Given both have a single field each that are not relevant to the other, I think that we can just combine them and document that.

RELEASE NOTES: none

### DIFF
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -184,66 +184,16 @@ const (
 )
 
 // ClientOptions contains the fields needed to be filled by the client.
-type ClientOptions struct {
-	// IdentityOptions is OPTIONAL on client side. This field only needs to be
-	// set if mutual authentication is required on server side.
-	IdentityOptions IdentityCertificateOptions
-	// AdditionalPeerVerification is a custom verification check after certificate signature
-	// check.
-	// If this is set, we will perform this customized check after doing the
-	// normal check(s) indicated by setting VerificationType.
-	AdditionalPeerVerification PostHandshakeVerificationFunc
-	// VerifyPeer is a custom verification check after certificate signature
-	// check.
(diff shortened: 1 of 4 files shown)


security/advancedtls/advancedtls.go
```diff
@@ -184,66 +184,18 @@ const (
...
 // ClientOptions contains the fields needed to be filled by the client.
-type ClientOptions struct {
+// Deprecated: use Options instead.
+type ClientOptions = Options
+
+// Options contains the fields needed to be filled by the server.
+// Deprecated: use Options instead.
+type ServerOptions = Options
+
+// Options contains the fields a user can configure when settings up TLS clients
```
dfawley, line 194: *setting up

security/advancedtls/advancedtls.go
```diff
@@ -261,9 +213,11 @@ type ServerOptions struct {
 	// RootOptions is OPTIONAL on server side. This field only needs to be set if
 	// mutual authentication is required(RequireClientCert is true).
 	RootOptions RootCertificateOptions
-	// If the server want the client to send certificates.
+	// If the server want the client to send certificates. This value is only
```
dfawley, line 216: want -> wants.  Or "requires" probably is more precise.

dfawley (approved):
LGTM, just a few minor things.

security/advancedtls/advancedtls.go
gtcooke94: Fixed

security/advancedtls/advancedtls.go
gtcooke94: Fixed

PR #7204, opened 2024-05-06
peer: remove change detector test

arvindbr8:
This change detector test seems asserts for an exact match and relies on the go std library print statements which can be subjected to change. 

RELEASE NOTES: none


zasweq (approved):
LGTM.

PR #7205, opened 2024-05-07
stats/opentelemetry: Add CSM Plugin Option

zasweq:
This PR adds support for the CSM Plugin Option, as speced in the design. The interface and the implementation live in OpenTelemetry's internal/. The next step is the usage of the CSM Plugin Option in OpenTelemetry and the global configuration.

RELEASE NOTES: N/A


stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	appendToLabelsFromMetadata(labels, "csm.remote_workload_type", "type", fields)
+	// The value of “csm.remote_workload_canonical_service” comes from
+	// MetadataExchange with the key “canonical_service”. (Note that this should
+	// be read even if the remote type is unknown.)
+	appendToLabelsFromMetadata(labels, "csm.remote_workload_canonical_service", "canonical_service", fields)
+
+	// Unset/unknown types, and types that aren't GKE or GCP return early with
+	// just local labels, remote_workload_type and
+	// remote_workload_canonical_service labels.
+	typeVal := labels["csm.remote_workload_type"]
```
dfawley, line 107: s/typeVal/workloadType/?

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	if val, ok := os.LookupEnv(envvar); ok {
+		envvarVal = val
+	}
+	labels[labelKey] = envvarVal
+}
+
+var (
+	// This function will be overridden in unit tests.
+	getAttrSetFromResourceDetector = func() *attribute.Set {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
dfawley, line 167: Only main may call `context.Background`...  Seems like the thing that creates the plugin needs to accept a context parameter and we plumb it here?

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	// metadataExchangeLabelsEncoded are the metadata exchange labels to be sent
+	// as the value of metadata key "x-envoy-peer-metadata" in proto wire format
+	// and base 64 encoded. This gets sent out from all the servers running in
+	// this process and for csm channels.
+	metadataExchangeLabelsEncoded string
+}
+
+// NewPluginOption returns a new PluginOption with local labels and metadata
+// exchange labels derived from the environment.
+func NewPluginOption() internal.PluginOption {
```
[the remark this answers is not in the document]
zasweq: This is in internal/csm, so the only thing able to use it is the OpenTelemetry instrumentation right? This has to be exported or else the external facing csm package, which I'll have at opentelemetry/csm. Nothing else can access this expect through this repo which we control?

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	"go.opentelemetry.io/contrib/detectors/gcp"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+)
+
+var logger = grpclog.Component("csm-observability-plugin")
+
+// AddLabels adds CSM labels to the provided context's metadata, as a encoded
+// protobuf Struct as the value of x-envoy-metadata.
+func (cpo *PluginOption) AddLabels(ctx context.Context) context.Context {
```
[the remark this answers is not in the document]
zasweq: Ah, good point. Done.

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	"google.golang.org/protobuf/types/known/structpb"
+
+	"go.opentelemetry.io/contrib/detectors/gcp"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+)
+
+var logger = grpclog.Component("csm-observability-plugin")
+
+// AddLabels adds CSM labels to the provided context's metadata, as a encoded
```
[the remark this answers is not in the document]
zasweq: Done for both.

stats/opentelemetry/internal/csm/pluginoption.go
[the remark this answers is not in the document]
zasweq: Right now I have this referenced in OpenTelemetry stats handler (pluginOption internal.PluginOption), which is created by calling this.

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// metadataExchangeKey is the key for HTTP metadata exchange.
+const metadataExchangeKey = "x-envoy-peer-metadata"
+
+// PluginOption emits CSM Labels from the environment and metadata exchange
+// for csm channels and all servers.
+//
+// Do not use this directly; use NewPluginOption instead.
+type PluginOption struct {
```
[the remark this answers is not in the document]
zasweq: Unexported, will this wokr if I call NewPluginOption from opentelemetry (for stats handler to hold onto this).

stats/opentelemetry/internal/pluginoption.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// This API is experimental. In the future, we hope to stabilize and expose this
+// API to allow pluggable plugin options to allow users to inject labels of
+// their choosing into metrics recorded.
+type PluginOption interface {
+	// AddLabels adds metadata exchange labels to the outgoing metadata of the
+	// context.
+	AddLabels(context.Context) context.Context
+	// NewLabelsMD creates metadata exchange labels as a new MD.
+	NewLabelsMD() metadata.MD
```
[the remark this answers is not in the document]
zasweq: This is what I talked to you about yesterday offline. Client side and server side have different data/object representations for the same gRPC/HTTP concept of "metadata". On client, the way we discussed was to simply append the metadata exchange labels to outgoing context in unary/streaming interceptors. On server side, I attach this to the wrapped stream in the interceptor as a metadata object to be appended to intercepted headers or trailers being sent on the underlying stream object. Thus, I have a simple constructor for MD. I used to have it append, but you mentioned there's no use of the passed in md, so it's best not to take it as a parameter. FWIW, C also has different representations for client and server so they have two different ways of adding as well.

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	// caller only passes in two potential xDS Optional Labels: service_name and
+	// service_namespace.
+	for k, v := range optionalLabels {
+		labels[k] = v
+	}
+	val := md.Get("x-envoy-peer-metadata")
+	// This can't happen if corresponding csm client because of proto wire
+	// format encoding, but since it is arbitrary off the wire be safe.
+	if len(val) != 1 {
+		return labels
```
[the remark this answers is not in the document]
zasweq: Ah, I had logs here but deleted them since I didn't want to log spam, but yeah this is probably useful information.

stats/opentelemetry/internal/pluginoption.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// PluginOption is the interface which represents a plugin option for the
+// OpenTelemetry instrumentation component. This plugin option emits labels from
+// metadata and also sets labels in different forms of metadata. These labels
+// are intended to be added to applicable OpenTelemetry metrics recorded in the
+// OpenTelemetry instrumentation component.
+//
+// This API is experimental. In the future, we hope to stabilize and expose this
+// API to allow pluggable plugin options to allow users to inject labels of
```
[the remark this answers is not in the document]
zasweq: Switched to that, but I feel like that is the behavior right now, but the user has no way to implement their own, which was what I was getting at. I guess that is implied by "stabilize and expose this API".

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+		if r != nil {
+			set = r.Set()
+		}
+		return set
+	}
+)
+
+// constructMetadataFromEnv creates local labels and labels to send to the peer
+// using metadata exchange based off resource detection and environment
+// variables.
```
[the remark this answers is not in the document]
zasweq: I had the thought for named returns while implementing this, I had an outdated comment talking about global setting for initalizeLocalAndMetadataLabels but not this one. I added a comment for this helper.

stats/opentelemetry/internal/csm/pluginoption.go
zasweq: Done.

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+		"csm.remote_workload_canonical_service": "unknown",
+	}
+	// Append the local labels.
+	for k, v := range cpo.localLabels {
+		labels[k] = v
+	}
+
+	// Append the optional labels. To avoid string comparisons, assume the
+	// caller only passes in two potential xDS Optional Labels: service_name and
+	// service_namespace.
```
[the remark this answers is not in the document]
zasweq: Previously, this layer was meant to "filter" or "eat" labels that weren't one of "service_name" or "service_namespace". I initially allowed all labels through CDS, however switched CDS to only process these two labels if set and if string value: https://github.com/grpc/grpc-go/commit/308dbc446644b2f783c29a7abc03f9bec60e8512#diff-976ef43c72aa5b948da568948f07eaf816e7e7d51704259e3ebafc1d7284d376R89. Way back in the o11y thread I had a back and forth with Yash about what to do in each layer. Eventually, Craig Yash and Mark decided to do this check in xDS Client rather than here to avoid costly string comparisons happening on the RPC path.

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+
+// appendToLabelsFromEnv appends an environment variable key value pair to the
+// labels passed in. It sets "unknown" if environment variable is unset, the
+// environment variable otherwise.
+func appendToLabelsFromEnv(labels map[string]string, labelKey string, envvar string) {
+	envvarVal := "unknown"
+	if val, ok := os.LookupEnv(envvar); ok {
+		envvarVal = val
+	}
+	labels[labelKey] = envvarVal
```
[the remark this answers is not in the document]
zasweq: Done for all 3. Good point.

stats/opentelemetry/internal/csm/pluginoption.go
zasweq: Hmm ok I guess our global csm plugin layer will also take a context in it's constructor.

zasweq: Added context flow through NewPluginOption() -> getAttrSetFromResourceDetector().

stats/opentelemetry/internal/csm/pluginoption.go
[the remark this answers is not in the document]
zasweq: Added warning log.

stats/opentelemetry/internal/csm/pluginoption.go
```diff
@@ -0,0 +1,341 @@
...
+	val := md.Get("x-envoy-peer-metadata")
+	// This can't happen if corresponding csm client because of proto wire
+	// format encoding, but since it is arbitrary off the wire be safe.
+	if len(val) != 1 {
+		return labels
+	}
+
+	protoWireFormat, err := base64.RawStdEncoding.DecodeString(val[0])
+	if err != nil {
+		return labels
```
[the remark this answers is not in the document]
zasweq: Added warning logs.

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package csm contains the implementation of the CSM Plugin Option.
```
dfawley, line 19: Does this make sense to users?  ("What's a 'plugin option' anyway?"  It's an unexported field in a configuration struct, so they should have no idea.)

stats/opentelemetry/internal/pluginoption.go
```diff
@@ -0,0 +1,39 @@
...
+	"google.golang.org/grpc/metadata"
+)
+
+// PluginOption is the interface which represents a plugin option for the
+// OpenTelemetry instrumentation component. This plugin option emits labels from
+// metadata and also creates metadata containing labels. These labels are
+// intended to be added to applicable OpenTelemetry metrics recorded in the
+// OpenTelemetry instrumentation component.
+//
+// This API is experimental. In the future, we hope to stabilize and expose this
```
dfawley, line 30: Since this is unexported, there's no need to say it's "experimental".  (It's not even that level.)

stats/opentelemetry/internal/pluginoption.go
```diff
@@ -0,0 +1,39 @@
...
+// API to allow plugins to inject labels of their choosing into metrics
+// recorded.
+type PluginOption interface {
+	// GetMetadata creates a MD with metadata exchange labels.
+	GetMetadata() metadata.MD
+	// GetLabels emits relevant labels from the metadata provided and the
+	// optional labels provided, alongside any relevant labels.
```
dfawley, line 37: Maybe mention where the metadata and labels come from?  The metadata is (all of?) the incoming RPC metadata, but I'm not sure about the labels.

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+	for k, v := range cpo.localLabels {
+		labels[k] = v
+	}
+
+	// Append the optional labels. To avoid string comparisons, assume the
+	// caller only passes in two potential xDS Optional Labels: service_name and
+	// service_namespace.
+	for k, v := range optionalLabels {
+		labels[k] = v
+	}
```
dfawley, line 97: Why don't we merge these labels externally, rather than passing them in here and merging them?

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+// getFromMetadata gets the value for the metadata key from the protobuf
+// metadata. Returns "unknown" if the metadata is not found in the protobuf
+// metadata, or if the value is not a string value. Returns the string value
+// otherwise.
+func getFromMetadata(metadataKey string, metadata map[string]*structpb.Value) string {
+	ret := "unknown"
+	if metadata != nil {
+		if metadataVal, ok := metadata[metadataKey]; ok {
+			if _, ok := metadataVal.GetKind().(*structpb.Value_StringValue); ok {
+				ret = metadataVal.GetStringValue()
```
dfawley, line 156: Nit: return here to reduce (eliminate) state.  You can delete `ret` then.

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// getFromResource gets the value for the resource key from the attribute set.
+// Returns "unknown" if the resourceKey is not found in the attribute set or is
+// not a string value, the string value otherwise.
+func getFromResource(resourceKey attribute.Key, set *attribute.Set) string {
+	ret := "unknown"
+	if set != nil {
+		if resourceVal, ok := set.Value(resourceKey); ok && resourceVal.Type() == attribute.STRING {
+			ret = resourceVal.AsString()
```
dfawley, line 170: As above.

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	return ret
+}
+
+// getEnv returns "unknown" if environment variable is unset, the environment
+// variable otherwise.
+func getEnv(name string) string {
+	ret := "unknown"
+	if val, ok := os.LookupEnv(name); ok {
+		ret = val
```
dfawley, line 181: As above.

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+		return false
+	}
+
+	if parsedTarget.Scheme == "xds" {
+		if parsedTarget.Host == "" {
+			return true // "In the cases where no authority is mentioned, the authority is assumed to be csm"
+		}
+		return parsedTarget.Host == "traffic-director-global.xds.googleapis.com"
+	}
+	return false
```
dfawley, line 335: Optional: FWIW this can be simplified:

```go
return parsedTarget.Scheme == "xds" && (parsedTarget.Host == "" || parsedTarget.Host == tdGlobal)
```

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+const metadataExchangeKey = "x-envoy-peer-metadata"
+
+func determineTargetCSM(target string) bool {
+	// On the client-side, the channel target is used to determine if a channel is a
+	// CSM channel or not. CSM channels need to have an “xds” scheme and a
+	// "traffic-director-global.xds.googleapis.com" authority. In the cases where no
+	// authority is mentioned, the authority is assumed to be CSM. MetadataExchange
+	// is performed only for CSM channels. Non-metadata exchange labels are detected
+	// as described below.
+	parsedTarget, err := url.Parse(target)
```
dfawley, line 321: Is this getting the unparsed target instead of the parsed target?  Would it be better to pass the parsed one?

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -0,0 +1,336 @@
...
+var (
+	// This function will be overridden in unit tests.
+	getAttrSetFromResourceDetector = func(ctx context.Context) *attribute.Set {
+		r, err := resource.New(ctx, resource.WithDetectors(gcp.NewDetector()))
+
+		if err != nil {
+			logger.Errorf("error reading OpenTelemetry resource: %v", err)
+		}
+		var set *attribute.Set
+		if r != nil {
```
dfawley, line 195: Can `r` be `nil` without an error?

I would have expected:
```go
r, err := ...
if err != nil {
	// log err
	return nil
}
// can safely access r since there was no error.
return r.Set()
```

stats/opentelemetry/internal/pluginoption.go
zasweq: I purposely left this semantically vague because this is general purpose. To have it in one function call I pass in all the incoming RPC metadata, optional labels come from xDS (but can be arbitrary, locality etc) and the relevant labels come from local labels derived at plugin option init time. Thus, these three sources for any plugin option was my goal here.

stats/opentelemetry/csm/pluginoption.go
zasweq: Yeah, this confused me too. This is a built in function of how this API should work (one API call that emits labels from the three sources described above). Thus, it is intended for this layer to receive this arbitrary map of optional labels and make the determination as to which it is interested in. For this case, we assume the caller has processed these service_name and service_namespace for the sake of avoiding string comparisons.

stats/opentelemetry/csm/pluginoption.go
zasweq: This is getting the parsed target in string form. I could change this to take the parsed target struct thing.

stats/opentelemetry/csm/pluginoption.go
zasweq: Oh nice one.

stats/opentelemetry/csm/pluginoption.go
zasweq: Ah yeah I see what you're saying. Let me look into code at return site.

stats/opentelemetry/csm/pluginoption.go
zasweq: Good point. Done.

stats/opentelemetry/csm/pluginoption.go
zasweq: Done.

stats/opentelemetry/csm/pluginoption.go
zasweq: Done.

stats/opentelemetry/csm/pluginoption.go
zasweq: r is actually never nil, but your codeblock still flows better. Switched.

stats/opentelemetry/csm/pluginoption.go
zasweq: Good point. This package will also be the "hook" user calls from main(). I switched to "Package csm contains utilities for Google Cloud Service Mesh observability", but I can delve a little deeper and talk about how it's only additional labels if desired.

stats/opentelemetry/internal/pluginoption.go
zasweq: Ah, ok, noted. Deleted segment about it being experimental.

stats/opentelemetry/csm/pluginoption.go
zasweq: Switched to take a *url.Url

stats/opentelemetry/csm/pluginoption.go
dfawley: So why not remove it from the API then, if we're just going to use whatever is passed in?

stats/opentelemetry/internal/pluginoption.go
dfawley: It's always going to be "incoming RPC metadata".  I don't think it makes sense to intentionally make this vague.

dfawley:
Looks like you still need to exclude these tests from go1.20 runs, too.

stats/opentelemetry/internal/pluginoption.go
zasweq: Ok incoming RPC metadata is fixed, other two buckets aren't.

zasweq:
Argh yeah, I started that but forget to wrap it up/send it out. Working on it now.

stats/opentelemetry/csm/pluginoption.go
yashykt: There can be optional labels other than service_name and service_namespace. Immediately, we have "locality". Additionally, the CSM Plugin Options changes the name of the label from "service_name" to "csm.service_name".

zasweq: Oh whoops I didn't prepend csm. for the two. Let me do that now.

dfawley: > Immediately, we have "locality".

Then why doesn't this need to filter out `locality` or ensure it only passes through `service_name[space]`?

zasweq: Hmmmm ok. I do plan on using the same API that I set xDS Labels with for the locality, and even though there will be orthogonal stats handlers for GCS it will still be in same call which will get merged with this, and call out into this. So I'll add a filter here.

zasweq: Just talked to Yash about this, the intended behavior is to log "unknown" for the xDS labels that don't show up. Even though we want to make this general purpose, C++ actually just takes a struct here with two xDS Labels and a service_name label. Would you prefer that to avoid filtering?

zasweq: Discussed offline with Yash; decided to do this in the OTel layer with an additional API which we will need for GCS SetOptionalLabel, and also prepend csm. in the xDS Client.

.github/workflows/testing.yml
```diff
@@ -103,10 +104,18 @@ jobs:
           go version
           go test ${{ matrix.testflags }} -cpu 1,4 -timeout 7m google.golang.org/grpc/...
           cd "${GITHUB_WORKSPACE}"
+          set -x
```
dfawley, line 107: Remove?

.github/workflows/testing.yml
```diff
@@ -103,10 +104,18 @@ jobs:
...
           for MOD_FILE in $(find . -name 'go.mod' | grep -Ev '^\./go\.mod'); do
             pushd "$(dirname ${MOD_FILE})"
-            go test ${{ matrix.testflags }} -cpu 1,4 -timeout 2m ./...
+            # Skip OpenTelemetry module if 3 releases ago, Go OpenTelemetry only supports
+            # previous two releases.
+            if [[ "${{ matrix.skipotel }}" == "true" && "${PWD}" =~ /stats/opentelemetry$ ]]; then
```
dfawley, line 112: Maybe we should name this something nicer so the value (which is an arbitrary string not a boolean) doesn't say "true".

I.e.

```
- type: tests
  goversion: '1.20'
  mode: skipotel
```

stats/opentelemetry/csm/pluginoption.go
```diff
@@ -311,26 +298,15 @@ func getNodeID() string {
...
 
-	if parsedTarget.Scheme == "xds" {
-		if parsedTarget.Host == "" {
-			return true // "In the cases where no authority is mentioned, the authority is assumed to be csm"
-		}
-		return parsedTarget.Host == "traffic-director-global.xds.googleapis.com"
-	}
-	return false
+	// "In the cases where no authority is mentioned, the authority is assumed
+	// to be csm".
```
dfawley, line 310: This is already above?

stats/opentelemetry/internal/pluginoption.go
```diff
@@ -27,13 +27,12 @@ import (
...
+// In the future, we hope to stabilize and expose this API to allow plugins to
+// inject labels of their choosing into metrics recorded.
 type PluginOption interface {
 	// GetMetadata creates a MD with metadata exchange labels.
 	GetMetadata() metadata.MD
-	// GetLabels emits relevant labels from the metadata provided and the
-	// optional labels provided, alongside any relevant labels.
-	GetLabels(metadata.MD, map[string]string) map[string]string
+	// GetLabels emits relevant labels from the incoming RPC metadata, alongside
+	// any relevant labels.
```
dfawley, line 36: Delete "alongside" stuff since that refs the labels you used to pass in?

zasweq: Alongside is semantically referring to the local labels I emit from the Get call unconditionally (derived at plugin option creation time).

.github/workflows/testing.yml
zasweq: Why not just keep this in to enable debugging? I'll delete though.

.github/workflows/testing.yml
zasweq: Done.

stats/opentelemetry/csm/pluginoption.go
zasweq: Oh whoops sorry this got left behind when the top level comment was far because I had a lot more lines before this single line.

dfawley (approved):

PR #7213, opened 2024-05-08
advancedTLS: Documentation

gtcooke94:
Add documentation to the main package, the `RootCertificateOptions` struct, and the `IdentityCertificateOptions` struct.

RELEASE NOTES: none

### DIFF
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -119,8 +119,8 @@ type GetRootCAsResults = RootCertificates
 
 // RootCertificateOptions contains options to obtain root trust certificates
 // for both the client and the server.
-// At most one option could be set. If none of them are set, we
-// use the system default trust certificates.
+// At most one option should be set. If none of them are set, we
+// use the system default trust certificates. Setting more than one will result in undefined behavior.
 type RootCertificateOptions struct {
 	// If RootCertificates is set, it will be used every time when verifying
 	// the peer certificates, without performing root certificate reloading.
--- a/security/advancedtls/advancedtls.go
+++ b/security/advancedtls/advancedtls.go
@@ -16,14 +16,15 @@
  *
  */
 
-// Package advancedtls is a library containing APIs for configuring grpc
-// connetions with TLS. The APIs here give the user more customizable control to
-// fit their security landscape, thus the "advanced" moniker. This package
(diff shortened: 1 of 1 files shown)


security/advancedtls/advancedtls.go
```diff
@@ -16,9 +16,14 @@
  *
  */
 
-// Package advancedtls is a utility library containing functions to construct
-// credentials.TransportCredentials that can perform credential reloading and
-// custom verification check.
+// Package advancedtls is a library containing APIs for configuring grpc
```
dfawley, line 19: Nit: all packages are libraries, and all libraries contain APIs, so I think you could remove some words here to make everything more to-the-point.  How about something like:

`Package advancedtls provides gRPC transport credentials that allow easy configuration of advanced TLS features.`

security/advancedtls/advancedtls.go
```diff
@@ -119,8 +124,9 @@ type GetRootCAsResults = RootCertificates
 
 // RootCertificateOptions contains options to obtain root trust certificates
 // for both the client and the server.
-// At most one option could be set. If none of them are set, we
-// use the system default trust certificates.
+// At most one option should be set. If none of them are set, we use the system
```
dfawley, line 127: What is an "option"?  Is each option a single field in this struct?  If so, maybe "at most one field should be set"?

security/advancedtls/advancedtls.go
```diff
@@ -153,18 +159,18 @@ func (o RootCertificateOptions) nonNilFieldCount() int {
 
 // IdentityCertificateOptions contains options to obtain identity certificates
 // for both the client and the server.
-// At most one option could be set.
```
dfawley, line 156: As above.

security/advancedtls/advancedtls.go
gtcooke94: Changed, I like that

security/advancedtls/advancedtls.go
gtcooke94: Changed, good call

security/advancedtls/advancedtls.go
gtcooke94: Changed

dfawley (approved):

PR #7218, opened 2024-05-09
Change version to 1.64.0

arvindbr8:


version.go
```diff
@@ -19,4 +19,4 @@
 package grpc
 
 // Version is the current grpc version.
-const Version = "1.64.0-dev"
```
purnesh42H, line 22: Just for my knowledge, why was it suffixed with -dev before?

arvindbr8: Discussed offline

dfawley (approved):

PR #7221, opened 2024-05-09
advancedTLS: Add in deprecated name for transitionary period

gtcooke94:
I changed `RevocationConfig` to `RevocationOptions` in https://github.com/grpc/grpc-go/pull/7151 and missed leaving the old name as a deprecated option so that migrating is easier. This PR adds it back so that we shouldn't be actively breaking code.

RELEASE NOTES: none


arvindbr8 (approved):
Thanks. LGTM!

dfawley (approved):

PR #7222, opened 2024-05-09
.*: update deps

arvindbr8:


dfawley:
All tests are failing; PTAL

arvindbr8:
I think we need to until the next release before we can update the deps. This is because the _grpc.pb.go have a compile time assertion to use gRPC-go v1.64 or higher (due to this change https://github.com/grpc/grpc-go/pull/7057/files#diff-51876cdb24ccecba4859f376f237b87dc813d32839ec4e76a02db28081b4d212)

arvindbr8:
I'll reopen this once I've figured this out

PR #7224, opened 2024-05-10
test: fix flaky test ClientSendsAGoAway

### LINKED ISSUE #7160: Flaky test: TestClientSendsAGoAway
This is the first time I'm seeing it: 
https://github.com/grpc/grpc-go/actions/runs/8787418013/job/24112571687?pr=7156

Might be worth to check it out.

aranjans:
Fixes #7160 
fix: `goaway_test/TestClientSendsAGoAway` didn't use to wait for channels to be ready and which is why sometimes it was not able to read the frame and eventually result in test failure. This PR change make sure we read the frame only when the channel's state becomes READY.

RELEASE NOTES: n/a

### DIFF
--- a/test/goaway_test.go
+++ b/test/goaway_test.go
@@ -800,6 +800,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 		for {
 			f, err := ct.fr.ReadFrame()
 			if err != nil {
+				t.Logf("error reading frame: %v", err)
 				return
 			}
 			switch fr := f.(type) {
@@ -808,6 +809,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 				if fr.ErrCode == http2.ErrCodeNo {
 					t.Logf("GoAway received from client")
 					close(goAwayReceived)
+					return
 				}
 			default:
 				t.Errorf("server tester received unexpected frame type %T", f)
@@ -816,6 +818,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 			}
 		}
 	}()
+	cc.WaitForStateChange(ctx, connectivity.Connecting)
 	cc.Close()
 	defer ct.conn.Close()
 	select {
--- a/test/goaway_test.go
+++ b/test/goaway_test.go
@@ -800,7 +800,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 		for {
 			f, err := ct.fr.ReadFrame()
 			if err != nil {
-				t.Logf("error reading frame: %v", err)
+				errCh <- fmt.Errorf("error reading frame: %v", err)
 				return
 			}
 			switch fr := f.(type) {
--- a/test/goaway_test.go
+++ b/test/goaway_test.go
@@ -770,33 +770,16 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error listening: %v", err)
 	}
-	ctCh := testutils.NewChannel()
+	defer lis.Close()
+	goAwayReceived := make(chan struct{})
+	errCh := make(chan error)
 	go func() {
 		conn, err := lis.Accept()
 		if err != nil {
 			t.Errorf("error in lis.Accept(): %v", err)
 		}
 		ct := newClientTester(t, conn)
-		ctCh.Send(ct)
-	}()
-	defer lis.Close()
-
-	cc, err := grpc.Dial(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
-	if err != nil {
-		t.Fatalf("error dialing: %v", err)
-	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
-	defer cancel()
-
-	val, err := ctCh.Receive(ctx)
-	if err != nil {
-		t.Fatalf("timeout waiting for client transport (should be given after http2 creation)")
-	}
-	ct := val.(*clientTester)
-	goAwayReceived := make(chan struct{})
-	errCh := make(chan error)
-	go func() {
+		defer ct.conn.Close()
 		for {
 			f, err := ct.fr.ReadFrame()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


test/goaway_test.go
```diff
@@ -816,6 +818,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 			}
 		}
 	}()
+	cc.WaitForStateChange(ctx, connectivity.Connecting)
```
purnesh42H, line 821: what would be the max wait here? Timeout for all tests is 7m. Is there a chance it can breach that?

purnesh42H:
Probably need some more description about the fix

test/goaway_test.go
aranjans: Max timeout is as long context is not timed out.

purnesh42H: shouldn't the state to wait be `connectivity.Ready`?

dfawley: I think it depends where the race is happening.

If there's just a race where we call `Close` before we even start creating the connection, then this is a problem with the test, and `Connecting` should fix that.

If this is a race where a connection is established and doesn't get a GOAWAY written to it (due to internal state in the `ClientConn`/`addrConn`, then `Connecting` will expose that race and we should fix it.

test/goaway_test.go
```diff
@@ -800,6 +800,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 		for {
```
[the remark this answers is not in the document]
aranjans: @dfawley we already have a same test with that logic here[https://github.com/grpc/grpc-go/blob/master/internal/transport/transport_test.go#L2666](url). And actually its checking for the same, do you feel we can remove this goaway_test/ClientSendsGoAway as we already have there in transport package or shall we replicate the same here which i don't think makes sense IMO?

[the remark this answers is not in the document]
dfawley: I am not sure I understand what you're getting at.

I'm just saying this test seems to do some unnecessary work and is more complicated than it needs to be.

The other test you added is inside the transport layer, and testing at the API layer, which this test does, is _generally_ preferred.

dfawley:
From chat offline:

> Actually, this structure of the test actually proves there is a bug here that I don't think you're fixing
> If a connection is created, we want it to reliably get a goaway.  We know a connection exists otherwise we won't get to this part of the code, because the channel passes it along

So I think the test is correct (though it can still be simplified), but there is a race in the code that prevents the GOAWAY from being sent in some cases.

test/goaway_test.go
aranjans: @purnesh42H WaitForStateChange waits until the connectivity.State of ClientConn changes from sourceState or ctx expires. A true value is returned in former case and false in latter. So we are providing sourceState as Connecting and hence it will wait until state changes.

test/goaway_test.go
```diff
@@ -800,6 +800,7 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
...
 			f, err := ct.fr.ReadFrame()
 			if err != nil {
+				t.Logf("error reading frame: %v", err)
```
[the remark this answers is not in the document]
aranjans: Updated

test/goaway_test.go
aranjans: I think race is detected when client conn is getting closed before we try to send goAway so believe just waiting for clientConn state change will fix this race condition.

test/goaway_test.go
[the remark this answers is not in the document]
aranjans: Updated

aranjans:
@dfawley updated the PR

aranjans:
Update: [this](https://github.com/grpc/grpc-go/blob/0020ccf9d9fc448fcfa099173df4ad330e96326c/clientconn.go#L1409) is the code where we are hardcode closing the transport. Will update the PR with the change where we send goAway frame in this case as well.

aranjans:
@dfawley As per our discussion offline, we won't be waiting for all the transport conn to close down. So this PR's change is ready to review as it just makes sure we are waiting for the clientconn to be READY.

test/goaway_test.go
```diff
@@ -816,8 +800,40 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 			}
 		}
 	}()
+	defer lis.Close()
```
dfawley, line 803: This belongs next to the creation of the listener.

test/goaway_test.go
```diff
@@ -816,8 +800,40 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
...
+
+	cc, err := grpc.Dial(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("error dialing: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	//go func() {
```
dfawley, line 813: Delete please

test/goaway_test.go
```diff
@@ -816,8 +800,40 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
...
+	//				return
+	//			}
+	//		default:
+	//			t.Errorf("server tester received unexpected frame type %T", f)
+	//			errCh <- fmt.Errorf("server tester received unexpected frame type %T", f)
+	//			close(errCh)
+	//		}
+	//	}
+	//}()
+	cc.WaitForStateChange(ctx, connectivity.Connecting)
```
dfawley, line 835: Should be `Ready` now.

aranjans: @dfawley `WaitForStateChange` waits until the connectivity.State of ClientConn changes from sourceState or ctx expires.  So aren't we already waiting for conn to be READY here as it would wait until state changes from Connecting?

test/goaway_test.go
```diff
@@ -816,8 +801,17 @@ func (s) TestClientSendsAGoAway(t *testing.T) {
 			}
 		}
 	}()
+
+	cc, err := grpc.Dial(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
```
purnesh42H, line 805: nit: Shall we use NewClient since Dial is in deprecation stage?

dfawley: +1

test/goaway_test.go
dfawley: You are correct.  Instead you should probably be using `AwaitState(Ready)`:

https://github.com/grpc/grpc-go/blob/master/internal/testutils/state.go#L58

test/goaway_test.go
aranjans: Updated.

test/goaway_test.go
aranjans: Updated.

dfawley (approved):

Issue #4615, opened 2021-07-23
Retry: grpc maxretry attempts is hardcoded


Leomax004:
Grpc Max Retry attempt is hardcoded and not allowing to retry more than 5

if rp.MaxAttempts > 5 {
		// TODO(retry): Make the max maxAttempts configurable.
		rp.MaxAttempts = 5
	}

dfawley:
Reference from gRFC:

https://github.com/grpc/proposal/blob/master/A6-client-retries.md#limits-on-retries-and-hedges

We never implemented this particular "channel arg" in Go.

gnvk:
How is this a feature and not a bug? This makes the whole retry config pretty much unusable. And it's not even mentioned in the documentation!

arvindbr8:
Feel free to send us a PR if you need this. I would consider missing features not bugs :)

PR #7229, opened 2024-05-12
client: implement maxAttempts for retryPolicy

imoore76:
Addresses https://github.com/grpc/grpc-go/issues/4615

RELEASE NOTES: none

### DIFF
--- a/dialoptions.go
+++ b/dialoptions.go
@@ -80,6 +80,7 @@ type dialOptions struct {
 	idleTimeout                 time.Duration
 	recvBufferPool              SharedBufferPool
 	defaultScheme               string
+	maxRetryAttempts            int
 }
 
 // DialOption configures how we set up the connection.
@@ -650,11 +651,12 @@ func defaultDialOptions() dialOptions {
 			UseProxy:        true,
 			UserAgent:       grpcUA,
 		},
-		bs:              internalbackoff.DefaultExponential,
-		healthCheckFunc: internal.HealthCheckFunc,
-		idleTimeout:     30 * time.Minute,
-		recvBufferPool:  nopBufferPool{},
-		defaultScheme:   "dns",
+		bs:               internalbackoff.DefaultExponential,
+		healthCheckFunc:  internal.HealthCheckFunc,
+		idleTimeout:      30 * time.Minute,
+		recvBufferPool:   nopBufferPool{},
+		defaultScheme:    "dns",
+		maxRetryAttempts: 5, // https://github.com/grpc/proposal/blob/master/A6-client-retries.md#limits-on-retries-and-hedges
 	}
 }
 
@@ -709,6 +711,19 @@ func WithResolvers(rs ...resolver.Builder) DialOption {
 func WithIdleTimeout(d time.Duration) DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
 		o.idleTimeout = d
+	})
+}
+
+// WithMaxRetryAttempts returns a DialOption that configures the maximum number
+// of retry attempts for the channel. Service owners may specify a higher value
+// for these parameters, but higher values will be treated as equal to the
+// maximum value by the client implementation. This mitigates security concerns
+// related to the service config being transferred to the client via DNS.
+//
+// A default value of 5 will be used if this dial option is not set.
+func WithMaxRetryAttempts(n int) DialOption {
+	return newFuncDialOption(func(o *dialOptions) {
+		o.maxRetryAttempts = n
 	})
 }
 
--- a/service_config.go
+++ b/service_config.go
@@ -285,10 +285,6 @@ func convertRetryPolicy(jrp *jsonRetryPolicy) (p *internalserviceconfig.RetryPol
 		BackoffMultiplier:    jrp.BackoffMultiplier,
 		RetryableStatusCodes: make(map[codes.Code]bool),
 	}
-	if rp.MaxAttempts > 5 {
-		// TODO(retry): Make the max maxAttempts configurable.
-		rp.MaxAttempts = 5
-	}
 	for _, code := range jrp.RetryableStatusCodes {
 		rp.RetryableStatusCodes[code] = true
 	}
--- a/test/retry_test.go
+++ b/test/retry_test.go
@@ -102,6 +102,64 @@ func (s) TestRetryUnary(t *testing.T) {
 	}
 }
 
+func (s) TestRetryMaxAttemptsUnary(t *testing.T) {
+	callCount := 0
+	ss := &stubserver.StubServer{
+		EmptyCallF: func(context.Context, *testpb.Empty) (r *testpb.Empty, err error) {
+			defer func() { t.Logf("server call %v returning err %v", callCount, err) }()
+			callCount++
+			return nil, status.New(codes.AlreadyExists, "retryable error").Err()
+		},
+	}
+	testCases := []struct {
+		serviceMaxAttempts int
+		clientMaxAttempts  int
+		expectedAttempts   int
+	}{
+		{serviceMaxAttempts: 9, clientMaxAttempts: 4, expectedAttempts: 4},
+		{serviceMaxAttempts: 9, clientMaxAttempts: 7, expectedAttempts: 7},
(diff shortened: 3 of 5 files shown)


imoore76:
Hi @aranjans! Is there something you are waiting for from me?

aranjans:
@imoore76 no, i think this PR LGTM! Approved.

aranjans (approved):

service_config.go
```diff
@@ -285,10 +285,6 @@ func convertRetryPolicy(jrp *jsonRetryPolicy) (p *internalserviceconfig.RetryPol
 		BackoffMultiplier:    jrp.BackoffMultiplier,
 		RetryableStatusCodes: make(map[codes.Code]bool),
 	}
-	if rp.MaxAttempts > 5 {
```
dfawley, line 288: Is it possible to keep the relevant logic here (where all the retry code is) by parameterizing the "max max attempts" setting, instead of moving it into the RPC path?

dialoptions.go
```diff
@@ -712,6 +714,19 @@ func WithIdleTimeout(d time.Duration) DialOption {
...
+// WithMaxRetryAttempts returns a DialOption that configures the maximum number
+// of retry attempts for the channel. Service owners may specify a higher value
+// for these parameters, but higher values will be treated as equal to the
+// maximum value by the client implementation. This mitigates security concerns
+// related to the service config being transferred to the client via DNS.
+//
+// A default value of 5 will be used if this dial option is not set.
+func WithMaxRetryAttempts(n int) DialOption {
+	return newFuncDialOption(func(o *dialOptions) {
+		o.maxRetryAttempts = n
```
dfawley, line 726: Are there no bounds that should be placed on this?

The `maxAttempts` field in the retry config must be >1, so that seems like a reasonable check for this as well.  (There's a separate setting to disable retry entirely.)

Since we can't return an error, maybe:

```
// A default value of 5 will be used if this dial option is not set or if n < 2.
```

?

test/retry_test.go
```diff
@@ -102,6 +102,64 @@ func (s) TestRetryUnary(t *testing.T) {
...
+	testCases := []struct {
+		serviceMaxAttempts int
+		clientMaxAttempts  int
+		expectedAttempts   int
+	}{
+		{serviceMaxAttempts: 9, clientMaxAttempts: 4, expectedAttempts: 4},
+		{serviceMaxAttempts: 9, clientMaxAttempts: 7, expectedAttempts: 7},
+		{serviceMaxAttempts: 3, clientMaxAttempts: 10, expectedAttempts: 3},
+		{serviceMaxAttempts: 8, clientMaxAttempts: -1, expectedAttempts: 5}, // 5 is default max
+		{serviceMaxAttempts: 3, clientMaxAttempts: 0, expectedAttempts: 1},
```
dfawley, line 123: How? :)

dialoptions.go
imoore76: The spec does not place any upper bound on it.

I think your comment and logic make sense. I'll update it.

service_config.go
imoore76: I can certainly try :)

test/retry_test.go
imoore76: I agree this visually doesn't make sense. IIRC, retry logic isn't hit until after the first attempt so you'll always get at least one attempt. I did not change any of the retry logic, I just added the test that exposed this behavior.

This test will change once I implement your suggested lower bound on the setting. The dial option name probably makes more sense as `maxAttempts` rather than `maxRetryAttempts`. I will change the name.

dfawley: I was mostly joking here and referencing my other comment.  :)

The only change I think the test will need is to remove the `if` around the place where you create the DialOption.

It's true that "`retry`" doesn't really explain this fully, but `MaxAttempts` seems too nebulous.  How about `WithMaxCallAttempts`?

imoore76: sgtm

service_config.go
imoore76: @dfawley Looking at all the places `parseServiceConfig` is called, If I parameterize this, we end up passing the max call setting around from dial options to the dns resolver, load balancer config, as well as here in service config. It would get messy real fast. Do you want me to do this?

dfawley: That doesn't sound right to me.  Resolvers should be calling `ParseConfig` on the `ClientConn`, and that should be adding the setting in at that time.  Maybe I'm missing something though?  Is it tests that are the problem?

imoore76: You are mostly right. There are a couple other places (aside from tests) where it is called as `internal.ParseServiceConfig` where it doesn't have access to `ClientConn`, but in those places `maxAttempts` seems inconsequential.

imoore76: I've updated as you've suggested.

dialoptions.go
```diff
@@ -736,16 +741,19 @@ func WithIdleTimeout(d time.Duration) DialOption {
 	})
 }
 
-// WithMaxRetryAttempts returns a DialOption that configures the maximum number
+// WithMaxCallAttempts returns a DialOption that configures the maximum number
 // of retry attempts for the channel. Service owners may specify a higher value
```
dfawley, line 745: How about:

`...configures the maximum number of attempts per call (including retries and hedging) using the channel`?

I don't especially like mentioning "hedging" here since we don't support it yet, but it seems better than forgetting it later...

service_config.go
```diff
@@ -164,9 +164,11 @@ type jsonSC struct {
 }
 
 func init() {
-	internal.ParseServiceConfig = parseServiceConfig
+	internal.ParseServiceConfig = func(js string) *serviceconfig.ParseResult {
+		return parseServiceConfig(js, defaultMaxCallAttempts)
+	}
 }
-func parseServiceConfig(js string) *serviceconfig.ParseResult {
+func parseServiceConfig(js string, maxAttempts int) *serviceconfig.ParseResult {
```
dfawley, line 171: This almost definitely should be a method on CC, but that's our problem to fix, and it probably won't be easy...  Just FYI only.

test/retry_test.go
```diff
@@ -137,9 +137,7 @@ func (s) TestRetryMaxAttemptsUnary(t *testing.T) {
 		  }
 		}]}`, tc.serviceMaxAttempts)),
 		}
-		if tc.clientMaxAttempts >= 0 {
-			clientOpts = append(clientOpts, grpc.WithMaxRetryAttempts(tc.clientMaxAttempts))
-		}
+		clientOpts = append(clientOpts, grpc.WithMaxCallAttempts(tc.clientMaxAttempts))
```
dfawley, line 140: Nit, optional: don't use `append`.  Could also eliminate `clientOpts` in theory, but it's possible it aids readability, too.

test/retry_test.go
```diff
@@ -574,9 +572,7 @@ func (s) TestRetryMaxAttemptsStreaming(t *testing.T) {
 				},
 			}
 
-			if tc.clientMaxAttempts >= 0 {
-				clientOpts = append(clientOpts, grpc.WithMaxRetryAttempts(tc.clientMaxAttempts))
-			}
+			clientOpts = append(clientOpts, grpc.WithMaxCallAttempts(tc.clientMaxAttempts))
```
dfawley, line 575: As above.

test/retry_test.go
```diff
@@ -477,6 +533,79 @@ func (s) TestRetryStreaming(t *testing.T) {
 	}
 }
 
+func (s) TestMaxCallAttemptsStreaming(t *testing.T) {
```
dfawley, line 536: Can this be folded into the other test as another dimension to avoid repeating all the setup and test cases?

dfawley (commented):
LGTM, thank you for the PR.  Just one thing that needs to change: the godoc comment, and one thing that would be nice-to-have (combining test cases) if it's possible, and one small optional thing that is a judgment call that I'll leave up to you.

test/retry_test.go
imoore76: Ah yes. The `append` is superfluous now that the condition is removed.

test/retry_test.go
imoore76: folded into one test now

PR #7234, opened 2024-05-14
grpc: Add perTargetDialOption type and global list

zasweq:
This PR adds an internal only (exported, but can only set through internal/) LateApplyDialOption type. This will be used in CSM Observability.

RELEASE NOTES: N/A

### DIFF
--- a/clientconn.go
+++ b/clientconn.go
@@ -153,22 +153,13 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error) 
 		opt.apply(&cc.dopts)
 	}
 
-	// Register ClientConn with channelz.
-	cc.channelzRegistration(target)
-	// TODO: Ideally it should be impossible to error from this function after
-	// channelz registration.  This will require removing some channelz logs
-	// from the following functions that can error.  Errors can be returned to
-	// the user, and successful logs can be emitted here, after the checks have
-	// passed and channelz is subsequently registered.
-
 	// Determine the resolver to use.
 	if err := cc.parseTargetAndFindResolver(); err != nil {
-		channelz.RemoveEntry(cc.channelz.ID)
 		return nil, err
 	}
 
-	for _, lateApplyOpt := range globalLateApplyDialOptions {
-		lateApplyOpt.DialOption(cc.parsedTarget.URL).apply(&cc.dopts)
+	for _, opt := range globalPerTargetDialOptions {
+		opt.DialOption(cc.parsedTarget.URL).apply(&cc.dopts)
 	}
 
 	chainUnaryClientInterceptors(cc)
@@ -188,9 +179,12 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error) 
 	cc.mkp = cc.dopts.copts.KeepaliveParams
 
 	if err = cc.determineAuthority(); err != nil {
-		channelz.RemoveEntry(cc.channelz.ID)
 		return nil, err
 	}
+
+	// Register ClientConn with channelz.
+	cc.channelzRegistration(target)
+	channelz.Infof(logger, cc.channelz, "Channel authority set to %q", cc.authority)
 
 	cc.csMgr = newConnectivityStateManager(cc.ctx, cc.channelz)
 	cc.pickerWrapper = newPickerWrapper(cc.dopts.copts.StatsHandlers)
@@ -1686,14 +1680,14 @@ func (cc *ClientConn) connectionError() error {
 //
 // Doesn't grab cc.mu as this method is expected to be called only at Dial time.
 func (cc *ClientConn) parseTargetAndFindResolver() error {
-	channelz.Infof(logger, cc.channelz, "original dial target is: %q", cc.target)
+	logger.Infof("original dial target is: %q", cc.target)
 
 	var rb resolver.Builder
 	parsedTarget, err := parseTarget(cc.target)
 	if err != nil {
-		channelz.Infof(logger, cc.channelz, "dial target %q parse failed: %v", cc.target, err)
+		logger.Infof("dial target %q parse failed: %v", cc.target, err)
 	} else {
-		channelz.Infof(logger, cc.channelz, "parsed dial target is: %#v", parsedTarget)
+		logger.Infof("parsed dial target is: %#v", parsedTarget)
 		rb = cc.getResolver(parsedTarget.URL.Scheme)
 		if rb != nil {
 			cc.parsedTarget = parsedTarget
@@ -1712,15 +1706,15 @@ func (cc *ClientConn) parseTargetAndFindResolver() error {
 		defScheme = resolver.GetDefaultScheme()
 	}
 
-	channelz.Infof(logger, cc.channelz, "fallback to scheme %q", defScheme)
+	logger.Infof("fallback to scheme %q", defScheme)
 	canonicalTarget := defScheme + ":///" + cc.target
 
 	parsedTarget, err = parseTarget(canonicalTarget)
 	if err != nil {
-		channelz.Infof(logger, cc.channelz, "dial target %q parse failed: %v", canonicalTarget, err)
+		logger.Infof("dial target %q parse failed: %v", canonicalTarget, err)
 		return err
 	}
-	channelz.Infof(logger, cc.channelz, "parsed dial target is: %+v", parsedTarget)
+	logger.Infof("parsed dial target is: %+v", parsedTarget)
 	rb = cc.getResolver(parsedTarget.URL.Scheme)
 	if rb == nil {
 		return fmt.Errorf("could not get resolver for default scheme: %q", parsedTarget.URL.Scheme)
@@ -1843,6 +1837,5 @@ func (cc *ClientConn) determineAuthority() error {
 	} else {
 		cc.authority = encodeAuthority(endpoint)
 	}
-	channelz.Infof(logger, cc.channelz, "Channel authority set to %q", cc.authority)
 	return nil
 }
--- a/clientconn.go
+++ b/clientconn.go
@@ -154,7 +154,7 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error) 
 	}
 
 	// Determine the resolver to use.
-	if err := cc.parseTargetAndFindResolver(); err != nil {
+	if err := cc.initParsedTargetAndResolverBuilder(); err != nil {
 		return nil, err
 	}
 
@@ -178,7 +178,7 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error) 
 	}
 	cc.mkp = cc.dopts.copts.KeepaliveParams
 
-	if err = cc.determineAuthority(); err != nil {
+	if err = cc.initAuthority(); err != nil {
 		return nil, err
 	}
 
@@ -588,11 +588,11 @@ type ClientConn struct {
 
 	// The following are initialized at dial time, and are read-only after that.
 	target          string            // User's dial target.
-	parsedTarget    resolver.Target   // See parseTargetAndFindResolver().
-	authority       string            // See determineAuthority().
+	parsedTarget    resolver.Target   // See initParsedTargetAndResolverBuilder().
+	authority       string            // See initAuthority().
 	dopts           dialOptions       // Default and user specified dial options.
 	channelz        *channelz.Channel // Channelz object.
-	resolverBuilder resolver.Builder  // See parseTargetAndFindResolver().
+	resolverBuilder resolver.Builder  // See initParsedTargetAndResolverBuilder().
 	idlenessMgr     *idle.Manager
 
 	// The following provide their own synchronization, and therefore don't
@@ -1674,14 +1674,14 @@ func (cc *ClientConn) connectionError() error {
 	return cc.lastConnectionError
 }
 
-// parseTargetAndFindResolver parses the user's dial target and stores the
-// parsed target in `cc.parsedTarget`.
+// initParsedTargetAndResolverBuilder parses the user's dial target and stores
+// the parsed target in `cc.parsedTarget`.
 //
 // The resolver to use is determined based on the scheme in the parsed target
 // and the same is stored in `cc.resolverBuilder`.
 //
 // Doesn't grab cc.mu as this method is expected to be called only at Dial time.
-func (cc *ClientConn) parseTargetAndFindResolver() error {
+func (cc *ClientConn) initParsedTargetAndResolverBuilder() error {
 	logger.Infof("original dial target is: %q", cc.target)
 
 	var rb resolver.Builder
@@ -1800,7 +1800,7 @@ func encodeAuthority(authority string) string {
 // credentials do not match the authority configured through the dial option.
 //
 // Doesn't grab cc.mu as this method is expected to be called only at Dial time.
-func (cc *ClientConn) determineAuthority() error {
+func (cc *ClientConn) initAuthority() error {
 	dopts := cc.dopts
 	// Historically, we had two options for users to specify the serverName or
 	// authority for a channel. One was through the transport credentials
--- a/clientconn.go
+++ b/clientconn.go
@@ -159,7 +159,7 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error) 
 	}
 
 	for _, opt := range globalPerTargetDialOptions {
-		opt.DialOption(cc.parsedTarget.URL).apply(&cc.dopts)
+		opt.DialOptionForTarget(cc.parsedTarget.URL).apply(&cc.dopts)
 	}
 
 	chainUnaryClientInterceptors(cc)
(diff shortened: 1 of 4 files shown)


clientconn.go
```diff
@@ -152,6 +152,25 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error)
...
+	// passed and channelz is subsequently registered.
+
+	// Determine the resolver to use.
+	if err := cc.parseTargetAndFindResolver(); err != nil {
+		channelz.RemoveEntry(cc.channelz.ID)
+		return nil, err
+	}
+
+	for _, lateApplyOpt := range globalLateApplyDialOptions {
+		lateApplyOpt.DialOption(cc.parsedTarget.URL).apply(&cc.dopts)
```
dfawley, line 171: Nit: `opt` (as above) or even `o` is preferred for very narrowly scoped vars like this.

clientconn.go
```diff
@@ -152,6 +152,25 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error)
 	for _, opt := range opts {
 		opt.apply(&cc.dopts)
 	}
+
+	// Register ClientConn with channelz.
+	cc.channelzRegistration(target)
+	// TODO: Ideally it should be impossible to error from this function after
```
dfawley, line 158: This diff makes the problem this TODO describes even worse.  And there are places where we error from this now that don't call `channelz.RemoveEntry`, leaking resources.

Please make `parseTargetBlah` stop doing channelz logs (maybe move some useful channelz messages into this function after channelz registration), and move channelz registration back down to where it was.

If you're feeling extra helpful, you could also extend that to the other initialization functions that might do channelz logs, and make sure we don't create the channelz representation of this channel until immediately before the function returns.

dialoptions.go
```diff
@@ -89,6 +96,14 @@ type DialOption interface {
 
 var globalDialOptions []DialOption
 
+// LateApplyDialOption takes a parsed target and returns a dial option to apply.
+type LateApplyDialOption interface {
```
dfawley, line 100: Unexport?  Move to `internal`?  I really don't want users seeing this hack.

internal/internal.go
```diff
@@ -106,6 +106,14 @@ var (
 	// This is used in the 1.0 release of gcp/observability, and thus must not be
 	// deleted or changed.
 	ClearGlobalDialOptions func()
+
+	// AddGlobalLateApplyDialOptions adds a slice of lateApplyDialOptions that
+	// will be configured for newly created ClientConns.
```
dfawley, line 111: This (or something) should explain what these magic things can or cannot do.

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,29 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
...
+
+func (do *testLateApplyDialOption) DialOption(parsedTarget url.URL) DialOption {
+	if parsedTarget.Scheme == "passthrough" {
+		return WithTransportCredentials(insecure.NewCredentials()) // credentials provided, should pass NewClient.
+	}
+	return WithReadBufferSize(10) // no credentials, should fail NewClient
+}
+
+func (s) TestGlobalLateApplyDialOption(t *testing.T) {
+	internal.AddGlobalLateApplyDialOptions.(func(opt ...LateApplyDialOption))(&testLateApplyDialOption{}) // Do I need a clear?
```
dfawley, line 88: Comment?

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,29 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
 	internal.ClearGlobalDialOptions()
 }
 
+type testLateApplyDialOption struct{}
+
+func (do *testLateApplyDialOption) DialOption(parsedTarget url.URL) DialOption {
+	if parsedTarget.Scheme == "passthrough" {
+		return WithTransportCredentials(insecure.NewCredentials()) // credentials provided, should pass NewClient.
+	}
+	return WithReadBufferSize(10) // no credentials, should fail NewClient
```
dfawley, line 84: Can we use `EmptyDialOption` instead?

dialoptions.go
```diff
@@ -89,6 +96,14 @@ type DialOption interface {
...
+	// DialOption returns a Dial Option to apply.
+	DialOption(parsedTarget url.URL) DialOption
+}
+
+var globalLateApplyDialOptions []LateApplyDialOption
```
dfawley, line 105: Should we ditch this name since they don't have a `lateApply` method?  How about `perTargetDialOption` or something?

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,29 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
...
+}
+
+func (s) TestGlobalLateApplyDialOption(t *testing.T) {
```
dfawley, line 87: It took a good bit of learning this to figure out what it's even doing.  A short comment to say "configure a global late dial option that produces TransportCredentials for channels using the passthrough scheme, which makes those channels valid but others invalid" would be really helpful.

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,29 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
...
+	internal.AddGlobalLateApplyDialOptions.(func(opt ...LateApplyDialOption))(&testLateApplyDialOption{}) // Do I need a clear?
+	noTSecStr := "no transport security set"
+	if _, err := NewClient("fake"); !strings.Contains(fmt.Sprint(err), noTSecStr) {
```
dfawley, line 90: An explicit `dns:///` would be helpful here.

dialoptions.go
zasweq: Will need to move to internal/ since the csm/ package needs to typecast the any in internal. But yeah makes sense I didn't want this to be in exported namespace (at least right now the user can't set it :)) but forgot I can just put in internal/.

clientconn.go
zasweq: Ah, right. Switched to opt.

default_dial_option_server_option_test.go
zasweq: Good point. Switched.

default_dial_option_server_option_test.go
zasweq: Good point. Done.

default_dial_option_server_option_test.go
zasweq: Deleted. The answer is duh, otherwise other tests would pick up globals :).

default_dial_option_server_option_test.go
zasweq: Added comment: "TestGlobalLateApplyDialOption configures a global dial option that produces transport credentials for channels using "passthrough" scheme. Channels that use the passthrough scheme should this be successfully created due to picking up transport credentials, whereas other channels should fail at creation due to not having transport credentials."

dialoptions.go
zasweq: Alright will switch to perTarget.

clientconn.go
zasweq: Switched to solution discussed offline; make parseTarget log at info, and log authority in NewClient body. Only create channelz after we know any errors can occur.

internal/internal.go
zasweq: I don't know what you mean by "can or cannot do". This is an arbitrary hack we added for one specific purpose, so I feel like scoping out what it's properties are and it's constraints doesn't apply? I added "This gets called after NewClient() parses the target, and allows per target
configuration set through a returned DialOption." to the PerTargetDialOption type (which I moved to internal/ to hide from users). Hopefully that's around what you meant by this comment?

dialoptions.go
zasweq: Discussed further offline; unexported and typecast from any in SetGlobalDialOptions.

clientconn.go
```diff
@@ -152,6 +152,16 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error)
 	for _, opt := range opts {
 		opt.apply(&cc.dopts)
 	}
+
+	// Determine the resolver to use.
+	if err := cc.parseTargetAndFindResolver(); err != nil {
```
dfawley, line 157: Would you mind converting this and `cc.determineAuthority` so they are not `cc` methods?  Otherwise they might forget they're not supposed to be using `cc.channelz`.

clientconn.go
```diff
@@ -168,25 +178,14 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error)
...
-	if err := cc.parseTargetAndFindResolver(); err != nil {
-		channelz.RemoveEntry(cc.channelz.ID)
-		return nil, err
-	}
 	if err = cc.determineAuthority(); err != nil {
-		channelz.RemoveEntry(cc.channelz.ID)
 		return nil, err
 	}
 
+	// Register ClientConn with channelz.
```
dfawley, line 185: Maybe something that says "Note that this is only done after channel creation cannot fail", so that nobody adds error cases after this.  Maybe also move immediately before the `return` to make it doubly sure.

clientconn.go
```diff
@@ -1681,14 +1680,14 @@ func (cc *ClientConn) connectionError() error {
 //
 // Doesn't grab cc.mu as this method is expected to be called only at Dial time.
 func (cc *ClientConn) parseTargetAndFindResolver() error {
-	channelz.Infof(logger, cc.channelz, "original dial target is: %q", cc.target)
+	logger.Infof("original dial target is: %q", cc.target)
```
dfawley, line 1683: The problem with this simple conversion is that the channelz version would prefix with info about the channel.

Let's clean things up here.

- This log message is not helpful, or if it is, it should be moved into `NewClient` and left as channelz-scoped.

clientconn.go
```diff
@@ -1681,14 +1680,14 @@ func (cc *ClientConn) connectionError() error {
...
 
 	var rb resolver.Builder
 	parsedTarget, err := parseTarget(cc.target)
 	if err != nil {
-		channelz.Infof(logger, cc.channelz, "dial target %q parse failed: %v", cc.target, err)
+		logger.Infof("dial target %q parse failed: %v", cc.target, err)
```
dfawley, line 1688: - This doesn't seem worth keeping.  We're going to try some more stuff that might make it valid, and something as simple as `hostname:port` will trigger this.

clientconn.go
```diff
@@ -1681,14 +1680,14 @@ func (cc *ClientConn) connectionError() error {
...
 	} else {
-		channelz.Infof(logger, cc.channelz, "parsed dial target is: %#v", parsedTarget)
+		logger.Infof("parsed dial target is: %#v", parsedTarget)
```
dfawley, line 1690: * Move to `NewClient` please.

clientconn.go
```diff
@@ -1707,15 +1706,15 @@ func (cc *ClientConn) parseTargetAndFindResolver() error {
...
 	canonicalTarget := defScheme + ":///" + cc.target
 
 	parsedTarget, err = parseTarget(canonicalTarget)
 	if err != nil {
-		channelz.Infof(logger, cc.channelz, "dial target %q parse failed: %v", canonicalTarget, err)
+		logger.Infof("dial target %q parse failed: %v", canonicalTarget, err)
 		return err
 	}
-	channelz.Infof(logger, cc.channelz, "parsed dial target is: %+v", parsedTarget)
+	logger.Infof("parsed dial target is: %+v", parsedTarget)
```
dfawley, line 1717: - Delete since we already do this in `NewClient` per the above.

clientconn.go
```diff
@@ -1707,15 +1706,15 @@ func (cc *ClientConn) parseTargetAndFindResolver() error {
 		defScheme = resolver.GetDefaultScheme()
 	}
 
-	channelz.Infof(logger, cc.channelz, "fallback to scheme %q", defScheme)
+	logger.Infof("fallback to scheme %q", defScheme)
```
dfawley, line 1709: * This can be deleted.

clientconn.go
```diff
@@ -1707,15 +1706,15 @@ func (cc *ClientConn) parseTargetAndFindResolver() error {
...
 	canonicalTarget := defScheme + ":///" + cc.target
 
 	parsedTarget, err = parseTarget(canonicalTarget)
 	if err != nil {
-		channelz.Infof(logger, cc.channelz, "dial target %q parse failed: %v", canonicalTarget, err)
+		logger.Infof("dial target %q parse failed: %v", canonicalTarget, err)
```
dfawley, line 1714: Seems unnecessary as we will return this error to the user.

clientconn.go
```diff
@@ -1838,6 +1837,5 @@ func (cc *ClientConn) determineAuthority() error {
 	} else {
 		cc.authority = encodeAuthority(endpoint)
 	}
-	channelz.Infof(logger, cc.channelz, "Channel authority set to %q", cc.authority)
```
dfawley, line 1841: LGTM

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,34 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
...
+func (do *testPerTargetDialOption) DialOption(parsedTarget url.URL) DialOption {
+	if parsedTarget.Scheme == "passthrough" {
+		return WithTransportCredentials(insecure.NewCredentials()) // credentials provided, should pass NewClient.
+	}
+	return EmptyDialOption{} // no credentials, should fail NewClient
+}
+
+// TestGlobalPerTargetDialOption configures a global per target dial option that
+// produces transport credentials for channels using "passthrough" scheme.
+// Channels that use the passthrough scheme should this be successfully created
```
dfawley, line 89: "This" makes no sense.  :laughing: 

(Delete "this")

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,34 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
...
+	internal.AddGlobalPerTargetDialOptions.(func(opt any))(&testPerTargetDialOption{})
+	noTSecStr := "no transport security set"
+	if _, err := NewClient("dns:///fake"); !strings.Contains(fmt.Sprint(err), noTSecStr) {
+		t.Fatalf("Dialing received unexpected error: %v, want error containing \"%v\"", err, noTSecStr)
+	}
+	cc, err := NewClient("passthrough:///nice")
+	if err != nil {
+		t.Fatalf("Dialing with insecure credentials failed: %v", err)
+	}
+	defer cc.Close()
```
dfawley, line 102: Nit: just `cc.Close()`?

default_dial_option_server_option_test.go
```diff
@@ -74,6 +75,34 @@ func (s) TestDisableGlobalOptions(t *testing.T) {
...
+	internal.ClearGlobalPerTargetDialOptions()
```
dfawley, line 103: Defer at the start, otherwise if this fails it impacts other tests.

dialoptions.go
```diff
@@ -89,6 +98,17 @@ type DialOption interface {
 
 var globalDialOptions []DialOption
 
+// perTargetDialOption takes a parsed target and returns a dial option to apply.
+//
+// This gets called after NewClient() parses the target, and allows per target
+// configuration set through a returned DialOption.
+type perTargetDialOption interface {
+	// DialOption returns a Dial Option to apply.
+	DialOption(parsedTarget url.URL) DialOption
```
dfawley, line 107: Bikeshed: `DialOptionForTarget`?

clientconn.go
zasweq: This reads and writes a lot of stuff from cc. Unless you want me to pass in the target, dial options, and return a bunch of things to NewClient to write to cc's stuff. For the sake of readability I think it's better to keep it on cc.

zasweq: This used to all be in the function body inline and Easwar rewrote it (which I reviewed) to be on cc to factor out functionality into readable helpers and keep this function body clean.

clientconn.go
zasweq: Added comment. I don't think I can move it right before return because the component creation reads the channelz pointer of the channel to log in future, and if I move to right before return would read off a nil pointer and not log anything.

clientconn.go
zasweq: Deleted (and switched conditional to err == nil.

clientconn.go
zasweq: Done.

clientconn.go
zasweq: Done.

clientconn.go
zasweq: Done, yeah because this now gets logged.

clientconn.go
zasweq: Good point, deleted.

default_dial_option_server_option_test.go
zasweq: Whoops haha, deleted.

default_dial_option_server_option_test.go
zasweq: Sure, changed.

default_dial_option_server_option_test.go
zasweq: Changed, and also changed the other tests in this file I wrote for gcp observability :).

dialoptions.go
zasweq: Sure, done.

clientconn.go
dfawley: I guess let's rename these to `initParsedTargetAndResolverBuilder` and `initAuthority` and then at least it's obvious they're being called before the cc is fully initialized.

internal/internal.go
dfawley: E.g. `WithResolvers` doesn't work.  Anything else?

dfawley (approved):

PR #7242, opened 2024-05-18
fix: deprecated code changes

yosuzzy:
"Dial" method is deprecated.

So use "NewClient" instead. Will be better for newbee

### DIFF
--- a/examples/gotutorial.md
+++ b/examples/gotutorial.md
@@ -278,17 +278,17 @@ In this section, we'll look at creating a Go client for our `RouteGuide` service
 
 ### Creating a stub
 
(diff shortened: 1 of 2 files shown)


purnesh42H:
Thanks for the PR. Would you be able to change all the occurrences of it in documentation and examples?

purnesh42H:
@yosuzzy you probably should first try searching for `grpc.Dial(` in [/examples](https://github.com/grpc/grpc-go/tree/master/examples) and[ /documentation](https://github.com/grpc/grpc-go/tree/master/Documentation) and replace to equivalent `NewClient` calls

dfawley:
I think I got these and more in #7248

PR #7256, opened 2024-05-22
stats/opentelemetry: CSM Observability client side component changes

zasweq:
This PR adds the CSM Observability logic for the client side of the OpenTelemetry component.

RELEASE NOTES: N/A

### DIFF
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -22,6 +22,8 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
+	istats "google.golang.org/grpc/internal/stats"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/stats"
 	"google.golang.org/grpc/status"
 
@@ -62,6 +64,14 @@ func (csh *clientStatsHandler) unaryInterceptor(ctx context.Context, method stri
 		method: csh.determineMethod(method, opts...),
 	}
 	ctx = setCallInfo(ctx, ci)
+
+	if csh.o.MetricsOptions.pluginOption != nil {
+		md := csh.o.MetricsOptions.pluginOption.GetMetadata()
+		val := md.Get("x-envoy-peer-metadata")
+		if len(val) == 1 {
+			ctx = metadata.AppendToOutgoingContext(ctx, metadataExchangeKey, val[0])
+		}
+	}
 
 	startTime := time.Now()
 	err := invoker(ctx, method, req, reply, cc, opts...)
@@ -123,8 +133,21 @@ func (csh *clientStatsHandler) HandleConn(context.Context, stats.ConnStats) {}
 
 // TagRPC implements per RPC attempt context management.
 func (csh *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	// Numerous stats handlers can be used for the same channel. The cluster
+	// impl balancer which writes to this will only write once, thus have this
+	// stats handler's per attempt scoped context point to the same optional
+	// labels map if set.
+	var labels *istats.Labels
+	if labels = istats.GetLabels(ctx); labels == nil {
+		labels = &istats.Labels{
+			TelemetryLabels: make(map[string]string),
+		} // Create optional labels map only if first stats handler in possible chain for a channel.
+	}
+	ctx = istats.SetLabels(ctx, labels)
 	mi := &metricsInfo{ // populates information about RPC start.
 		startTime: time.Now(),
+		xDSLabels: labels.TelemetryLabels,
+		method:    info.FullMethodName,
 	}
 	ri := &rpcInfo{
 		mi: mi,
@@ -155,9 +178,20 @@ func (csh *clientStatsHandler) processRPCEvent(ctx context.Context, s stats.RPCS
 		atomic.AddInt64(&mi.sentCompressedBytes, int64(st.CompressedLength))
 	case *stats.InPayload:
 		atomic.AddInt64(&mi.recvCompressedBytes, int64(st.CompressedLength))
+	case *stats.InHeader:
+		csh.getLabelsFromPluginOption(mi, st.Header)
+	case *stats.InTrailer:
+		csh.getLabelsFromPluginOption(mi, st.Trailer)
 	case *stats.End:
 		csh.processRPCEnd(ctx, mi, st)
 	default:
+	}
+}
+
+func (csh *clientStatsHandler) getLabelsFromPluginOption(mi *metricsInfo, incomingMetadata metadata.MD) {
+	if !mi.labelsReceived && csh.o.MetricsOptions.pluginOption != nil {
+		mi.labels = csh.o.MetricsOptions.pluginOption.GetLabels(incomingMetadata)
+		mi.labelsReceived = true
 	}
 }
 
@@ -174,7 +208,23 @@ func (csh *clientStatsHandler) processRPCEnd(ctx context.Context, mi *metricsInf
 		st = canonicalString(s.Code())
 	}
 
-	clientAttributeOption := metric.WithAttributes(attribute.String("grpc.method", ci.method), attribute.String("grpc.target", ci.target), attribute.String("grpc.status", st))
+	attributes := []attribute.KeyValue{
+		attribute.String("grpc.method", ci.method),
+		attribute.String("grpc.target", ci.target),
+		attribute.String("grpc.status", st),
+	}
+
+	for k, v := range mi.labels {
+		attributes = append(attributes, attribute.String(k, v))
+	}
+
+	for _, o := range csh.o.MetricsOptions.OptionalLabels {
+		if val, ok := mi.xDSLabels[o]; ok {
+			attributes = append(attributes, attribute.String(o, val))
+		}
+	}
+
+	clientAttributeOption := metric.WithAttributes(attributes...)
 	csh.clientMetrics.attemptDuration.Record(ctx, latency, clientAttributeOption)
 	csh.clientMetrics.attemptSentTotalCompressedMessageSize.Record(ctx, atomic.LoadInt64(&mi.sentCompressedBytes), clientAttributeOption)
 	csh.clientMetrics.attemptRcvdTotalCompressedMessageSize.Record(ctx, atomic.LoadInt64(&mi.recvCompressedBytes), clientAttributeOption)
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -200,7 +200,7 @@ func (csh *clientStatsHandler) processRPCEvent(ctx context.Context, s stats.RPCS
 }
 
 func (csh *clientStatsHandler) setLabelsFromPluginOption(ai *attemptInfo, incomingMetadata metadata.MD) {
-	if ai.pluginOptionLabels != nil && csh.o.MetricsOptions.pluginOption != nil {
+	if ai.pluginOptionLabels == nil && csh.o.MetricsOptions.pluginOption != nil {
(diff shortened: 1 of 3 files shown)


stats/opentelemetry/client_metrics.go
```diff
@@ -63,6 +65,14 @@ func (csh *clientStatsHandler) unaryInterceptor(ctx context.Context, method stri
 	}
 	ctx = setCallInfo(ctx, ci)
 
+	if csh.o.MetricsOptions.pluginOption != nil {
+		md := csh.o.MetricsOptions.pluginOption.GetMetadata()
+		val := md.Get("x-envoy-peer-metadata")
```
dfawley, line 70: Range over all instead as discussed offline.

stats/opentelemetry/client_metrics.go
```diff
@@ -123,8 +133,21 @@ func (csh *clientStatsHandler) HandleConn(context.Context, stats.ConnStats) {}
...
 func (csh *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	// Numerous stats handlers can be used for the same channel. The cluster
+	// impl balancer which writes to this will only write once, thus have this
+	// stats handler's per attempt scoped context point to the same optional
+	// labels map if set.
+	var labels *istats.Labels
+	if labels = istats.GetLabels(ctx); labels == nil {
+		labels = &istats.Labels{
+			TelemetryLabels: make(map[string]string),
+		} // Create optional labels map only if first stats handler in possible chain for a channel.
```
dfawley, line 144: Put this above the `if`?  Combine with previous comment?

stats/opentelemetry/client_metrics.go
```diff
@@ -123,8 +133,21 @@ func (csh *clientStatsHandler) HandleConn(context.Context, stats.ConnStats) {}
...
+	}
+	ctx = istats.SetLabels(ctx, labels)
```
dfawley, line 146: Doesn't this need to be inside the `if`?

Do you have testing with multiple stats handlers to confirm they're working correctly?

stats/opentelemetry/client_metrics.go
```diff
@@ -155,12 +178,23 @@ func (csh *clientStatsHandler) processRPCEvent(ctx context.Context, s stats.RPCS
 		atomic.AddInt64(&mi.sentCompressedBytes, int64(st.CompressedLength))
 	case *stats.InPayload:
 		atomic.AddInt64(&mi.recvCompressedBytes, int64(st.CompressedLength))
+	case *stats.InHeader:
+		csh.getLabelsFromPluginOption(mi, st.Header)
```
dfawley, line 182: Maybe this should be `set` since it doesn't seem to `get` (return) anything?

stats/opentelemetry/opentelemetry.go
```diff
@@ -220,6 +231,10 @@ type metricsInfo struct {
 
 	startTime time.Time
 	method    string
+
+	labelsReceived bool
+	labels         map[string]string // labels to attach to metrics emitted
```
dfawley, line 236: Nit: `pluginOptionLabels`?

stats/opentelemetry/opentelemetry.go
```diff
@@ -220,6 +231,10 @@ type metricsInfo struct {
 
 	startTime time.Time
 	method    string
+
+	labelsReceived bool
```
dfawley, line 235: Should we replace this with a `nil` check on `labels`?

stats/opentelemetry/client_metrics.go
```diff
@@ -174,7 +208,23 @@ func (csh *clientStatsHandler) processRPCEnd(ctx context.Context, mi *metricsInf
 		st = canonicalString(s.Code())
 	}
 
-	clientAttributeOption := metric.WithAttributes(attribute.String("grpc.method", ci.method), attribute.String("grpc.target", ci.target), attribute.String("grpc.status", st))
+	attributes := []attribute.KeyValue{
```
dfawley, line 211: Can we maybe import the otel packages very-generic-names as `otel*` like `otelattribute` and `otelmetric`?  That would help with readability for me.

stats/opentelemetry/client_metrics.go
```diff
@@ -174,7 +208,23 @@ func (csh *clientStatsHandler) processRPCEnd(ctx context.Context, mi *metricsInf
...
+		attribute.String("grpc.method", ci.method),
```
dfawley, line 212: Should `metricsInfo` be named `attemptInfo` to distinguish it better from `callInfo`?  It took me awhile to figure this out while reviewing this change.

stats/opentelemetry/opentelemetry.go
```diff
@@ -220,6 +231,10 @@ type metricsInfo struct {
...
+	labels         map[string]string // labels to attach to metrics emitted
+	xDSLabels      map[string]string
```
dfawley, line 237: Nit: `xdsLabels`

stats/opentelemetry/opentelemetry.go
```diff
@@ -27,11 +27,15 @@ import (
...
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/internal"
+	otelinternal "google.golang.org/grpc/stats/opentelemetry/internal"
 
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/metric/noop"
 )
 
+// metadataExchangeKey is the key for HTTP metadata exchange.
+const metadataExchangeKey = "x-envoy-peer-metadata"
```
dfawley, line 37: This can be deleted with the range, I think.

stats/opentelemetry/opentelemetry.go
```diff
@@ -126,6 +130,13 @@ type MetricsOptions struct {
 	// grpc.StaticMethodCallOption as a call option into Invoke or NewStream.
 	// This only applies for server side metrics.
 	MethodAttributeFilter func(string) bool
+
+	// OptionalLabels are labels received from xDS that this component should
```
dfawley, line 134: Maybe this shouldn't mention xDS?  "labels received from LB policies"?

stats/opentelemetry/client_metrics.go
zasweq: Deleted since I guess already stated in previous comment.

stats/opentelemetry/client_metrics.go
zasweq: Moved. Will add a test in #7259.

stats/opentelemetry/client_metrics.go
zasweq: Done. Kept the length check since this is a slice of values.

stats/opentelemetry/client_metrics.go
zasweq: Switched.

stats/opentelemetry/client_metrics.go
zasweq: Done. Will do this in server side PR once we get to that too.

stats/opentelemetry/client_metrics.go
zasweq: It's scoped to call server side, and it uses the same. I'd like to change it, but do you have a better name that works for both sides?

stats/opentelemetry/opentelemetry.go
zasweq: Deleted.

stats/opentelemetry/opentelemetry.go
zasweq: Done. Yeah xDS Configures LB Policies, but LB Policies pass it to this.

stats/opentelemetry/opentelemetry.go
zasweq: The only change would be if GetLabels on a user set plugin option returns nil it would call it twice, but then there's no point of the plugin option. Function call in rare case worth code simplicitly.

stats/opentelemetry/opentelemetry.go
zasweq: Switched.

stats/opentelemetry/opentelemetry.go
zasweq: Done. I know the policy is acronym or ACRONYM based off first letter but I guess this felt undefined because lowercaseUPPERCASEUPPERCASE in the usual usage. I guess the first letter is what everything in acronym should be even if acronym is a mix in standard usage.

stats/opentelemetry/client_metrics.go
dfawley: Attempts and calls are the same thing server-side.

Do we use both of these on the server or just the one?

stats/opentelemetry/opentelemetry.go
dfawley: Hmmm...would this mean that it could generate labels from the trailers in the case of an RPC with both headers and trailers?  Would that "work correctly" - i.e. all metrics would get the labels that came out of the second call?

stats/opentelemetry/client_metrics.go
```diff
@@ -67,9 +67,10 @@ func (csh *clientStatsHandler) unaryInterceptor(ctx context.Context, method stri
 
 	if csh.o.MetricsOptions.pluginOption != nil {
 		md := csh.o.MetricsOptions.pluginOption.GetMetadata()
-		val := md.Get("x-envoy-peer-metadata")
-		if len(val) == 1 {
-			ctx = metadata.AppendToOutgoingContext(ctx, metadataExchangeKey, val[0])
+		for k, v := range md {
+			if len(v) == 1 {
```
dfawley, line 71: Why are we imposing this restriction here?

stats/opentelemetry/opentelemetry.go
zasweq: The CSM Plugin Option specifically will never return nil labels if set. It'll always return at least the local labels, which will be "unknown" if nothing set in env.

stats/opentelemetry/client_metrics.go
zasweq: We only use this on the server side, CallInfo is only used client side. If call info is the same as AttemptInfo then I can Switch this name to AttemptInfo.

dfawley: Ohhh, I'm sorry, I thought this was storing per-attempt data on the client side, vs. per-call data.  So it took me awhile to figure out and I got it wrong...

Maybe they should be `clientCallInfo` and `serverCallInfo` then?

stats/opentelemetry/opentelemetry.go
dfawley: Right but it's using a "generic" (not exported) interface that anything else can implement.  So I'm fine with either:

1. Documenting that it must never return `nil` and (probably?) validating that it never returns nil.
2. Not documenting that but validating that it's always non-nil.
3. Actually allowing it, and letting the labels get determined by either one at the choice of the PluginOption.  It seems like it will actually "work correctly" in this case.

Of these I'm probably leaning toward (2).  I don't think we can make errors happen or anything, but we could replace `nil` with the empty map (`map[string]string{}`).

stats/opentelemetry/client_metrics.go
zasweq: This is storing per attempt info. The Client Call Info has a specific type: https://github.com/grpc/grpc-go/pull/7256/files#diff-6cb099cd7cc7853ea6b069e48ce1493ae735d7a56f7e36a8a828b20d82470eddR176

stats/opentelemetry/client_metrics.go
zasweq: Changed to for each over the values and writing all values.

stats/opentelemetry/opentelemetry.go
zasweq: Implemented 2.

dfawley (approved):

PR #7271, opened 2024-05-27
ringhash: port e2e tests from c-core

atollena:
Fixes https://github.com/grpc/grpc-go/issues/6072.

RELEASE NOTES: none


xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +142,336 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+		Endpoints: []*endpointv3.ClusterLoadAssignment{ep1, ep2},
+		Clusters:  []*v3clusterpb.Cluster{cluster, cluster1, cluster2},
+		Routes:    []*v3routepb.RouteConfiguration{route},
+		Listeners: []*v3listenerpb.Listener{listener},
+	})
+	if err != nil {
+		t.Fatalf("failed to update xDS resources: %v", err)
+	}
+
+	// Here we use the hostname field in the response proto to verify which
```
easwars, line 304: Is it possible to the Peer call option to figure out where the request lands instead of relying on the HostName field in the response? Would that work better?

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +142,336 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	conn, err := grpc.NewClient("xds:///test.server", grpc.WithResolvers(xdsResolver), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("failed to create client: %s", err)
+	}
+	conn.Connect() // force triggering resource fetch. This isn't great, see comment below.
+	defer conn.Close()
+	client := testgrpc.NewTestServiceClient(conn)
+
+	// Wait to give time to the LB policy to request the child clusters.
```
easwars, line 283: If we can get the server to respond to a subset of resources using this: https://github.com/grpc/grpc-go/blob/master/internal/testutils/xds/e2e/server.go#L82, can we rely on channel state or some other state instead of sleeping?

atollena: I think allowing resource subset works fine for this case actually, and we don't need any time-sensitive code. I've updated the code to use it.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -20,21 +20,28 @@ package ringhash_test
...
 import (
 	"context"
+	"strconv"
 	"testing"
 	"time"
 
+	v3clusterpb "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	endpointv3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	v3listenerpb "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	v3routepb "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
```
easwars, line 30: Nit: Could you please group these proto imports with other proto imports (currently on line 41). Thanks.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -62,6 +69,7 @@ func (*testService) EmptyCall(context.Context, *testpb.Empty) (*testpb.Empty, er
 // ring contains a single subConn, and verifies that when the server goes down,
 // the LB policy on the client automatically reconnects until the subChannel
 // moves out of TRANSIENT_FAILURE.
+// XXX do we need to keep this one?
```
easwars, line 72: If I remember correctly, we added a special case in the code for this logic. So, unless this is covered by the 24 tests from c-core, we should continue to hold on to this.

Also, worthwhile might to get test coverage from before and after your tests. I believe we have a codecov bot and @arvindbr8 would know the details.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +142,318 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
 
 	// An RPC at this point is expected to fail.
 	if _, err := client.EmptyCall(ctx, &testpb.Empty{}); err != nil {
 		t.Fatalf("rpc EmptyCall() failed: %v", err)
 	}
 }
+
+// XXX this is copy pasted from
+// https://github.com/grpc/grpc-go/blob/9199290ff8fee7c37283586929af39ba23dbab4b/xds/internal/balancer/clusterresolver/e2e_test/eds_impl_test.go#L93
+// Perhaps we should move it to a common package?
```
easwars, line 156: We *probably could* move it to a common package. I would assume then that the StubServer would need to implement all methods on the `TestService` to some acceptable degree.

But given that this is already going to a big change, I'm totally OK with having some repetitive test helper code spread across our code base.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +142,318 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	}
+}
+
+// makeNonExistingBackends starts numBackends backends and stops them
+// immediately to simulate a server that is not reachable.
+func makeNonExistingBackends(t *testing.T, numBackends int) []*stubserver.StubServer {
+	var servers []*stubserver.StubServer
+	for i := 0; i < numBackends; i++ {
+		servers = append(servers, stubserver.StartTestService(t, nil))
+		servers[i].Stop()
```
easwars, line 183: Can this lead to test flakes at all, if the port released here (after the server is closed) is used by another test backend spun up by the same test binary?

easwars: Put another way, is there a different/better way to simulate a server that is not reachable?

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +142,318 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	}
+	return servers
+}
+
+// backendOptions returns a slice of e2e.BackendOptions for the given stub
+// servers.
+func backendOptions(t *testing.T, servers []*stubserver.StubServer) []e2e.BackendOptions {
```
easwars, line 190: Here and in other helper functions, I don't see `t` being used for anything other than calling `t.Helper()`. Can we not pass `t` in these cases instead?

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +142,318 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	return backendOpts
+}
+
+// TestRingHash_AggregateClusterFallBackFromRingHashAtStartup tests that when
+// an aggregate cluster is configured with ring hash policy, and the first
+// cluster is in transient failure, all RPCs are sent to the second cluster
+// using the ring hash policy.
+func (s) TestRingHash_AggregateClusterFallBackFromRingHashAtStartup(t *testing.T) {
+	// origin: https://github.com/grpc/grpc/blob/083bbee4805c14ce62e6c9535fe936f68b854c4f/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc#L97
+
```
easwars, line 207: We can probably get rid of this comment.

easwars (commented):
I think the overall approach looks good. So, I will wait for you to move it out of draft before I give any further comments. Thanks.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: It works equally well and removes some clutter, so I took your suggestion.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
arvindbr8: tsk, Codecov job run seems to have [silently failed](https://github.com/grpc/grpc-go/actions/runs/9273303440/job/25513084598?pr=7271#step:5:32) - seems to be a throttling issue. I will fix it.  ugh, It looked strange to me that the coverage report was missing.

atollena:
I implemented 8 out of the 23 tests in https://github.com/grpc/grpc/blob/master/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc.

I'd suggest we try to get those in and add more tests in a separate commit.

Regarding potential flakyness: I ran those tests with -count 1000 which didn't yield any failure.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +154,794 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+		}},
+	})
+	cluster := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		ClusterName: clusterName,
+		ServiceName: clusterName,
+	})
+	// Increasing min ring size for random distribution.
+	config := testutils.MarshalAny(t, &v3ringhashpb.RingHash{
+		HashFunction:    v3ringhashpb.RingHash_XX_HASH,
+		MinimumRingSize: &wrapperspb.UInt64Value{Value: 100000},
```
atollena, line 816: Actually this doesn't work because raising the minimum ring size is gated behind an environment variable. I'll look into:
- whether increasing the ring size really makes the distribution more uniform
- how to set the environment variable in tests (e.g. via mutating envconfig perhaps?)

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +154,794 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+			t.Errorf("Got %v RPCs routed to a backend, want %v", got, want)
+		}
+	}
+
+}
+
+// computeIdealNumberOfRPCs computes the ideal number of RPCs to send so that
+// we can observe an event happening with probability p, and the result will
+// have value p with the given error tolerance.
+func computeIdealNumberOfRPCs(p, errorTolerance float64) int {
```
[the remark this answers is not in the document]
atollena: OK, that was actually due to some rounding errors and not properly using ring hash 🤦.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: I ended up setting envconfig in a helper here.

easwars:
I see that this is still assigned to you. Please assign it to me when you think it is ready for me to take a pass. Thanks.

atollena:
> I see that this is still assigned to you. Please assign it to me when you think it is ready for me to take a pass. Thanks.

This is ready for review.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	defer stop()
+
+	// Stop the servers that we want to be unreachable.
+	stopNonExistant()
+
+	newCluster1Name := "new_cluster_1"
+	newEdsService1Name := "new_eds_service_1"
+	newCluster2Name := "new_cluster_2"
+	newEdsService2Name := "new_eds_service_2"
+	clusterName := "aggregate_cluster"
```
easwars, line 253: How about: `primaryClusterName`, `primaryServiceName`, `secondaryClusterName`, `secondaryServiceName` and `aggregateClusterName` as names for these? And also, these should be const instead of var. Thanks.

Also, in the same vein, the resource names could `primaryCluster`, `primaryEndpoints` etc

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	})
+	cluster2 := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		ClusterName: newCluster2Name,
+		ServiceName: newEdsService2Name,
+	})
+	cluster := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		ClusterName: clusterName,
+		Type:        e2e.ClusterTypeAggregate,
+		// TODO: when "A75: xDS Aggregate Cluster Behavior Fixes" is merged, the
+		// policy will have to be set on the child clusters.
```
easwars, line 283: Nit: Technically, not when A75 is merged, but when support for that is merged in the grpc-go repo.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	err := xdsServer.Update(context.Background(), e2e.UpdateOptions{
+		NodeID:    nodeID,
+		Endpoints: []*v3endpointpb.ClusterLoadAssignment{ep1, ep2},
+		Clusters:  []*v3clusterpb.Cluster{cluster, cluster1, cluster2},
+		Routes:    []*v3routepb.RouteConfiguration{route},
+		Listeners: []*v3listenerpb.Listener{listener},
+	})
+	if err != nil {
+		t.Fatalf("failed to update xDS resources: %v", err)
```
easwars, line 298: Nit: All these test failure error strings can be capitalized. The style guide recommendation to not capitalize error strings does not apply to test failures, log messages etc.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	gotPerBackend := checkRPCSendOK(t, ctx, client, 100)
+	if len(gotPerBackend) != 1 {
+		t.Errorf("Got RPCs routed to %v backends, want %v", len(gotPerBackend), 1)
+	}
+	for _, count := range gotPerBackend {
+		if count != 100 {
+			t.Errorf("Got %v RPCs routed to a backend, want %v", count, 100)
+		}
+	}
```
easwars, line 319: Can we tighten this check?

At this point, it just verifies that all RPCs are sent to one backend. It does not verify that it is actually sent to *one* of the secondary backends. Although, I do agree that adding aggregate clusters into the mix here for ring_hash testing seems excessive. But given that we are following what C-core has in terms of test cases, we should actually verify that RPCs in fact are routed to the backends in the cluster that we expect RPCs to go to.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+		_, err := client.EmptyCall(ctx, &testpb.Empty{}, grpc.Peer(&remote))
+		if err != nil {
+			t.Fatalf("rpc UnaryCall() failed: %v", err)
+		}
+		backendCount[remote.Addr.String()]++
+	}
+	return backendCount
+}
+
+// TestRingHash_AggregateClusterFallBackFromRingHashAtStartup tests that when
```
easwars, line 232: I agree we have test comments that include the test name (just like docstings for regular functions/methods). But in subsequent reviews, we came to the conclusion that that is not really required. So, newer test code can just say `//Tests <foooooo baaaaar>` instead of `//Test_Fooooo_Baaaaaar tests <fooooo barrrrrr>`

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+// addresses as keys and number of RPCs sent to this address as value. Abort the
+// test if any RPC fails.
+func checkRPCSendOK(t *testing.T, ctx context.Context, client testpb.TestServiceClient, num int) map[string]int {
+	t.Helper()
+
+	backendCount := make(map[string]int)
+	for i := 0; i < num; i++ {
+		var remote peer.Peer
+		_, err := client.EmptyCall(ctx, &testpb.Empty{}, grpc.Peer(&remote))
+		if err != nil {
```
easwars, line 224: Can be one line.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+			Weight: 1,
+		})
+	}
+	return backendOpts
+}
+
+// channelIDHashRoute returns a RouteConfiguration with a hash policy that hashes using
+// the provided key.
+func channelIDHashRoute(clusterName string) *v3routepb.RouteConfiguration {
+	route := e2e.DefaultRouteConfig("new_route", "test.server", clusterName)
```
easwars, line 201: Can we either accept listener and route config names along with cluster name as arguments to this func, or accept a route config object that we modify here. Hard-coding the listener and route config resource names here introduces too much coupling, I feel.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+		backendOpts = append(backendOpts, e2e.BackendOptions{
+			Port:   testutils.ParsePort(t, server.Address),
+			Weight: 1,
+		})
+	}
+	return backendOpts
+}
+
+// channelIDHashRoute returns a RouteConfiguration with a hash policy that hashes using
+// the provided key.
```
easwars, line 199: Comment needs a bit of update. The hash policy configured here actually hashes based on the grpc channel ID.

easwars (changes requested):
Initial pass. 

Many of these comments inside of the test apply to multiple tests instead of just the one where I've made the comment.

Thanks.

easwars (changes requested):
Initial pass. 

Many of these comments inside of the test apply to multiple tests instead of just the one where I've made the comment.

Thanks.

internal/testutils/blocking_context_dialer.go
```diff
@@ -0,0 +1,56 @@
...
+	select {
+	case <-d.blockCh:
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+	return d.dialer.DialContext(ctx, "tcp", addr)
+}
+
+// Resume unblocks the dialer. Resume can only be called once.
+func (d *BlockingDialer) Resume() {
```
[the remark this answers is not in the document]
atollena: Added a comment. I anticipate that I will have to change BlockingDialer to offer the same functionalities as c-core connection injector for the rest of the tests though (hold a connection, wait, resume, fail, and add a delay). I'll do it in the next PR, if you're OK with it.

internal/testutils/xds/e2e/clientresources.go
```diff
@@ -708,7 +712,7 @@ type EndpointOptions struct {
 func DefaultEndpoint(clusterName string, host string, ports []uint32) *v3endpointpb.ClusterLoadAssignment {
 	var bOpts []BackendOptions
 	for _, p := range ports {
-		bOpts = append(bOpts, BackendOptions{Port: p})
+		bOpts = append(bOpts, BackendOptions{Port: p, Weight: 1})
```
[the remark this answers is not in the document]
atollena: Done.

internal/testutils/xds/e2e/setup_management_server.go
```diff
@@ -110,3 +111,30 @@ func DefaultBootstrapContents(nodeID, serverURI string) ([]byte, error) {
...
+// addresses of the test backends.
+//
+// Returns the following:
+//   - a channel onto which the DNS target being resolved is written to by the
+//     mock DNS resolver
+//   - a channel to notify close of the DNS resolver
+//   - a channel to notify re-resolution requests to the DNS resolver
+//   - a manual resolver which is used to mock the actual DNS resolution
+//   - a cleanup function which re-registers the original DNS resolver
+func SetupDNS() (chan resolver.Target, chan struct{}, chan resolver.ResolveNowOptions, *manual.Resolver, func()) {
```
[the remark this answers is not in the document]
atollena: Done. I am questioning the ergonomy of this function though. Would it be enough to just return the manual resolver, for most test cases?
Asserting for the exact names that were resolves sounds like it would be useful to test the exact interaction between the channel and the resolver, but I wonder how generally useful that is, and perhaps it encourages brittle tests?

internal/testutils/xds/e2e/setup_management_server.go
```diff
@@ -110,3 +111,30 @@ func DefaultBootstrapContents(nodeID, serverURI string) ([]byte, error) {
 	}
 	return bs, nil
 }
+
+// SetupDNS unregisters the DNS resolver and registers a manual resolver for the
```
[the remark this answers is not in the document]
atollena: Hmm, interesting. Instead of using the passthrough resolver, did you consider mocking the results for the DNS name of the xDS server instead? To me, what you ran into works as intended...

Another thing that may be useful to the facility to mock DNS is to pass local addresses through, which may also have solved your problem.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
 		}
 	}
 	if err := ctx.Err(); err != nil {
-		t.Fatalf("Timeout waiting for channel to reach READT after server restart: %v", err)
+		t.Fatalf("Timeout waiting for channel to reach READY after server restart: %v", err)
 	}
 
 	// An RPC at this point is expected to fail.
```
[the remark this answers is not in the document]
atollena: Ha yes, I had not read this one.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
 		}
 	}
 	if err := ctx.Err(); err != nil {
-		t.Fatalf("Timeout waiting for channel to reach READT after server restart: %v", err)
+		t.Fatalf("Timeout waiting for channel to reach READY after server restart: %v", err)
```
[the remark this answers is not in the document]
atollena: I replaced this usage.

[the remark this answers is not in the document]
atollena: Addressed this dupe comment in https://github.com/grpc/grpc-go/pull/7271#discussion_r1624934144.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -48,6 +67,8 @@ func Test(t *testing.T) {
 const (
 	defaultTestTimeout      = 10 * time.Second
 	defaultTestShortTimeout = 10 * time.Millisecond // For events expected to *not* happen.
+
+	errorTolerance = .05 // For tests that rely on statistical significance.
```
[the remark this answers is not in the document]
atollena: switched to use stubserver everywhere.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +155,797 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	}
+	return backendCount
+}
+
+// TestRingHash_AggregateClusterFallBackFromRingHashAtStartup tests that when
+// an aggregate cluster is configured with ring hash policy, and the first
+// cluster is in transient failure, all RPCs are sent to the second cluster
+// using the ring hash policy.
+func (s) TestRingHash_AggregateClusterFallBackFromRingHashAtStartup(t *testing.T) {
+	xdsServer, nodeID, _, xdsResolver, stop := e2e.SetupManagementServer(t, e2e.ManagementServerOptions{
```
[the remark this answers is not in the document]
atollena: fixed in all tests.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: Done (I had kept the names from c-core).

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: I replaced "merged" with "implemented".

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: Right, fixed.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: I checked that we address the other backend, I had taken that from the c-core tests. I'm keeping a list of mistakes in the c-core tests, I'll open a ticket in grpc/grpc when I'm done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: Done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: Fixed this and a few other occurrences.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: Good point, I added the two as parameters.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: Fixed.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
easwars: It looks like splitting the imports into more than two groups (stdlib and project related imports) is more like a soft recommendation, and the onus is on the project maintainers to keep this consistent across the codebase. See: https://google.github.io/styleguide/go/decisions#import-grouping

I try to do this whenever I see this, but this is not a hard requirement and feel free to push back on this.

But if you think a linter rule can be added in our `vet.sh` for this, I would be thrilled. Not so sure about @dfawley though :)

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
easwars: Yeah, we don't run tests in parallel. That would require us to move some mountains (or maybe not, maybe its just that our leak checker doesn't like parallel tests).

internal/testutils/blocking_context_dialer.go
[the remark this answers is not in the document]
easwars: Sure, sounds good. Thanks.

internal/testutils/xds/e2e/clientresources.go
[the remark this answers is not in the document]
easwars: I don't see it being done, but I don't mind/care at this point :)

internal/testutils/xds/e2e/setup_management_server.go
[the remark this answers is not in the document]
easwars: Yes, we have been having discussions about keeping these setup-y kind of test functions, to be as simple as possible, and keep setup code close to test logic, so that it makes reading the test easier.

Yeah, I do agree that this function does more than would be required for most tests. Some tests use the returned channels (one or all of the returned ones), while some tests don't care about the returned channels, and only want the manual resolver.

I would be fine at this point with whatever approach you want to take:
- Completely get rid of this, and inline the mocking part in each of your tests. Even though that would be duplicate code, that seems to be the recommended approach for tests, instead of having complicated setup code that lives outside of the test logic
- Leave it as it and maybe add a TODO to perform cleanup later.

easwars (changes requested):
Some of the comments apply to all tests. I will continue reviewing from `headerHashRoute` tomorrow.

internal/testutils/xds/e2e/setup_management_server.go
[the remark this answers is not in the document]
atollena: I did option 1: duplicate DNS setup in tests that need it, and removed unused return values.

internal/testutils/xds/e2e/setup_management_server.go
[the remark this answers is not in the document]
atollena: My point was that from a testing perspective, if you're going to intercept all DNS calls, then it makes a lot of sense to have to do it for the management server as well...

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+		t.Fatalf("Failed to create client: %s", err)
+	}
+	defer conn.Close()
+	client := testgrpc.NewTestServiceClient(conn)
+
+	// TODO: c-core injects a 500ms RPC delay here: https://github.com/grpc/grpc/blob/master/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc#L220
+	// Should we try to do the same? do we have a way to do that in Go? Perhaps using something like https://github.com/grpc/grpc-go/blob/04ea82009cdb9ecdefc6289f4c93ec919a10b3b6/benchmark/latency/latency.go#L76-L75?
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
```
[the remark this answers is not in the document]
atollena: Done. I haven't seen such a linter rule, but I think that would be useful. I opened https://github.com/grpc/grpc-go/issues/7304 and set it as P2/Feature. I attached the full list of usage of context.Background, curious whether you think we should go over it and fix it.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	edsCluster := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		ClusterName: edsClusterName,
+		ServiceName: edsClusterName,
+	})
+
+	logicalDNSCluster := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		Type:        e2e.ClusterTypeLogicalDNS,
+		ClusterName: logicalDNSClusterName,
+		DNSHostName: "server.example.com",
+		DNSPort:     443,
```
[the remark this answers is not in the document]
atollena: Nope, added a comment about it (resource is nacked if they're missing).

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	defer cancel()
+
+	gotPerBackend := checkRPCSendOK(t, ctx, client, 100)
+
+	// Since this is using ring hash with the channel ID as the key, all RPCs
+	// are routed to the same backend of the secondary locality.
+	if len(gotPerBackend) != 1 {
+		t.Errorf("Got RPCs routed to %v backends, want %v", len(gotPerBackend), 1)
+	}
+	for backend, got := range gotPerBackend {
```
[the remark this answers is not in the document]
atollena: Wow, that's a trick. I'm not particularly convinced from a readability standpoint, but I'm willing to adopt it if that a common idiom. Also, gofmt reformats this as:

```
var backend string
var numRPCs int
for backend, numRPCs := range gotPerBackend {
}
// The rest of the test logic.
```

which is one more line.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	// TODO: c-core injects a 500ms RPC delay here: https://github.com/grpc/grpc/blob/master/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc#L220
+	// Should we try to do the same? do we have a way to do that in Go? Perhaps using something like https://github.com/grpc/grpc-go/blob/04ea82009cdb9ecdefc6289f4c93ec919a10b3b6/benchmark/latency/latency.go#L76-L75?
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	if _, err = client.EmptyCall(ctx, &testpb.Empty{}); err != nil {
+		t.Errorf("rpc UnaryCall() failed: %v", err)
+	}
```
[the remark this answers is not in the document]
atollena: Actually, this is the same comment as you gave me for the first test -- c-core doesn't make this check, presumably because there is a single listening backend, so there's no way RPC succeed if they go to the wrong backend. I added the check though, I think it makes sense to have it.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+			Weight:   1,
+			Backends: backendOptions(t, nonExistingBackend1),
+			Priority: 0,
+		}, {
+			Name:     "locality1",
+			Weight:   1,
+			Backends: backendOptions(t, nonExistingBackend2),
+			Priority: 1,
+		}},
+	})
```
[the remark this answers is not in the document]
atollena: Again I took that from c-core. The fact the non existant backends are in separate localities with different priorities test a bit more, but arguably that's irrelevant to this test. I removed it.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+			// Increase backoff time, so that subconns stay in TRANSIENT_FAILURE
+			// for long enough to trigger potential problems.
+			Backoff: backoff.Config{
+				BaseDelay: defaultTestTimeout,
+			},
+			MinConnectTimeout: 0,
+		}))
+	if err != nil {
+		t.Fatalf("Failed to create client: %s", err)
+	}
```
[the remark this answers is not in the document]
atollena: Done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	wg := sync.WaitGroup{}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	wg.Add(1)
+	go func() {
+		if _, err := client.EmptyCall(ctx, &testpb.Empty{}); err != nil {
+			t.Errorf("First rpc UnaryCall() failed: %v", err)
+		}
+		wg.Done()
```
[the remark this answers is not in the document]
atollena: Done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -134,11 +154,794 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	dnsR.InitialState(resolver.State{Addresses: []resolver.Address{{Addr: backends[0].Address}}})
+
+	conn, err := grpc.NewClient("xds:///test.server", grpc.WithResolvers(xdsResolver), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("failed to create client: %s", err)
+	}
+	defer conn.Close()
+	client := testgrpc.NewTestServiceClient(conn)
+
+	// TODO: c-core injects a 500ms RPC delay here: https://github.com/grpc/grpc/blob/master/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc#L220
```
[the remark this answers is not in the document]
atollena: OK, I read the connection injector stuff from c-core a bit more in details, and it also only adds delay to connection attempts. It would be easy to add this feature to the blocking dialer that I added in this PR.

The delay feature is only used in this test in c-core for ring hash, and I also don't feel like it should be changing the expected behavior in any way. I removed this TODO.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: OK, let's wait until we have consensus with the team before spending time doing the linter thing or even opening a ticket. We can discuss this offline. In the mean time I grouped the imports.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,775 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	conn, err := grpc.NewClient("xds:///test.server", grpc.WithResolvers(xdsResolver), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("Failed to create client: %s", err)
+	}
+	defer conn.Close()
+	client := testgrpc.NewTestServiceClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
```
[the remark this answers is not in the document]
atollena: Addressed in https://github.com/grpc/grpc-go/pull/7271#discussion_r1629119693.

easwars:
Looks like one of the tests added in this PR is flaky here: https://github.com/grpc/grpc-go/actions/runs/9398809643/job/25884944530?pr=7271.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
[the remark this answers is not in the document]
easwars: Thanks.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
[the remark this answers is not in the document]
easwars: I didn't want to force this option on you, and hence had marked the comment optional. When I was new to the team and saw this pattern for the first time, I didn't like it either. But have grown to like it over time.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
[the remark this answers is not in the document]
easwars: Maybe we can throw an "Help-Wanted" on it, and see if someone is interested to fix it. We are far off from october, but we can still hope for the best :)

easwars (approved):
LGTM, modulo minor nits. Thanks again for adding these tests.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,801 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	// Note each type of RPC contains a header value that will always be hashed
+	// to a specific backend as the header value matches the value used to
+	// create the entry in the ring.
+	for _, backend := range backends {
+		ctx := metadata.NewOutgoingContext(ctx, metadata.Pairs("address_hash", backend.Address+"_0"))
+		reqPerBackend := checkRPCSendOK(t, ctx, client, 1)
+		if reqPerBackend[backend.Address] != 1 {
+			t.Errorf("Got RPC routed to backend %v, want %v", reqPerBackend, backend.Address)
+		}
```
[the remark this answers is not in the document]
atollena: I think that what you suggest tests something different than this test. This test makes sure that the algorithm we use for ring hash, which is deterministic, sends the request to the right backend for specific values of the header.

Perhaps what you are suggesting is a different test: generate random hashes, and make sure for a given hash it always lands on the same backend. I think there's value in what you propose, but it wasn't part of the c-core suite.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,801 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	if want := 1; len(gotPerBackend) != want {
+		t.Errorf("Got RPCs routed to %v backends, want %v", len(gotPerBackend), want)
+	}
+	var got int
+	for _, got = range gotPerBackend {
+	}
+	if want := 400; got != want {
+		t.Errorf("Got %v RPCs routed to a backend, want %v", got, want)
+	}
+
```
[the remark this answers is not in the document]
atollena: Done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,801 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	for _, count = range received {
+	}
+	if count != 100 {
+		t.Errorf("Got %v RPCs routed to a backend, want %v", count, 100)
+	}
+}
+
+// headerHashRoute creates a RouteConfiguration with a hash policy that uses the
+// provided header.
+func headerHashRoute(clusterName, header string) *v3routepb.RouteConfiguration {
```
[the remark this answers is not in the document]
atollena: Done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,801 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	})
+	edsCluster := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		ClusterName: edsClusterName,
+		ServiceName: edsClusterName,
+	})
+
+	logicalDNSCluster := e2e.ClusterResourceWithOptions(e2e.ClusterOptions{
+		Type:        e2e.ClusterTypeLogicalDNS,
+		ClusterName: logicalDNSClusterName,
+		DNSHostName: "server.example.com",
```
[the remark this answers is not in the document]
atollena: Done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,801 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+	backends, stop := startTestServiceBackends(t, 1)
+	defer stop()
+
+	nonExistingBackend1, stop1 := startTestServiceBackends(t, 2)
+	nonExistingBackend2, stop2 := startTestServiceBackends(t, 2)
+	stop1()
+	stop2()
+
+	endpoints := e2e.EndpointResourceWithOptions(e2e.EndpointOptions{
```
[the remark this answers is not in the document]
atollena: Yes, done.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,801 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+
+// computeIdealNumberOfRPCs computes the ideal number of RPCs to send so that
+// we can observe an event happening with probability p, and the result will
+// have value p with the given error tolerance.
+//
+// See https://github.com/grpc/grpc/blob/4f6e13bdda9e8c26d6027af97db4b368ca2b3069/test/cpp/end2end/xds/xds_end2end_test_lib.h#L941
+// for an explanation of the formula.
+func computeIdealNumberOfRPCs(p, errorTolerance float64) int {
+	if p < 0 || p > 1 {
+		panic("p must be in (0, 1)")
```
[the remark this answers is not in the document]

atollena:
> Looks like one of the tests added in this PR is flaky here: https://github.com/grpc/grpc-go/actions/runs/9398809643/job/25884944530?pr=7271.

I looked into this, and I think there is a race which is not obvious.
Here is what may be happening:
1. We create a stubserver and close it immediately, to simulate a server that is unreachable
1. The stub server calls `Serve` in a separate goroutine spawned at https://github.com/grpc/grpc-go/blob/e40eb2e2c1bd31663ca4210fa51a3cfd0a059dba/internal/stubserver/stubserver.go#L181 -- if this goroutine runs immediately, then the gRPC server is started properly. Calling `Stop()` on it will make sure the listener is closed synchronously at https://github.com/grpc/grpc-go/blob/e40eb2e2c1bd31663ca4210fa51a3cfd0a059dba/server.go#L1886-L1892.
1. However, if `server.Stop()` is called before `server.Serve()` is scheduled, then the listener is not yet registered and the listener is not closed. This is what happened in this test run, and in my tests in most cases if you do:
```
lis, err := net.Listen
server := grpc.NewServer()
go server.Serve(lis)
server.Stop()
```
then `Stop` will run before `Serve`.
Even when this happens, in most cases the listener is then closed by `Serve` (behavior implemented at https://github.com/grpc/grpc-go/blob/e40eb2e2c1bd31663ca4210fa51a3cfd0a059dba/server.go#L843-L848).

However, since that happens in a separate goroutine, there's no guarantee that closing the listener will run before the first connections, later on in a test.

I think the best way to work around this is to directly create listeners instead of creating stubservers. This avoids the asynchronous close behavior described above, while still reserving an available port and making sure it's not used by another server. I did that in the last changes.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -128,17 +136,811 @@ func (s) TestRingHash_ReconnectToMoveOutOfTransientFailure(t *testing.T) {
...
+	defer stop()
+
+	servers, stop := startTestServiceBackends(t, 2)
+	defer stop()
+
+	primaryClusterName := "new_cluster_1"
+	primaryServiceName := "new_eds_service_1"
+	secondaryClusterName := "new_cluster_2"
+	secondaryServiceName := "new_eds_service_2"
+	clusterName := "aggregate_cluster"
```
easwars, line 251: Ah, sorry, should have caught this much earlier. Could we please use consts here, and in other tests as well. Thanks.

easwars (approved):
Thanks for taking care of all the comments.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
atollena: I fixed that in the last commit.

easwars (approved):
LGTM

Yay!!!

arvindbr8 (approved):

arvindbr8:
Thanks @atollena for taking care of this!

PR #7273, opened 2024-05-28
advancedtls: use realistic ciphersuite in test

rolandshoemaker:
Instead of 3DES, something which should basically never be used in production. Go is removing default support for 3DES is Go 1.24, requiring new modules to opt into support for this cipher.

This fixes an internal Google test failure, see b/343028077.

RELEASE NOTES: none


gtcooke94 (approved):
LGTM

PR #7277, opened 2024-05-29
stats/opentelemetry: Add CSM Observability API

zasweq:
RELEASE NOTES: N/A

### DIFF
--- a/stats/opentelemetry/csm/observability.go
+++ b/stats/opentelemetry/csm/observability.go
@@ -28,19 +28,7 @@ import (
 	otelinternal "google.golang.org/grpc/stats/opentelemetry/internal"
 )
 
-var clientSideOTelWithCSM grpc.DialOption
-var clientSideOTel grpc.DialOption
-
-// Observability sets up CSM Observability for the binary globally. It sets up
-// two client side OpenTelemetry instrumentation components configured with the
-// provided options, one with a CSM Plugin Option configured, one without.
-// Created channels will pick up one of these dependent on whether channels are
-// CSM Channels or not, which is derived from the Channel's parsed target after
-// dialing.
-//
-// It sets up a server side OpenTelemetry instrumentation component configured
-// with options provided alongside a CSM Plugin Option to be registered globally
-// and picked up for every server.
+// EnableObservability sets up CSM EnableObservability for the binary globally.
 //
 // The CSM Plugin Option is instantiated with local labels and metadata exchange
 // labels pulled from the environment, and emits metadata exchange labels from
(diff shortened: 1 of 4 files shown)


stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,95 @@
...
+var clientSideOTelWithCSM grpc.DialOption
+var clientSideOTel grpc.DialOption
+
+// Observability sets up CSM Observability for the binary globally. It sets up
+// two client side OpenTelemetry instrumentation components configured with the
+// provided options, one with a CSM Plugin Option configured, one without.
+// Created channels will pick up one of these dependent on whether channels are
+// CSM Channels or not, which is derived from the Channel's parsed target after
+// dialing.
+//
```
dfawley, line 40: Is much/all of this implementation details the user does not care about?  I'm not sure but I was thinking so.

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,95 @@
...
+// and picked up for every server.
+//
+// The CSM Plugin Option is instantiated with local labels and metadata exchange
+// labels pulled from the environment, and emits metadata exchange labels from
+// the peer and local labels. Context timeouts do not trigger an error, but set
+// certain labels to "unknown".
+//
+// This function is not thread safe, and should only be invoked once in main
+// before any channels or servers are created. Returns a cleanup function to be
+// deferred in main.
```
dfawley, line 52: What's the reason to return a cleanup function if it needs to be called in main?  Can we just remove it?

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,95 @@
...
+func Observability(ctx context.Context, options opentelemetry.Options) func() {
```
dfawley, line 53: Nit: `EnableObservability`?  `Observability` is a noun but it doesn't return an object.

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,95 @@
...
+
+// replace grpc module and otel module I guess
+
+type perTargetDialOption struct {}
+
+func (perTargetDialOption) DialOptionForTarget(parsedTarget url.URL) grpc.DialOption {
+	if determineTargetCSM(&parsedTarget) {
+		return clientSideOTelWithCSM
+	}
+	return clientSideOTel
```
dfawley, line 83: Can we store these in the `perTargetDialOption` instead of having globals?

stats/opentelemetry/csm/observability.go
dfawley, line 53: Why are we eating the gcp detector's error instead of passing it back to the user?  That seems wrong, especially if there is a problem with the context.  We have a way to return it (from here), so it seems like we should.

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,95 @@
...
+func Observability(ctx context.Context, options opentelemetry.Options) func() {
+	csmPluginOption := newPluginOption(ctx)
+	clientSideOTelWithCSM = dialOptionWithCSMPluginOption(options, csmPluginOption)
+	clientSideOTel = opentelemetry.DialOption(options)
+
+	serverSideOTelWithCSM := serverOptionWithCSMPluginOption(options, csmPluginOption)
+
+	internal.AddGlobalPerTargetDialOptions.(func(opt any))(perTargetDialOption{})
+
+	internal.AddGlobalServerOptions.(func(opt ...grpc.ServerOption))(serverSideOTelWithCSM)
```
dfawley, line 62: Nit about organization:

```go
// do client-side initialization

// do server-side initialization
```

Is nicer than inter-splicing them with lots of extra blank lines everywhere.

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return func() {
+		internal.ClearGlobalServerOptions()
+		internal.ClearGlobalPerTargetDialOptions()
+	}
+} // fail, can run it locally and on cloudtop, replace to point to this stats/opentelemetry (replace grpc or something else)?
```
dfawley, line 68: ?

Delete all these notes please

dfawley (commented):
Looks pretty good - mostly just nits.

stats/opentelemetry/csm/observability.go
zasweq: Doesn't need to clean up resources? It will leave globals around but I guess this is fine if your binary is exiting? I returned this to be like gcp/observability.

stats/opentelemetry/csm/observability.go
zasweq: Based this off gcp/observability. Will switch to Enable.

stats/opentelemetry/csm/observability.go
zasweq: Good point. Deleted implemetationy part of comment.

stats/opentelemetry/csm/observability.go
zasweq: Oh great point. Switched.

stats/opentelemetry/csm/observability.go
zasweq: Whoops, deleted.

stats/opentelemetry/csm/observability.go
dfawley: Seems like we're keeping this since we might be doing other kinds of exporting in the future that will need to be flushed for correctness.

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"net/url"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/internal"
+	"google.golang.org/grpc/stats/opentelemetry"
+	otelinternal "google.golang.org/grpc/stats/opentelemetry/internal"
+)
+
+// EnableObservability sets up CSM EnableObservability for the binary globally.
```
dfawley, line 31: CSM EnableObservability? What's that?

stats/opentelemetry/csm/observability.go
```diff
@@ -0,0 +1,80 @@
...
+//
+// The CSM Plugin Option is instantiated with local labels and metadata exchange
```
dfawley, line 33: The CSM Stats Plugin?  Metrics exporter?  ???  Let's not put "plugin option" in our docstrings, though, I think.

dfawley (approved):

stats/opentelemetry/csm/observability.go
zasweq: Lol it switched this docstring.

stats/opentelemetry/csm/observability.go
zasweq: Choose "Stats Plugin".

PR #7279, opened 2024-05-29
stats/opentelemetry: Add e2e testing for CSM Observability

zasweq:
This PR adds the e2e testing for CSM Observability. It refactors common logic shared with OpenTelemetry e2e tests into internal/testingutils.

Alongside this, this PR deletes the Target Attribute filter and adds a unit test for the method attribute filter. The target attribute filter used to be part of the OTel API, but was deleted as it was realized it was too much complexity for a use case we don't even know if users want: https://github.com/grpc/proposal/pull/431. The method attribute filter I simply forgot to unit test.

RELEASE NOTES:
* stats/opentelemetry/csm: Add CSM Observability.

### DIFF
--- a/stats/opentelemetry/csm/observability_test.go
+++ b/stats/opentelemetry/csm/observability_test.go
@@ -37,7 +37,7 @@ import (
 	testpb "google.golang.org/grpc/interop/grpc_testing"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/stats/opentelemetry"
-	"google.golang.org/grpc/stats/opentelemetry/internal/testingutils"
+	"google.golang.org/grpc/stats/opentelemetry/internal/testutils"
 )
 
 // setupEnv configures the environment for CSM Observability Testing. It returns
@@ -126,7 +126,7 @@ func (s) TestCSMPluginOption(t *testing.T) {
 		// the interceptor level that can plumb metadata exchange header in.
 		unaryCallFunc     func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error)
 		streamingCallFunc func(stream testgrpc.TestService_FullDuplexCallServer) error
-		opts              testingutils.MetricDataOptions
+		opts              testutils.MetricDataOptions
 	}{
 		// Different permutations of operations that should all trigger csm md
 		// exchange labels to be written on the wire.
@@ -145,7 +145,7 @@ func (s) TestCSMPluginOption(t *testing.T) {
 					}
 				}
 			},
-			opts: testingutils.MetricDataOptions{
+			opts: testutils.MetricDataOptions{
 				CSMLabels:            csmLabels,
 				UnaryMessageSent:     true,
 				StreamingMessageSent: false,
@@ -164,7 +164,7 @@ func (s) TestCSMPluginOption(t *testing.T) {
 					}
 				}
 			},
-			opts: testingutils.MetricDataOptions{
+			opts: testutils.MetricDataOptions{
 				CSMLabels:            csmLabels,
 				UnaryMessageSent:     false,
 				StreamingMessageSent: false,
@@ -189,7 +189,7 @@ func (s) TestCSMPluginOption(t *testing.T) {
 					}
 				}
 			},
-			opts: testingutils.MetricDataOptions{
+			opts: testutils.MetricDataOptions{
 				CSMLabels:            csmLabels,
 				UnaryMessageSent:     true,
 				StreamingMessageSent: false,
@@ -213,7 +213,7 @@ func (s) TestCSMPluginOption(t *testing.T) {
 					}
 				}
 			},
-			opts: testingutils.MetricDataOptions{
+			opts: testutils.MetricDataOptions{
 				CSMLabels:            csmLabels,
 				UnaryMessageSent:     true,
 				StreamingMessageSent: false,
@@ -237,7 +237,7 @@ func (s) TestCSMPluginOption(t *testing.T) {
 					}
 				}
 			},
-			opts: testingutils.MetricDataOptions{
+			opts: testutils.MetricDataOptions{
 				CSMLabels:            csmLabels,
 				UnaryMessageSent:     true,
 				StreamingMessageSent: true,
@@ -294,8 +294,8 @@ func (s) TestCSMPluginOption(t *testing.T) {
 
 			opts := test.opts
 			opts.Target = ss.Target
-			wantMetrics := testingutils.MetricData(opts)
-			testingutils.CompareGotWantMetrics(ctx, t, mr, gotMetrics, wantMetrics)
+			wantMetrics := testutils.MetricData(opts)
+			testutils.CompareGotWantMetrics(ctx, t, mr, gotMetrics, wantMetrics)
 		})
 	}
 }
@@ -435,7 +435,7 @@ func (s) TestxDSLabels(t *testing.T) {
 					{
 						Attributes: attribute.NewSet(unaryMethodClientSideEnd...),
 						Count:      1,
-						Bounds:     testingutils.DefaultLatencyBounds,
+						Bounds:     testutils.DefaultLatencyBounds,
 					},
 				},
 				Temporality: metricdata.CumulativeTemporality,
@@ -450,7 +450,7 @@ func (s) TestxDSLabels(t *testing.T) {
 					{
 						Attributes:   attribute.NewSet(unaryMethodClientSideEnd...),
 						Count:        1,
-						Bounds:       testingutils.DefaultSizeBounds,
+						Bounds:       testutils.DefaultSizeBounds,
 						BucketCounts: unaryBucketCounts,
 						Min:          unaryExtrema,
 						Max:          unaryExtrema,
@@ -469,7 +469,7 @@ func (s) TestxDSLabels(t *testing.T) {
 					{
 						Attributes:   attribute.NewSet(unaryMethodClientSideEnd...),
 						Count:        1,
-						Bounds:       testingutils.DefaultSizeBounds,
+						Bounds:       testutils.DefaultSizeBounds,
 						BucketCounts: unaryBucketCounts,
 						Min:          unaryExtrema,
 						Max:          unaryExtrema,
@@ -488,7 +488,7 @@ func (s) TestxDSLabels(t *testing.T) {
 					{
 						Attributes: attribute.NewSet(unaryMethodAttr, targetAttr, unaryStatusAttr),
 						Count:      1,
-						Bounds:     testingutils.DefaultLatencyBounds,
+						Bounds:     testutils.DefaultLatencyBounds,
 					},
 				},
 				Temporality: metricdata.CumulativeTemporality,
@@ -496,7 +496,7 @@ func (s) TestxDSLabels(t *testing.T) {
 		},
 	}
 
-	testingutils.CompareGotWantMetrics(ctx, t, mr, gotMetrics, wantMetrics)
+	testutils.CompareGotWantMetrics(ctx, t, mr, gotMetrics, wantMetrics)
 }
 
 // TestObservability tests that Observability global function compiles and runs
--- a/stats/opentelemetry/csm/observability_test.go
+++ b/stats/opentelemetry/csm/observability_test.go
@@ -1,0 +1,508 @@
+/*
+ *
+ * Copyright 2024 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package csm
+
+import (
+	"context"
+	"errors"
+	"io"
+	"os"
+	"testing"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/encoding/gzip"
+	istats "google.golang.org/grpc/internal/stats"
+	"google.golang.org/grpc/internal/stubserver"
+	"google.golang.org/grpc/internal/testutils/xds/bootstrap"
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+	testpb "google.golang.org/grpc/interop/grpc_testing"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/stats/opentelemetry"
+	"google.golang.org/grpc/stats/opentelemetry/internal/testingutils"
+)
+
+// setupEnv configures the environment for CSM Observability Testing. It returns
+// a cleanup function to be invoked to clear the environment.
+func setupEnv(t *testing.T, resourceDetectorEmissions map[string]string, nodeID string, csmCanonicalServiceName string, csmWorkloadName string) func() {
+	clearEnv()
+
+	cleanup, err := bootstrap.CreateFile(bootstrap.Options{
+		NodeID:    nodeID,
+		ServerURI: "xds_server_uri",
+	})
+	if err != nil {
+		t.Fatalf("failed to create bootstrap: %v", err)
+	}
+	os.Setenv("CSM_CANONICAL_SERVICE_NAME", csmCanonicalServiceName)
+	os.Setenv("CSM_WORKLOAD_NAME", csmWorkloadName)
+
+	var attributes []attribute.KeyValue
+	for k, v := range resourceDetectorEmissions {
+		attributes = append(attributes, attribute.String(k, v))
+	}
+	// Return the attributes configured as part of the test in place
+	// of reading from resource.
+	attrSet := attribute.NewSet(attributes...)
+	origGetAttrSet := getAttrSetFromResourceDetector
+	getAttrSetFromResourceDetector = func(context.Context) *attribute.Set {
+		return &attrSet
+	}
+
+	return func() {
+		cleanup()
+		os.Unsetenv("CSM_CANONICAL_SERVICE_NAME")
+		os.Unsetenv("CSM_WORKLOAD_NAME")
+		getAttrSetFromResourceDetector = origGetAttrSet
+	}
+}
+
+// TestCSMPluginOption tests the CSM Plugin Option and labels. It configures the
+// environment for the CSM Plugin Option to read from. It then configures a
+// system with a gRPC Client and gRPC server with the OpenTelemetry Dial and
+// Server Option configured with a CSM Plugin Option with certain unary and
+// streaming handlers set to induce different ways of setting metadata exchange
+// labels, and makes a Unary RPC and a Streaming RPC. These two RPCs should
+// cause certain recording for each registered metric observed through a Manual
+// Metrics Reader on the provided OpenTelemetry SDK's Meter Provider. The CSM
+// Labels emitted from the plugin option should be attached to the relevant
+// metrics.
+func (s) TestCSMPluginOption(t *testing.T) {
+	resourceDetectorEmissions := map[string]string{
+		"cloud.platform":     "gcp_kubernetes_engine",
+		"cloud.region":       "cloud_region_val", // availability_zone isn't present, so this should become location
+		"cloud.account.id":   "cloud_account_id_val",
+		"k8s.namespace.name": "k8s_namespace_name_val",
+		"k8s.cluster.name":   "k8s_cluster_name_val",
+	}
+	nodeID := "projects/12345/networks/mesh:mesh_id/nodes/aaaa-aaaa-aaaa-aaaa"
+	csmCanonicalServiceName := "csm_canonical_service_name"
+	csmWorkloadName := "csm_workload_name"
+	cleanup := setupEnv(t, resourceDetectorEmissions, nodeID, csmCanonicalServiceName, csmWorkloadName)
+	defer cleanup()
+
+	attributesWant := map[string]string{
+		"csm.workload_canonical_service": csmCanonicalServiceName, // from env
+		"csm.mesh_id":                    "mesh_id",               // from bootstrap env var
+
+		// No xDS Labels - this happens in a test below.
+
+		"csm.remote_workload_type":              "gcp_kubernetes_engine",
+		"csm.remote_workload_canonical_service": csmCanonicalServiceName,
+		"csm.remote_workload_project_id":        "cloud_account_id_val",
+		"csm.remote_workload_cluster_name":      "k8s_cluster_name_val",
+		"csm.remote_workload_namespace_name":    "k8s_namespace_name_val",
+		"csm.remote_workload_location":          "cloud_region_val",
+		"csm.remote_workload_name":              csmWorkloadName,
+	}
+
+	var csmLabels []attribute.KeyValue
+	for k, v := range attributesWant {
+		csmLabels = append(csmLabels, attribute.String(k, v))
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	tests := []struct {
+		name string
+		// To test the different operations for Unary and Streaming RPC's from
+		// the interceptor level that can plumb metadata exchange header in.
+		unaryCallFunc     func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error)
+		streamingCallFunc func(stream testgrpc.TestService_FullDuplexCallServer) error
+		opts              testingutils.MetricDataOptions
+	}{
+		// Different permutations of operations that should all trigger csm md
+		// exchange labels to be written on the wire.
+		{
+			name: "normal-flow",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			streamingCallFunc: func(stream testgrpc.TestService_FullDuplexCallServer) error { // this is trailers only - no messages or headers sent
+				for {
+					_, err := stream.Recv()
+					if err == io.EOF {
+						return nil
+					}
+				}
+			},
+			opts: testingutils.MetricDataOptions{
+				CSMLabels:            csmLabels,
+				UnaryMessageSent:     true,
+				StreamingMessageSent: false,
+			},
+		},
+		{
+			name: "trailers-only-unary-streaming",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				return nil, errors.New("some error") // return an error and no message - this triggers trailers only - no messages or headers sent
+			},
+			streamingCallFunc: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+				for {
+					_, err := stream.Recv()
+					if err == io.EOF {
+						return nil
+					}
+				}
+			},
+			opts: testingutils.MetricDataOptions{
+				CSMLabels:            csmLabels,
+				UnaryMessageSent:     false,
+				StreamingMessageSent: false,
+				UnaryCallFailed:      true,
+			},
+		},
+		{
+			name: "set-header-client-server-side",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				grpc.SetHeader(ctx, metadata.New(map[string]string{"some-metadata": "some-metadata-val"}))
+
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			streamingCallFunc: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+				stream.SetHeader(metadata.New(map[string]string{"some-metadata": "some-metadata-val"}))
+				for {
+					_, err := stream.Recv()
+					if err == io.EOF {
+						return nil
+					}
+				}
+			},
+			opts: testingutils.MetricDataOptions{
+				CSMLabels:            csmLabels,
+				UnaryMessageSent:     true,
+				StreamingMessageSent: false,
+			},
+		},
+		{
+			name: "send-header-client-server-side",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				grpc.SendHeader(ctx, metadata.New(map[string]string{"some-metadata": "some-metadata-val"}))
+
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			streamingCallFunc: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+				stream.SendHeader(metadata.New(map[string]string{"some-metadata": "some-metadata-val"}))
+				for {
+					_, err := stream.Recv()
+					if err == io.EOF {
+						return nil
+					}
+				}
+			},
+			opts: testingutils.MetricDataOptions{
+				CSMLabels:            csmLabels,
+				UnaryMessageSent:     true,
+				StreamingMessageSent: false,
+			},
+		},
+		{
+			name: "send-msg-client-server-side",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			streamingCallFunc: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+				stream.Send(&testpb.StreamingOutputCallResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}})
+				for {
+					_, err := stream.Recv()
+					if err == io.EOF {
+						return nil
+					}
+				}
+			},
+			opts: testingutils.MetricDataOptions{
+				CSMLabels:            csmLabels,
+				UnaryMessageSent:     true,
+				StreamingMessageSent: true,
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			mr, ss := setup(ctx, t, test.unaryCallFunc, test.streamingCallFunc, true, nil)
+			defer ss.Stop()
+
+			var request *testpb.SimpleRequest
+			if test.opts.UnaryMessageSent {
+				request = &testpb.SimpleRequest{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}
+			}
+
+			// Make two RPC's, a unary RPC and a streaming RPC. These should cause
+			// certain metrics to be emitted, which should be able to be observed
+			// through the Metric Reader.
+			ss.Client.UnaryCall(ctx, request, grpc.UseCompressor(gzip.Name))
+			stream, err := ss.Client.FullDuplexCall(ctx, grpc.UseCompressor(gzip.Name))
+			if err != nil {
+				t.Fatalf("ss.Client.FullDuplexCall failed: %f", err)
+			}
+
+			if test.opts.StreamingMessageSent {
+				if err := stream.Send(&testpb.StreamingOutputCallRequest{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}); err != nil {
+					t.Fatalf("stream.Send failed")
+				}
+				if _, err := stream.Recv(); err != nil {
+					t.Fatalf("stream.Recv failed with error: %v", err)
+				}
+			}
+
+			stream.CloseSend()
+			if _, err = stream.Recv(); err != io.EOF {
+				t.Fatalf("unexpected error: %v, expected an EOF error", err)
+			}
+
+			rm := &metricdata.ResourceMetrics{}
+			mr.Collect(ctx, rm)
+
+			gotMetrics := map[string]metricdata.Metrics{}
+			for _, sm := range rm.ScopeMetrics {
+				for _, m := range sm.Metrics {
+					gotMetrics[m.Name] = m
+				}
+			}
+
+			opts := test.opts
+			opts.Target = ss.Target
+			wantMetrics := testingutils.MetricData(opts)
+			testingutils.CompareGotWantMetrics(ctx, t, mr, gotMetrics, wantMetrics)
+		})
+	}
+}
+
+// setup creates a stub server with the provided unary call and full duplex call
+// handlers, alongside with OpenTelemetry component with a CSM Plugin Option
+// configured on client and server side based off bool. It also takes in a unary
+// interceptor to configure. It returns a reader for metrics emitted from the
+// OpenTelemetry component and the stub server.
+func setup(ctx context.Context, t *testing.T, unaryCallFunc func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error), streamingCallFunc func(stream testgrpc.TestService_FullDuplexCallServer) error, serverOTelConfigured bool, clientUnaryInterceptor grpc.UnaryClientInterceptor) (*metric.ManualReader, *stubserver.StubServer) { // specific for plugin option
+	reader := metric.NewManualReader()
+	provider := metric.NewMeterProvider(
+		metric.WithReader(reader),
+	)
+	ss := &stubserver.StubServer{
+		UnaryCallF:      unaryCallFunc,
+		FullDuplexCallF: streamingCallFunc,
+	}
+
+	po := newPluginOption(ctx)
+	var sopts []grpc.ServerOption
+	if serverOTelConfigured {
+		sopts = append(sopts, serverOptionWithCSMPluginOption(opentelemetry.Options{
+			MetricsOptions: opentelemetry.MetricsOptions{
+				MeterProvider: provider,
+				Metrics:       opentelemetry.DefaultMetrics,
+			}}, po))
+	}
+	dopts := []grpc.DialOption{dialOptionWithCSMPluginOption(opentelemetry.Options{
+		MetricsOptions: opentelemetry.MetricsOptions{
+			MeterProvider:  provider,
+			Metrics:        opentelemetry.DefaultMetrics,
+			OptionalLabels: []string{"csm.service_name", "csm.service_namespace_name"}, // should be a no-op unless receive labels through optional labels mechanism
+		},
+	}, po)}
+	if clientUnaryInterceptor != nil {
+		dopts = append(dopts, grpc.WithUnaryInterceptor(clientUnaryInterceptor))
+	}
+	if err := ss.Start(sopts, dopts...); err != nil {
+		t.Fatalf("Error starting endpoint server: %v", err)
+	}
+	return reader, ss
+}
+
+func unaryInterceptorAttachxDSLabels(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	ctx = istats.SetLabels(ctx, &istats.Labels{
+		TelemetryLabels: map[string]string{
+			// mock what the cluster impl would write here ("csm." xDS Labels)
+			"csm.service_name":      "service_name_val",
+			"csm.service_namespace_name": "service_namespace_val",
+		},
+	})
+
+	// TagRPC will just see this in the context and set it's xDS Labels to point
+	// to this map on the heap.
+	return invoker(ctx, method, req, reply, cc, opts...)
+}
+
+// TestxDSLabels tests that xDS Labels get emitted from OpenTelemetry metrics.
+// This test configures OpenTelemetry with the CSM Plugin Option, and xDS
+// Optional Labels turned on. It then configures an interceptor to attach
+// labels, representing the cluster_impl picker. It then makes a unary RPC, and
+// expects xDS Labels labels to be attached to emitted relevant metrics. Full
+// xDS System alongside OpenTelemetry will be tested with interop. (there is
+// a test for xDS -> Stats handler and this tests -> OTel -> emission).
+func (s) TestxDSLabels(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	mr, ss := setup(ctx, t, func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+		return &testpb.SimpleResponse{Payload: &testpb.Payload{
+			Body: make([]byte, 10000),
+		}}, nil
+	}, nil, false, unaryInterceptorAttachxDSLabels)
+	defer ss.Stop()
+	ss.Client.UnaryCall(ctx, &testpb.SimpleRequest{Payload: &testpb.Payload{
+		Body: make([]byte, 10000),
+	}}, grpc.UseCompressor(gzip.Name))
+
+	rm := &metricdata.ResourceMetrics{}
+	mr.Collect(ctx, rm)
+
+	gotMetrics := map[string]metricdata.Metrics{}
+	for _, sm := range rm.ScopeMetrics {
+		for _, m := range sm.Metrics {
+			gotMetrics[m.Name] = m
+		}
+	}
+
+	unaryMethodAttr := attribute.String("grpc.method", "grpc.testing.TestService/UnaryCall")
+	targetAttr := attribute.String("grpc.target", ss.Target)
+	unaryStatusAttr := attribute.String("grpc.status", "OK")
+
+	serviceNameAttr := attribute.String("csm.service_name", "service_name_val")
+	serviceNamespaceAttr := attribute.String("csm.service_namespace_name", "service_namespace_val")
+	meshIDAttr := attribute.String("csm.mesh_id", "unknown")
+	workloadCanonicalServiceAttr := attribute.String("csm.workload_canonical_service", "unknown")
+	remoteWorkloadTypeAttr := attribute.String("csm.remote_workload_type", "unknown")
+	remoteWorkloadCanonicalServiceAttr := attribute.String("csm.remote_workload_canonical_service", "unknown")
+
+	unaryMethodClientSideEnd := []attribute.KeyValue{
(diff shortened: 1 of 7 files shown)


stats/opentelemetry/internal/testingutils/testingutils.go
easwars, line 1: Nit: In all other places in our code base we use `testutils`. Any specific reason for this being `testingutils`?

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest"
+)
+
+var (
+	// DefaultLatencyBounds are the default bounds for latency metrics.
+	DefaultLatencyBounds = []float64{0, 0.00001, 0.00005, 0.0001, 0.0003, 0.0006, 0.0008, 0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.008, 0.01, 0.013, 0.016, 0.02, 0.025, 0.03, 0.04, 0.05, 0.065, 0.08, 0.1, 0.13, 0.16, 0.2, 0.25, 0.3, 0.4, 0.5, 0.65, 0.8, 1, 2, 5, 10, 20, 50, 100} // provide "advice" through API, SDK should set this too
```
easwars, line 34: I don't follow what the trailing comment says. Can it be improved? Can it be deleted?

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest"
+)
+
+var (
+	// DefaultLatencyBounds are the default bounds for latency metrics.
```
easwars, line 33: How were these arrived at? Does the reader of the test need to know what these mean?

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+	DefaultLatencyBounds = []float64{0, 0.00001, 0.00005, 0.0001, 0.0003, 0.0006, 0.0008, 0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.008, 0.01, 0.013, 0.016, 0.02, 0.025, 0.03, 0.04, 0.05, 0.065, 0.08, 0.1, 0.13, 0.16, 0.2, 0.25, 0.3, 0.4, 0.5, 0.65, 0.8, 1, 2, 5, 10, 20, 50, 100} // provide "advice" through API, SDK should set this too
+	// DefaultSizeBounds are the default bounds for metrics which record size.
+	DefaultSizeBounds = []float64{0, 1024, 2048, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864, 268435456, 1073741824, 4294967296}
+)
+
+// waitForServerCompletedRPCs waits until the unary and streaming stats.End
+// calls are finished processing. It does this by waiting for the expected
+// metric triggered by stats.End to appear through the passed in metrics reader.
+func waitForServerCompletedRPCs(ctx context.Context, reader metric.Reader, wantMetric metricdata.Metrics, t *testing.T) (map[string]metricdata.Metrics, error) {
```
easwars, line 42: If there is a `context.Context` parameter. It always has to be the first.
If there is a `testing.T` parameter, it always has to be the first, unless there is a `context.Context` parameter as well. So, here the `testing.T` needs to be the second parameter and not the last one.

I couldn't find the specific guideline in the style guide. But I clearly remember this. So, take my word for it.

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+		for _, sm := range rm.ScopeMetrics {
+			for _, m := range sm.Metrics {
+				gotMetrics[m.Name] = m
+			}
+		}
+		val, ok := gotMetrics[wantMetric.Name]
+		if !ok {
+			continue
+		}
+		if !metricdatatest.AssertEqual(t, wantMetric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars()) {
```
easwars, line 56: Hmm .. clear violation of the style guide in the otel library API. go/go-style/decisions#assert. Not much we can do here though.

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+var (
+	// DefaultLatencyBounds are the default bounds for latency metrics.
+	DefaultLatencyBounds = []float64{0, 0.00001, 0.00005, 0.0001, 0.0003, 0.0006, 0.0008, 0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.008, 0.01, 0.013, 0.016, 0.02, 0.025, 0.03, 0.04, 0.05, 0.065, 0.08, 0.1, 0.13, 0.16, 0.2, 0.25, 0.3, 0.4, 0.5, 0.65, 0.8, 1, 2, 5, 10, 20, 50, 100} // provide "advice" through API, SDK should set this too
+	// DefaultSizeBounds are the default bounds for metrics which record size.
+	DefaultSizeBounds = []float64{0, 1024, 2048, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864, 268435456, 1073741824, 4294967296}
+)
+
+// waitForServerCompletedRPCs waits until the unary and streaming stats.End
+// calls are finished processing. It does this by waiting for the expected
+// metric triggered by stats.End to appear through the passed in metrics reader.
```
easwars, line 41: What does the return value contain and how do callers use it?

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+			continue
+		}
+		return gotMetrics, nil
+	}
+	return nil, fmt.Errorf("error waiting for metric %v: %v", wantMetric, ctx.Err())
+}
+
+// assertDataPointWithinFiveSeconds asserts the metric passed in contains
+// a histogram with dataPoints that fall within buckets that are <=5.
+func assertDataPointWithinFiveSeconds(metric metricdata.Metrics) error {
```
easwars, line 66: Maybe drop the `assert` prefix and replace it with something since this is not `assert`ing in the true sense, but is returning an error.

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+		}
+		if !foundPoint {
+			return fmt.Errorf("no data point found for metric")
+		}
+	}
+	return nil
+}
+
+// MetricDataOptions are the options used to configure the metricData emissions
+// of expected metrics data from NewMetricData. (rename function? this feels
```
easwars, line 96: Handle the second half of this comment?

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+			},
+		},
+	}
+}
+
+// CompareGotWantMetrics asserts wantMetrics are what we expect. It polls for
+// eventual server metrics (not emitted synchronously with client side rpc
+// returning), and for duration metrics makes sure the data point is within
+// possible testing time (five seconds from context timeout).
+func CompareGotWantMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader, gotMetrics map[string]metricdata.Metrics, wantMetrics []metricdata.Metrics) { // return an error instead of t...
```
easwars, line 396: Do you want to handle the trailing comment here?

stats/opentelemetry/internal/testingutils/testingutils.go
```diff
@@ -0,0 +1,430 @@
...
+	for _, metric := range wantMetrics {
+		if metric.Name == "grpc.server.call.sent_total_compressed_message_size" || metric.Name == "grpc.server.call.rcvd_total_compressed_message_size" {
+			// Sync the metric reader to see the event because stats.End is
+			// handled async server side. Thus, poll until metrics created from
+			// stats.End show up.
+			var err error
+			if gotMetrics, err = waitForServerCompletedRPCs(ctx, mr, metric, t); err != nil { // move to shared helper
```
easwars, line 403: Wouldn't this overwrite the argument `gotMetrics`? Is that what you want?

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Haha I believe you. Switched.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: No reason. testutils sgtm. Switched.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Ah this is to guide users of the OTel component. Doesn't make sense in tests. Deleted.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: These are the default bounds scoped in https://github.com/grpc/proposal/blob/master/A66-otel-stats.md#metrics-schema. These are needed as the metrics are expected to be emitted with these bounds.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Oh whoops this was left behind when I was trying to figure out if I should take a testing.T and fail it, or return an error an fail it in caller. I chose the former.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Yes, to poll for the new metrics. This is a sync point, so it makes sure it gets the async recorded metrics before continuing.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Oh sorry forgot to delete.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Switched to check. Switched comment.

stats/opentelemetry/internal/testingutils/testingutils.go
zasweq: Added "Returns a new gotMetrics map containing the metric data being polled for, or
an error if failed to wait for metric."

zasweq:
Thanks for the pass! Got to all comments.

easwars:
> Alongside this, this PR deletes the Target Attribute filter and adds a unit test for the method attribute filter.

Why are we doing this?

stats/opentelemetry/e2e_test.go
```diff
@@ -94,23 +69,16 @@ func setup(t *testing.T, tafOn bool, maf func(string) bool) (*metric.ManualReade
...
 			MeterProvider:         provider,
 			Metrics:               DefaultMetrics,
-			TargetAttributeFilter: taf,
 			MethodAttributeFilter: maf,
 		}})}, DialOption(Options{
 		MetricsOptions: MetricsOptions{
 			MeterProvider:         provider,
 			Metrics:               DefaultMetrics,
-			TargetAttributeFilter: taf,
 			MethodAttributeFilter: maf,
```
easwars, line 82: Does the `MethodAttributeFilter` make sense on the client side? Why is it included in the dial option?

stats/opentelemetry/e2e_test.go
```diff
@@ -46,35 +46,10 @@ func Test(t *testing.T) {
...
-		return gotMetrics, nil
-	}
-	return nil, fmt.Errorf("error waiting for metric %v: %v", wantMetric, ctx.Err())
-}
-
 // setup creates a stub server with OpenTelemetry component configured on client
 // and server side. It returns a reader for metrics emitted from OpenTelemetry
 // component and the server.
-func setup(t *testing.T, tafOn bool, maf func(string) bool) (*metric.ManualReader, *stubserver.StubServer) {
+func setup(t *testing.T, maf func(string) bool) (*metric.ManualReader, *stubserver.StubServer) {
```
easwars, line 52: There is a lot of setup in this function, but it is not very readable.

- what is `maf`? For someone new to the code base, it is not obvious that it the method attribute filter func
- Is it possible to have the stub server initialized within the test, so that the RPC handler functionality is close to the test logic. See: go/go-style/best-practices#t-common-setup-scope
- initialization of the `reader` and the `provider` and one liners, but are done here. And this means that when one reads the test logic, it is not clear what the `reader` returned from this function is, and they will have to navigate to here.

stats/opentelemetry/e2e_test.go
```diff
@@ -127,12 +95,11 @@ func (s) TestMethodTargetAttributeFilter(t *testing.T) {
...
 		return str != "/grpc.testing.TestService/UnaryCall"
 	}
-	// pull out setup into a helper
-	reader, ss := setup(t, true, maf)
+	reader, ss := setup(t, maf)
 	defer ss.Stop()
 
-	// make a single RPC (unary rpc), and filter out the target and method
-	// that would correspond.
+	// make a single RPC (unary rpc), and filter out the method that would
```
easwars, line 101: One unary and one streaming RPC is being made here.

zasweq:
The target attribute filter used to part of the OTel API, but was deleted as it was realized it was too much complexity: https://github.com/grpc/proposal/pull/431. The method attribute filter I simply forgot to unit test.

stats/opentelemetry/e2e_test.go
```diff
@@ -18,7 +18,6 @@ package opentelemetry
```
easwars, line 18: The name of the file is `e2e_test.go`, but is defined in the same package `opentelemetry`. So, there is no way to guarantee that this is actually an E2E test, because it can use unexported symbols from the package that it is testing.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,433 @@
...
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest"
+)
+
+var (
+	// DefaultLatencyBounds are the default bounds for latency metrics.
+	DefaultLatencyBounds = []float64{0, 0.00001, 0.00005, 0.0001, 0.0003, 0.0006, 0.0008, 0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.008, 0.01, 0.013, 0.016, 0.02, 0.025, 0.03, 0.04, 0.05, 0.065, 0.08, 0.1, 0.13, 0.16, 0.2, 0.25, 0.3, 0.4, 0.5, 0.65, 0.8, 1, 2, 5, 10, 20, 50, 100}
+	// DefaultSizeBounds are the default bounds for metrics which record size.
+	DefaultSizeBounds = []float64{0, 1024, 2048, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864, 268435456, 1073741824, 4294967296}
```
easwars, line 36: Why do these have to be defined here, instead of being able to refer to the one defined in the `opentelemetry` package. If the reason is a cyclic dependency, can we instead use the `internal` package trick and not have to redefine it here.

If using the internal package trick makes the Go doc for the opentelemetry package less readable, this is fine. In that case, a comment here as to why this is being redefined here would be appropriate.

stats/opentelemetry/e2e_test.go
```diff
@@ -127,12 +95,11 @@ func (s) TestMethodTargetAttributeFilter(t *testing.T) {
 		// Will allow duplex/any other type of RPC.
 		return str != "/grpc.testing.TestService/UnaryCall"
```
easwars, line 96: I believe the generated code defines a const for the method names. Could we use that instead?

stats/opentelemetry/e2e_test.go
```diff
@@ -127,12 +95,11 @@ func (s) TestMethodTargetAttributeFilter(t *testing.T) {
 		// Will allow duplex/any other type of RPC.
```
easwars, line 95: This test is only testing the method attribute filter now. The test name and comment needs updating.

stats/opentelemetry/e2e_test.go
```diff
@@ -151,6 +118,12 @@ func (s) TestMethodTargetAttributeFilter(t *testing.T) {
...
 	rm := &metricdata.ResourceMetrics{}
 	reader.Collect(ctx, rm)
+	gotMetrics := map[string]metricdata.Metrics{}
+	for _, sm := range rm.ScopeMetrics {
+		for _, m := range sm.Metrics {
+			gotMetrics[m.Name] = m
+		}
+	}
 
 	wantMetrics := []metricdata.Metrics{
```
easwars, line 128: Is there a reason to not use `testutils.MetricData` here?

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,433 @@
...
+			},
+		},
+	}
+}
+
+// CompareGotWantMetrics asserts wantMetrics are what we expect. It polls for
+// eventual server metrics (not emitted synchronously with client side rpc
+// returning), and for duration metrics makes sure the data point is within
+// possible testing time (five seconds from context timeout).
+func CompareGotWantMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader, gotMetrics map[string]metricdata.Metrics, wantMetrics []metricdata.Metrics) {
```
easwars, line 399: CompareMetrics is a nice enough name I feel.

easwars, line 399: Is it possible to make `gotMetrics` and `wantMetrics` be of the same type?

easwars:
> The target attribute filter used to part of the OTel API, but was deleted as it was realized it was too much complexity: [grpc/proposal#431](https://github.com/grpc/proposal/pull/431). The method attribute filter I simply forgot to unit test.

Please specify in the PR description.

stats/opentelemetry/e2e_test.go
zasweq: Ah, good point. See https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/opentelemetry.go#L135 this comment. This only applies to server side. Deleted from here.

stats/opentelemetry/e2e_test.go
zasweq: Ah, good point. Switched.

stats/opentelemetry/e2e_test.go
zasweq: Ah, good point. Switched.

stats/opentelemetry/e2e_test.go
zasweq: Added comment: "Make a Unary and Streaming RPC. The Unary RPC should be filtered by the method attribute filter, and the Full Duplex (Streaming) RPC should not."

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Switched.

stats/opentelemetry/internal/testutils/testutils.go
zasweq: https://pkg.go.dev/google.golang.org/grpc/stats/opentelemetry#pkg-variables yes the default bounds show up in the external go doc, and it will create a cyclic dependency if I reach into opentelemetry from here. Added comment.

stats/opentelemetry/e2e_test.go
zasweq: Yeah, the creation function specifies the full emitted metric atoms, and this one is a lighter weight check on one metric from client and one metric from server. Related to that, do you like my metrics data creation function? I added so many options to it (like our xDS Resources) and I had a ton of branches in the helper (and coupled some logic there with callsite like messages sent which need to be observed on peer). I don't really see a way around it, but let me know if you've seen a cleaner way to do this.

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Hmmm good point. In order to avoid o(n^2) iteration times, I think both should be a map. However, thinking about it, I actually ordered the sync point that recording the server call duration comes before the compressed message metrics, so I only added a compressed message metric sync point, as once that records server duration records: https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/server_metrics.go#L250 so I would need to add server duration sync point as well. Thinking about it, the main reason I didn't do a map though is the duplication of the metric.Name (it is part of the assertion as well) as it would be map[string]metricdata.Metric{"metric_name": &metricdata.Metric{ Name: "metric_name"...} so I would prefer not to unless you feel strongly.

zasweq:
Thank you for the detailed pass!

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+// environment for the CSM Plugin Option to read from. It then configures a
+// system with a gRPC Client and gRPC server with the OpenTelemetry Dial and
+// Server Option configured with a CSM Plugin Option with certain unary and
+// streaming handlers set to induce different ways of setting metadata exchange
+// labels, and makes a Unary RPC and a Streaming RPC. These two RPCs should
+// cause certain recording for each registered metric observed through a Manual
+// Metrics Reader on the provided OpenTelemetry SDK's Meter Provider. The CSM
+// Labels emitted from the plugin option should be attached to the relevant
+// metrics.
+func (s) TestCSMPluginOption(t *testing.T) {
```
easwars, line 88: What do you think about splitting this into two tests, one for unary and one for streaming. I see that because both of these are tested in the same test:
- the test logic is complicated
- because the cross-product of the two options are tested, each one is repeated a few times without much value
- the subtest names aren't very descriptive or readable

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+		})
+	}
+}
+
+// setup creates a stub server with the provided unary call and full duplex call
+// handlers, alongside with OpenTelemetry component with a CSM Plugin Option
+// configured on client and server side based off bool. It also takes in a unary
+// interceptor to configure. It returns a reader for metrics emitted from the
+// OpenTelemetry component and the stub server.
+func setup(ctx context.Context, t *testing.T, unaryCallFunc func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error), streamingCallFunc func(stream testgrpc.TestService_FullDuplexCallServer) error, serverOTelConfigured bool, clientUnaryInterceptor grpc.UnaryClientInterceptor) (*metric.ManualReader, *stubserver.StubServer) { // specific for plugin option
```
easwars, line 308: This function would be much simpler (or even better, completely eliminated and inlined in the test) if we split up the above test into two as I suggested in the previous comment.

Sharing code between tests is all good, but making these helpers complicated to accommodate different tests makes the tests very hard to read. See: https://engdoc.corp.google.com/eng/doc/tott/episodes/643.md?cl=head

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+	reader := metric.NewManualReader()
+	provider := metric.NewMeterProvider(
+		metric.WithReader(reader),
+	)
```
easwars, line 312: This could be a single line.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+	defer cancel()
+	mr, ss := setup(ctx, t, func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+		return &testpb.SimpleResponse{Payload: &testpb.Payload{
+			Body: make([]byte, 10000),
+		}}, nil
+	}, nil, false, unaryInterceptorAttachxDSLabels)
+	defer ss.Stop()
+	ss.Client.UnaryCall(ctx, &testpb.SimpleRequest{Payload: &testpb.Payload{
+		Body: make([]byte, 10000),
+	}}, grpc.UseCompressor(gzip.Name))
```
easwars, line 375: Do we support uncompressed bytes as a metric and do we have tests for that?

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+
+	testutils.CompareMetrics(ctx, t, mr, gotMetrics, wantMetrics)
+}
+
+// TestObservability tests that Observability global function compiles and runs
+// without error. The actual functionality of this function will be verified in
+// interop tests.
+func (s) TestObservability(t *testing.T) {
+	cleanup := EnableObservability(context.Background(), opentelemetry.Options{})
+	defer cleanup()
```
easwars, line 507: Nit: get rid of the defer and just invoke the `cleanup` function.

stats/opentelemetry/e2e_test.go
easwars: s/The method and target filter set should bucket/The method filter set should bucket/

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,436 @@
...
+		foundPoint := false
+		for i, bucket := range dataPoint.BucketCounts {
+			if i >= boundWithFive {
+				return fmt.Errorf("data point not found in bucket <=5 seconds")
+			}
+			if bucket == 1 {
+				foundPoint = true
+				break
+			}
+		}
```
easwars, line 94: This is very hard to read. Both `for` loops define a `bucket` variable, but clearly they mean different things. And what is special about `5`? Why is `5` being used?

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,436 @@
...
+}
+
+// MetricData returns a metricsDataSlice for A66 metrics for client and server
+// with a unary RPC and streaming RPC with certain compression and message flow
+// sent. If csmAttributes is set to true, the corresponding CSM Metrics (not
+// client side call metrics, or started on client and server side).
+func MetricData(options MetricDataOptions) []metricdata.Metrics {
+	unaryMethodAttr := attribute.String("grpc.method", "grpc.testing.TestService/UnaryCall")
+	duplexMethodAttr := attribute.String("grpc.method", "grpc.testing.TestService/FullDuplexCall")
+
```
easwars, line 133: Nit: nix these newlines, which don't seem to server much purpose.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,436 @@
...
+	unaryMethodClientSideEnd = append(unaryMethodClientSideEnd, options.CSMLabels...)
+	streamingMethodClientSideEnd = append(streamingMethodClientSideEnd, options.CSMLabels...)
+	unaryMethodServerSideEnd = append(unaryMethodServerSideEnd, options.CSMLabels...)
+	streamingMethodServerSideEnd = append(streamingMethodServerSideEnd, options.CSMLabels...)
+	unaryCompressedBytesSentRecv := int64(0)
+	unaryBucketCounts := []uint64{0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
+	unaryExtrema := metricdata.NewExtrema(int64(0))
+	if options.UnaryMessageSent {
+		unaryCompressedBytesSentRecv = 57 // Fixed 10000 bytes with gzip assumption.
```
easwars, line 170: This introduces very tight coupling between the test logic and this utility package. Can we do better?

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,436 @@
...
+		unaryMethodAttr,
+		unaryStatusAttr,
+	}
+
+	streamingMethodServerSideEnd := []attribute.KeyValue{
+		duplexMethodAttr,
+		streamingStatusAttr,
+	}
+
+	unaryMethodClientSideEnd = append(unaryMethodClientSideEnd, options.CSMLabels...)
```
easwars, line 162: Also, would it make sense to simplify this function by splitting this into two, one for unary and one for streaming. And for tests which want to test both, there should be a way for the test to call both the functions and somehow combine the emissions into one single `wantMetrics`.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,436 @@
...
+	streamingMethodClientSideEnd = append(streamingMethodClientSideEnd, options.CSMLabels...)
+	unaryMethodServerSideEnd = append(unaryMethodServerSideEnd, options.CSMLabels...)
+	streamingMethodServerSideEnd = append(streamingMethodServerSideEnd, options.CSMLabels...)
+	unaryCompressedBytesSentRecv := int64(0)
+	unaryBucketCounts := []uint64{0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
```
easwars, line 167: Would it make sense to make a helper function that creates those bucket counts. That way, the test can call that helper with a slice of message sizes, and the return value would be bucket counts corresponding to that.

stats/opentelemetry/e2e_test.go
easwars: Left some comments there :)
I feel like there is room for simplification. A good start though.

easwars (changes requested):

stats/opentelemetry/e2e_test.go
easwars: I'm not going to ask you to refactor this as part of this PR, but something to keep in mind: https://engdoc.corp.google.com/eng/doc/tott/episodes/643.md?cl=head

stats/opentelemetry/csm/observability_test.go
zasweq: We had some for OpenCensus but not for OpenTelemetry: https://github.com/grpc/proposal/blob/master/A66-otel-stats.md#client-per-attempt-instruments. Thus I just test the compressed bytes. However, for OpenCensus, I do tet both: https://github.com/grpc/grpc-go/blob/master/stats/opencensus/e2e_test.go#L682 (they are represented by a histogram, so by a different bucket count).

stats/opentelemetry/csm/observability_test.go
zasweq: Done.

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Yeah sure good point. These bucket counts (besides latency) are only for message sizes, but in the future we could add more histogram metrics, so my helper takes a slice of counts and also the bounds for extensibility in the future.

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Maybe I should make this a knob.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+	attrSet := attribute.NewSet(attributes...)
+	origGetAttrSet := getAttrSetFromResourceDetector
+	getAttrSetFromResourceDetector = func(context.Context) *attribute.Set {
+		return &attrSet
+	}
+
+	return func() {
+		cleanup()
+		os.Unsetenv("CSM_CANONICAL_SERVICE_NAME")
+		os.Unsetenv("CSM_WORKLOAD_NAME")
```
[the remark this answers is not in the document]
zasweq: Yeah I thought about this too, and decided it was fine since this is the behavior the rest of the codebase does. I'll switch it to old calue. I think the issue was the distinction between unset vs. set with a zero value.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+	"google.golang.org/grpc/internal/stubserver"
+	"google.golang.org/grpc/internal/testutils/xds/bootstrap"
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+	testpb "google.golang.org/grpc/interop/grpc_testing"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/stats/opentelemetry"
+	"google.golang.org/grpc/stats/opentelemetry/internal/testutils"
+)
+
+// setupEnv configures the environment for CSM Observability Testing. It returns
```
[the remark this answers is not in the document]
zasweq: Good point, done. Comment became "setupEnv configures the environment for CSM Observability Testing. It sets the bootstrap env var to a bootstrap file with a nodeID provided. It sets CSM Env Vars as well, and mocks the resource detector's returned attribute set to simulate the environment. It returns a cleanup function to be invoked to clear the environment."

stats/opentelemetry/internal/testutils/testutils.go
zasweq: I guess bucket semantically means a->b, so it doesn't apply to either of these for loops. Switched first bucket to bound, and second bucket to count.

stats/opentelemetry/e2e_test.go
zasweq: Argh whoops haha. Switched.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,508 @@
...
+	}
+	// Return the attributes configured as part of the test in place
+	// of reading from resource.
+	attrSet := attribute.NewSet(attributes...)
+	origGetAttrSet := getAttrSetFromResourceDetector
+	getAttrSetFromResourceDetector = func(context.Context) *attribute.Set {
+		return &attrSet
+	}
+
+	return func() {
```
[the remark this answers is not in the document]
zasweq: Done. Seems to be useful for parallezation since it guarantees ordering between threads: https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests.

stats/opentelemetry/csm/observability_test.go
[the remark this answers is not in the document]
zasweq: Switched to:
```
oldCSMCanonicalServiceName, csmCanonicalServiceNamePresent := os.LookupEnv("CSM_CANONICAL_SERVICE_NAME")
oldCSMWorkloadName, csmWorkloadNamePresent := os.LookupEnv("CSM_WORKLOAD_NAME")
	os.Setenv("CSM_CANONICAL_SERVICE_NAME", csmCanonicalServiceName)
os.Setenv("CSM_WORKLOAD_NAME", csmWorkloadName)
```

and cleanup:

```
if csmCanonicalServiceNamePresent {
			os.Setenv("CSM_CANONICAL_SERVICE_NAME", oldCSMCanonicalServiceName)
		} else {
			os.Unsetenv("CSM_CANONICAL_SERVICE_NAME")
		}
		if csmWorkloadNamePresent {
			os.Setenv("CSM_WORKLOAD_NAME", oldCSMWorkloadName)
		} else {
			os.Unsetenv("CSM_WORKLOAD_NAME")
		}
```
We should keep this in mind for future tests, since I saw a few places in our codebase that did this and I thought same thing but thought it was fine since elsewhere did it :).

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Done.

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Made a knob, and called a helper for the counts :). Great idea.

zasweq:
Thank you so much for the detailed pass, looks a lot cleaner now. Got to all comments.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,798 @@
...
+// time (five seconds from context timeout).
+func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader, gotMetrics map[string]metricdata.Metrics, wantMetrics []metricdata.Metrics) {
+	for _, metric := range wantMetrics {
+		if metric.Name == "grpc.server.call.sent_total_compressed_message_size" || metric.Name == "grpc.server.call.rcvd_total_compressed_message_size" {
+			// Sync the metric reader to see the event because stats.End is
+			// handled async server side. Thus, poll until metrics created from
+			// stats.End show up.
+			var err error
+			if gotMetrics, err = waitForServerCompletedRPCs(ctx, t, mr, metric); err != nil { // move to shared helper
+				t.Fatalf("error waiting for sent total compressed message size for metric %v: %v", metric.Name, err)
```
easwars, line 772: The error returned by `waitForServerCompletedRPCs` also starts with `error waiting for metric`.

Also, error strings in logging and testing failures need not follow the capitalization rules for error strings. So, here and else where in this PR, strings passed to `t.Error/t.Fatal(f)` should begin with upper case wherever appropriate.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+	"google.golang.org/grpc/stats/opentelemetry/internal/testutils"
+)
+
+// setupEnv configures the environment for CSM Observability Testing. It sets
+// the bootstrap env var to a bootstrap file with a nodeID provided. It sets CSM
+// Env Vars as well, and mocks the resource detector's returned attribute set to
+// simulate the environment. It registers a cleanup function on the provided t
+// to restore the environment to it's original state.
+func setupEnv(t *testing.T, resourceDetectorEmissions map[string]string, nodeID string, csmCanonicalServiceName string, csmWorkloadName string) {
+	clearEnv()
```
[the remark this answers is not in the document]
zasweq: Ah sorry I changed it below in the t.Cleanup but read the old ones after this. Yeah you're right I no longer need this. Deleted.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+	"google.golang.org/grpc/stats/opentelemetry"
+	"google.golang.org/grpc/stats/opentelemetry/internal/testutils"
+)
+
+// setupEnv configures the environment for CSM Observability Testing. It sets
+// the bootstrap env var to a bootstrap file with a nodeID provided. It sets CSM
+// Env Vars as well, and mocks the resource detector's returned attribute set to
+// simulate the environment. It registers a cleanup function on the provided t
+// to restore the environment to it's original state.
+func setupEnv(t *testing.T, resourceDetectorEmissions map[string]string, nodeID string, csmCanonicalServiceName string, csmWorkloadName string) {
```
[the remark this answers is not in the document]
zasweq: Done.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+	resourceDetectorEmissions := map[string]string{
+		"cloud.platform":     "gcp_kubernetes_engine",
+		"cloud.region":       "cloud_region_val", // availability_zone isn't present, so this should become location
+		"cloud.account.id":   "cloud_account_id_val",
+		"k8s.namespace.name": "k8s_namespace_name_val",
+		"k8s.cluster.name":   "k8s_cluster_name_val",
+	}
+	nodeID := "projects/12345/networks/mesh:mesh_id/nodes/aaaa-aaaa-aaaa-aaaa"
+	csmCanonicalServiceName := "csm_canonical_service_name"
+	csmWorkloadName := "csm_workload_name"
```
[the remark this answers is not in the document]
zasweq: Done.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+			}
+			defer ss.Stop()
+
+			var request *testpb.SimpleRequest
+			if test.opts.UnaryCompressedMessageSize != 0 {
+				request = &testpb.SimpleRequest{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}
+			}
+			// Make two RPC's, a unary RPC and a streaming RPC. These should cause
```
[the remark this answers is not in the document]
zasweq: Argh, whoops. Switched.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+			name: "send-msg",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			opts: testutils.MetricDataOptions{
+				CSMLabels:                  csmLabels,
+				UnaryCompressedMessageSize: float64(57),
+			},
```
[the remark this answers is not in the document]
zasweq: Discussed offline; The reason for the streaming/unary inputs of operations is the most intricate piece of this was correctly attaching the metadata exchange labels from server side. Server side can be headers + msg + trailers || trailers only. So the different logic of what streaming/unary operations would induce metadata exchange is the main thing I wanted to test here. There's different sets of operations that can induce the metadata exchange labels to be written to header or trailers, so I test all code paths that can trigger md exchange. The major thing being tested is that the metadata exchange labels a. successfully get picked up, whether in headers or trailers. b. don't get added twice, which would cause client to error (in headers or trailers - two values say from SetHeader and then SendHeader).

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+				grpc.SendHeader(ctx, metadata.New(map[string]string{"some-metadata": "some-metadata-val"}))
+
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			opts: testutils.MetricDataOptions{
+				CSMLabels:                  csmLabels,
+				UnaryCompressedMessageSize: float64(57),
+			},
```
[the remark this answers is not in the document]
zasweq: As above.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+				grpc.SetHeader(ctx, metadata.New(map[string]string{"some-metadata": "some-metadata-val"}))
+
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
+				}}, nil
+			},
+			opts: testutils.MetricDataOptions{
+				CSMLabels:                  csmLabels,
+				UnaryCompressedMessageSize: float64(57),
+			},
```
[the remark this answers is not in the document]
zasweq: As above.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+		// To test the different operations for Unary RPC's from the interceptor
+		// level that can plumb metadata exchange header in.
+		unaryCallFunc func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error)
+		opts          testutils.MetricDataOptions
+	}{
+		{
+			name: "normal-flow",
+			unaryCallFunc: func(ctx context.Context, in *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+				return &testpb.SimpleResponse{Payload: &testpb.Payload{
+					Body: make([]byte, 10000),
```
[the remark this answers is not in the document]
zasweq: Ah good point. Done.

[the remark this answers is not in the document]
zasweq: I was debating between two things: allocating new memory such as 
```
return &testpb.SimpleResponse{Payload: &testpb.Payload{
	Body: make([]byte, len(in.GetPayload().GetBody())),
}}, nil
```

or just taking the pointer off the request and sticking it in the response.

```
return &testpb.SimpleResponse{Payload: &testpb.Payload{
	Body: in.GetPayload().GetBody(),
}}, nil
```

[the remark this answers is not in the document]
zasweq: I chose the first because even though I haven't triaged it explicitly, I feel like it's bad to take the memory of the proto passed into the application layer and then send it back down into the transport to be compressed etc. Let me know if you have any thoughts about this. I did this in both observability test and OpenTelemetry e2e test.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+					MetricsOptions: opentelemetry.MetricsOptions{
+						MeterProvider: provider,
+						Metrics:       opentelemetry.DefaultMetrics,
+					}}, po),
+			}
+			dopts := []grpc.DialOption{dialOptionWithCSMPluginOption(opentelemetry.Options{
+				MetricsOptions: opentelemetry.MetricsOptions{
+					MeterProvider:  provider,
+					Metrics:        opentelemetry.DefaultMetrics,
+					OptionalLabels: []string{"csm.service_name", "csm.service_namespace_name"}, // should be a no-op unless receive labels through optional labels mechanism
```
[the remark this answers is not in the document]
zasweq: This knob in the OTel options doesn't do anything expect when it receives labels through the optional labels mechanism (i.e. right now only from an xDS LB Policy). Due to this information being part of this Knob's docstring, I will delete these comments.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+	resourceDetectorEmissions := map[string]string{
+		"cloud.platform":     "gcp_kubernetes_engine",
+		"cloud.region":       "cloud_region_val", // availability_zone isn't present, so this should become location
+		"cloud.account.id":   "cloud_account_id_val",
+		"k8s.namespace.name": "k8s_namespace_name_val",
+		"k8s.cluster.name":   "k8s_cluster_name_val",
+	}
+	nodeID := "projects/12345/networks/mesh:mesh_id/nodes/aaaa-aaaa-aaaa-aaaa"
+	csmCanonicalServiceName := "csm_canonical_service_name"
+	csmWorkloadName := "csm_workload_name"
```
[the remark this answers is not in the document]
zasweq: Done.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+		// interceptor level that can plumb metadata exchange header in.
+		streamingCallFunc func(stream testgrpc.TestService_FullDuplexCallServer) error
+		opts              testutils.MetricDataOptions
+	}{
+		{
+			name: "trailers-only",
+			streamingCallFunc: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+				for {
+					_, err := stream.Recv()
+					if err == io.EOF {
```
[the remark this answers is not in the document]
zasweq: Done.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+						return nil
+					}
+				}
+			},
+			opts: testutils.MetricDataOptions{
+				CSMLabels: csmLabels,
+			},
+		},
+		{
+			name: "set-header",
```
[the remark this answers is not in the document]
zasweq: Same as above.

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+}
+
+// TestxDSLabels tests that xDS Labels get emitted from OpenTelemetry metrics.
+// This test configures OpenTelemetry with the CSM Plugin Option, and xDS
+// Optional Labels turned on. It then configures an interceptor to attach
+// labels, representing the cluster_impl picker. It then makes a unary RPC, and
+// expects xDS Labels labels to be attached to emitted relevant metrics. Full
+// xDS System alongside OpenTelemetry will be tested with interop. (there is
+// a test for xDS -> Stats handler and this tests -> OTel -> emission).
+func (s) TestxDSLabels(t *testing.T) {
```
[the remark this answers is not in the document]
zasweq: Ah, done.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -0,0 +1,798 @@
...
+		for _, sm := range rm.ScopeMetrics {
+			for _, m := range sm.Metrics {
+				gotMetrics[m.Name] = m
+			}
+		}
+		val, ok := gotMetrics[wantMetric.Name]
+		if !ok {
+			continue
+		}
+		if !metricdatatest.AssertEqual(t, wantMetric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars()) {
```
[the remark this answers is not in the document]
zasweq: Hmmmm good point. Removed the conditional and continue, and just let it fail the test. I do think I saw this come up a few times where it hit this during testing iterations.

stats/opentelemetry/internal/testutils/testutils.go
zasweq: Ah right this repeats it. I just will fail directly with the error here.

zasweq: Done for Capitalization strings in all testing files under stats/opentelemetry (Expect in cases that start with a local ref, such as stream or ss). Will note for future.

stats/opentelemetry/csm/observability_test.go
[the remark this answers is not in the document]
easwars: We don't run our tests in parallel, because that would require us to change the way our `leakcheck` works. But using `t.Cleanup` to run cleanup functions from helper routines instead of returning a cancel function is preferable unless the caller needs control of *when* the cleanup needs to be done. In the same vein, we don't use `t.Cleanup` when we allocate some resources in the main test goroutine since `defer` is esthetically simpler in most cases when compared to `t.Cleanup`. 

Also `t.Cleanup` and `defer` use separate function stacks. So, if we can want to guarantee some sort of ordering between the cleanup functions pushed on these two stacks, then it is a problem, and it is best to use a single one in that case.

stats/opentelemetry/csm/observability_test.go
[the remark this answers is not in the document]
easwars: Sounds reasonable.

easwars (approved):

stats/opentelemetry/csm/observability_test.go
```diff
@@ -0,0 +1,617 @@
...
+
+			opts := test.opts
+			opts.Target = ss.Target
+			wantMetrics := testutils.MetricDataStreaming(opts)
+			testutils.CompareMetrics(ctx, t, reader, gotMetrics, wantMetrics)
+		})
+	}
+}
+
+func unaryInterceptorAttachxDSLabels(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
```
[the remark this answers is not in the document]
zasweq: Whoops, missed this one.

stats/opentelemetry/csm/observability_test.go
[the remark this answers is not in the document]
zasweq: Ah ok noted.

zasweq:
Thank you for the detailed passes on this; this looks much better than when it started.

PR #7280, opened 2024-05-29
interop/xds: Interop client and server changes for CSM Observability

zasweq:
This PR adds the CSM Observability flags for interop testing. It makes the xDS interop client and server their own go.mod to avoid upgrading the minimum required go version of the top level grpc/ go.mod to require go 1.21, as the client and server take a transitive OpenTelemetry dependency through the csm package, which requires go 1.21.

RELEASE NOTES: N/A


interop/xds/client/Dockerfile
```diff
@@ -19,19 +19,19 @@
 FROM golang:1.21-alpine as build
 
 # Make a grpc-go directory and copy the repo into it.
-WORKDIR /go/src/grpc-go
+WORKDIR /go/src/grpc-go/
```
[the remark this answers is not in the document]
zasweq: Deleted.

interop/xds/client/Dockerfile
```diff
@@ -19,19 +19,19 @@
...
 # the Go compiler and gRPC repo from the earlier stage. This significantly
 # reduces the docker image size.
 FROM alpine
-COPY --from=build /go/src/grpc-go/client .
+COPY --from=build /go/src/grpc-go/interop/xds/client/client .
 ENV GRPC_GO_LOG_VERBOSITY_LEVEL=99
 ENV GRPC_GO_LOG_SEVERITY_LEVEL="info"
 ENV GRPC_GO_LOG_FORMATTER="json"
-ENTRYPOINT ["./client"]
+ENTRYPOINT ["./client"]
```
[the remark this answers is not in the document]
zasweq: Done. I think this came from copy and paste from my branch on my fork.

interop/xds/client/client.go
```diff
@@ -35,15 +36,21 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/xds"
 	"google.golang.org/grpc/grpclog"
+	_ "google.golang.org/grpc/interop/xds"
```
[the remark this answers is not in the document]
zasweq: Oh, whoops. Deleted.

interop/xds/client/client.go
```diff
@@ -368,6 +378,22 @@ func parseRPCMetadata(rpcMetadataStr string, rpcs []string) []*rpcConfig {
...
 	flag.Parse()
+	if *enableCSMObservability {
+		exporter, err := prometheus.New()
+		if err != nil {
+			logger.Fatalf("Failed to start prometheus exporter: %v", err)
+		}
+		provider := metric.NewMeterProvider(
+			metric.WithReader(exporter),
+		)
+		go http.ListenAndServe("0.0.0.0:9464", promhttp.Handler())
```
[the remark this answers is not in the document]
zasweq: This is a defined number from the testing design: "opentelemetry::exporter::metrics::PrometheusExporterOptions opts;
// default was "localhost:9464" which causes connection issue
// across GKE pods
opts.url = "0.0.0.0:9464";" 

"When the client / server GKE pods are running, the Prometheus metrics should be available at the port :9464."

If this port is in use the test will fail. This is unexpected, as I think the tests run synchronously with defined client/server image behavior so nothing will collide. If the port collides, that will be an infra issue I think.

interop/xds/server/Dockerfile
```diff
@@ -24,14 +24,14 @@ COPY . .
...
 # the Go compiler and gRPC repo from the earlier stage. This significantly
 # reduces the docker image size.
 FROM alpine
-COPY --from=build /go/src/grpc-go/server .
+COPY --from=build /go/src/grpc-go/interop/xds/server/server .
 ENV GRPC_GO_LOG_VERBOSITY_LEVEL=99
 ENV GRPC_GO_LOG_SEVERITY_LEVEL="info"
 ENV GRPC_GO_LOG_FORMATTER="json"
-ENTRYPOINT ["./server"]
+ENTRYPOINT ["./server"]
```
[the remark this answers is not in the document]
zasweq: Ah whoops added newline.

interop/xds/client/client.go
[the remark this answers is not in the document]
dfawley: The 0s don't seem correct to me, though, because that limits us to IPv4?  I think you want `:9464` to attach the listener onto all networks.  But also this seems like something that should be a flag?

interop/xds/server/go.mod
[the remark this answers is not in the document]
zasweq: Done.

interop/xds/client/go.mod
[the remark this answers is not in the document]
zasweq: Done.

interop/xds/client/client.go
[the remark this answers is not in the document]
zasweq: "Stanley Cheung, Now
Sergii and I talked about this and we are fine with leaving this hard coded for now. We are aware of this and if there's a need we can pass that as a flag. Just don't see a need right now"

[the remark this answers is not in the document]
zasweq: "Stanley Cheung, Now
I somehow ran into a problem for the c++ exporter that the default URL doesn't work and I have to override it with [0.0.0.0:9464](http://0.0.0.0:9464/) even though the default port is 9464"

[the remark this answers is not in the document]
zasweq: "Let's do that in Go too. If there's a problem then at least we know all of the lang clients have the same thing to change"

interop/xds/go.mod
```diff
@@ -1,16 +1,19 @@
-module google.golang.org/interop/test/server
+module google.golang.org/interop/xds
 
 go 1.21.0
 
-replace google.golang.org/grpc => ../../..
+replace google.golang.org/grpc => ../..
 
-replace google.golang.org/grpc/stats/opentelemetry => ../../../stats/opentelemetry
+replace google.golang.org/grpc/interop/xds => ./
```
dfawley, line 7: What does this even mean? :)

zasweq: Replace this with the current directory

zasweq:
Thanks for the pass sorry for the hassle; interop still works on latest commit

interop/xds/client/client
dfawley, line 1: Please remove.

interop/xds/client/client.go
```diff
@@ -368,6 +378,22 @@ func parseRPCMetadata(rpcMetadataStr string, rpcs []string) []*rpcConfig {
...
 	flag.Parse()
+	if *enableCSMObservability {
+		exporter, err := prometheus.New()
+		if err != nil {
+			logger.Fatalf("Failed to start prometheus exporter: %v", err)
+		}
+		provider := metric.NewMeterProvider(
+			metric.WithReader(exporter),
+		)
+		go http.ListenAndServe(":9464", promhttp.Handler())
```
dfawley, line 389: Please add port flag and use it here.

dfawley, line 389: Should we do something like `go func() { if err := ListenAndServe; err != nil { logger.Fatalf } }()`, so if the port is taken we get a noisy error instead of just things not working right?

interop/xds/client/client.go
```diff
@@ -368,6 +378,22 @@ func parseRPCMetadata(rpcMetadataStr string, rpcs []string) []*rpcConfig {
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+		defer cancel()
+		cleanup := csm.EnableObservability(ctx, opentelemetry.Options{MetricsOptions: opentelemetry.MetricsOptions{MeterProvider: provider}})
+		defer cleanup()
```
dfawley, line 394: If we're going to need to flush stuff in a cleanup function one day maybe, is it possible it will also need to accept a context so that can apply a timeout/cancelation?  @ejona86 @yashykt  (Note that if a context is needed, we aren't supposed to ever substitute our own; it is supposed to _always, only_ come from `main()`.)

interop/xds/client/client.go
dfawley: *or environment variable, whatever is decided.

interop/xds/client/client.go
ejona86: That wouldn't be unreasonable, although any I/O is probably already happening and this is just waiting for it to complete.

What did you do for gcp-observability?

zasweq: https://github.com/grpc/grpc-go/blob/master/gcp/observability/observability.go#L89 no context. IIRC remember it flushes the data to the exporter? So idk how timeouts work I think it's just a write somewhere.

interop/xds/client/client
zasweq: Whoops, deleted.

interop/xds/client/client.go
zasweq: Done for both client and server.

zasweq: Decided to do env var "OTEL_EXPORTER_PROMETHEUS_PORT" with default of 9464. Interop test won't set and will default.

interop/xds/client/client.go
dfawley: The downstream APIs we needed at the time (e.g. https://github.com/grpc/grpc-go/blob/24e90243755430b73804d2bf13f0abcfa19361e5/gcp/observability/exporting.go#L90) didn't require it, and I probably didn't think about this as being a possibility, either.  But since we don't know what we're targeting, we also don't know whether other future downstream APIs might.

I'm fine to not pass a context here since we don't think we'll need it, and just make our own context if we have to in the future, even if it's in violation of Go's ideals.

zasweq:
Verified recent commits with another interop pass.

interop/xds/go.mod
dfawley: "Replace myself with myself".  :)

Glad we figured out what was wrong and that it was just a simple mistake.

interop/xds/client/Dockerfile
```diff
@@ -24,13 +24,13 @@ COPY . .
 
 # Build a static binary without cgo so that we can copy just the binary in the
 # final image, and can get rid of Go compiler and gRPC-Go dependencies.
-RUN go build -tags osusergo,netgo interop/xds/client/client.go
+RUN cd interop/xds/client && go build -tags osusergo,netgo client.go
```
dfawley, line 27: How about `.`?

I think if you specify the file here and we add another file, it will break.

interop/xds/server/Dockerfile
```diff
@@ -24,13 +24,13 @@ COPY . .
 
 # Build a static binary without cgo so that we can copy just the binary in the
 # final image, and can get rid of the Go compiler and gRPC-Go dependencies.
-RUN go build -tags osusergo,netgo interop/xds/server/server.go
+RUN cd interop/xds/server && go build -tags osusergo,netgo server.go
```
dfawley, line 27: As above, `build .` is probably better

dfawley (approved):

interop/xds/go.mod
zasweq: Haha :). Thanks for the help!

interop/xds/client/Dockerfile
zasweq: Done.

interop/xds/server/Dockerfile
zasweq: Done.

zasweq:
Continues to pass; merging.

PR #7282, opened 2024-05-30
*: update deps

arvindbr8:
I got the feeling that #7274 was not complete. For that PR, I ran `go mod tidy` with `--compat=1.19` to update the deps per our release manual. However, `toolchain` declarative got added pointing to `go1.22.1`. I think this is based on the go env on my local. 

Taking into we have already drop support for 1.19 and 1.20, I updated all references to older go versions from the repo including:
1. `go.mod` s
2. `staticheck` in vet.sh
3. `go mod tidy` in vet.sh

This PR contains the changes that includes everything i said above plus, the additional changes that were required to remove deprecated usages of 
1. [rand.Seed](https://pkg.go.dev/math/rand#Seed)
2. [RevocationList.RevokedCertificates](https://pkg.go.dev/crypto/x509#RevocationList.RevokedCertificates)

RELEASE NOTES: none


cmd/protoc-gen-go-grpc/go.mod
```diff
@@ -1,5 +1,5 @@
 module google.golang.org/grpc/cmd/protoc-gen-go-grpc
 
-go 1.19
+go 1.21
```
dfawley, line 3: Did this happen because of a dependency?  Is there some way to find out which one?  Maybe a `go` command?

go.mod
```diff
@@ -1,6 +1,6 @@
 module google.golang.org/grpc
 
-go 1.19
+go 1.21
```
dfawley, line 3: Similar question here.

arvindbr8:
Per offline discussion yday: 

There seems to be enough reasons for gRPC-Go to require go1.21, but is not limited to:"

1. To remove grpcrand and use math/rand directly
2. stats/opentelemetry module pulls in dependency which requires go1.21+

dfawley (approved):

go.mod
ldemailly: I do have a question about this, I don't think you use any feature of the language that isn't in 1.19, and that change then propagates to my library which, while I personally build binary with 1.22.5, I'd rather not force upgrade yet for users (and I was planning on a minor patch level bump but if my go.mod changes from 1.19 to 1.21 I'd have to make a new minor)

ldemailly: nvm
> stats/opentelemetry module pulls in dependency which requires go1.21+
so I guess I would need to track down why _they_ bumped it... possibly unecessarily

dfawley: I think this is the new normal for the Go ecosystem, particularly after the changes made to go.mod in 1.21.  We won't bump this unless we have to, but it will become more and more likely that we'll need to shortly after any new Go version is released.

caldempsey:
I think one of the best things about Go is the communities cadence to not sit on old versions of Golang. It is inconvenient to force an upgrade, yes, but we also have to learn from history. We also have to learn from Node.

caldempsey:
Opinion piece: I think one of the best things about Go is the communities cadence to not sit on old versions of Golang. It is inconvenient to force an upgrade, yes, but we also have to learn from history. We also have to learn from TypeScript. The principle of least pain suggests to me that staying current is a greater wisdom than falling so far behind developers have to be juggling four versions of Node.

PR #7303, opened 2024-06-05
[advancedTLS] Removed deprecated APIs in advancedTLS

gtcooke94:
This PR removes the following deprecated APIs in preparation for a 1.0 release of advancedTLS:
The first list were simply renamed
```
RootCertificateOptions.RootCACerts
GetRootCAsParams
GetRootCAsResults
ClientOptions
ServerOptions
Options.MinVersion
Options.MaxVersion
RevocationConfig
Options.RevocationConfig
AllowUndetermined
CustomVerificationFunc
Options.VType
Options.VerifyPeer
```

The second list are fully deprecated features in the CRL stack - use CRL providers instead. For reading a directory, the `FileWatcherCRLProvider` should be a relatively simple replacement.
```
Cache
RevocationOptions.RootDir
RevocationOptions.Cache
```

RELEASE NOTES: None

### DIFF
--- a/security/advancedtls/crl.go
+++ b/security/advancedtls/crl.go
@@ -23,20 +23,13 @@ package advancedtls
 
 import (
 	"bytes"
-	"crypto/sha1"
-	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/asn1"
-	"encoding/binary"
-	"encoding/hex"
 	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
-	"strings"
-	"time"
 
 	"golang.org/x/crypto/cryptobyte"
 	cbasn1 "golang.org/x/crypto/cryptobyte/asn1"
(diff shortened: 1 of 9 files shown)


matthewstevenson88 (commented):
Please fix the failing tests, otherwise LGTM.

security/advancedtls/crl.go
```diff
@@ -319,7 +183,7 @@ func fetchCRL(c *x509.Certificate, crlVerifyCrt []*x509.Certificate, cfg Revocat
 		}
 		return crl, nil
 	}
-	return fetchIssuerCRL(c.RawIssuer, crlVerifyCrt, cfg)
+	return nil, fmt.Errorf("trying to fetch CRL but CRLProvider is nil")
```
erm-g, line 186: Should we move this check to become the fist thing in this function?

gtcooke94: Good call, changed

erm-g (approved):
LGTM

arvindbr8:
Note: the rel notes in the PR description is only for grpc-Go releases. We would need to come up with rel notes for the advancedTLS release.

Also since the fields were deprecated before the 1.0 release, i would rather not mention anything in the rel notes for 1.0

gtcooke94:
That's a good point - I removed release notes

arvindbr8 (approved):
LGTM. 

As for 1.0 release, I think we should just audit everything once again. For eg: I see we have a separate module under `security/advancedtls/examples/` which has a replace directive to the advancedTLS module. So after release, we should also update the go.mod to reference to the right release

PR #7305, opened 2024-06-06
vet: remove --quiet from git grep when output is expected

atollena:
We pass `--quiet` to git grep in places where we re-grep the output, which never yields any result because `--quiet` suppresses all output from stdout, even when -l is passed.

Remove `--quiet` from all grep in vet.

RELEASE NOTES: none


arvindbr8 (approved):
You are right.  errr, the change I added the `--quiet` in -- I remember I was tackling another problem. But everytime I run vet in Darwin it keeps hanging in an editor window when executing the git grep command. 

My bad. Approved.

atollena:
> But everytime I run vet in Darwin it keeps hanging in an editor window when executing the git grep command.

Hmm. fwiw I also use Darwin, if you run into issues we can try to collaborate on them.

PR #7349, opened 2024-06-25
experimental/stats: Add metrics registry

zasweq:
This PR adds the instrument registry for Non Per Call Metrics. It also moves Metrics symbol to external, and defined a MetricRecorder type for non per call metrics.

The next PR will be the OpenTelemetry component usage of this global registry and implementation of MetricRecorder, including the merging of default metrics by shifting default metrics to a function as per Yashes suggestion.

RELEASE NOTES: N/A


zasweq:
Got to offline discussion items, and moved symbols around to where we discussed.

experimental/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+
+// Package instrumentregistry contains the experimental instrument registry data
+// structures exposed to users.
+package instrumentregistry
+
+import "google.golang.org/grpc/stats"
+
+// Int64CountHandle is a typed handle for a int count instrument. This handle is
```
dfawley, line 25: *an int - throughout.

experimental/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,61 @@
...
+// passed at the recording point in order to know which instrument to record on.
+type Int64CountHandle struct {
+	Index int
```
dfawley, line 28: Does this need to be exported?

experimental/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,61 @@
...
+// Int64GaugeHandle is a typed handle for a int gauge instrument. This handle is
+// passed at the recording point in order to know which instrument to record on.
+type Int64GaugeHandle struct {
+	Index int
+}
+
+// DefaultMetrics are the default metrics registered through global instruments
+// registry. This is written to at initialization time only, and is read
+// only after initialization.
+var DefaultMetrics = make(map[stats.Metric]bool)
```
dfawley, line 61: I don't think we want users to have direct access to this map?  I think they just need a `Register` type function exposed.

internal/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package instrumentregistry is a instrument registry that gRPC components can
```
dfawley, line 19: *an

internal/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,199 @@
...
+package instrumentregistry
+
+import (
+	"log"
+
+	"google.golang.org/grpc/experimental/stats/instrumentregistry"
+	"google.golang.org/grpc/stats"
+)
+
+// InstrumentDescriptor is a data of a registered instrument (metric).
```
dfawley, line 30: "the data for"?

internal/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,199 @@
...
+type InstrumentDescriptor struct {
+	// Name is the name of this metric.
+	Name string
+	// Description is the description of this metric.
+	Description string
+	// Unit is the unit of this metric.
```
dfawley, line 36: Can we delete these comments? Or make them less tautological?

internal/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,199 @@
...
+
+import (
+	"log"
+
+	"google.golang.org/grpc/experimental/stats/instrumentregistry"
+	"google.golang.org/grpc/stats"
+)
+
+// InstrumentDescriptor is a data of a registered instrument (metric).
+type InstrumentDescriptor struct {
```
dfawley, line 31: Aren't these things that third-party plugins will eventually need to work with?  (I.e. should be in `experimental`?)

internal/stats/instrumentregistry/instrument_registry.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+// RegisterInt64Count registers the int count instrument description onto the
+// global registry. It returns a typed handle to use when recording data.
+//
+// NOTE: this function must only be called during initialization time (i.e. in
+// an init() function), and is not thread-safe. If multiple instruments are
+// registered with the same name, this function will panic.
+func RegisterInt64Count(name string, desc string, unit string, labels []string, optionalLabels []string, def bool) instrumentregistry.Int64CountHandle {
```
dfawley, line 101: How about having the user pass the `InstrumentDescriptor` directly?

experimental/stats/instrumentregistry/instrument_registry.go
zasweq: Yes, both the OpenTelemetry component and the MetricsRecorderList (the next two PRs) read this index out. (In MetricsRecorderList, it uses the index to verify the labels/optionalLabels (which will log at error and eat call, not panic like Java as discussed offline). OpenTelemetry uses this index as the invariant of instrument registry that it creates metrics from and needs the Index to know which created metric to record. (In the code I have, I create noops for unregistered metrics as well, and the index will match to that and be eaten).

experimental/stats/instrumentregistry/instrument_registry.go
zasweq: We discussed this offline; I thought one of the requirements you gave me was to explicitly expose this, so users that write their own stats plugins/non per call interface can have access to it.

internal/stats/instrumentregistry/instrument_registry.go
zasweq: Oh hmmmm you're right. So the only things kept internal are what a gRPC component calls into, and this will need to be read by external users in their stats plugins so they can create instruments based off it.

zasweq: I asked about this in the chat; thinking about it do I need to move the whole instrument registry into experimental actually? A user can also write their own balancer/resolver, which should be able to register instruments to eventually record on so I think that all the symbols/logic need to be external (in experimental). What do you think?

internal/stats/instrumentregistry/instrument_registry.go
zasweq: Hmm ok. Thinking about it that would allow scaling up fields if needed in a backward compatible way if we decide we want to persist another piece of information here. Underlying though I'll persist around a copy of the struct not the same pointer as a user.

internal/stats/instrumentregistry/instrument_registry.go
zasweq: Discussed in chat with Yash; decided to move this whole instrument registry to experimental and make externally visible.

experimental/stats/instrumentregistry/instrument_registry.go
dfawley: They need to read the default metrics but not directly access a set of them.

Also: why not use the `Metrics` set type for this?

internal/stats/instrumentregistry/instrument_registry.go
dfawley: You can pass by value instead of pointer to avoid the need to manually copy.

internal/stats/instrumentregistry/instrument_registry.go
zasweq: Switched.

internal/stats/instrumentregistry/instrument_registry.go
zasweq: Yeah these are just for vet, is there a way to ignore vet for exported fields like this that are self explanatory?

internal/stats/instrumentregistry/instrument_registry.go
zasweq: Done.

internal/stats/instrumentregistry/instrument_registry.go
zasweq: I was thinking about it too now that I use a metrics set for defaults and there's a metrics type I might just have the inst desc take a Metric, and have the user pass in a Metric for explicit types and no need to typecast in this registry.

experimental/stats/instrumentregistry/instrument_registry.go
zasweq: Oh ok good point. I did add Merge() so yeah this should be a MetricsSet, is having the exported type be a MetricsSet good enough for your point of reading but not accessing it: "They need to read the default metrics but not directly access a set of them."

zasweq: This merge will be called at runtime in OpenTelemetry for the default metrics as discussed offline.

stats/stats.go
```diff
@@ -347,3 +348,70 @@ func OutgoingTrace(ctx context.Context) []byte {
 	b, _ := ctx.Value(outgoingTraceKey{}).([]byte)
 	return b
 }
+
+// Metric is an identifier for a metric.
+type Metric string
```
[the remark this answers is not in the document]
zasweq: Ah, good point. Done.

stats/stats.go
```diff
@@ -347,3 +348,70 @@ func OutgoingTrace(ctx context.Context) []byte {
...
+
+// Metrics is a set of metrics to record. Once created, Metrics is immutable,
+// however Add and Remove can make copies with specific metrics added or
+// removed, respectively.
+//
+// Do not construct directly; use NewMetrics instead.
+type Metrics struct {
+	// Metrics are the set of Metrics to initialize.
+	Metrics map[Metric]bool
```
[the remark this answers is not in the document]
zasweq: I export this to read this field in OTel construction (determines whether a metric is created or not)...do you want me to expose a getter for a map and copy? That would prevent the user from doing operations on this?

experimental/stats/instrumentregistry/instrument_registry.go
zasweq: Done.

internal/stats/instrumentregistry/instrument_registry.go
dfawley: You can always improve them to make them more useful.  Like give examples for units, explain what labels are used for, etc.

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package stats
+
+import (
+	"log"
```
dfawley, line 22: We don't use `"log"`

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+// MetricDescriptor is the data for a registered metric.
+type MetricDescriptor struct {
+	// Name is the name of this metric. This name must be unique across whole
+	// binary (including any per call metrics). See
+	// https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md#metric-instrument-naming-conventions
+	// for metric naming conventions.
+	Name stats.Metric
+	// Description is the description of this metric.
+	Description string
+	// Unit is the unit of this metric (e.g. entries, milliseconds).
```
dfawley, line 41: Isn't OTel opposed to using "milliseconds" and instead prefer to use "seconds" for most/all things time related?

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+// DefaultMetrics are the default metrics registered through global metrics
+// registry. This is written to at initialization time only, and is read only
+// after initialization.
+var DefaultMetrics = stats.NewMetrics()
+
+// MetricDescriptor is the data for a registered metric.
+type MetricDescriptor struct {
+	// Name is the name of this metric. This name must be unique across whole
```
dfawley, line 34: Docstrings on fields don't follow this convention like types/functions.  You can just say "The name of the metric".  Throughout.  Maybe that's why it seems so redundant to me "Name is the name".  Just say "The name".

https://google.github.io/styleguide/go/decisions#doc-comments

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+	// Type is the type of metric. This is set by the metric registry, and not
+	// intended to be set by a component registering a metric.
+	Type MetricType
+}
+
+// Int64CountHandle is a typed handle for a float count metric. This handle
+// is passed at the recording point in order to know which metric to record
+// on.
+type Int64CountHandle struct {
+	MetricDescriptor *MetricDescriptor
```
dfawley, line 61: Unexport this field throughout?  Or maybe `type Int64CountHandle MetricDescriptor`?  Or `type Int64CountHandle *MetricDescriptor`?

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Record records the int64 count value on the metrics recorder provided.
+func (h Int64CountHandle) Record(recorder MetricsRecorder, incr int64, labels ...string) {
```
dfawley, line 65: Should these just be passed around by pointer everywhere?  Why not?

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Record records the int64 histo value on the metrics recorder provided.
+func (h Int64GaugeHandle) Record(recorder MetricsRecorder, incr int64, labels ...string) {
+	recorder.RecordIntGauge(h, incr, labels...)
+}
+
+// registeredMetrics are the registered metric descriptor names.
+var registeredMetrics = make(map[stats.Metric]bool)
+
+// MetricsRegistry are all of the registered metrics.
```
dfawley, line 117: "Registry is" or "contains", not "are".

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+//
+// This is written to only at init time, and read only after that.
+var MetricsRegistry = make(map[stats.Metric]*MetricDescriptor)
```
dfawley, line 120: Why is this exported by the functions that access it aren't?  Shouldn't it be the other way around?

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,240 @@
...
+	MetricTypeIntGauge
+)
+
+// clearMetricsRegistryForTesting clears the global data of the metrics
+// registry. It returns a closure to be invoked that sets the metrics registry
+// to its original state. Only called in testing functions.
+func clearMetricsRegistryForTesting() func() {
+	oldDefaultMetrics := DefaultMetrics
+	oldRegisteredMetrics := registeredMetrics
+	oldMetricsRegistry := MetricsRegistry
```
dfawley, line 229: This shouldn't clear.  It should observe how things are before the test runs only, then reset the state to that when the test is done.

Also can we pass `testing.T` here and use `t.Cleanup`?

experimental/stats/stats.go
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package stats contains experimental metrics/stats API's.
+package stats
+
+// MetricsRecorder records on metrics derived from instrument registry.
```
dfawley, line 22: We don't seem to use "instrument" anymore right?

Also I think this should be in "metrics.go" or "metricregistry.go"?

experimental/stats/stats.go
```diff
@@ -0,0 +1,39 @@
...
+type MetricsRecorder interface {
+	// RecordIntCount records the measurement alongside labels on the int count
+	// associated with the provided handle.
+	RecordIntCount(Int64CountHandle, int64, ...string)
```
dfawley, line 26: It's nice to name your parameters as a documentation aide.

stats/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+ *
+ */
+
+package stats
+
+import "maps"
+
+// Metric is an identifier for a metric.
+type Metric string
```
dfawley, line 24: Don't we want all these to be in experimental for now, too?

stats/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Metrics is a set of metrics to record. Once created, Metrics is immutable,
+// however Add and Remove can make copies with specific metrics added or
+// removed, respectively.
+//
+// Do not construct directly; use NewMetrics instead.
+type Metrics struct {
+	// Metrics are the set of Metrics to initialize.
+	Metrics map[Metric]bool
```
dfawley, line 33: We should hide fields to prevent direct accesses, I think.

stats/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+		newMetrics[metric] = true
+	}
+	return &Metrics{
+		Metrics: newMetrics,
+	}
+}
+
+// Join joins the metrics passed in with the metrics set, and returns a new copy
+// with the merged metrics.
+func (m *Metrics) Join(metrics *Metrics) *Metrics { // Use this API...
```
dfawley, line 65: Can this function body be simplified to `m.Add(m.Values()/*or whatever*/)`?

dfawley (commented):
If we check this in without any code using it, then we should probably be prepared to make a lot of changes here.

experimental/stats/stats.go
zasweq: Ah whoops missed this file (I went on a crusade to switch all instances in other files). Switched.

experimental/stats/metricregistry.go
zasweq: Discussed offline; tests should be robust against prepopulated state in the instrument registry as that simulates real world scenario, clear only if absolutely needed.

experimental/stats/metricregistry.go
zasweq: Ah I completely forgot. I was using this in OpenTelemetry and wondering why we had a map in the first place since the name was in the values, but yes if we return a getter that gets a pointer for a metric name which will create a better API, we do need the map as we discussed previously. Thank you for pointing this out.

experimental/stats/metricregistry.go
zasweq: Sure we could. This is a copy of a that x bit pointer but pointer works too.

experimental/stats/metricregistry.go
zasweq: Hmm ok so we could do an alias too? I needed this exported for OpenTelemetry to read the pointer to look into it's data structures it builds out for metrics. (I guess the metrics recorder list could live here for the assertion on labels/optional labels length and access unexported fields).

experimental/stats/metricregistry.go
zasweq: Maybe I'll just delete these comments.

experimental/stats/metricregistry.go
zasweq: Oh yeah I thought the A66 metrics were /milliseconds, but they are defined as /s (https://github.com/grpc/proposal/blob/master/A66-otel-stats.md#client-per-attempt-instruments) which is why we had those bucket counts.

stats/metrics.go
zasweq: This field is read at OTel creation time. I could add a to slice function which would solve the function body comment below, but then it wouldn't be constant accesses at otel creation time when determining whether to build a metric or not. Let me see what looks best.

experimental/stats/stats.go
zasweq: Ah, yeah I was wondering this too. Named the parameters handle, incr, and labels.

experimental/stats/metricregistry.go
zasweq: Choose contains.

experimental/stats/metricregistry.go
zasweq: Actually switched comments for field to // The Field...

experimental/stats/metricregistry.go
zasweq: If I make it a pointer alias; I can't define functions on it. I tried making it a alias to metricDescriptor but the problem is if I typecast the descriptor into a xxxHandle it changes address. How do we keep it pointers so it can link alongside calling methods on handle returned?

stats/metrics.go
zasweq: The stats package is entirely experimental as per top level package comment. I think it's fine to leave this here, since configures unstable OTel which lives in this package too.

experimental/stats/metricregistry.go
zasweq: Ah whoops I had this for panic formatting. Switched to logger.Fatalf().

stats/metrics.go
zasweq: I added a getter for this an unexported so we can use that map to check for presence in constant time.

stats/metrics.go
zasweq: I left it as is since I added a getter for map only not to convert to slice.

experimental/stats/stats.go
zasweq: Let's leave this here for now until we wrap up discussion on whether metrics.go should be experimental or not.

stats/metrics.go
dfawley: If you're going with that philosophy then why have `experimental/stats` for the other stuff?  Let's keep it together.

> since configures unstable OTel which lives in this package too

OTel should not have anything in the stats package, right?

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,231 @@
...
+var logger = grpclog.Component("metrics-registry")
+
+// DefaultMetrics are the default metrics registered through global metrics
+// registry. This is written to at initialization time only, and is read only
+// after initialization.
+var DefaultMetrics = stats.NewMetrics() // loop through this set,, export metrisdescriptor passed upward for pointer and labels/optional lables...
+
+// MetricDescriptor is the data for a registered metric.
+type MetricDescriptor struct {
+	// The name. This name must be unique across whole binary (including any per
```
dfawley, line 37: "The name...of this metric"?

And "across _the_ whole binary" reads a bit better.

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,231 @@
...
+	"google.golang.org/grpc/grpclog"
+	"google.golang.org/grpc/stats"
+)
+
+var logger = grpclog.Component("metrics-registry")
+
+// DefaultMetrics are the default metrics registered through global metrics
+// registry. This is written to at initialization time only, and is read only
+// after initialization.
+var DefaultMetrics = stats.NewMetrics() // loop through this set,, export metrisdescriptor passed upward for pointer and labels/optional lables...
```
dfawley, line 33: Delete comment?

experimental/stats/metricregistry.go
dfawley: Maybe you can point to a playground link illustrating the things you tried?  The code as-is is still a struct so I can't see what you did.

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,231 @@
...
+
+// metricsRegistry contains all of the registered metrics.
+//
+// This is written to only at init time, and read only after that.
+var metricsRegistry = make(map[stats.Metric]*MetricDescriptor) // so OTel just loops through set provided by user into this thing...and calls get
+
+// GetMetric returns the MetricDescriptor from the global registry.
+//
+// Returns nil if MetricDescriptor not present.
+func GetMetric(metric stats.Metric) *MetricDescriptor {
```
dfawley, line 128: `GetMetricDescriptor` maybe?

dfawley: Or `DescriptorForMetric`?

experimental/stats/metricregistry.go
zasweq: Done for both.

experimental/stats/metricregistry.go
zasweq: Sorry missed this.

experimental/stats/metricregistry.go
zasweq: Chose DescriptorForMetric.

stats/metrics.go
zasweq: K will move.

zasweq:
Thanks for your help; got to all comments and fixed tests.

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,230 @@
...
+	// The unit (e.g. entries, seconds) of this metric.
+	Unit string
+	// The required label keys for this metric. These are intended to
+	// metrics emitted from a stats handler.
+	Labels []string
+	// The optional label keys for this metric. These are intended to attached
+	// to metrics emitted from a stats handler if configured.
+	OptionalLabels []string
+	// Whether this metric is on by default.
+	Default bool
```
dfawley, line 53: Something to think about for later when you're implementing the things that actually use all of this: should we remove this field from the descriptor since it (I think?) only affects the initial registration and whether to put it into the `DefaultMetrics` set?

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Int64CountHandle is a typed handle for a int count metric. This handle
+// is passed at the recording point in order to know which metric to record
+// on.
+type Int64CountHandle MetricDescriptor
+
+// Record records the int64 count value on the metrics recorder provided.
+func (h *Int64CountHandle) Record(recorder MetricsRecorder, incr int64, labels ...string) {
+	recorder.RecordIntCount(h, incr, labels...)
```
dfawley, line 66: Why does `Int64CountHandle` have a size on it but the `RecordIntCount` method does not?  It seems like either both should say 64, or neither should.

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+}
+
+// RegisterInt64Count registers the metric description onto the global registry.
+// It returns a typed handle to use to recording data.
+//
+// NOTE: this function must only be called during initialization time (i.e. in
+// an init() function), and is not thread-safe. If multiple metrics are
+// registered with the same name, this function will panic.
+func RegisterInt64Count(descriptor MetricDescriptor) *Int64CountHandle {
```
dfawley, line 137: Let's make sure the "numbers vs. no numbers" matches here too.

experimental/stats/metricregistry.go
```diff
@@ -0,0 +1,230 @@
...
+// MetricType is the type of metric.
+type MetricType int
+
+const (
+	MetricTypeIntCount MetricType = iota
+	MetricTypeFloatCount
+	MetricTypeIntHisto
+	MetricTypeFloatHisto
+	MetricTypeIntGauge
+)
```
dfawley, line 210: Nit: this should maybe move to immediately before or after the `MetricDescriptor` type definition.

experimental/stats/metricregistry_test.go
```diff
@@ -0,0 +1,260 @@
...
+	desc := MetricDescriptor{
+		// Type is not expected to be set from the registerer, but meant to be
+		// set by the metric registry.
+		Name:        "simple counter",
+		Description: "number of times recorded on tests",
+		Unit:        "calls",
+	}
+	RegisterInt64Count(desc)
+	RegisterInt64Gauge(desc)
+}
```
dfawley, line 56: This test doesn't fail if a panic doesn't occur.

experimental/stats/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+	newMetrics := make(map[Metric]bool)
+	for _, metric := range metrics {
+		newMetrics[metric] = true
+	}
+	return &Metrics{
+		metrics: newMetrics,
+	}
+}
+
+// Metrics returns the metrics set.
```
dfawley, line 67: State that the returned map should be read-only and must not be modified, please.

dfawley (approved):

dfawley:
This should be good to go modulo the small nits/test issue.  Let's be ready to iterate on the registry stuff if you find it isn't working well while implementing the things that use it.

experimental/stats/metricregistry.go
zasweq: Yeah I've thought about this persistence before, I wanted to delete this since this information is encoded in DefaultSet, but this is how the user registers through API now (since we persist a copy of what user sets). Do you see a way to delete the field but keep the API?

experimental/stats/metricregistry.go
zasweq: Ah yeah good point. I'd prefer leaving the 64 bit identifier as OpenTelemetry and Java/C specify 64. This is because OpenTelemetry has both int/float(32||64) instruments so if we scale this registry up in the future it more aligns with OpenTelemetry.

experimental/stats/metricregistry.go
zasweq: See above comment; chose numbers route.

experimental/stats/metricregistry.go
zasweq: Good point. Chose after since Type MetricType field is at the last field.

experimental/stats/metrics.go
zasweq: Done. It's weird since it says metrics set and returned map but I think those are aliases of each other, and functionally this does return a map, which is treated as a set.

experimental/stats/metricregistry_test.go
zasweq: recover() returns nil if the test is not panicing. Then I typecast to string, and that won't contain the want right? Will my typecast of a nil to a string error?

dfawley: Oh, duh, the defer will still run, sorry!

experimental/stats/metricregistry.go
dfawley: Two obvious other options:

1. `RegisterFooMetric()` and `RegisterDefaultFooMetric()`
2. `RegisterFooMetric(md MetricDescriptor, default bool)`

I think this is okay, too, but we should consider the above since this data is effectively discarded immediately after registration.

dfawley (approved):

experimental/stats/metricregistry.go
zasweq: I'd rather do 2. Let's keep this in mind for the future since I'm sure this metrics registry will iterate organically as we merge more layers.

PR #7379, opened 2024-07-02
fix: Add a package-level mutex protecting SetLoggerV2 to avoid DATA RACE errors

ppicom:
Hi there! 👋 

I have an application running e2e tests for a gRPC server (AKA my service). That service also uses other gRPC clients to call other services. That eventually leads to `grpc_logrus.ReplaceGrpcLogger` being called concurrently due to the server and the clients constructors trying to replace and use the logger at the same time.

It's failing my test suite due to DATA ERRORS and I was wondering if protecting SetLoggerV2 with a Mutex was such a bad idea to avoid this 🤔 

Thanks!

### DIFF
--- a/grpclog/loggerv2.go
+++ b/grpclog/loggerv2.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 
 	"google.golang.org/grpc/internal/grpclog"
 )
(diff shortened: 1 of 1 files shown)


arjan-bal:
Hi @ppicom, the recommendation is to only call SetLoggerV2 only in the `init()` function of modules. The Go runtime would then ensure that `SetLoggerV2` calls are serialized.

Based on the above, I don't see the need for the mutex to ensure synchronization, so I'm closing this PR. Feel free to open an issue if you have a different usecase and we can discuss it further.

PR #7380, opened 2024-07-02
Change version to 1.63.3

arvindbr8:


arvindbr8:
Yes. I think it's always a case that the upstream protos are updated in older releases. But I dont think we worry about that as much.

Issue #7365, opened 2024-06-28
Flaky Test/AuthorityRevive in google.golang.org/grpc/xds/internal/xdsclient/tests


atollena:
https://github.com/grpc/grpc-go/actions/runs/9711071699/job/26803095072?pr=7364

```
--- FAIL: Test (14.05s)
    --- FAIL: Test/AuthorityRevive (0.00s)
        authority_test.go:76: Created new snapshot cache...
        tlogger.go:116: INFO server.go:685 [core] [Server #518]Server created  (t=+259.644µs)
        authority_test.go:76: Registered Aggregated Discovery Service (ADS)...
        authority_test.go:76: xDS management server serving at: 127.0.0.1:44757...
        authority_test.go:79: Created new snapshot cache...
        tlogger.go:116: INFO server.go:685 [core] [Server #519]Server created  (t=+316.09µs)
        authority_test.go:79: Registered Aggregated Discovery Service (ADS)...
        authority_test.go:79: xDS management server serving at: 127.0.0.1:41611...
        tlogger.go:116: INFO server.go:881 [core] [Server #518 ListenSocket #520]ListenSocket created  (t=+404.195µs)
        tlogger.go:116: INFO server.go:881 [core] [Server #519 ListenSocket #521]ListenSocket created  (t=+480.408µs)
        tlogger.go:116: INFO bootstrap.go:571 [xds] [xds-bootstrap] Bootstrap config for creating xds-client: &{xDSServers:[0xc00076abd0] cpcs:map[] serverListenerResourceNameTemplate: clientDefaultListenerResourceNameTemplate:%s authorities:map[test-authority1:0xc00076d830 test-authority2:0xc00076d860 test-authority3:0xc00076d890] node:{ID:4974a10e-6dfc-42ee-b520-e37a6fc2948f Cluster: Locality:{Region: Zone: SubZone:} Metadata:<nil> userAgentName:gRPC Go userAgentVersionType:{UserAgentVersion:1.66.0-dev} clientFeatures:[envoy.lb.does_not_support_overprovisioning xds.config.resource-in-sotw]} certProviderConfigs:map[]}  (t=+538.716µs)
        tlogger.go:116: INFO client_new.go:87 [xds] [xds-client 0xc0004ea0a0] Created client to xDS management server: 127.0.0.1:44757-insecure-  (t=+573.17µs)
        server.go:229: Created new resource snapshot...
        server.go:235: Updated snapshot cache with resource snapshot...
[...]

Issue #6914, opened 2024-01-10
Flaky test package: test/xds


zasweq:
Alongside #6913 and #6912, I have ran the test/xds suite on master since I added tests to it for my xDS Server fix #6889. I have encountered numerous flakes on g3, particularly those outlined in custom lb tests for distribution #6601. However, I have encountered almost every client and server side xDS test flake with a context timeout for a RPC expected to proceed. Each has different logs/events preceeding it's timeout, but every test seems susceptible to timeout. The flakes are generally rare, but due to the number of tests in the test suite you can successfully trigger by running the full test suite enough times. My initial inkling tells me there's some synchronization needed or something gets stuck in the management server/testing xDS Client flow. This also manifests in rare flakes for my xDS Server fix, where I expect something like an err that represents Accept and Close, and I get a context timeout instead.

arvindbr8:
another one: https://github.com/grpc/grpc-go/actions/runs/7480796959/job/20361025267?pr=6916

zasweq:
https://github.com/grpc/grpc-go/actions/runs/7618526481/job/20749870802?pr=6933

arvindbr8:
https://github.com/grpc/grpc-go/actions/runs/7716359885/job/21032969470?pr=6949

arvindbr8:
https://github.com/grpc/grpc-go/actions/runs/7791646589/job/21248143511?pr=6965

zasweq:
https://github.com/grpc/grpc-go/actions/runs/8546803299/job/23417838293?pr=7085

arjan-bal:
https://github.com/grpc/grpc-go/actions/runs/9790522733/job/27032314481?pr=7390

arjan-bal:
@zasweq I investigated this and the problem seems to be due to the xDS management server getting stuck while writing to this buffered channel
https://github.com/grpc/grpc-go/blob/d27ddb5eb5940c949f88bc2cb21eed9254f8be75/test/xds/xds_server_certificate_providers_test.go#L249

In the logs of failing runs for `TestServerSideXDS_WithValidAndInvalidSecurityConfiguration`, I noticed that the resource snapshot update request is sent to the xds management server before the xds client is able to connect to the xds server. This somehow results in more than 1 Listener requests being sent to the xds server which get stuck waiting to write to the buffered channel.

[...]

PR #7390, opened 2024-07-04
grpc: hold ac.mu while calling resetTransport to prevent concurrent connection attempts

arjan-bal:
This change ensures that caller of `resetTransport()` keep holding the mutex instead of releasing it and having `resetTransport()` re-acquire it. This ensures that no concurrent requests are able to start once caller of `resetTransport` does some validation and calls `resetTransport`.

See https://github.com/grpc/grpc-go/issues/7365#issuecomment-2208258873 for more details.

## Tested
Verified that Test/AuthorityRevive no longer flakes for 100000 attempts with the change.

Fixes: https://github.com/grpc/grpc-go/issues/7365

RELEASE NOTES:
- client: fix race that could lead to orphaned connections and associated resources.

### DIFF
--- a/clientconn.go
+++ b/clientconn.go
@@ -1228,6 +1228,9 @@ func (ac *addrConn) adjustParams(r transport.GoAwayReason) {
 	}
 }
 
+// resetTransportAndUnlock unconditionally connects the addrConn.
+//
+// ac.mu must be held by the caller, and this function will guarantee it is released.
 func (ac *addrConn) resetTransportAndUnlock() {
 	acCtx := ac.ctx
 	if acCtx.Err() != nil {
--- a/clientconn.go
+++ b/clientconn.go
@@ -918,6 +918,9 @@ func (ac *addrConn) connect() error {
 		ac.mu.Unlock()
 		return nil
 	}
+	// Update the state to ensure no concurrent requests start once the lock
+	// is released.
+	ac.updateConnectivityState(connectivity.Connecting, nil)
 	ac.mu.Unlock()
 
 	ac.resetTransport()
--- a/clientconn.go
+++ b/clientconn.go
@@ -918,12 +918,8 @@ func (ac *addrConn) connect() error {
 		ac.mu.Unlock()
 		return nil
 	}
-	// Update the state to ensure no concurrent requests start once the lock
-	// is released.
-	ac.updateConnectivityState(connectivity.Connecting, nil)
-	ac.mu.Unlock()
 
-	ac.resetTransport()
+	ac.resetTransportAndUnlock()
 	return nil
 }
 
@@ -995,11 +991,9 @@ func (ac *addrConn) updateAddrs(addrs []resolver.Address) {
 		ac.updateConnectivityState(connectivity.Idle, nil)
 	}
 
-	ac.mu.Unlock()
-
 	// Since we were connecting/connected, we should start a new connection
 	// attempt.
-	go ac.resetTransport()
+	go ac.resetTransportAndUnlock()
 }
 
 // getServerName determines the serverName to be used in the connection
@@ -1234,8 +1228,7 @@ func (ac *addrConn) adjustParams(r transport.GoAwayReason) {
 	}
 }
 
-func (ac *addrConn) resetTransport() {
-	ac.mu.Lock()
+func (ac *addrConn) resetTransportAndUnlock() {
 	acCtx := ac.ctx
 	if acCtx.Err() != nil {
 		ac.mu.Unlock()


clientconn.go
```diff
@@ -918,6 +918,9 @@ func (ac *addrConn) connect() error {
 		ac.mu.Unlock()
 		return nil
 	}
+	// Update the state to ensure no concurrent requests start once the lock
+	// is released.
+	ac.updateConnectivityState(connectivity.Connecting, nil)
```
purnesh42H, line 923: We should probably file a bug for this otherwise it will be a behavior change?

arjan-bal: Can you look at https://github.com/grpc/grpc-go/issues/7365#issuecomment-2208258873 for the details of this bug?

What change in behaviour are you concerned about?

purnesh42H: Oh I was just asking with respect to release notes because currently the bug points to test flake. Anyways, I just checked it doesn't matter because release notes refer to the fix PR and not the issue. Although in the release notes, we should prefix the package

`balancer: Fix race condition that could lead to multiple transports being created in parallel`

purnesh42H: Also, it looks like without your fix, there is a case where resetTransport can error out and return without updating the connectivity state https://github.com/grpc/grpc-go/blob/bdd707e642e40cf75db5ac3f0f6af48077f48368/clientconn.go#L1237

May be we can make the resetTransport() in the same critical section instead of releasing lock and aquiring again in resetTransport()?

arjan-bal: I don't see any benefit in adding the `acCtcx.Err` check in `connect` because even resetTransport sets the the state to `Connecting` and releases the lock:
https://github.com/grpc/grpc-go/blob/bdd707e642e40cf75db5ac3f0f6af48077f48368/clientconn.go#L1262-L1263

This means that the context can be cancelled (and subsequently addrConn shutdown) after the channel is in `connecting` state even without the change.

IIUC we just need to ensure that we don't set connecting state after the channel enters shutdown.

The test for shutdown state on top should be enough protection to ensure shutdown state comes only after we enter `connecting`.

arjan-bal: > May be we can make the resetTransport() in the same critical section instead of releasing lock and aquiring again in resetTransport()?

We could rename `resetTransport` to `resetTransportLocked` and expect the callers to hold the lock while calling this method. However, `resetTransport` releases the lock temporarily. Add to this that ac.updateAddrs calles `resetTransport` in a new go routine, it feels a little risky to make that change. I don't know for sure, but I feel we could end up in a situation when the lock is not released correctly and end in a deadlock. 

I don't want do make that change as the first option.

purnesh42H: > I don't see any benefit in adding the acCtcx.Err check in connect because even resetTransport sets the the state to Connecting and releases the lock

From the code it looks like in case of `acCtcx.Err`, resetTransport() doesn't update the state and return so state will be still idle but after your fix in case of `acCtcx.Err` state will be updated to `connecting`. Am I missing something?

purnesh42H:
Release notes needs to be prefixed with "<pacakge name>:" in this case `balancer:`?

clientconn.go
arjan-bal: From my understanding, `ac.ctx` is used to control the creation of remote connections while `ac.state` is used to synchronize all the state transitions for the `addrConn`. `ac.connect()` doesn't deal with creating remote connections, so it doesn't need to check `ac.ctx.Err()`. It needs to ensure the transition to `Connecting` is valid, which it does by locking the mutex and verifying that `ac.state != Shutdown`

`ac.ctx` is used to avoid doing throw away work which takes significant time (creating a remote conn).

Please let me know if your understanding is different.

arjan-bal:
> Release notes needs to be prefixed with ":" in this case `balancer:`?

This is a change in the outermost `grpc` package. Added the package name in the release notes.

purnesh42H:
is `Test/ServerSideXDS_WithValidAndInvalidSecurityConfiguration` failure related to change?

purnesh42H:
is `Test/ServerSideXDS_WithValidAndInvalidSecurityConfiguration` failure related to change or just a flake?

arjan-bal:
> is `Test/ServerSideXDS_WithValidAndInvalidSecurityConfiguration` failure related to change or just a flake?

It's a flake, there's an existing issue for this and I've commented on the issue about this failure: https://github.com/grpc/grpc-go/issues/6914

clientconn.go
purnesh42H: Discussed offline: it doesn't matter if restTransport() returns error after state being updated to `connecting`

purnesh42H (approved):
lgtm

dfawley:
> RELEASE NOTES:
> * Fix race condition that could lead to multiple transports being created in parallel.

What is the user-visible symptom here?  A memory leak?  Or just an extra connection that was attempted, but that will quickly go away on its own anyway?  (Or will the extra connection stick around until the channel is closed?)

clientconn.go
```diff
@@ -1231,8 +1228,7 @@ func (ac *addrConn) adjustParams(r transport.GoAwayReason) {
 	}
 }
 
-func (ac *addrConn) resetTransport() {
-	ac.mu.Lock()
+func (ac *addrConn) resetTransportAndUnlock() {
```
dfawley, line 1231: Please add a short comment here:
```go
// resetTransportAndUnlock unconditionally connects the addrConn.
//
// ac.mu must be held by the caller, and this function will guarantee it is released.
```

dfawley (approved):

clientconn.go
purnesh42H: > ac.mu must be held by the caller, and this function will guarantee it is released

should we have code check for this as well?

arjan-bal: Add the doc comment.

arjan-bal: > should we have code check for this as well?
We don't verify if the mutex is locked in other functions in the code which assume the caller has the lock. These functions have the suffix `Locked` in the name. 
* There is a `TryLock()` method on `sync/mutex`, but its use is discouraged. 
* This is a private method and we run tests with race detector so we can catch incorrect usage.
* When the `resetTransportAndUnlock` method tries to unlock a mutex that isn't locked, it will panic.

purnesh42H: i meant just having a doc doesn't enforce the mutex is locked. Unlocking a mutex that is not locked in Go will result in a runtime panic.

arjan-bal:
> What is the user-visible symptom here? A memory leak? Or just an extra connection that was attempted, but that will quickly go away on its own anyway? (Or will the extra connection stick around until the channel is closed?)

What I'm seeing in the test is that only one transport is closed when the subConn is updated (`ac.transport`) and the other transports are orphaned. The orphaned transports get closed when the server is shutdown at the end of the test. Updated the release notes.

clientconn.go
arjan-bal: Can you suggest how to enforce the locking? 
How is the caller expected to handle the error if `resetTransportAndUnlock` doesn't panic?
Since its a private method, doesn't a panic make the failure more visible and ensure incorrect usages are caught by tests?

purnesh42H: yeah it will probably require to implement custom mutex with some boolean field but since we have precedences of methods (e.g https://github.com/grpc/grpc-go/blob/master/clientconn.go#L728) unlocking mutex, devs will be aware of these

dfawley: I think the name of the function and the comment should be sufficient for this.

PR #7391, opened 2024-07-04
refactor:rename clientside dialoption field 'maxAttempts'

No-SilverBullet:
RELEASE NOTES:
- Renamed the variable `maxCallAttempts` to `maxRetryAttempts` to improve clarity and better represent its purpose.

Recently, I noticed a bug with the RPC client's setting of the maximum number of reconnections. I was very surprised to find that it has been solved in this PR [client: implement maxAttempts for retryPolicy](https://github.com/grpc/grpc-go/pull/7229) .  However, I think the meaning of this variable 'maxCallAttempts' are not clear enough, so I submitted this PR to rename this variable.

### DIFF
--- a/clientconn.go
+++ b/clientconn.go
@@ -173,7 +173,7 @@ func NewClient(target string, opts ...DialOption) (conn *ClientConn, err error) 
 	}
 
 	if cc.dopts.defaultServiceConfigRawJSON != nil {
-		scpr := parseServiceConfig(*cc.dopts.defaultServiceConfigRawJSON, cc.dopts.maxCallAttempts)
+		scpr := parseServiceConfig(*cc.dopts.defaultServiceConfigRawJSON, cc.dopts.maxRetryAttempts)
 		if scpr.Err != nil {
 			return nil, fmt.Errorf("%s: %v", invalidDefaultServiceConfigErrPrefix, scpr.Err)
 		}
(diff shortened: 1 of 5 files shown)


arjan-bal:
Hi @No-SilverBullet, I agree that the name could be better. The API that users interact with is the `WithMaxCallAttempt` dialoption.
https://github.com/grpc/grpc-go/blob/bdd707e642e40cf75db5ac3f0f6af48077f48368/dialoptions.go#L752-L759

If we were to change the name of `WithMaxCallAttempt`, it would be a breaking change. To handle it gracefully, we would need to create a dial option with the new name, deprecate the old one and then completely remove the old one in a future release. This doesn't seem to be worth the effort so I'm closing this PR.

No-SilverBullet:
> Hi @No-SilverBullet, I agree that the name could be better. The API that users interact with is the `WithMaxCallAttempt` dialoption.
> 
> https://github.com/grpc/grpc-go/blob/bdd707e642e40cf75db5ac3f0f6af48077f48368/dialoptions.go#L752-L759
> 
> If we were to change the name of `WithMaxCallAttempt`, it would be a breaking change. To handle it gracefully, we would need to create a dial option with the new name, deprecate the old one and then completely remove the old one in a future release. This doesn't seem to be worth the effort so I'm closing this PR.
> 
> Since the external method is named as such, IMO keeping the private variable names consistent with it seems to desired bahaviour.

Got it! Thanks for reply

PR #7396, opened 2024-07-08
test/channelz: change channelz_test to use write data

printchard:
Remove the call to `WriteRawFrame` in order to remove that method from the new Framer API.

This particular test was calling `WriteRawFrame` to write a Data Frame, which could be replaced by `WriteData` method provided by the x/net/http2 framer.

RELEASE NOTES: none


arvindbr8 (approved):
LGTM! Congrats on the first PR.  cc'ing @dfawley to take another quick pass since he is most familiar with the channelz tests.

dfawley (approved):

PR #7409, opened 2024-07-12
scripts: regenerate pbs with caching deps to a fixed tmp folder

### LINKED ISSUE #7400: scripts: fixes/improvements to regenerate.sh / install_protoc.sh
There's some weird interactions going on, and some things I might like to see changed:

- `install_protoc.sh` checks the protoc version in your path before downloading, and refuses to download if your path's protoc is a different version than you'd like to install
- A function called `download_binary` shouldn't do other checks; those checks should be moved elsewhere.
- This line makes no sense: https://github.com/grpc/grpc-go/blob/45d44a736ec6cdcb73f9411bf6a1c2d1abea1956/scripts/install_protoc.sh#L63
- This line needs quotes: https://github.com/grpc/grpc-go/blob/45d44a736ec6cdcb73f9411bf6a1c2d1abea1956/scripts/regenerate.sh#L54
- This line probably needs `export` in case a `GOBIN` isn't already exported: https://github.com/grpc/grpc-go/blob/45d44a736ec6cdcb73f9411bf6a1c2d1abea1956/scripts/regenerate.sh#L27 (`go install` needs to be able to read it.)
- `regenerate.sh` calls `install_protoc.sh` after several other things have been done.
- `regenerate.sh` apparently tries to "save your `$PATH`", but since it shouldn't be executed via ` source` there's no need to do any of this.
- Super nit-picky: `install_protoc.sh` sources `vet_common.sh` but is not a `vet*` script.  Maybe rename `vet_common.sh` to `common.sh`?

IMO one of three models for these scripts seems pretty reasonable:

1. Always download everything into a temp directory.
  - Downside: needs to download/recompile every time regenerate is run
2. Check the versions in your path and print help/commands to update them to the needed versions.
  - Downside: it's hard to know where the user would want protoc to live.
3. Carve out a directory like `$HOME/grpc-go-tools` or `$HOME/tmp/grpc-go-tools` or something and download the tools there if the versions don't match.
  - Downside: you're potentially polluting the user's directory.

Of these, I think I prefer (2) but am not strongly opposed to any of them.  It looks like the scripts were already trying to implement (1), which is okay but is my least favorite option.

cc @arvindbr8 @aranjans

arvindbr8:
Fixes #7400 

@dfawley I went with something that's more along the lines of option no.3 in the issue.  Now the deps are installed in a 
"well-known" directory which is `/tmp/grpc-go-tools`.  This will ensure not requiring to re-download all deps when running regenerate.

Changes: 

install-protoc.sh
1. Now requires an INSTALL_PATH to be specified
	- This is a behavior change.  Previously INSTALL_PATH was optional and would put it in your GOBIN/GOPATH.  We dont want to be polluting the users GO dir with non-golang binaries.
2. Now checks if `protoc` in the required version is present in INSTALL_PATH and not $PATH

regenerate.sh
1. Installs everything to `/tmp/grpc-go-tools` instead of mktmp
2. Only download deps if not present
3. Other shell linter changes

cc: @aranjans 

RELEASE NOTES: none

### DIFF
--- a/scripts/install-protoc.sh
+++ b/scripts/install-protoc.sh
@@ -65,9 +65,19 @@ download_binary() {
   fi
 
   DOWNLOAD_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-${OS}-${ARCH}.zip"
-  # For cURL, -L follows redirects. 
-  # For tar, --exclude excludes files, and -C specifies the directory. 
-  curl -L "${DOWNLOAD_URL}" | tar -xz --exclude="readme.txt" -C "${INSTALL_PATH}"
+  # -L follows redirects.
+  # -O writes output to a file.
+  curl -LO "${DOWNLOAD_URL}"
+  # Unzip the downloaded file and except readme.txt.
+  # The file structure should look like:
+  # INSTALL_PATH
+  # ├── bin
+  # │   └── protoc
+  # └── include
+  #     └── other files...
+  unzip "protoc-${PROTOC_VERSION}-$2-$1.zip" -d "${INSTALL_PATH}" -x "readme.txt"
+  rm "protoc-${PROTOC_VERSION}-$2-$1.zip"
+  # Make the protoc binary executable. ¯\_(ツ)_/¯  crazy, right?
   chmod +x "${INSTALL_PATH}/bin/protoc"
 }
 
--- a/scripts/install-protoc.sh
+++ b/scripts/install-protoc.sh
@@ -39,13 +39,13 @@ main() {
   if [[ "$#" -ne 1 ]]; then
     die "Usage: $0 INSTALL_PATH"
   fi
-  
+
   INSTALL_PATH="${1}"
 
   if [[ ! -d "${INSTALL_PATH}" ]]; then
     die "INSTALL_PATH (${INSTALL_PATH}) does not exist."
   fi
-  
+
   echo "Installing protoc version $PROTOC_VERSION to ${INSTALL_PATH}..."
 
   # Detect the hardware platform.
@@ -55,7 +55,7 @@ main() {
     "arm64")    ARCH="aarch_64";;
     *)          die "Install unsuccessful. Hardware platform not supported by installer: $1";;
   esac
-  
+
   # Detect the Operating System.
   case "$(uname -s)" in
     "Darwin") OS="osx";;
@@ -72,11 +72,11 @@ main() {
   fi
 
   DOWNLOAD_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-${OS}-${ARCH}.zip"
-  
+
   # -L follows redirects.
   # -O writes output to a file.
   curl -LO "${DOWNLOAD_URL}"
-  
+
   # Unzip the downloaded file and except readme.txt.
   # The file structure should look like:
   # INSTALL_PATH
@@ -86,7 +86,7 @@ main() {
   #     └── other files...
   unzip "protoc-${PROTOC_VERSION}-${OS}-${ARCH}.zip" -d "${INSTALL_PATH}" -x "readme.txt"
   rm "protoc-${PROTOC_VERSION}-${OS}-${ARCH}.zip"
-  
+
   # Make the protoc binary executable. ¯\_(ツ)_/¯  crazy, right?
   chmod +x "${INSTALL_PATH}/bin/protoc"
 }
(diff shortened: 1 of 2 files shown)


scripts/install-protoc.sh
```diff
@@ -39,53 +35,66 @@ source "$(dirname $0)/common.sh"
...
-          "https://github.com/protocolbuffers/protobuf/releases/ and add to PATH" ;;
-  esac
+  if [[ ! -d "${INSTALL_PATH}" ]]; then
+    die "INSTALL_PATH (${INSTALL_PATH}) does not exist."
+  fi
+  echo "Installing protoc version $PROTOC_VERSION to ${INSTALL_PATH}."
+  # Detect the hardware platform.
+  ARCH="$(uname -m)"
+  # Detect the Operating System.
+  OS="$(uname -s)"
```
dfawley, line 96: I'd probably delete these vars and do these things in `download_binary` instead.  Note that you're setting `ARCH` twice to different things, so that's a little confusing, and you're passing these as positional parameters, so separating their usage from their definition.

scripts/regenerate.sh
```diff
@@ -15,14 +16,10 @@
...
-
-function finish {
-  rm -rf "$WORKDIR"
-}
-trap finish EXIT
+WORKDIR="/tmp/grpc-go-tools"
+mkdir -p "${WORKDIR}"
 
-./scripts/install-protoc.sh "${WORKDIR}"
+"./$(dirname "${0}")/install-protoc.sh" ${WORKDIR}
```
dfawley, line 22: I think you want to remove `./` at the start?  Try executing as `/abs/path/to/scripts/regenerate.sh` and I think it will fail?

scripts/regenerate.sh
```diff
@@ -31,7 +28,7 @@ mkdir -p "${GOBIN}"
 echo "remove existing generated files"
```
dfawley, line 28: Nit: `"Removing existing generated files."`

scripts/regenerate.sh
```diff
@@ -31,7 +28,7 @@ mkdir -p "${GOBIN}"
...
 # grpc_testing_not_regenerate/*.pb.go is not re-generated,
 # see grpc_testing_not_regenerate/README.md for details.
-rm -f $(find . -name '*.pb.go' | grep -v 'grpc_testing_not_regenerate')
+find . -name '*.pb.go' | grep -v 'grpc_testing_not_regenerate' | xargs rm -f || true
 
 echo "go install google.golang.org/protobuf/cmd/protoc-gen-go"
```
dfawley, line 33: Maybe `"Executing: go install _______"` (etc)?

scripts/regenerate.sh
```diff
@@ -40,10 +37,18 @@ echo "go install cmd/protoc-gen-go-grpc"
 (cd cmd/protoc-gen-go-grpc && go install .)
 
 echo "git clone https://github.com/grpc/grpc-proto"
-git clone --quiet https://github.com/grpc/grpc-proto "${WORKDIR}/grpc-proto"
+if [ -d "${WORKDIR}/grpc-proto" ]; then
+  (cd "${WORKDIR}/grpc-proto" && git pull)
```
dfawley, line 41: Why parens here (and below)?

dfawley (approved):
LGTM, just some nits.

scripts/install-protoc.sh
arvindbr8: Yes you are right. I dont know why we needed a function to be defined like this. I've moved everything into main() now. Looks cleaner this way.

scripts/regenerate.sh
arvindbr8: Yea, for whatever reason i've always run executables with a leading `./`. Every time I do a shell PR I learn something new.

I also did not want to put the `abs/path` since that would require that you run the script from the `gRPC-go/` dir

scripts/regenerate.sh
arvindbr8: Done

scripts/regenerate.sh
arvindbr8: Done. Let me know how you feel about the new echo-s

scripts/regenerate.sh
arvindbr8: so that I dont have to `cd -` after the command :)

scripts/regenerate.sh
dfawley: > for whatever reason i've always run executables with a leading `./`.

If you're running one in the current directory, then that's important.  If it's in an absolute path, you'd use `/usr/bin/ls`, e.g.  If it's in a relative path, then either `./bin/foo` or `bin/foo` work.

scripts/regenerate.sh
dfawley: Fair enough

dfawley (approved):

PR #7410, opened 2024-07-13
stats/opentelemetry: Add usage of metrics registry

zasweq:
This PR adds the OpenTelemetry usage of the metrics registry. It creates OpenTelemetry instruments and now implements the MetricsRecorder interface to record on those instruments.

RELEASE NOTES: N/A

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -215,6 +215,12 @@ var (
 
 	// SetConnectedAddress sets the connected address for a SubConnState.
 	SetConnectedAddress any // func(scs *SubConnState, addr resolver.Address)
+
+	// SnapshotMetricsRegistryForTesting snapshots the global data of the
+	// metrics registry. Registers a cleanup function on the provided testing.T
+	// that sets the metrics registry to its original state. Only called in
+	// testing functions.
+	SnapshotMetricRegistryForTesting any // func(t *testing.T)
 )
 
 // HealthChecker defines the signature of the client-side LB channel health
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -216,10 +216,10 @@ var (
 	// SetConnectedAddress sets the connected address for a SubConnState.
 	SetConnectedAddress any // func(scs *SubConnState, addr resolver.Address)
 
-	// SnapshotMetricsRegistryForTesting snapshots the global data of the
-	// metrics registry. Registers a cleanup function on the provided testing.T
-	// that sets the metrics registry to its original state. Only called in
-	// testing functions.
+	// SnapshotMetricRegistryForTesting snapshots the global data of the metric
+	// registry. Registers a cleanup function on the provided testing.T that
+	// sets the metric registry to its original state. Only called in testing
+	// functions.
 	SnapshotMetricRegistryForTesting any // func(t *testing.T)
 )
 
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -33,9 +33,8 @@ import (
 )
 
 type clientStatsHandler struct {
-	registryMetrics
-	options Options
-
+	estats.MetricsRecorder
+	options       Options
 	clientMetrics clientMetrics
 }
 
@@ -62,37 +61,11 @@ func (h *clientStatsHandler) initializeMetrics() {
 	h.clientMetrics.attemptRcvdTotalCompressedMessageSize = createInt64Histogram(metrics.Metrics(), "grpc.client.attempt.rcvd_total_compressed_message_size", meter, otelmetric.WithUnit("By"), otelmetric.WithDescription("Compressed message bytes received per call attempt."), otelmetric.WithExplicitBucketBoundaries(DefaultSizeBounds...))
 	h.clientMetrics.callDuration = createFloat64Histogram(metrics.Metrics(), "grpc.client.call.duration", meter, otelmetric.WithUnit("s"), otelmetric.WithDescription("Time taken by gRPC to complete an RPC from application's perspective."), otelmetric.WithExplicitBucketBoundaries(DefaultLatencyBounds...))
 
-	h.registryMetrics = registryMetrics{
-		intCounts:   make(map[*estats.MetricDescriptor]otelmetric.Int64Counter),
-		floatCounts: make(map[*estats.MetricDescriptor]otelmetric.Float64Counter),
-		intHistos:   make(map[*estats.MetricDescriptor]otelmetric.Int64Histogram),
-		floatHistos: make(map[*estats.MetricDescriptor]otelmetric.Float64Histogram),
-		intGauges:   make(map[*estats.MetricDescriptor]otelmetric.Int64Gauge),
-
+	rm := &registryMetrics{
 		optionalLabels: h.options.MetricsOptions.OptionalLabels,
 	}
-
-	for metric := range metrics.Metrics() {
-		desc := estats.DescriptorForMetric(metric)
-		if desc == nil {
-			// Either the metric was per call or the metric is not registered.
-			// Thus, if this component ever receives the desc as a handle in
-			// record it will be a no-op.
-			continue
-		}
-		switch desc.Type {
-		case estats.MetricTypeIntCount:
-			h.registryMetrics.intCounts[desc] = createInt64Counter(metrics.Metrics(), desc.Name, meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
-		case estats.MetricTypeFloatCount:
-			h.registryMetrics.floatCounts[desc] = createFloat64Counter(metrics.Metrics(), desc.Name, meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
-		case estats.MetricTypeIntHisto:
-			h.registryMetrics.intHistos[desc] = createInt64Histogram(metrics.Metrics(), desc.Name, meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description), otelmetric.WithExplicitBucketBoundaries(desc.Bounds...))
-		case estats.MetricTypeFloatHisto:
-			h.registryMetrics.floatHistos[desc] = createFloat64Histogram(metrics.Metrics(), desc.Name, meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description), otelmetric.WithExplicitBucketBoundaries(desc.Bounds...))
-		case estats.MetricTypeIntGauge:
-			h.registryMetrics.intGauges[desc] = createInt64Gauge(metrics.Metrics(), desc.Name, meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
-		}
-	}
+	h.MetricsRecorder = rm
+	rm.registerMetrics(metrics, meter)
 }
 
 func (h *clientStatsHandler) unaryInterceptor(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -30,6 +30,7 @@ import (
 
 	otelattribute "go.opentelemetry.io/otel/attribute"
 	otelmetric "go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
 )
 
 type clientStatsHandler struct {
@@ -52,7 +53,7 @@ func (h *clientStatsHandler) initializeMetrics() {
 
 	metrics := h.options.MetricsOptions.Metrics
 	if metrics == nil {
-		metrics = DefaultMetrics
+		metrics = DefaultMetrics()
 	}
 
 	h.clientMetrics.attemptStarted = createInt64Counter(metrics.Metrics(), "grpc.client.attempt.started", meter, otelmetric.WithUnit("attempt"), otelmetric.WithDescription("Number of client call attempts started."))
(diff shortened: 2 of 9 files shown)


stats/opentelemetry/opentelemetry.go
```diff
@@ -184,65 +194,123 @@ type attemptInfo struct {
...
+	}
+	ret, err := meter.Int64Gauge(string(metricName), options...)
+	if err != nil {
+		logger.Errorf("failed to register metric \"%v\", will not record: %v", metricName, err)
+		return noop.Int64Gauge{}
+	}
+	return ret
+}
+
+func createAttributeOptionFromLabels(labelKeys []string, optionalLabelKeys []string, optionalLabels []string, labelVals ...string) otelmetric.MeasurementOption {
```
dfawley, line 295: `optionFromLables` maybe to save some columns?

stats/opentelemetry/client_metrics.go
```diff
@@ -52,14 +53,119 @@ func (h *clientStatsHandler) initializeMetrics() {
...
+	}
+}
+
+func (h *clientStatsHandler) RecordInt64Count(handle *estats.Int64CountHandle, incr int64, labels ...string) {
+	desc := (*estats.MetricDescriptor)(handle)
+	ao := createAttributeOptionFromLabels(desc.Labels, desc.OptionalLabels, h.options.MetricsOptions.OptionalLabels, labels...)
+	h.getInt64Counter(desc).Add(h.options.MetricsOptions.Context, incr, ao)
+}
+
+func (h *clientStatsHandler) getInt64Counter(desc *estats.MetricDescriptor) otelmetric.Int64Counter {
```
dfawley, line 112: s/get// throughout probably?

https://google.github.io/styleguide/go/decisions#getters

But.. this function is only ever called once, and probalby only ever will be called once?  Maybe we should just inline it and delete the `noop` stuff?

stats/opentelemetry/opentelemetry.go
```diff
@@ -96,6 +100,9 @@ type MetricsOptions struct {
 // configured for an individual metric turned on, the API call in this component
 // will create a default view for that metric.
 func DialOption(o Options) grpc.DialOption {
+	if o.MetricsOptions.Context == nil {
+		o.MetricsOptions.Context = context.Background()
+	}
```
dfawley, line 105: Shouldn't this be passed in through the first function parameter, and be mandatory, instead of through a struct?

Or maybe we should replace the places where we access this with a `context.TODO` since we don't really know how it should work?  Or even invent our own `emptyContext` or something that implements `context.Context` but is effectively the same thing to avoid linters, since maybe we're actually sure it's impossible to ever pass a context here?

stats/opentelemetry/opentelemetry.go
```diff
@@ -184,65 +194,123 @@ type attemptInfo struct {
...
+	attemptStarted otelmetric.Int64Counter
 	// "grpc.client.attempt.duration"
-	attemptDuration metric.Float64Histogram
+	attemptDuration otelmetric.Float64Histogram
 	// "grpc.client.attempt.sent_total_compressed_message_size"
-	attemptSentTotalCompressedMessageSize metric.Int64Histogram
+	attemptSentTotalCompressedMessageSize otelmetric.Int64Histogram
 	// "grpc.client.attempt.rcvd_total_compressed_message_size"
-	attemptRcvdTotalCompressedMessageSize metric.Int64Histogram
+	attemptRcvdTotalCompressedMessageSize otelmetric.Int64Histogram
```
dfawley, line 203: I know we talked about it already, but...can these be moved into the maps, too?

stats/opentelemetry/opentelemetry.go
```diff
@@ -253,6 +321,14 @@ var (
...
 	// DefaultSizeBounds are the default bounds for metrics which record size.
 	DefaultSizeBounds = []float64{0, 1024, 2048, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864, 268435456, 1073741824, 4294967296}
-	// DefaultMetrics are the default metrics provided by this module.
-	DefaultMetrics = estats.NewMetrics(ClientAttemptStarted, ClientAttemptDuration, ClientAttemptSentCompressedTotalMessageSize, ClientAttemptRcvdCompressedTotalMessageSize, ClientCallDuration, ServerCallStarted, ServerCallSentCompressedTotalMessageSize, ServerCallRcvdCompressedTotalMessageSize, ServerCallDuration)
+	// defaultPerCallMetrics are the default metrics provided by this module.
+	defaultPerCallMetrics = estats.NewMetrics(ClientAttemptStarted, ClientAttemptDuration, ClientAttemptSentCompressedTotalMessageSize, ClientAttemptRcvdCompressedTotalMessageSize, ClientCallDuration, ServerCallStarted, ServerCallSentCompressedTotalMessageSize, ServerCallRcvdCompressedTotalMessageSize, ServerCallDuration)
 )
+
+// DefaultMetrics returns a set of default metrics for per call joined with any
+// default metrics registered through metrics registry.
```
dfawley, line 329: Is a user going to understand what it means to be "metrics for per call joined with ... metrics registered through metrics registry"?  If I'm a user, I'm probably thinking "what's a 'metrics registry'?"  How about just "... returns the default open telemetry metrics"?

stats/opentelemetry/server_metrics.go
```diff
@@ -51,13 +52,119 @@ func (h *serverStatsHandler) initializeMetrics() {
...
+			fh := createFloat64Histogram(metrics.Metrics(), desc.Name, meter, opts...)
+			h.serverMetrics.floatHistos[desc] = fh
+		case estats.MetricTypeIntGauge:
+			ig := createInt64Gauge(metrics.Metrics(), estats.Metric(desc.Name), meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
+			h.serverMetrics.intGauges[desc] = ig
+		}
+	}
+}
+
+func (h *serverStatsHandler) RecordInt64Count(handle *estats.Int64CountHandle, incr int64, labels ...string) {
```
dfawley, line 105: These are all identical as the client.  Can we wrap them in another struct that encapsulates all this behavior?

stats/opentelemetry/client_metrics.go
```diff
@@ -52,14 +53,119 @@ func (h *clientStatsHandler) initializeMetrics() {
...
+			ic := createInt64Counter(metrics.Metrics(), estats.Metric(desc.Name), meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
+			h.clientMetrics.intCounts[desc] = ic
+		case estats.MetricTypeFloatCount:
+			fc := createFloat64Counter(metrics.Metrics(), estats.Metric(desc.Name), meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
+			h.clientMetrics.floatCounts[desc] = fc
+		case estats.MetricTypeIntHisto:
+			opts := []otelmetric.Int64HistogramOption{otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description)}
+			if len(desc.Bounds) != 0 {
+				opts = append(opts, otelmetric.WithExplicitBucketBoundaries(desc.Bounds...))
+			}
```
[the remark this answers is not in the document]
zasweq: Yeah I thought about this, the intended behavior is to fallback to default OTel bounds. So if you set empty (calling that option with no actual bounds) my gut told me it wouldn't pick up the defaults. But I didn't actually triage this, let me triage now.

stats/opentelemetry/client_metrics.go
```diff
@@ -52,14 +53,119 @@ func (h *clientStatsHandler) initializeMetrics() {
...
+	for metric := range metrics.Metrics() {
+		desc := estats.DescriptorForMetric(metric)
+		// Per call or metrics not registered, when gets a handle corresponding
+		// it will be a no-op record.
+		if desc == nil {
+			continue
+		}
+		switch desc.Type {
+		case estats.MetricTypeIntCount:
+			ic := createInt64Counter(metrics.Metrics(), estats.Metric(desc.Name), meter, otelmetric.WithUnit(desc.Unit), otelmetric.WithDescription(desc.Description))
```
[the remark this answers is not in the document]
zasweq: Done (this moved to the new embedded type).

stats/opentelemetry/client_metrics.go
zasweq: Deleted these getters, inlined and got rid of no-op.

stats/opentelemetry/opentelemetry.go
zasweq: Discussed offline; decided context.TODO and ask OpenTelemetry team the technical reasons their records on instruments take a context (which is ignored) whereas no context in Java and C.

stats/opentelemetry/opentelemetry.go
zasweq: Discussed offline; decided not worth the extra complexity.

stats/opentelemetry/opentelemetry.go
zasweq: Done - optionFromLabels :).

stats/opentelemetry/opentelemetry.go
zasweq: The metrics registry metrics aren't nesscarily opentelemetry, but I guess that doesn't make the rewritten comment incorrect, the generic defaults happen to be used as default opentelemetry metrics as well.

stats/opentelemetry/server_metrics.go
zasweq: Done, moved all the persisted data/these methods into an embedded struct in both, great suggestion!

stats/opentelemetry/client_metrics.go
```diff
@@ -52,14 +53,119 @@ func (h *clientStatsHandler) initializeMetrics() {
...
+	h.clientMetrics.intCounts = make(map[*estats.MetricDescriptor]otelmetric.Int64Counter)
+	h.clientMetrics.floatCounts = make(map[*estats.MetricDescriptor]otelmetric.Float64Counter)
+	h.clientMetrics.intHistos = make(map[*estats.MetricDescriptor]otelmetric.Int64Histogram)
+	h.clientMetrics.floatHistos = make(map[*estats.MetricDescriptor]otelmetric.Float64Histogram)
+	h.clientMetrics.intGauges = make(map[*estats.MetricDescriptor]otelmetric.Int64Gauge)
+
+	for metric := range metrics.Metrics() {
+		desc := estats.DescriptorForMetric(metric)
+		// Per call or metrics not registered, when gets a handle corresponding
+		// it will be a no-op record.
```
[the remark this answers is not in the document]
zasweq: Moved to inside the if. Switched comment to:

"Either the metric was per call or the metric is not registered. Thus, if this component ever receives the desc as a handle in record it will be a no-op."

Let me know if you want me to just delete the second sentence.

stats/opentelemetry/client_metrics.go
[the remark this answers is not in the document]
zasweq: Reading the code, the options are functionality equivalent, good point. If I call it with no args it sets the bucket bounds to nil, which is the same as if I don't call the option at all. So regardless of what the API implementation does (docstring for comment: "This option is considered "advisory", and may be ignored by API implementations") it will pass the implementation the same option. We want the "fallback" if unset to be the default OpenTelemetry bounds, which we eventually want to merge to exponential bounds, which will determine bounds dynamically based off the data recorded.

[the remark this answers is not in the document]
zasweq: Actually, I thought about this more and there's a distinction in what's passed down whether you pass a nil or an empty slice as variadic args: https://go.dev/play/p/fN87rHmymW7. However after looking deep into the *OpenTelemetry SDK* the conditional is 

```
if histAgg, ok := agg.(AggregationExplicitBucketHistogram); ok && len(boundaries) > 0 {
			histAgg.Boundaries = boundaries
			agg = histAgg
		}
```

thus nil and empty are functionality equivalent. This isn't necessarily the case according to the docstring since it's advisory, and it could be treated as distinct (fallback to default or one big bucket). But I'm fine removing this for now since through the SDK (the main implementation) these are functionally equivalent.

stats/opentelemetry/opentelemetry.go
zasweq: "They are currently used to record exemplars (here: https://github.com/open-telemetry/opentelemetry-go/blob/99f830ba01c23f5b0cdc55357f1d5c82bb83e344/sdk/metric/internal/exemplar/storage.go#L64)."

"If you call Record() or Add() with the context returned by tracer.Start(), you will get exemplars from the spans you record on the metrics collected." - David, so since these don't apply to traces I think ignoring makes sense here.

zasweq:
Thanks so much for the pass; got to all comments!

stats/opentelemetry/metricsregistry_test.go
```diff
@@ -0,0 +1,311 @@
...
+	// so only those should show up in metrics emissions. All the required
+	// labels should show up in metrics emissions.
+	mo := MetricsOptions{
+		Metrics:        DefaultMetrics().Add("int-counter-3"),
+		OptionalLabels: []string{"float counter optional label key", "float histo optional label key"},
+		MeterProvider:  clientProvider,
+	}
+	// Don't have access to clientStatsHandler in DialOption, so just create
+	// one directly.
+	csh := &clientStatsHandler{options: Options{MetricsOptions: mo}}
```
dfawley, line 239: If you have a `newClientStatsHandler(MetricsOptions)` function (and server equivalent) then you can remove some duplicated code around constructing the reader/provider -- those can be created inside the loop along with MetricsOptions, then the test cases just do:

```go
for _, constructor := range []func(MetricsOptions)otelmetric.MetricsRecorder{newClientStatsHandler,newServerStatsHandler} {
	t.Run(...)
}
```

stats/opentelemetry/metricsregistry_test.go
```diff
@@ -0,0 +1,311 @@
...
+			metricsRecorder: csh,
+			reader:          clientReader,
+		},
+		{
+			metricsRecorder: ssh,
+			reader:          serverReader,
+		},
+	}
+
+	for _, test := range tests {
```
dfawley, line 268: Prefer immediate `t.Run(___, func(t *testing.T) { /* body of loop */})` inside a loop like this.

stats/opentelemetry/opentelemetry.go
```diff
@@ -184,65 +185,188 @@ type attemptInfo struct {
...
+		for _, optLabel := range optionalLabels { // o(n) could build out a set but n is currently capped at < 5
+			if label == optLabel {
+				attributes = append(attributes, otelattribute.String(label, labelVals[i+len(labelKeys)]))
+			}
+		}
+	}
+	return otelmetric.WithAttributes(attributes...)
+}
+
+type registryMetrics struct {
```
dfawley, line 289: Comment e.g.:
```go
// registryMetrics implements MetricsRecorder for the client and server stats handlers.
```

dfawley (approved):

internal/internal.go
```diff
@@ -215,6 +215,12 @@ var (
 
 	// SetConnectedAddress sets the connected address for a SubConnState.
 	SetConnectedAddress any // func(scs *SubConnState, addr resolver.Address)
+
+	// SnapshotMetricsRegistryForTesting snapshots the global data of the
```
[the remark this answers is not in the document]
zasweq: Hmm I thought about this writing the other PRs. I agree MetricsRegistry sounds better but I think for the sake of consistency the non plural makes more sense here.

stats/opentelemetry/opentelemetry.go
zasweq: Done verbatim.

stats/opentelemetry/metricsregistry_test.go
zasweq: Great suggestion! I had to scale up the interface for initializeMetrics but otherwise took away code duplication.

stats/opentelemetry/metricsregistry_test.go
zasweq: Ah right I forgot about this. Switched.

PR #7416, opened 2024-07-15
outlierdetection: Disable outlier detection for pickfirst balancer

arjan-bal:
As part of the dual stack changes, the pick first LB policy will start creating 1 subConn for each address. As a result, outlier detection (OD) will start working on subConns created by pickfirst (https://github.com/grpc/grpc-go/pull/7384).

Based on the comment in c-core we don't want to support outlier detection for pick first: https://github.com/grpc/grpc/issues/32967#issuecomment-1574427837
>We are planning to make some significant structural changes as part of the forthcoming design for supporting dualstack IPv4/IPv6 backends, and as part of that work, we are going to be modeling outlier detection as a health signal instead of a connectivity signal. This will make it similar to what we support for client-side health checking (see [gRFC A17](https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md)), which we [explicitly do not support with pick_first](https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md#pick_first). If we start supporting outlier detection with pick_first now, it will cause complications for this dualstack design.

Similar change from c-core: https://github.com/grpc/grpc/pull/33336

Before this change, OD worked for pick first when there was only one address. This isn't very useful since there would be no addresses left after ejection.

RELEASE NOTES:
* Outlier detection is disabled when using pick first child policy. Earlier it would work only when there was only one address provided to pick first.

### DIFF
--- a/balancer/pickfirst/pickfirst.go
+++ b/balancer/pickfirst/pickfirst.go
@@ -46,6 +46,11 @@ const (
 	// Name is the name of the pick_first balancer.
 	Name      = "pick_first"
 	logPrefix = "[pick-first-lb %p] "
+	// TODO(arjan-bal): This is a hack to disable outlier_detection when used
+	// with pick_first, for the reasons described in
+	// https://github.com/grpc/grpc/issues/32967.  Remove this when
+	// implementing the dualstack design.
+	DisableOutlierDetectionAttributeName = "disable_outlier_detection"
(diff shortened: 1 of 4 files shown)


arjan-bal:
Closing while I figure out the correct sequence of making of the changes for Dual Stack.

PR #7417, opened 2024-07-15
xds/balancer/priority: Unlock mutex before returning

arjan-bal:
Release the mutex that is locked a couple of lines above.

https://github.com/grpc/grpc-go/blob/d27ddb5eb5940c949f88bc2cb21eed9254f8be75/xds/internal/balancer/priority/balancer.go#L271-L274

This could subsequent calls to `UpdateClientConnState` to get stuck.

RELEASE NOTES: None


dfawley (approved):

dfawley:
> This could subsequent calls to UpdateClientConnState to get stuck.

Calling a `Close`d `Balancer` would be illegal, and isn't something we should ever do anywhere for any reason.

Issue #7327, opened 2024-06-16
Proxy connection buffer necessary?


flowchartsman:
While attempting to implement a proxied connection via configuration value (as opposed to as provided by env), I noticed the following type in `internal/transport/proxy.go`:

```go
// To read a response from a net.Conn, http.ReadResponse() takes a bufio.Reader.
// It's possible that this reader reads more than what's need for the response and stores
// those bytes in the buffer.
// bufConn wraps the original net.Conn and the bufio.Reader to make sure we don't lose the
// bytes in the buffer.
type bufConn struct {
	net.Conn
	r io.Reader
}

func (c *bufConn) Read(b []byte) (int, error) {
	return c.r.Read(b)
}
```

The description seems to indicate that this type exists due to a concern that the `*bufio.Reader` wrapping the `net.Conn` will read more bytes than is necessary to fulfill the call to `http.ReadResponse` in [`doHTTPConnectHandshake`](https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go#L98) and potentially consume some of the initial bytes of TLS exchange; however the code that serves the same purpose in `http.Transport` [asserts that this won't happen](https://github.com/golang/go/blob/go1.22.4/src/net/http/transport.go#L1744-L1747) since the TLS host will not respond until the conn is written to again, making the `*bufio.Reader` safe to throw away.

Has anyone ever actually observed the reader getting more than just the `CONNECT` response, or is this just an overabundance of caution? I may have missed something, but this doesn't seem too different from the way it's done in `http.Transport`.

I don't know how much it would affect throughput or anything, but it's certainly wasteful to have another buffered reader in between you and the `net.Conn` if you don't need it.

I removed the wrapper in favor of returning the `conn` directly from `doHTTPConnectHandshake`, and all of the tests still pass, though that's not necessarily compelling proof for a negative.  I can try and make a stronger case if needed--just have to figure out if it can be effectively tested.

aranjans:
Hi @flowchartsman, thanks for writing in. We are currently looking to this issue and will have update soon!

arjan-bal:
I'll try to figure out why this buffer was introduced and if we still need it by going though the code and commit history.

arjan-bal:
[...]

PR #7424, opened 2024-07-18
transport: Discard the buffer when empty after http connect handshake

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/7327

The buffer used to read the http connect response could contain extra bytes from the target server, so we can't discard it. However, in many cases where the server waits for the client to send the first message (e.g. when TLS is being used), the buffer will be empty, so we can avoid the overhead of reading through this buffer.

This PR also adds a unit test to ensure the buffer is not discarded when it has unread data.

RELEASE NOTES:
* Double buffering is avoided when using an http connect proxy and the target server waits for client to send the first message.

### DIFF
--- a/internal/transport/proxy_test.go
+++ b/internal/transport/proxy_test.go
@@ -117,7 +117,13 @@ func (p *proxyServer) stop() {
 	}
 }
 
-func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxyReqCheck func(*http.Request) error, serverMessage []byte) {
+type testArgs struct {
+	proxyURLModify func(*url.URL) *url.URL
+	proxyReqCheck  func(*http.Request) error
+	serverMessage  []byte
+}
+
+func testHTTPConnect(t *testing.T, args testArgs) {
 	plis, err := net.Listen("tcp", "localhost:0")
 	if err != nil {
 		t.Fatalf("failed to listen: %v", err)
@@ -125,9 +131,9 @@ func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxy
 	p := &proxyServer{
 		t:            t,
 		lis:          plis,
-		requestCheck: proxyReqCheck,
+		requestCheck: args.proxyReqCheck,
 	}
-	go p.run(len(serverMessage) > 0)
+	go p.run(len(args.serverMessage) > 0)
 	defer p.stop()
 
 	blis, err := net.Listen("tcp", "localhost:0")
@@ -145,14 +151,14 @@ func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxy
 			return
 		}
 		defer in.Close()
-		in.Write(serverMessage)
+		in.Write(args.serverMessage)
 		in.Read(recvBuf)
 		done <- nil
 	}()
 
 	// Overwrite the function in the test and restore them in defer.
 	hpfe := func(req *http.Request) (*url.URL, error) {
-		return proxyURLModify(&url.URL{Host: plis.Addr().String()}), nil
+		return args.proxyURLModify(&url.URL{Host: plis.Addr().String()}), nil
 	}
 	defer overwrite(hpfe)()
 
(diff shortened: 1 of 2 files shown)


internal/transport/proxy_test.go
```diff
@@ -157,6 +176,18 @@ func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxy
...
 	}
+
+	c.SetReadDeadline(time.Now().Add(20 * time.Millisecond))
+
+	gotServerMessage := make([]byte, len(serverMessage))
+	// This call will return a deadline exceeded error if the server has nothing
+	// to send. This is expected.
+	if _, err := c.Read(gotServerMessage); err != nil {
+		t.Logf("Got error while reading message from server: %v", err)
+	}
```
dfawley, line 187: Similar; I don't really like adding a magic deadline to things that we know will timeout sometimes.

Probably we should only do this read if `len(serverMessage) != 0`, and then we can apply a longer deadline while not impacting code that doesn't use it?

internal/transport/proxy_test.go
```diff
@@ -100,7 +118,7 @@ func (p *proxyServer) stop() {
 	}
 }
 
-func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxyReqCheck func(*http.Request) error) {
+func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxyReqCheck func(*http.Request) error, serverMessage []byte) {
```
dfawley, line 121: We probably should add a struct about now to hold all the parameters (unless we decide to test this code differently per the above comment).

internal/transport/proxy_test.go
```diff
@@ -84,7 +85,24 @@ func (p *proxyServer) run() {
...
+	// the response to the connect request and proxied packets from the
+	// destination server when it reads the transport.
+	out.SetReadDeadline(time.Now().Add(20 * time.Millisecond))
+	b := make([]byte, 50)
+	bytesRead, err := out.Read(b)
+	// The read is expected to fail with deadline exceeded if the server doesn't
+	// have a message to send.
+	if err != nil {
+		p.t.Logf("Got error while reading message from server: %v", err)
+	}
```
dfawley, line 101: Is there another way to test this functionality that we can use instead of setting a magic deadline?

Maybe a more directed unit test of the `doHTTPConnectHandshake` function?

Or parameterize the proxyServer so it knows to do this step / skip it, and set a longer deadline and fail if it errors?

internal/transport/proxy_test.go
arjan-bal: Changed, not the proxy is configured to wait for server hello.

internal/transport/proxy_test.go
arjan-bal: Parameterize proxy server to wait for the server hello.

internal/transport/proxy_test.go
arjan-bal: Done.

internal/transport/proxy_test.go
```diff
@@ -157,33 +181,62 @@ func testHTTPConnect(t *testing.T, proxyURLModify func(*url.URL) *url.URL, proxy
 	if string(recvBuf) != string(msg) {
 		t.Fatalf("received msg: %v, want %v", recvBuf, msg)
 	}
+
+	if len(args.serverMessage) > 0 {
+		c.SetReadDeadline(time.Now().Add(defaultTestTimeout))
```
dfawley, line 186: There's now 3 places where we're setting this deadline.

Maybe instead when the conn is created in `proxyDial` we should just do `c.SetDeadline(....defaultTestTimeout)` (note `Deadline` not `Read` to also abort writes)?

dfawley (approved):

internal/transport/proxy_test.go
arjan-bal: During testing, I noticed that the test can hang in `proxyDial` if the proxy server waits for a `server hello` when the server doesn't intend to send one. This was due to a logical error on my part which I fixed.

To be safe, I added read deadlines in both the places where reads could hang.

> Maybe instead when the conn is created in proxyDial we should just do c.SetDeadline(....defaultTestTimeout) (note Deadline not Read to also abort writes)?

Do you mean adding a deadline in the actual `proxyDial()` implementation by introducing a timeout that only the tests use?
Or are you referring to adding a deadline on the conn returned by `proxyDial` in the test code?

The l latter would not prevent `proxyDial` from hanging when the proxy server is stuck reading. The former would work but it involves changing non-test code, which I was trying to avoid.

dfawley: Sorry, I didn't mean _in_ `proxyDial`... I meant the connection it returns, we could immediately, always, do a `SetDeadline(...)`, instead of doing it later and conditionally.

arjan-bal: Changed `SetReadDeadline` to `SetDeadline`. Moved the SetDeadline immediately after the `proxyDial` call. The `SetDeadline` in `proxyServer.run()` is still present to avoid proxyDial from hanging indefinitely in case of test failures.

dfawley (approved):

PR #7428, opened 2024-07-19
internal/stats: Add metrics recorder list and usage in ClientConn 

zasweq:
This PR adds a metrics recorder list component as a top level component of the ClientConn. It also forwards this to balancer by scaling up balancer.BuildOptions.

The Server Side equivalent will come when needed to pass down to xDS Client, but no need for now.

RELEASE NOTES: N/A


internal/stats/metrics_recorder_list.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+	return &MetricsRecorderList{
+		metricsRecorders: mrs,
+	}
+}
+
+func (l *MetricsRecorderList) RecordInt64Count(handle *estats.Int64CountHandle, incr int64, labels ...string) {
+	if got, want := len(handle.Labels)+len(handle.OptionalLabels), len(labels); got != want {
+		logger.Infof("length of labels passed to RecordInt64Count incorrect got: %v, want: %v", got, want)
```
dfawley, line 54: This isn't a test and probably shouldn't be formatted like one since it's a bit awkward looking re: got/want messages.  Also logs do not follow the same rule as errors of not starting with a capital letter.  They are typically complete sentences and should be formatted as such.

Also: should be Error level since this is a programming bug, or at least Warning.

And lastly: we just log the problem and still call Record.  Is calling Record OK if the parameter length doesn't match?  Don't we want to protect the MetricsRecorders from needing to deal with bad data?

internal/stats/metrics_recorder_list.go
```diff
@@ -0,0 +1,100 @@
...
+		logger.Infof("length of labels passed to RecordInt64Histo incorrect got: %v, want: %v", got, want)
+	}
+
+	for _, metricRecorder := range l.metricsRecorders {
+		metricRecorder.RecordInt64Histo(handle, incr, labels...)
+	}
+}
+
+func (l *MetricsRecorderList) RecordFloat64Histo(handle *estats.Float64HistoHandle, incr float64, labels ...string) {
+	if got, want := len(handle.Labels)+len(handle.OptionalLabels), len(labels); got != want {
```
dfawley, line 83: Maybe don't repeat the validation logic at each of these `Record` points, and instead have a helper function do it instead?

internal/stats/test/metrics_recorder_list_test.go
```diff
@@ -0,0 +1,133 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package test implements an e2e test for the Metrics Recorder List component
+// of the Client Conn, and a TestMetricsRecorder utility.
+package test
```
dfawley, line 21: `test_test` or else it doesn't run with `go test` I think?

But why put it in a separate directory instead of just putting it in `stats` with package `stats_test`?

zasweq: Discussed offline; move metrics_recorder_list_test and recording_load_balancer to internal/stats as stats_test package and test_metrics_recorder to internal/testutils/stats.

internal/stats/metrics_recorder_list.go
zasweq: Discussed offline; decided to make it panic in a helper.

internal/stats/metrics_recorder_list.go
zasweq: Done, refactored to a helper which panics.

zasweq:
Thanks for the pass; got to all comments.

internal/stats/metrics_recorder_list.go
```diff
@@ -49,50 +49,46 @@ func NewMetricsRecorderList(shs []stats.Handler) *MetricsRecorderList {
 	}
 }
 
-func (l *MetricsRecorderList) RecordInt64Count(handle *estats.Int64CountHandle, incr int64, labels ...string) {
-	if got, want := len(handle.Labels)+len(handle.OptionalLabels), len(labels); got != want {
-		logger.Infof("length of labels passed to RecordInt64Count incorrect got: %v, want: %v", got, want)
+func verifyLabels(labels []string, optionalLabels []string, labelsRecv ...string) {
```
dfawley, line 52: Could this pass a `*MetricDescriptor` instead of the labels/optionallabels fields?

internal/stats/metrics_recorder_list.go
```diff
@@ -49,50 +49,46 @@ func NewMetricsRecorderList(shs []stats.Handler) *MetricsRecorderList {
...
+	if got, want := len(labelsRecv), len(labels)+len(optionalLabels); got != want {
+		logger.Fatalf("Length of labels passed to Record incorrect, got: %v, want: %v.", got, want)
```
dfawley, line 54: IMO delete the logger now and call `panic(fmt.Sprintf)` here.

Also if we pass the descriptor let's print the metric name with `%q` as well.  Maybe lose the `got, want` formatting which I believe is unique to Go test assertions?

internal/stats/metrics_recorder_list.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	estats "google.golang.org/grpc/experimental/stats"
+	"google.golang.org/grpc/grpclog"
+	"google.golang.org/grpc/stats"
+)
+
+var logger = grpclog.Component("metrics-recorder-list")
+
+// MetricsRecorderList forwards Record calls to all of it's metricsRecorders.
```
dfawley, line 27: its*.  Check throughout please.

internal/testutils/stats/test_metrics_recorder.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+// TestMetricsRecorder is a MetricsRecorder to be used in tests. It sends
+// recording events on channels and provides helpers to check if certain events
+// have taken place. It also persists metrics data keyed on the metrics
+// descriptor.
+type TestMetricsRecorder struct {
+	t *testing.T
+
+	// TODO: scale this persisted storage up with label key values.
```
dfawley, line 39: Sorry, I'm not sure what this means?

dfawley (commented):
Looks good overall, just some minor things..

internal/stats/metrics_recorder_list.go
zasweq: Done.

internal/stats/metrics_recorder_list.go
zasweq: Got rid of got, want. Recorded the metric name with %q from desc.

internal/testutils/stats/test_metrics_recorder.go
zasweq: Discussed offline; decided to get rid of these maps entirely and add these when needed.

internal/stats/metrics_recorder_list.go
zasweq: Done. I control fed and it seems like this is the only instance.

dfawley: I would like to propose that you never write "it's" and use either "its" or "it is" from now on.  No more apostrophes for you!  ;)

internal/stats/metrics_recorder_list.go
```diff
@@ -49,46 +48,46 @@ func NewMetricsRecorderList(shs []stats.Handler) *MetricsRecorderList {
...
-		logger.Fatalf("Length of labels passed to Record incorrect, got: %v, want: %v.", got, want)
+func verifyLabels(desc *estats.MetricDescriptor, labelsRecv ...string) {
+	if len(labelsRecv) != len(desc.Labels)+len(desc.OptionalLabels) {
+		panic(fmt.Sprintf("Length of labels passed to Record incorrect for metric %q.", desc.Name))
 	}
 }
 
 func (l *MetricsRecorderList) RecordInt64Count(handle *estats.Int64CountHandle, incr int64, labels ...string) {
-	verifyLabels(handle.Labels, handle.OptionalLabels, labels...)
+	verifyLabels((*estats.MetricDescriptor)(handle), labels...)
```
dfawley, line 58: Oh..I was thinking we wouldn't need to do the cast here for some reason...  I wonder if it's worth adding a method to the handles like:

```go
func (h handle) Descriptor() *estats.MetricDescriptor {
	return (*estats.MetricDescriptor)(h)
}
```

...in order to keep the ugliness out of the places we do the cast.

internal/stats/metrics_recorder_list.go
dfawley: Oh I didn't mean to remove the counts..  Just to turn it into a more english-y thing than a Go test error message.  E.g.

Like `Received %d labels in call to record metric %q, but expected %d`

internal/stats/metrics_recorder_list_test.go
```diff
@@ -223,7 +223,7 @@ func (s) TestMetricRecorderListPanic(t *testing.T) {
 	})
 	mrl := istats.NewMetricsRecorderList(nil)
 
-	want := "Length of labels passed to Record incorrect, got: 1, want: 2."
+	want := `Length of labels passed to Record incorrect for metric "simple counter".`
```
dfawley, line 226: It's not super important to check the string, especially in its entirety, if you don't want to keep adjusting this.

dfawley (approved):

internal/stats/metrics_recorder_list.go
zasweq: Yeah, these are also cast like this in OTel, we could add but then that populates exported namespace. Although if a user does use the metrics registry and writes their own stats handlers, they will need to key on a property of the descriptor, and perhaps they will use the pointer as well.

internal/stats/metrics_recorder_list.go
zasweq: Ah ok noted switched to that.

internal/stats/metrics_recorder_list.go
zasweq: I'll keep it as a separate commit and you can see how it looks.

internal/stats/metrics_recorder_list_test.go
zasweq: Left it I like it.

internal/stats/metrics_recorder_list.go
zasweq: Sent out a separate commit: https://github.com/grpc/grpc-go/pull/7428/commits/619b20e3640e89ca427550dd3b90410dc4d541c6, let me know what you think.

PR #7432, opened 2024-07-22
mem: introduce `mem` package to facilitate memory reuse

PapaCharlie:
This package is required for https://github.com/grpc/grpc-go/pull/7356 but is introduced early here to allow additional features to be built on top of it. It provides a mechanism to reuse and recycle buffers to reduce GC pressure.

RELEASE NOTES:
- TBD


arvindbr8:
cc: @printchard

mem/buffers_test.go
```diff
@@ -0,0 +1,44 @@
+package mem
```
easwars, line 1: Copyright notice is missing. Please fix. Thanks.

mem/buffers.go
```diff
@@ -0,0 +1,156 @@
...
+// ReadOnlyData returns the underlying byte slice. Note that it is undefined
+// behavior to modify the contents of this slice in any way.
+func (b *Buffer) ReadOnlyData() []byte {
+	if b.freed {
+		panic("Cannot read freed buffer")
+	}
+	return b.data
+}
+
+// Ref returns a new reference to this Buffer's underlying buffer. Calling Free
```
easwars, line 79: If we mention that the free function will only be called after all references are released in the docstring for the `Buffer` type, we wont have to mention that in specific methods.

mem/buffers.go
```diff
@@ -0,0 +1,156 @@
...
+
+	b.refs.Add(1)
+	return &Buffer{
+		data: b.data,
+		refs: b.refs,
+		free: b.free,
+	}
+}
+
+// Free decrements this Buffer's reference counter. It then frees the underlying
```
easwars, line 95: Same here. If we mention that access to the underlying data will panic after a call to Free() in the docstring of the `Buffer` type, we wont have to mention that here.

mem/buffers.go
```diff
@@ -0,0 +1,156 @@
...
+// buffer if the counter reaches 0 as a result of this call, does nothing
+// otherwise, or if Free has already been called. All subsequent calls to
+// ReadOnlyData will panic. Noop if receiver is nil.
+func (b *Buffer) Free() {
+	if b == nil || b.freed {
```
easwars, line 100: Why do we handle the `nil` receiver in this method alone?

mem/buffers.go
```diff
@@ -0,0 +1,156 @@
...
+	if b.freed {
+		panic("Cannot split freed buffer")
+	}
+
+	b.refs.Add(1)
+
+	data := b.data
+	var free func([]byte)
+	if f := b.free; f != nil {
+		free = func(_ []byte) {
```
easwars, line 136: Nit: the `_` could be omitted here.

mem/buffers.go
```diff
@@ -0,0 +1,156 @@
...
+			f(data)
+		}
+	}
+
+	b.data = data[:n]
+	b.free = free
```
easwars, line 142: Isn't this statement a no-op? Can this be removed?

mem/buffers.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return &Buffer{
+		data: data[n:],
+		refs: b.refs,
+		free: free,
+	}
+}
+
+// String returns a string representation of the buffer that contains the address
```
easwars, line 151: I would prefer for this docstring to simply say:
```
// String returns a string representation of the buffer. Maybe used for debugging purposes.
```
instead of describing exactly how it will look.

And also, we should avoid saying things like `used to uniquely identify the underlying buffer`, since very quickly our users will start relying on the output of this method for wildly critical things, and will complain when we change the implementation.

easwars (changes requested):

mem/buffers_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestSplit(t *testing.T) {
```
easwars, line 8: These tests should be methods on the `grpctest.Tester` type. Otherwise, our vet will fail.

mem/buffers_test.go
easwars, line 1: Also, if possible, could we move these to package `mem_test` to ensure that we testing the API and not relying on anything internal to the package. Thanks.

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
+package mem
```
easwars, line 1: Copyright notice missing here too.

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"io"
+)
+
+// BufferSlice offers a means to represent data that spans one or more Buffer
+// instances.
```
easwars, line 8: It might be worth mentioning here that BufferSlice methods could panic if any of the underlying Buffer instances have been freed and there is an attempt to access data represented by that Buffer.

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+type BufferSlice []*Buffer
```
easwars, line 9: Also, worth mentioning that a BufferSlice is meant to be immutable, hence methods have value receiver type.

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+// input slice.
+func (s BufferSlice) Ref() BufferSlice {
+	out := make(BufferSlice, len(s))
+	for i, b := range s {
+		out[i] = b.Ref()
+	}
+	return out
+}
+
+// Free invokes Buffer.Free on each Buffer in the slice.
```
easwars, line 35: Nit: 
```
// Free invokes Free() on each Buffer in the slice.
```

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+		data: s,
+		len:  s.Len(),
+	}
+}
+
+var _ io.ReadCloser = (*Reader)(nil)
+
+// Reader exposes a BufferSlice's data as an io.Reader, allowing it to interface
+// with other parts systems. It also provides an additional convenience method
+// Remaining which returns the number of unread bytes remaining in the slice.
```
easwars, line 85: Nit: s/`Remaining`/`Remaining(),`/

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+			r.idx = 0
+		} else {
+			r.idx += copied
+		}
+		n += copied
+	}
+
+	if n == 0 {
+		return 0, io.EOF
+	}
```
easwars, line 126: `n` could be `0` even when the passed in `buf` has zero capacity. So, I think we should handle this at the top of the method:
```golang
if r.len == 0 {
  return 0, io.EOF
}
```

And here after the `for` loop`, we don't need a check for `n == 0`.

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+var _ io.ReadCloser = (*Reader)(nil)
+
+// Reader exposes a BufferSlice's data as an io.Reader, allowing it to interface
+// with other parts systems. It also provides an additional convenience method
+// Remaining which returns the number of unread bytes remaining in the slice.
+//
+// Note that reading data from the reader does not free the underlying buffers!
+// Only calling Close once all data is read will free the buffers.
```
easwars, line 88: Should the `Reader` take a ref to all the underlying buffers instead? With the current approach, we will `panic` if `Free` is called on the `BufferSlice` *and* `Close` is called on the `Reader`. With the approach that I'm suggesting, both methods would have to be called, else it would leak the buffers. Did you pick this option over the other for some reason?

mem/buffer_slice.go
```diff
@@ -0,0 +1,155 @@
...
+type Reader struct {
+	data         BufferSlice
+	len          int
+	dataIdx, idx int
```
easwars, line 92: One of these is a index into the slice of buffers, while the other is an index into a single byte slice. I feel that we need better names to clearly disambiguate these two. What do you think?

mem/buffer_pool.go
```diff
@@ -0,0 +1,184 @@
...
+		// Ignore buffers that are too small to fit in the pool. Otherwise, when Get is
+		// called it will panic as it tries to index outside the bounds of the buffer.
+		return
+	}
+	buf = buf[:cap(buf)]
+	clear(buf)
+	p.pool.Put(&buf)
+}
+
+func newBufferPool(size int) *sizedBufferPool {
```
easwars, line 133: s/newBufferPool/newSizedBufferPool/ since we have many buffer pools in this file.

mem/buffer_pool.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (p *sizedBufferPool) Get(size int) []byte {
+	bs := *p.pool.Get().(*[]byte)
+	return bs[:size]
+}
+
+func (p *sizedBufferPool) Put(buf []byte) {
+	if cap(buf) < p.defaultSize {
+		// Ignore buffers that are too small to fit in the pool. Otherwise, when Get is
+		// called it will panic as it tries to index outside the bounds of the buffer.
```
easwars, line 125: Would it instead be better for `Get` to handle the case where the buffer returned from `p.pool.Get()` is of size less than `size`, by returning `nil`. Because IIUC, if `Get` returns `nil`, then `sync.Pool` would call `New()`. That seems to me like a better behavior than panicking.

mem/buffer_pool.go
```diff
@@ -0,0 +1,184 @@
...
+	return &tieredBufferPool{
+		sizedPools: pools,
+	}
+}
+
+// tieredBufferPool implements the BufferPool interface with multiple tiers of
+// buffer pools for different sizes of buffers.
+type tieredBufferPool struct {
+	sizedPools   []*sizedBufferPool
+	fallbackPool simpleBufferPool
```
easwars, line 84: This field is no longer initialized in `NewBufferPool`.

easwars (changes requested):

easwars:
@PapaCharlie : Could you please reply to individual comments with how you addressed them instead of marking them as resolved, since the latter approach means that the reviewer has to open all the resolved comments to see what that comment was in the first place and whether it was addressed to their satisfaction. Thanks.

PapaCharlie:
Ah sorry about that. I did reply, but I didn't publish them since I was in review mode still.. I just published my replies

mem/buffer_pool.go
easwars: Thanks for the explanation.

mem/buffer_pool.go
easwars: Thanks. I do the see docs being update on the `sizedBufferPool` type. That should be good.

mem/buffer_slice.go
easwars: From that comment I see that a reader will be used in the `loopyWriter` to read buffers from a `BufferSlice` inside of the `dataFrame`, and write it out via the framer. So, I get the point that we cannot free the buffers after reading them, because that would make it impossible to retry the write.

My suggestion is that:
- `Reader` takes a reference to the whole `BufferSlice`
- The `dataFrame` or any other entity that owns the `BufferSlice` also retains a reference
- Once the `Read` operation is complete, someone would have to call `Close` on the `Reader`
- And someone would also have to release the original reference to the `BufferSlice`

The reason I prefer this approach is that it ensures that the `BufferSlice` is not freed from under the `Reader`. Do you see any issues with this approach? Thanks.

easwars:
I've pushed a few more commits with more tests and in the process of writing them, I found a bug in the `Reader` and also fixed it. PTAL.

easwars:
At this point, I'm mostly OK with how this package looks.

@PapaCharlie : Could you please take a look at my commits to see if you are Ok with those changes.

@dfawley : Could you make a pass. Thanks.

easwars:
This one comment is still unaddressed though: https://github.com/grpc/grpc-go/pull/7432#discussion_r1687257265

I don't have a strong opinion there, but I was wondering if one is a preferred way of doing things over the other, and if so why? Thanks.

mem/buffers_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Verify that the free function is invoked when all references are freed.
+	buf.Free()
+	select {
+	case err := <-errCh:
+		if err != nil {
+			t.Fatal(err)
+		}
+	case <-time.After(defaultTestTimeout):
+		t.Fatalf("Timeout waiting for Buffer to freed")
```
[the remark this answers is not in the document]
easwars: Done.

mem/buffer_pool.go
```diff
@@ -0,0 +1,194 @@
...
+	return *defaultBufferPool.Load()
+}
+
+// SetDefaultBufferPoolForTesting updates the default buffer pool, for testing
+// purposes.
+//
+// # Testing Only
+//
+// This function should ONLY be used for testing purposes.
+func SetDefaultBufferPoolForTesting(pool BufferPool) {
```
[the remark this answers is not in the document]
easwars: Done.

mem/buffer_pool.go
```diff
@@ -0,0 +1,194 @@
...
+
+package mem
+
+import (
+	"sort"
+	"sync"
+	"sync/atomic"
+)
+
+// BufferPool is a pool of buffers that can be shared, resulting in
```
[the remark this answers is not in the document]
easwars: Done.

mem/buffer_pool.go
```diff
@@ -0,0 +1,184 @@
...
+var defaultBufferPoolSizes = []int{
+	256,
+	4 << 10,  // 4KB (go page size)
+	16 << 10, // 16KB (max HTTP/2 frame size used by gRPC)
+	32 << 10, // 32KB (default buffer size for io.Copy)
+	// TODO: Report the buffer sizes requested with Get to tune this properly.
+	1 << 20, // 1MB
+}
+
+var defaultBufferPool = func() *atomic.Pointer[BufferPool] {
```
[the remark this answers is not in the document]
easwars: Ok, went ahead and changed it along with removing the atomic.

mem/buffer_pool.go
```diff
@@ -0,0 +1,194 @@
...
+	ptr := new(atomic.Pointer[BufferPool])
+	ptr.Store(&pool)
+	return ptr
+}()
+
+// DefaultBufferPool returns the current default buffer pool. It is a BufferPool
+// created with NewBufferPool that uses a set of default sizes optimized for
+// expected workflows.
+func DefaultBufferPool() BufferPool {
+	return *defaultBufferPool.Load()
```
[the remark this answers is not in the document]
easwars: Done.

mem/buffer_pool.go
```diff
@@ -0,0 +1,194 @@
...
+// # Testing Only
+//
+// This function should ONLY be used for testing purposes.
+func SetDefaultBufferPoolForTesting(pool BufferPool) {
+	defaultBufferPool.Store(&pool)
+}
+
+// NewBufferPool returns a BufferPool implementation that uses multiple
+// underlying pools of the given pool sizes.
+func NewBufferPool(poolSizes ...int) BufferPool {
```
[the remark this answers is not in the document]
easwars: Done.

mem/buffer_pool.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (p *sizedBufferPool) Put(buf []byte) {
+	if cap(buf) < p.defaultSize {
+		// Ignore buffers that are too small to fit in the pool. Otherwise, when
+		// Get is called it will panic as it tries to index outside the bounds
+		// of the buffer.
+		return
+	}
+	buf = buf[:cap(buf)]
+	clear(buf)
```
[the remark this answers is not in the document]
easwars: I remember making the same comment in one of the older PRs. Paul had some rationale for it, but I don't remember it now. Will wait for Paul on this.

mem/buffer_slice.go
```diff
@@ -0,0 +1,197 @@
...
+
+// NewWriter wraps the given BufferSlice and BufferPool to implement the
+// io.Writer interface. Every call to Write copies the contents of the given
+// buffer into a new Buffer pulled from the given pool and the Buffer is added to
+// the given BufferSlice. For example, in the context of a http.Handler, the
+// following code can be used to copy the contents of a request into a
+// BufferSlice:
+//
+//	var out mem.BufferSlice
+//	n, err := io.Copy(mem.NewWriter(&out, pool), req.Body)
```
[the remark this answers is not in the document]
easwars: Added an example test for this.

mem/buffer_slice.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	out := make([]byte, l)
+	s.CopyTo(out)
+	return out
+}
+
+// LazyMaterialize functions like Materialize except that it writes the data to a
+// single Buffer pulled from the given BufferPool. As a special case, if the
+// input BufferSlice only actually has one Buffer, this function has nothing to
+// do and simply returns said Buffer, hence it being "lazy".
```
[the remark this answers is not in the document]
easwars: I wasn't very convinced with the term "lazy" here as well and had made a comment on one of the previous PRs. See: https://github.com/grpc/grpc-go/pull/7356#discussion_r1663070728

mem/buffers.go
```diff
@@ -0,0 +1,149 @@
...
+// NewBuffer creates a new Buffer from the given data, initializing the
+// reference counter to 1. The given free function is called when all references
+// to the returned Buffer are released.
+//
+// Note that the backing array of the given data is not copied.
+func NewBuffer(data []byte, free func([]byte)) *Buffer {
+	b := &Buffer{data: data, refs: new(atomic.Int32)}
+	if free != nil {
+		b.free = func() { free(data) }
+	}
```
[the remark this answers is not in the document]
PapaCharlie: This is actually important because the `Split` function changes `b.data`, so capturing this here avoids potential pitfalls down the line. It was initially set up that way but became too complicated to manage

mem/buffer_pool.go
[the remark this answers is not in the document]
PapaCharlie: I'll throw a test around this. It seems like a good idea to have since anything that remains in the buffer can leak when it's reused.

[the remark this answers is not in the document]
PapaCharlie: Added test

mem/buffer_slice.go
```diff
@@ -0,0 +1,197 @@
...
+
+// Free invokes Buffer.Free() on each Buffer in the slice.
+func (s BufferSlice) Free() {
+	for _, b := range s {
+		b.Free()
+	}
+}
+
+// CopyTo copies each of the underlying Buffer's data into the given buffer,
+// returning the number of bytes copied.
```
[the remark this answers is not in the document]
PapaCharlie: Done

mem/buffer_slice.go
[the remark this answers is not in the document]
PapaCharlie: Changed to `Concatenate`. I think it makes sense it's pretty descriptive that it's creating a new big buffer if needed but leaves room for the smart case to be smart.

mem/buffers.go
```diff
@@ -0,0 +1,149 @@
...
+	free  func()
+	freed bool
+}
+
+// NewBuffer creates a new Buffer from the given data, initializing the
+// reference counter to 1. The given free function is called when all references
+// to the returned Buffer are released.
+//
+// Note that the backing array of the given data is not copied.
+func NewBuffer(data []byte, free func([]byte)) *Buffer {
```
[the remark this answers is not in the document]
PapaCharlie: Sure. Do you also want me to rename the field in `Buffer`?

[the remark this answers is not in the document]
dfawley: Probably would be better, but that (or both, really) can be done as a cleanup later.

easwars (approved):

dfawley (approved):

PR #7434, opened 2024-07-23
stats: Add optional locality label in cluster_impl picker

zasweq:
This PR adds the setting of the optional locality label in the cluster_impl picker for per call metrics. It also sets the default of "" for this label in OpenTelemetry.

RELEASE NOTES: N/A

### DIFF
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -150,8 +150,9 @@ func (h *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo)
 	var labels *istats.Labels
 	if labels = istats.GetLabels(ctx); labels == nil {
 		labels = &istats.Labels{
-			// The defaults for all the per call optional labels that this
-			// OpenTelemetry component currently supports.
+			// The defaults for all the per call label from a plugin that
+			// executes on callpath that this OpenTelemetry component currently
+			// supports.
 			TelemetryLabels: map[string]string{
 				"grpc.lb.locality": "",
 			},
@@ -240,7 +241,6 @@ func (h *clientStatsHandler) processRPCEnd(ctx context.Context, ai *attemptInfo,
 		// CSM Plugin Option layer by adding an optional labels API.
 		if val, ok := ai.xdsLabels[o]; ok {
 			attributes = append(attributes, otelattribute.String(o, val))
-			continue
 		}
 	}
 
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -144,7 +144,11 @@ func (h *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo)
 	var labels *istats.Labels
 	if labels = istats.GetLabels(ctx); labels == nil {
 		labels = &istats.Labels{
-			TelemetryLabels: make(map[string]string),
+			// The defaults for all the per call optional labels that this
+			// OpenTelemetry component currently supports.
+			TelemetryLabels: map[string]string{
+				"grpc.lb.locality": "",
+			},
 		}
 		ctx = istats.SetLabels(ctx, labels)
 	}
@@ -226,8 +230,11 @@ func (h *clientStatsHandler) processRPCEnd(ctx context.Context, ai *attemptInfo,
 	}
 
 	for _, o := range h.options.MetricsOptions.OptionalLabels {
+		// TODO: Add a filter for converting to unknown if not present in the
+		// CSM Plugin Option layer by adding an optional labels API.
 		if val, ok := ai.xdsLabels[o]; ok {
 			attributes = append(attributes, otelattribute.String(o, val))
+			continue
 		}
 	}
 
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -150,9 +150,9 @@ func (h *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo)
 	var labels *istats.Labels
 	if labels = istats.GetLabels(ctx); labels == nil {
 		labels = &istats.Labels{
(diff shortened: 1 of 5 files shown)


stats/opentelemetry/client_metrics.go
```diff
@@ -232,8 +236,11 @@ func (h *clientStatsHandler) processRPCEnd(ctx context.Context, ai *attemptInfo,
 	}
 
 	for _, o := range h.options.MetricsOptions.OptionalLabels {
+		// TODO: Add a filter for converting to unknown if not present in the
+		// CSM Plugin Option layer by adding an optional labels API.
 		if val, ok := ai.xdsLabels[o]; ok {
 			attributes = append(attributes, otelattribute.String(o, val))
+			continue
```
dfawley, line 243: Why a continue when you're already at the end of the loop?

test/xds/xds_telemetry_labels_test.go
```diff
@@ -43,6 +43,9 @@ const serviceNamespaceKeyCSM = "csm.service_namespace_name"
 const serviceNameValue = "grpc-service"
 const serviceNamespaceValue = "grpc-service-namespace"
 
+const localityKey = "grpc.lb.locality"
+const localityValue = "{\"region\":\"region-1\",\"zone\":\"zone-1\",\"subZone\":\"subzone-1\"}"
```
dfawley, line 47: Please use backticks to quote so you don't have to escape all the double-quotes.

test/xds/xds_telemetry_labels_test.go
```diff
@@ -132,7 +135,9 @@ func (fsh *fakeStatsHandler) HandleRPC(ctx context.Context, rs stats.RPCStats) {
 		if label, ok := fsh.labels.TelemetryLabels[serviceNamespaceKeyCSM]; !ok || label != serviceNamespaceValue {
 			fsh.t.Fatalf("for telemetry label %v, want: %v, got: %v", serviceNamespaceKeyCSM, serviceNamespaceValue, label)
 		}
-
+		if label, ok := fsh.labels.TelemetryLabels[localityKey]; !ok || label != localityValue {
```
dfawley, line 138: Now that there's 3 of these...can we build up our expectation and `cmp.Diff`?

xds/internal/balancer/clusterimpl/picker.go
```diff
@@ -156,6 +156,12 @@ func (d *picker) Pick(info balancer.PickInfo) (balancer.PickResult, error) {
 		return pr, err
 	}
 
+	if info.Ctx != nil {
+		if labels := stats.GetLabels(info.Ctx); labels != nil && labels.TelemetryLabels != nil {
```
dfawley, line 160: This is the second call to `if info.Ctx != nil { ... GetLabels() ...` in this function..can we try to combine so we aren't duplicating this logic?

maybe

```go
func telLabels(ctx context.Context) map[string]string {
	if ctx == nil {
		return nil
	}
	labels := stats.GetLabels(info.Ctx)
	if labels == nil {
		return nil
	}
	return labels.TelemetryLabels
}

// ...

if labels := telLabels(info.Ctx); labels != nil {
	labels["grpc.lb.locality"] = lIDStr
}
```

dfawley (commented):
LGTM just nits.

stats/opentelemetry/client_metrics.go
```diff
@@ -150,7 +150,11 @@ func (h *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo)
 	var labels *istats.Labels
 	if labels = istats.GetLabels(ctx); labels == nil {
 		labels = &istats.Labels{
-			TelemetryLabels: make(map[string]string),
+			// The defaults for all the per call optional labels that this
+			// OpenTelemetry component currently supports.
```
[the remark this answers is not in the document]
zasweq: I don't think so. I thought since we discussed how all per call operations get this context, to just unify the mechanism to this. Perhaps I can specify this.

xds/internal/balancer/clusterimpl/picker.go
zasweq: Ah yeah I like refactoring this into a helper. I wanted to combine the writes, but I think where the two writes (unconditional for xDS Labels and here for locality) are the right place to put them.

stats/opentelemetry/client_metrics.go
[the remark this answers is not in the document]
zasweq: "Per call label from a plugin that executes on callpath" discussed offline.

stats/opentelemetry/client_metrics.go
zasweq: Oh whoops, this is a remnant of when I had the default logic below. Deleted.

test/xds/xds_telemetry_labels_test.go
zasweq: Done.

test/xds/xds_telemetry_labels_test.go
zasweq: Ah, good point. Done.

zasweq:
Got to all comments; no approval yet though so assigning back to you.

stats/opentelemetry/client_metrics.go
```diff
@@ -150,7 +150,12 @@ func (h *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo)
 	var labels *istats.Labels
 	if labels = istats.GetLabels(ctx); labels == nil {
 		labels = &istats.Labels{
-			TelemetryLabels: make(map[string]string),
+			// The defaults for all the per call label from a plugin that
```
dfawley, line 153: Nits:

"labelS"

"on _the_ call path"

dfawley (approved):

stats/opentelemetry/client_metrics.go
zasweq: Done for both.

PR #7436, opened 2024-07-23
resolver_wrapper: early return addChannelzTraceEvent on !channelz.IsOn()

jsm:
Note: this is a degradation introduced in this PR: https://github.com/grpc/grpc-go/pull/5192

Which [used to wrap calls](https://github.com/grpc/grpc-go/commit/a73725f42db97964eb538a5cf4d302f760357dbe#diff-f328282147b29224bf45cfe72630d84fa5d9b43b3b976ebb82077b3a322daa53L101) to `addChannelzTraceEvent` in a `if channelz.IsOn() {`

# Problem

Large and frequent calls to this method in UpdateState can cause an extremely high usage of CPU time and Memory due to underlying json.Marshal calls that are not used or needed if channelz is not on.

See alternative approach here: https://github.com/grpc/grpc-go/pull/7437

# Pprofs

image

image

### DIFF
--- a/resolver_wrapper.go
+++ b/resolver_wrapper.go
@@ -177,6 +177,9 @@ func (ccr *ccResolverWrapper) ParseServiceConfig(scJSON string) *serviceconfig.P
 // addChannelzTraceEvent adds a channelz trace event containing the new
 // state received from resolver implementations.
 func (ccr *ccResolverWrapper) addChannelzTraceEvent(s resolver.State) {
+	if !channelz.IsOn() {
+		return
+	}
 	var updates []string
 	var oldSC, newSC *ServiceConfig
 	var oldOK, newOK bool


jsm:
Note: this is a degradation introduced in this PR: https://github.com/grpc/grpc-go/pull/5192


Which [used to wrap calls](https://github.com/grpc/grpc-go/commit/a73725f42db97964eb538a5cf4d302f760357dbe#diff-f328282147b29224bf45cfe72630d84fa5d9b43b3b976ebb82077b3a322daa53L101) to `addChannelzTraceEvent` in a `if channelz.IsOn() {`

dfawley:
@jsm,

> Large and frequent calls to this method in UpdateState

Do you understand why this would be happening?  These types of calls are generally expected to be infrequent, and are often expensive as they result in a full state update of the channel.

jsm:
> @jsm,
> 
> > Large and frequent calls to this method in UpdateState
> 
> Do you understand why this would be happening? These types of calls are generally expected to be infrequent, and are often expensive as they result in a full state update of the channel.

We call this every 10 seconds to update the list of addresses, since they change frequently. That seems like a reasonable usage of this, but perhaps I'm missing something?

jsm:
> These types of calls are generally expected to be infrequent, and are often expensive as they result in a full state update of the channel.

I think regardless, my CPU profile shows that most of it's expense is coming from the addChannelZTraceEvent call (99.7% of the UpdateState call's CPU time in this particular example to be exact), rather than from actually updating the state?

arjan-bal:
Responded in https://github.com/grpc/grpc-go/pull/7437, we can move forward with the other PR.

dfawley:
> We call this every 10 seconds to update the list of addresses, since they change frequently. That seems like a reasonable usage of this, but perhaps I'm missing something?

That doesn't seem particularly bad.

> my CPU profile shows that most of it's expense is coming from the addChannelZTraceEvent call

That is quite bad, but out of curiosity, what is the total effect of this on the binary when you call it every 10s?  Are you seeing it causing other problems, or does total CPU usage spike to a visible amount, etc?

dfawley:
> You can see CPU utilization almost tripled for one of our services

But being 30+ versions apart, these graphs don't help prove that the log you are concerned about is related.

PR #7437, opened 2024-07-23
resolver_wrapper: add early return in addChannelzTraceEvent

jsm:
Note: this is a degradation introduced in this PR: https://github.com/grpc/grpc-go/pull/5192

Which [used to wrap calls](https://github.com/grpc/grpc-go/commit/a73725f42db97964eb538a5cf4d302f760357dbe#diff-f328282147b29224bf45cfe72630d84fa5d9b43b3b976ebb82077b3a322daa53L101) to `addChannelzTraceEvent` in a `if channelz.IsOn() {`

# Problem

Large and frequent calls to this method in UpdateState can cause an extremely high usage of CPU time and Memory due to underlying json.Marshal calls that are not used or needed if were not Info level logging as it is.

Alternative approach to: https://github.com/grpc/grpc-go/pull/7436

# Pprofs

image

image

RELEASE NOTES: none

### DIFF
--- a/resolver_wrapper.go
+++ b/resolver_wrapper.go
@@ -177,6 +177,9 @@ func (ccr *ccResolverWrapper) ParseServiceConfig(scJSON string) *serviceconfig.P
 // addChannelzTraceEvent adds a channelz trace event containing the new
 // state received from resolver implementations.
 func (ccr *ccResolverWrapper) addChannelzTraceEvent(s resolver.State) {
+	if !logger.V(0) && !channelz.IsOn() {
+		return
+	}
 	var updates []string
 	var oldSC, newSC *ServiceConfig
 	var oldOK, newOK bool


jsm:
Note: this is a degradation introduced in this PR: https://github.com/grpc/grpc-go/pull/5192


Which [used to wrap calls](https://github.com/grpc/grpc-go/commit/a73725f42db97964eb538a5cf4d302f760357dbe#diff-f328282147b29224bf45cfe72630d84fa5d9b43b3b976ebb82077b3a322daa53L101) to `addChannelzTraceEvent` in a `if channelz.IsOn() {`

arjan-bal:
Hi @jsm, thanks for reporting the issue and contributing the fix!
The method that is called through `addChannelzTraceEvent` is `AddTraceEvent`:
https://github.com/grpc/grpc-go/blob/0b33bfe7863c548658efdb4eac6e742ef2c8f3de/internal/channelz/trace.go#L182-L204

AddTraceEvent logs the event even if channelz is disabled. To retain the same behaviour, you should combine the conditions from both the PRs: `if !logger.V(0) && !channelz.IsOn()` before returning early.

dfawley (approved):
Logging is typically enabled.  Maybe we should also be looking to remove the use of `pretty.ToJSON` for the purpose of printing this struct?  Would `%+v` or `%#v` not work just as well?

dfawley:
I suspect `%v` is faster than JSON marshalling, but I haven't benchmarked it.  It also should be able to avoid any formatting unless the logging is actually happening.

PR #7439, opened 2024-07-25
balancer/weightedroundrobin: Add emissions of metrics through metrics registry

zasweq:
This PR adds emissions of weighted round robin metrics defined in A78.

RELEASE NOTES: N/A

### DIFF
--- a/balancer/weightedroundrobin/balancer.go
+++ b/balancer/weightedroundrobin/balancer.go
@@ -48,35 +48,35 @@ import (
 const Name = "weighted_round_robin"
 
 var (
-	rrFallbackHandle = estats.RegisterInt64Count(estats.MetricDescriptor{
+	rrFallbackMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
 		Name:           "grpc.lb.wrr.rr_fallback",
-		Description:    "This metric is experimental. Number of scheduler updates in which there were not enough endpoints with valid weight, which caused the WRR policy to fall back to RR behavior.",
+		Description:    "EXPERIMENTAL. Number of scheduler updates in which there were not enough endpoints with valid weight, which caused the WRR policy to fall back to RR behavior.",
 		Unit:           "update",
 		Labels:         []string{"grpc.target"},
 		OptionalLabels: []string{"grpc.lb.locality"},
 		Default:        false,
 	})
 
-	endpointWeightNotYetUsableHandle = estats.RegisterInt64Count(estats.MetricDescriptor{
+	endpointWeightNotYetUsableMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
 		Name:           "grpc.lb.wrr.endpoint_weight_not_yet_usable",
-		Description:    "This metric is experimental. Number of endpoints from each scheduler update that don't yet have usable weight information (i.e., either the load report has not yet been received, or it is within the blackout period).",
+		Description:    "EXPERIMENTAL. Number of endpoints from each scheduler update that don't yet have usable weight information (i.e., either the load report has not yet been received, or it is within the blackout period).",
 		Unit:           "endpoint",
 		Labels:         []string{"grpc.target"},
 		OptionalLabels: []string{"grpc.lb.locality"},
 		Default:        false,
 	})
 
-	endpointWeightStaleHandle = estats.RegisterInt64Count(estats.MetricDescriptor{
+	endpointWeightStaleMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
 		Name:           "grpc.lb.wrr.endpoint_weight_stale",
-		Description:    "This metric is experimental. Number of endpoints from each scheduler update whose latest weight is older than the expiration period.",
+		Description:    "EXPERIMENTAL. Number of endpoints from each scheduler update whose latest weight is older than the expiration period.",
 		Unit:           "endpoint",
 		Labels:         []string{"grpc.target"},
 		OptionalLabels: []string{"grpc.lb.locality"},
 		Default:        false,
 	})
-	endpointWeightsHandle = estats.RegisterFloat64Histo(estats.MetricDescriptor{
+	endpointWeightsMetric = estats.RegisterFloat64Histo(estats.MetricDescriptor{
 		Name:           "grpc.lb.wrr.endpoint_weights",
-		Description:    "This metric is experimental. Weight of each endpoint, recorded on every scheduler update. Endpoints without usable weights will be recorded as weight 0.",
+		Description:    "EXPERIMENTAL. Weight of each endpoint, recorded on every scheduler update. Endpoints without usable weights will be recorded as weight 0.",
 		Unit:           "endpoint",
 		Labels:         []string{"grpc.target"},
 		OptionalLabels: []string{"grpc.lb.locality"},
@@ -143,7 +143,7 @@ func (bb) Name() string {
 
 // wrrBalancer implements the weighted round robin LB policy.
 type wrrBalancer struct {
-	// These fields are set during construction and read only after.
+	// The following fields are immutable.
 	cc              balancer.ClientConn
 	logger          *grpclog.PrefixLogger
 	target          string
@@ -171,7 +171,7 @@ func (b *wrrBalancer) UpdateClientConnState(ccs balancer.ClientConnState) error 
 	}
 
 	b.cfg = cfg
-	b.locality = weightedtarget.GetLocality(ccs.ResolverState)
+	b.locality = weightedtarget.LocalityFromResolverState(ccs.ResolverState)
 	b.updateAddresses(ccs.ResolverState.Addresses)
 
 	if len(ccs.ResolverState.Addresses) == 0 {
@@ -394,18 +394,17 @@ type picker struct {
 	cfg       *lbConfig          // active config when picker created
 	subConns  []*weightedSubConn // all READY subconns
 
-	// These fields are built out at creation time and read only after, so no
-	// need for synchronization.
+	// The following fields are immutable.
 	target          string
 	locality        string
 	metricsRecorder estats.MetricsRecorder
 }
 
-func (p *picker) scWeights() []float64 {
+func (p *picker) scWeights(recordMetrics bool) []float64 {
 	ws := make([]float64, len(p.subConns))
 	now := internal.TimeNow()
 	for i, wsc := range p.subConns {
-		ws[i] = wsc.weight(now, time.Duration(p.cfg.WeightExpirationPeriod), time.Duration(p.cfg.BlackoutPeriod), true)
+		ws[i] = wsc.weight(now, time.Duration(p.cfg.WeightExpirationPeriod), time.Duration(p.cfg.BlackoutPeriod), recordMetrics)
 	}
 
 	return ws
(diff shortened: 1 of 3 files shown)


balancer/weightedroundrobin/balancer.go
```diff
@@ -45,6 +47,43 @@ import (
 // Name is the name of the weighted round robin balancer.
 const Name = "weighted_round_robin"
 
+var (
+	rrFallbackHandle = estats.RegisterInt64Count(estats.MetricDescriptor{
```
dfawley, line 51: Nit: `rrFallbackMetric` or `Instrument` or something more concrete?  `Handle` doesn't describe it.

balancer/weightedroundrobin/balancer.go
```diff
@@ -45,6 +47,43 @@ import (
...
+		Name:           "grpc.lb.wrr.rr_fallback",
+		Description:    "This metric is experimental. Number of scheduler updates in which there were not enough endpoints with valid weight, which caused the WRR policy to fall back to RR behavior.",
```
dfawley, line 53: FWIW I preferred Java's simpler, all-caps `EXPERIMENTAL. ___`

balancer/weightedroundrobin/balancer.go
```diff
@@ -339,16 +393,21 @@ type picker struct {
 	v         uint32             // incrementing value used by the scheduler; accessed atomically
 	cfg       *lbConfig          // active config when picker created
 	subConns  []*weightedSubConn // all READY subconns
+
+	// These fields are built out at creation time and read only after, so no
+	// need for synchronization.
```
dfawley, line 398: `// The following fields are immutable`?

balancer/weightedroundrobin/balancer.go
```diff
@@ -339,16 +393,21 @@ type picker struct {
...
+	target          string
+	locality        string
+	metricsRecorder estats.MetricsRecorder
 }
 
-// scWeights returns a slice containing the weights from p.subConns in the same
-// order as p.subConns.
 func (p *picker) scWeights() []float64 {
```
dfawley, line 404: This should take the boolean and pass it through, probably, since it will cause metrics to be recorded, but doesn't seem like it would from its name.

balancer/weightedroundrobin/balancer.go
```diff
@@ -404,8 +464,13 @@ func (p *picker) Pick(info balancer.PickInfo) (balancer.PickResult, error) {
 // When needed, it also tracks connectivity state, listens for metrics updates
 // by implementing the orca.OOBListener interface and manages that listener.
 type weightedSubConn struct {
+	// These fields are set at construction time and read only after, so no need
+	// for synchronization.
```
dfawley, line 468: `// The following fields are immutable.`?

balancer/weightedroundrobin/balancer.go
```diff
@@ -527,21 +592,35 @@ func (w *weightedSubConn) updateConnectivityState(cs connectivity.State) connect
 
 // weight returns the current effective weight of the subconn, taking into
 // account the parameters.  Returns 0 for blacked out or expired data, which
-// will cause the backend weight to be treated as the mean of the weights of
-// the other backends.
-func (w *weightedSubConn) weight(now time.Time, weightExpirationPeriod, blackoutPeriod time.Duration) float64 {
+// will cause the backend weight to be treated as the mean of the weights of the
+// other backends. If forScheduler is set to true, this function will emit
+// metrics through the mtrics registry.
+func (w *weightedSubConn) weight(now time.Time, weightExpirationPeriod, blackoutPeriod time.Duration, forScheduler bool) (weight float64) {
```
dfawley, line 598: Name the variable `recordWeightMetrics`?

balancer/weightedroundrobin/balancer.go
```diff
@@ -527,21 +592,35 @@ func (w *weightedSubConn) updateConnectivityState(cs connectivity.State) connect
...
 	w.mu.Lock()
 	defer w.mu.Unlock()
+
+	weight = 0
+	if forScheduler {
+		defer endpointWeightsHandle.Record(w.metricsRecorder, weight, []string{w.target, w.locality}...)
```
dfawley, line 604: I don't think this works?  I think `weight` gets captured immediately.  You would need to wrap in a `func()` if you want to do it this way.

balancer/weightedroundrobin/balancer.go
```diff
@@ -527,21 +592,35 @@ func (w *weightedSubConn) updateConnectivityState(cs connectivity.State) connect
...
 	// If the most recent update was longer ago than the expiration period,
 	// reset nonEmptySince so that we apply the blackout period again if we
 	// start getting data again in the future, and return 0.
 	if now.Sub(w.lastUpdated) >= weightExpirationPeriod {
+		if forScheduler {
+			endpointWeightStaleHandle.Record(w.metricsRecorder, 1, []string{w.target, w.locality}...)
+		}
 		w.nonEmptySince = time.Time{}
-		return 0
+		return
```
dfawley, line 615: Please (almost?) never use bare returns from functions that return values.  It's too hard to read.  Use `return weight` or, even better, keep as `return 0`.

balancer/weightedroundrobin/balancer.go
dfawley, line 604: You shouldn't ever do
```go
f(x, y, []string{a, b}...)
```

It should be:

```go
f(x, y, a, b)
```

balancer/weightedtarget/weightedtarget.go
```diff
@@ -84,6 +84,17 @@ type weightedTargetBalancer struct {
 	targets map[string]Target
 }
 
+type localityKeyType string
+
+const localityKey = localityKeyType("locality")
+
+// GetLocality returns the locality from the resolver.State provided, or an
+// empty string if not present.
+func GetLocality(state resolver.State) string {
```
dfawley, line 93: Does this belong inside this LB policy or elsewhere?  Is it strictly a part of `weightedtarget`'s behavior or might other policies want to be able to set it?

Also let's avoid `Get`.

zasweq: I think it has to be part of this LB policy, as per this line in the gRFC: "To support the locality label in the WRR metrics, we will extend the weighted_target LB policy (see [A28](https://github.com/grpc/proposal/blob/master/A28-xds-traffic-splitting-and-routing.md)) to define a resolver attribute that indicates the name of its child."

balancer/weightedroundrobin/balancer.go
zasweq: Ah sounds good noted. Switched for all callsites.

balancer/weightedroundrobin/balancer.go
zasweq: Haha, ok switched to that.

balancer/weightedroundrobin/balancer.go
zasweq: I'll return weight so reader doesn't need to keep track of two pieces of information (named return value which gets logged and the actual return value).

balancer/weightedroundrobin/balancer.go
zasweq: Oh wait there actually is a correctness issue here since I don't set weight before returning for the case it passes the two conditionals that mean weight = 0 (blackout/weight expiration period).

balancer/weightedtarget/weightedtarget.go
zasweq: Renamed LocalityFromResolverState to keep it inline with what we called our helper to get XDSClient from resolver state.

balancer/weightedroundrobin/balancer.go
zasweq: Ah haha yeah this is wrong wrapped with a func().

balancer/weightedroundrobin/balancer.go
zasweq: Done.

balancer/weightedroundrobin/balancer.go
zasweq: Sure this is how I see it documented throughout the codebase in other places but it's equivalent to yours so switched.

balancer/weightedroundrobin/balancer.go
zasweq: Sure this is how I see it documented throughout the codebase in other places but it's equivalent to yours so switched for all three structs I added this comment for.

balancer/weightedroundrobin/balancer.go
zasweq: Chose the former, xxxMetric for all 4 handles.

zasweq:
Thanks for the pass :)! Got to all comments!

balancer/weightedroundrobin/balancer.go
dfawley: IMO this is worse because now you have to go backtrack and figure out what `weight` might be while reading the code.  If it just says `return 0` then it's obvious.

Note that named return values are set upon returning.  I.e.

```go
func f() (x int) {
	x = 5
	defer func() {
		fmt.Println(x)
	}()

	return 10
}
```

Will print `10`.

balancer/weightedroundrobin/balancer.go
```diff
@@ -595,32 +593,34 @@ func (w *weightedSubConn) updateConnectivityState(cs connectivity.State) connect
...
 // metrics through the mtrics registry.
-func (w *weightedSubConn) weight(now time.Time, weightExpirationPeriod, blackoutPeriod time.Duration, forScheduler bool) (weight float64) {
+func (w *weightedSubConn) weight(now time.Time, weightExpirationPeriod, blackoutPeriod time.Duration, recordMetrics bool) float64 {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
-	weight = 0
-	if forScheduler {
-		defer endpointWeightsHandle.Record(w.metricsRecorder, weight, []string{w.target, w.locality}...)
+	weight := float64(0)
```
dfawley, line 600: _Named return values are totally fine_ if you want to keep it.  Just don't do a bare `return`, since it can be confusing (for a minute I actually thought you removed the return value when I saw the bare return, and had to go back to see what was happening).

(But you don't need to initialize them to 0 as in the previous code; Go always initializes everything to zero.)

balancer/weightedroundrobin/balancer.go
```diff
@@ -595,32 +593,34 @@ func (w *weightedSubConn) updateConnectivityState(cs connectivity.State) connect
...
 	if blackoutPeriod != 0 && (w.nonEmptySince == (time.Time{}) || now.Sub(w.nonEmptySince) < blackoutPeriod) {
-		endpointWeightNotYetUsableHandle.Record(w.metricsRecorder, 1, []string{w.target, w.locality}...)
+		endpointWeightNotYetUsableMetric.Record(w.metricsRecorder, 1, w.target, w.locality)
 
-		return
+		return weight
 	}
-
-	return w.weightVal
+	weight = w.weightVal
```
dfawley, line 624: Yeah...I'd prefer to use what you had before.  Use the named return value, and don't ever assign to `weight` at all.  Just read it in the defer to find out what was returned.

balancer/weightedroundrobin/balancer.go
zasweq: Ah ok I didn't realize named return values were set upon returning. Switched to incorporate all your suggestions.

balancer/weightedroundrobin/balancer.go
zasweq: Switched to named return value.

balancer/weightedroundrobin/balancer.go
zasweq: Switched to named return value.

dfawley (approved):

PR #7450, opened 2024-07-29
examples/retry: remove waitForReady from service config

### LINKED ISSUE #7440: Is waitForReady required for retry policies?
Hi there! I'm having a little bit of a hard time wrapping my head around the retry policies. One thing I've noticed is that all of the examples seem to set `waitForReady` to true in the method config. Is this required for retries to occur? Without setting this to true, all of my (expectedly broken) connection attempts seem to immediately fail otherwise.

Is the idea that, with `waitForReady` enabled, any configured retries are occurring internally while I wait for the RPC connection? And if I use `waitForReady` without a retry policy and an error occurs, must I wait for a timeout or explicit cancellation before I'd see that error?

Thanks!

dfawley:
This setting is irrelevant, so remove it to avoid any confusion.

Fixes #7440 

RELEASE NOTES: none

### DIFF
--- a/examples/features/retry/README.md
+++ b/examples/features/retry/README.md
@@ -41,7 +41,6 @@ RetryableStatusCodes: Retry only when receiving these status codes.
             "methodConfig": [{
                 // config per method or all methods under service
                 "name": [{"service": "grpc.examples.echo.Echo"}],
-                "waitForReady": true,
 
                 "retryPolicy": {
                     "MaxAttempts": 4,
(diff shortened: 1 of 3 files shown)


easwars (approved):

purnesh42H:
We should update in these two places as well 
- https://github.com/grpc/grpc-go/blob/master/test/retry_test.go
- https://github.com/grpc/grpc-go/blob/3eb01450ff91253bfc67a6b5b3c258ea2fc245c8/examples/features/retry/README.md?plain=1#L44  

Also, might be good to put a note in documentation that this setting is not relevant for retry policies

dfawley:
IIRC, in the tests, WFR is used as a trick to detect that the service config was properly applied by the client.  I'll update the corresponding doc for the example and look at the WFR docstring.

purnesh42H (approved):
lgtm

Issue #2248, opened 2018-07-31
http2_client is missing tests


jadekler:
The following cases are untested, and should be tested:

- https://github.com/grpc/grpc-go/blob/fb6867e42b4743663f07f4c7ce6c29c8d01247e1/internal/transport/http2_client.go#L275-L278

- https://github.com/grpc/grpc-go/blob/fb6867e42b4743663f07f4c7ce6c29c8d01247e1/internal/transport/http2_client.go#L279-L282

- https://github.com/grpc/grpc-go/blob/fb6867e42b4743663f07f4c7ce6c29c8d01247e1/internal/transport/http2_client.go#L298-L301

- https://github.com/grpc/grpc-go/blob/fb6867e42b4743663f07f4c7ce6c29c8d01247e1/internal/transport/http2_client.go#L304-L307

e.g. changes in https://github.com/grpc/grpc-go/pull/2219 were almost merged that should not have because of lack of the aforementioned test coverage.

v-sreejith:
@ginayeh I would like to contribute here

arvindbr8:
Sure! Assigning it to you. 

The original issue description might not be relevant anymore. I would re-evaluate the missed coverage of the file [here](https://app.codecov.io/gh/grpc/grpc-go/blob/master/internal%2Ftransport%2Fhttp2_server.go).

arvindbr8:
@v-sreejith -- ping.. Are you still actively tracking this?

hanut19:
@purnesh42H assign this issue to me

hanut19:
@purnesh42H, for reference which file contains other test cases for this function "func newHTTP2Client(connectCtx, ctx context.Context, addr resolver.Address, opts ConnectOptions, onClose func(GoAwayReason)) (_ *http2Client, err error) "
eg. name these other test cases are.

hanut19:
@purnesh42H 
In order to cover unit test for below condition (ref. screenshot). I need to mock the “Write” function which is defined “net.go” of net package. 

<img width="1681" alt="6" src="https://github.com/grpc/grpc-go/assets/50198451/0e9ff033-99cc-4cc8-ba06-03a6b2a42890">



I have written below unit test for cover above condition. I tried to mock “Write” function with below approach. 

 
<img width="1681" alt="5" src="https://github.com/grpc/grpc-go/assets/50198451/4069b615-76d0-4aa8-9d37-3f7daf9f7af5">


Using that approach unable to mock the “Write”. Get below error
[...]

PR #7462, opened 2024-07-31
transport: add missing tests for http2_client

hanut19:
Fix for: https://github.com/grpc/grpc-go/issues/2248

RELEASE NOTES: None

### DIFF
--- a/internal/transport/http2_client_test.go
+++ b/internal/transport/http2_client_test.go
@@ -53,7 +53,7 @@ type framerWindowUpdateConn struct {
 
 func (cp *clientPrefaceConn) Write(b []byte) (n int, err error) {
 	if string(b) == string(ClientPreface) {
-		return 0, errors.New("preface write error")
+		return 0, errors.New("force error for preface write")
 	}
 
 	// Normally write the bytes if they don't match the ClientPreface
@@ -93,7 +93,7 @@ func dialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error)
 
 func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
 	if string(b) == framerWriteSettingsString {
-		return 0, errors.New("intended test case string detected: forced write error")
+		return 0, errors.New("force error for framerWriteSettings")
 	}
 	framerValue := 9
 	n, err = fws.Conn.Write(b)
@@ -115,7 +115,7 @@ func dialerFramerWriteSettings(_ context.Context, addr string) (net.Conn, error)
 
 func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {
 	if string(b) == framerWindowUpdateString {
-		return 0, errors.New("intended test case string detected: forced write error")
+		return 0, errors.New("force error for framerWindowUpdate")
 	}
 	// Simulate a WINDOW_UPDATE frame's value (window size in bytes)
 	windowUpdateValue := 13
@@ -145,7 +145,7 @@ func (s) TestNewHTTP2ClientTarget(t *testing.T) {
 		{
 			name:     "client-preface-write",
 			opts:     ConnectOptions{Dialer: dialerClientPrefaceWrite},
-			expected: "connection error: desc = \"transport: failed to write client preface: preface write error\"",
+			expected: "connection error: desc = \"transport: failed to write client preface: force error for preface write\"",
 		},
 		{
 			name:     "client-preface-length",
--- a/internal/transport/http2_client_test.go
+++ b/internal/transport/http2_client_test.go
@@ -76,7 +76,7 @@ func dialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error)
 
 func (hc *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
 	n, err = hc.Conn.Write(b)
-	//compare framer value
+	// Compare the number of bytes written with the framer value
 	if n == 9 {
 		return 0, errors.New("Framer write setting error")
 	}
@@ -94,7 +94,7 @@ func dialerFramerWriteSettings(_ context.Context, addr string) (net.Conn, error)
 func (hc *framerWriteWindowUpdateConn) Write(b []byte) (n int, err error) {
 
 	n, err = hc.Conn.Write(b)
-	// compare for windowupdate value
+	// Compare the number of bytes written with the windowupdate value
 	if n == 13 {
 		return 0, errors.New("Framer write windowupdate error")
 	}
--- a/internal/transport/http2_client_test.go
+++ b/internal/transport/http2_client_test.go
@@ -27,27 +27,27 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
-type clientPrefaceWrite struct {
+type clientPrefaceConn struct {
 	net.Conn
 }
 
-type clientPrefaceLength struct {
+type clientPrefaceLengthConn struct {
 	net.Conn
 }
 
-type framerWriteSettings struct {
+type framerWriteSettingsConn struct {
 	net.Conn
 }
 
-type framerWriteWindowUpdate struct {
+type framerWriteWindowUpdateConn struct {
 	net.Conn
 }
 
-func (hc *clientPrefaceWrite) Write(b []byte) (n int, err error) {
+func (hc *clientPrefaceConn) Write(b []byte) (n int, err error) {
 	return 0, errors.New("preface write error")
 }
 
-func (hc *clientPrefaceWrite) Close() error {
+func (hc *clientPrefaceConn) Close() error {
 	return hc.Conn.Close()
 }
 
@@ -56,10 +56,10 @@ func dialerClientPrefaceWrite(_ context.Context, addr string) (net.Conn, error) 
 	if err != nil {
 		return nil, err
 	}
-	return &clientPrefaceWrite{Conn: conn}, nil
+	return &clientPrefaceConn{Conn: conn}, nil
 }
 
-func (hc *clientPrefaceLength) Write(b []byte) (n int, err error) {
+func (hc *clientPrefaceLengthConn) Write(b []byte) (n int, err error) {
 
 	incorrectPreface := "INCORRECT PREFACE\r\n\r\n"
 	n, err = hc.Conn.Write([]byte(incorrectPreface))
@@ -71,10 +71,10 @@ func dialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error)
 	if err != nil {
 		return nil, err
 	}
-	return &clientPrefaceLength{Conn: conn}, nil
+	return &clientPrefaceLengthConn{Conn: conn}, nil
 }
 
-func (hc *framerWriteSettings) Write(b []byte) (n int, err error) {
+func (hc *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
 	n, err = hc.Conn.Write(b)
 	//compare framer value
 	if n == 9 {
@@ -88,10 +88,10 @@ func dialerFramerWriteSettings(_ context.Context, addr string) (net.Conn, error)
 	if err != nil {
 		return nil, err
 	}
-	return &framerWriteSettings{Conn: conn}, nil
+	return &framerWriteSettingsConn{Conn: conn}, nil
 }
 
-func (hc *framerWriteWindowUpdate) Write(b []byte) (n int, err error) {
+func (hc *framerWriteWindowUpdateConn) Write(b []byte) (n int, err error) {
 
 	n, err = hc.Conn.Write(b)
 	// compare for windowupdate value
@@ -106,105 +106,56 @@ func dialerFramerWriteWindowUpdate(_ context.Context, addr string) (net.Conn, er
 	if err != nil {
 		return nil, err
 	}
-	return &framerWriteWindowUpdate{Conn: conn}, nil
+	return &framerWriteWindowUpdateConn{Conn: conn}, nil
 }
 
-func TestNewHTTP2ClientPrefaceFailure(t *testing.T) {
-
-	// Create a server.
-	lis, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatalf("Error while listening: %v", err)
+func (s) TestNewHTTP2ClientTarget(t *testing.T) {
+	tests := []struct {
+		name     string
+		opts     ConnectOptions
+		expected string
+	}{
+		{
+			name:     "client-preface-write",
+			opts:     ConnectOptions{Dialer: dialerClientPrefaceWrite},
+			expected: "connection error: desc = \"transport: failed to write client preface: preface write error\"",
+		},
+		{
+			name:     "client-preface-length",
+			opts:     ConnectOptions{Dialer: dialerClientPrefaceLength},
+			expected: "connection error: desc = \"transport: preface mismatch, wrote 21 bytes; want 24\"",
+		},
+		{
+			name:     "framer-write-settings",
+			opts:     ConnectOptions{Dialer: dialerFramerWriteSettings},
+			expected: "connection error: desc = \"transport: failed to write initial settings frame: Framer write setting error\"",
+		},
+		{
+			name:     "framer-write-windowUpdate",
+			opts:     ConnectOptions{Dialer: dialerFramerWriteWindowUpdate, InitialConnWindowSize: 80000},
+			expected: "connection error: desc = \"transport: failed to write window update: Framer write windowupdate error\"",
+		},
 	}
-	defer lis.Close()
-	defer func() {
-		if r := recover(); r != nil {
-			t.Fatalf("TestNewHTTP2ClientPrefaceFailure panicked: %v", r)
-		}
-	}()
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
-	defer cancel()
-	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerClientPrefaceWrite}, func(GoAwayReason) {})
-	if err == nil {
-		t.Error("Expected an error, but got nil")
-	} else {
-		if err.Error() != "connection error: desc = \"transport: failed to write client preface: preface write error\"" {
-			t.Fatalf("Error while creating client transport: %v", err)
-		}
-	}
-}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
 
-func TestNewHTTP2ClientPrefaceLengthFailure(t *testing.T) {
-	// Create a server.
-	lis, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatalf("Error while listening: %v", err)
-	}
-	defer lis.Close()
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
-	defer cancel()
-	defer func() {
-		if r := recover(); r != nil {
-			t.Fatalf("TestNewHTTP2ClientPrefaceLengthFailure panicked: %v", r)
-		}
-	}()
+			// Create a server.
+			lis, err := net.Listen("tcp", "localhost:0")
+			if err != nil {
+				t.Fatalf("Error while listening: %v", err)
+			}
+			defer lis.Close()
+			ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+			defer cancel()
 
-	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerClientPrefaceLength}, func(GoAwayReason) {})
-	if err == nil {
-		t.Errorf("Expected an error, but got nil")
-	} else {
-		if err.Error() != "connection error: desc = \"transport: preface mismatch, wrote 21 bytes; want 24\"" {
-			t.Fatalf("Error while creating client transport: %v", err)
-		}
-	}
-}
-
-func TestNewHTTP2ClientFramerWriteSettingsFailure(t *testing.T) {
-	// Create a server.
-	lis, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatalf("Error while listening: %v", err)
-	}
-	defer lis.Close()
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
-	defer cancel()
-	defer func() {
-		if r := recover(); r != nil {
-			t.Fatalf("TestNewHTTP2ClientFramerWriteSettingsFailure panicked: %v", r)
-		}
-	}()
-
-	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerFramerWriteSettings}, func(GoAwayReason) {})
-	if err == nil {
-		t.Errorf("Expected an error, but got nil")
-	} else {
-		if err.Error() != "connection error: desc = \"transport: failed to write initial settings frame: Framer write setting error\"" {
-			t.Fatalf("Error while creating client transport: %v", err)
-		}
-	}
-}
-
-func TestNewHTTP2ClientFramerWriteWindowUpdateFailure(t *testing.T) {
-	// Create a server.
-	lis, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatalf("Error while listening: %v", err)
-	}
-	defer lis.Close()
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
-	defer cancel()
-	defer func() {
-		if r := recover(); r != nil {
-			t.Fatalf("TestNewHTTP2ClientFramerWriteWindowUpdateFailure panicked: %v", r)
-		}
-	}()
-
-	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerFramerWriteWindowUpdate, InitialConnWindowSize: 80000}, func(GoAwayReason) {})
-	if err == nil {
-		t.Errorf("Expected an error, but got nil")
-	} else {
-		if err.Error() != "connection error: desc = \"transport: failed to write window update: Framer write windowupdate error\"" {
-			t.Fatalf("Error while creating client transport: %v", err)
-		}
+			_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, test.opts, func(GoAwayReason) {})
+			if err == nil {
+				t.Errorf("Expected an error, but got nil")
+			} else {
+				if err.Error() != test.expected {
+					t.Fatalf("TestNewHTTP2ClientTarget() = %v, want %s", err.Error(), test.expected)
+				}
+			}
+		})
 	}
 }
--- a/internal/transport/http2_client_test.go
+++ b/internal/transport/http2_client_test.go
@@ -1,3 +1,20 @@
+/*
+ *
+ * Copyright 2024 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
 package transport
 
 import (
@@ -10,24 +27,36 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
-type hangingClientPreface struct {
+type clientPrefaceWrite struct {
 	net.Conn
 }
 
-func (hc *hangingClientPreface) Write(b []byte) (n int, err error) {
+type clientPrefaceLength struct {
+	net.Conn
+}
+
+type framerWriteSettings struct {
+	net.Conn
+}
+
+type framerWriteWindowUpdate struct {
+	net.Conn
+}
+
+func (hc *clientPrefaceWrite) Write(b []byte) (n int, err error) {
 	return 0, errors.New("preface write error")
 }
 
-func (hc *hangingClientPreface) Close() error {
+func (hc *clientPrefaceWrite) Close() error {
 	return hc.Conn.Close()
 }
 
-func hangingDialerClientPreface(_ context.Context, addr string) (net.Conn, error) {
+func dialerClientPrefaceWrite(_ context.Context, addr string) (net.Conn, error) {
 	conn, err := net.Dial("tcp", addr)
 	if err != nil {
 		return nil, err
 	}
-	return &hangingClientPreface{Conn: conn}, nil
+	return &clientPrefaceWrite{Conn: conn}, nil
 }
 
 func TestNewHTTP2ClientPrefaceFailure(t *testing.T) {
@@ -45,35 +74,32 @@ func TestNewHTTP2ClientPrefaceFailure(t *testing.T) {
 	}()
 	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
 	defer cancel()
-	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: hangingDialerClientPreface}, func(GoAwayReason) {})
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerClientPrefaceWrite}, func(GoAwayReason) {})
+	if err == nil {
+		t.Error("Expected an error, but got nil")
+	}
 	if err != nil {
 		if err.Error() != "connection error: desc = \"transport: failed to write client preface: preface write error\"" {
 			t.Fatalf("Error while creating client transport: %v", err)
 		}
 	}
-	if err == nil {
-		t.Error("Expected an error, but got nil")
-	}
 }
 
-type hangingClientPrefaceLength struct {
-	net.Conn
-}
-
-func (hc *hangingClientPrefaceLength) Write(b []byte) (n int, err error) {
+func (hc *clientPrefaceLength) Write(b []byte) (n int, err error) {
 
 	incorrectPreface := "INCORRECT PREFACE\r\n\r\n"
 	n, err = hc.Conn.Write([]byte(incorrectPreface))
 	return n, err
 }
 
-func hangingDialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error) {
+func dialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error) {
 	conn, err := net.Dial("tcp", addr)
 	if err != nil {
 		return nil, err
 	}
-	return &hangingClientPrefaceLength{Conn: conn}, nil
+	return &clientPrefaceLength{Conn: conn}, nil
 }
+
 func TestNewHTTP2ClientPrefaceLengthFailure(t *testing.T) {
 	// Create a server.
 	lis, err := net.Listen("tcp", "localhost:0")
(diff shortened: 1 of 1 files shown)


internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func hangingDialerFramerWriteWindowUpdate(_ context.Context, addr string) (net.Conn, error) {
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &hangingFramerWriteWindowUpdate{Conn: conn}, nil
+}
+func TestNewHTTP2ClientFramerWriteWindowUpdateFailure(t *testing.T) {
```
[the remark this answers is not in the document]
hanut19: gaped between functions done

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"context"
+	"errors"
+	"net"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc/resolver"
+)
+
+type hangingClientPreface struct {
```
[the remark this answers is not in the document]
hanut19: name with hanging removed

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+type hangingFramerWriteSettings struct {
+	net.Conn
+}
+
+func (hc *hangingFramerWriteSettings) Write(b []byte) (n int, err error) {
+
+	n, err = hc.Conn.Write(b)
+	if n == 9 {
```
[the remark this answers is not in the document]
hanut19: inline comment added.

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+type hangingFramerWriteWindowUpdate struct {
+	net.Conn
+}
+
+func (hc *hangingFramerWriteWindowUpdate) Write(b []byte) (n int, err error) {
+
+	n, err = hc.Conn.Write(b)
+	if n == 13 {
```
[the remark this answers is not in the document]
hanut19: inline comment added and struct moved to top

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}()
+
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: hangingDialerFramerWriteSettings}, func(GoAwayReason) {})
+	if err != nil {
+		if err.Error() != "connection error: desc = \"transport: failed to write initial settings frame: Framer write setting error\"" {
+			t.Fatalf("Error while creating client transport: %v", err)
+		}
+	}
+	if err == nil {
```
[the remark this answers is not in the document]
hanut19: done

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,214 @@
...
+			t.Fatalf("TestNewHTTP2ClientPrefaceFailure panicked: %v", r)
+		}
+	}()
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+	defer cancel()
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerClientPrefaceWrite}, func(GoAwayReason) {})
+	if err == nil {
+		t.Error("Expected an error, but got nil")
+	}
+	if err != nil {
```
purnesh42H, line 81: nit: you don't have to check for err being not nil because you already checked that above

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,214 @@
...
+		if err.Error() != "connection error: desc = \"transport: failed to write client preface: preface write error\"" {
+			t.Fatalf("Error while creating client transport: %v", err)
+		}
+	}
+}
+
+func (hc *clientPrefaceLength) Write(b []byte) (n int, err error) {
```
purnesh42H, line 88: please move all the struct and their methods up before the test functions

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	n, err = hc.Conn.Write(b)
+	// compare for windowupdate value
+	if n == 13 {
+		return 0, errors.New("Framer write windowupdate error")
+	}
+	return n, err
+}
+
+func dialerFramerWriteWindowUpdate(_ context.Context, addr string) (net.Conn, error) {
```
purnesh42H, line 182: same: All helper methods should also be defined up before tests

internal/transport/http2_client_test.go
hanut19: @purnesh42H  I am not able to get this can you explain a bit more. Here we are handling error from NewClientTransport.

internal/transport/http2_client_test.go
hanut19: methods moved

internal/transport/http2_client_test.go
hanut19: helper methos moved

internal/transport/http2_client_test.go
purnesh42H: on line 78, you have already checked `if err == nil` so you don't have to check for `nil` again on line 81

hanut19: change done

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,210 @@
...
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &framerWriteWindowUpdate{Conn: conn}, nil
+}
+
+func TestNewHTTP2ClientPrefaceFailure(t *testing.T) {
```
purnesh42H, line 112: it looks like the test cases are mostly same except the net conn difference and error assertions. Can you convert the tests to table driven format like here https://github.com/grpc/grpc-go/blob/e6b6318ad90847fa8b5ed987b964f1494179c89f/clientconn_test.go#L811

See https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks for more details

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}()
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+	defer cancel()
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerClientPrefaceWrite}, func(GoAwayReason) {})
+	if err == nil {
+		t.Error("Expected an error, but got nil")
+	} else {
+		if err.Error() != "connection error: desc = \"transport: failed to write client preface: preface write error\"" {
+			t.Fatalf("Error while creating client transport: %v", err)
```
purnesh42H, line 132: line 132 has to follow assertion convention. The important part is got before want. See https://google.github.io/styleguide/go/decisions#got-before-want

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,210 @@
...
+			t.Fatalf("TestNewHTTP2ClientPrefaceLengthFailure panicked: %v", r)
+		}
+	}()
+
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerClientPrefaceLength}, func(GoAwayReason) {})
+	if err == nil {
+		t.Errorf("Expected an error, but got nil")
+	} else {
+		if err.Error() != "connection error: desc = \"transport: preface mismatch, wrote 21 bytes; want 24\"" {
+			t.Fatalf("Error while creating client transport: %v", err)
```
purnesh42H, line 157: same as line 132. Error message should state got before want

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,210 @@
...
+			t.Fatalf("TestNewHTTP2ClientFramerWriteSettingsFailure panicked: %v", r)
+		}
+	}()
+
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerFramerWriteSettings}, func(GoAwayReason) {})
+	if err == nil {
+		t.Errorf("Expected an error, but got nil")
+	} else {
+		if err.Error() != "connection error: desc = \"transport: failed to write initial settings frame: Framer write setting error\"" {
+			t.Fatalf("Error while creating client transport: %v", err)
```
purnesh42H, line 182: same as above

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,210 @@
...
+			t.Fatalf("TestNewHTTP2ClientFramerWriteWindowUpdateFailure panicked: %v", r)
+		}
+	}()
+
+	_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, ConnectOptions{Dialer: dialerFramerWriteWindowUpdate, InitialConnWindowSize: 80000}, func(GoAwayReason) {})
+	if err == nil {
+		t.Errorf("Expected an error, but got nil")
+	} else {
+		if err.Error() != "connection error: desc = \"transport: failed to write window update: Framer write windowupdate error\"" {
+			t.Fatalf("Error while creating client transport: %v", err)
```
purnesh42H, line 207: same as above

internal/transport/http2_client_test.go
hanut19: done

internal/transport/http2_client_test.go
hanut19: done

internal/transport/http2_client_test.go
hanut19: done

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"context"
+	"errors"
+	"net"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc/resolver"
+)
+
+type clientPrefaceWrite struct {
```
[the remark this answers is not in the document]
hanut19: suffix added

internal/transport/http2_client_test.go
hanut19: error change done

internal/transport/http2_client_test.go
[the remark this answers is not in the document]
purnesh42H: `//compare framer value` doesn't explain much. Please be explicit. Also, don't resolve comments by yourself. We follow the practice of reviewer resolving comments

internal/transport/http2_client_test.go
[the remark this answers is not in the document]
purnesh42H: suggestion: `throw error if frame is a window update`

internal/transport/http2_client_test.go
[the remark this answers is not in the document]
hanut19: linline comment added

purnesh42H:
Please modify the test cases mock conn’s to throw error only in case of test condition like client preface

internal/transport/http2_client_test.go
```diff
@@ -74,10 +73,12 @@ func dialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error)
 	return &clientPrefaceLengthConn{Conn: conn}, nil
 }
 
-func (hc *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
-	n, err = hc.Conn.Write(b)
+func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
+	framerValue := 9
+	n, err = fws.Conn.Write(b)
```
purnesh42H, line 78: @hanut19 please do `string(b)` and compare the actual string and throw the error if intended test case string is sent for write.

Please do this for all the 4 test cases

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,164 @@
...
+type framerWriteSettingsConn struct {
+	net.Conn
+}
+
+type framerWindowUpdateConn struct {
+	net.Conn
+}
+
+func (cp *clientPrefaceConn) Write(b []byte) (n int, err error) {
+	return 0, errors.New("preface write error")
```
purnesh42H, line 47: In this case you need to check if `b` is a byte slice representing `ClientPreface` string and throw the error only then

internal/transport/http2_client_test.go
hanut19: done

internal/transport/http2_client_test.go
hanut19: please check done

internal/transport/http2_client_test.go
```diff
@@ -27,6 +27,11 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
+const IntendedTestCaseString = "This is a test case string"
+
+// ClientPreface is the HTTP/2 client preface string.
+var ClientPreface = []byte("PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n")
```
purnesh42H, line 33: For ClientPreface assignment you can do simply this since the string is already defined https://github.com/grpc/grpc-go/blob/ced812e3287e15a009eab5b271c25750050a2f82/internal/transport/http_util.go#L50

internal/transport/http2_client_test.go
```diff
@@ -27,6 +27,11 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
+const IntendedTestCaseString = "This is a test case string"
```
purnesh42H, line 30: Why are defining your own string? Is this the expected string for which you will force error? If yes, name it exactly as the passed string testcase. For example, if it is for `framerWriteSettings`, just name it as `framerWriteSettingsString`. 

Also, don't you need 4 such strings i.e. one per test case? Right now you only have 2

internal/transport/http2_client_test.go
```diff
@@ -44,7 +49,15 @@ type framerWindowUpdateConn struct {
 }
 
 func (cp *clientPrefaceConn) Write(b []byte) (n int, err error) {
-	return 0, errors.New("preface write error")
+	if string(b) == IntendedTestCaseString {
+		return 0, errors.New("intended test case string detected: forced write error")
+	}
+
+	if len(b) >= len(ClientPreface) && string(b[:len(ClientPreface)]) == string(ClientPreface) {
```
purnesh42H, line 56: why is another condition needed? For first test case you want to just fail for client preface write, Ideally it should simple code like below

```
if string(b) == ClientPreface {
		return 0, errors.New("forced write error for client preface string")
	}

// Normally write the bytes if they don't match the ClientPreface
return cp.Conn.Write(b)
```
	
	
```

Comment (purnesh42H) [reply internal/transport/http2_client_test.go:56]: You have to similar thing for all 4 cases

File: internal/transport/http2_client_test.go (hunk 17 above)
Comment (hanut19) [reply internal/transport/http2_client_test.go]: change done

File: internal/transport/http2_client_test.go (hunk 18 above)
Comment (hanut19) [reply internal/transport/http2_client_test.go]: change done

File: internal/transport/http2_client_test.go (hunk 19 above)
Comment (hanut19) [reply internal/transport/http2_client_test.go]: change done please check

File: internal/transport/http2_client_test.go [hunk 20]
```diff
@@ -0,0 +1,185 @@
...
+func dialerClientPrefaceLength(_ context.Context, addr string) (net.Conn, error) {
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &clientPrefaceLengthConn{Conn: conn}, nil
+}
+
+func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
+	if string(b) != string(ClientPreface) {
```
purnesh42H, line 91: You need to throw error only` if string(b) == string(framerWriteSettings)`. `if string(b) != string(ClientPreface)` will throw error for everything except ClientPreface

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,185 @@
...
+func dialerFramerWriteSettings(_ context.Context, addr string) (net.Conn, error) {
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &framerWriteSettingsConn{Conn: conn}, nil
+}
+
+func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {
+	if string(b) != string(ClientPreface) {
```
purnesh42H, line 113: same as above to check for framerWindowUpdate only

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,185 @@
...
+			lis, err := net.Listen("tcp", "localhost:0")
+			if err != nil {
+				t.Fatalf("Listen() = _, %v, want _, <nil>", err)
+			}
+			defer lis.Close()
+			ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+			defer cancel()
+
+			_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, test.opts, func(GoAwayReason) {})
+			if err == nil {
```
purnesh42H, line 176:
```
if err == nil {
    t.Errorf("Expected an error, but got nil")
}

if err.Error() != test.expected {
	t.Fatalf("TestNewHTTP2ClientTarget() = %s, want %s", err.Error(), test.expected)
}
```

internal/transport/http2_client_test.go
hanut19: changes done please check

internal/transport/http2_client_test.go
hanut19: changes done please check

internal/transport/http2_client_test.go
hanut19: done please check

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,187 @@
...
+func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {
+	if string(b) == framerWindowUpdateString {
+		return 0, errors.New("intended test case string detected: forced write error")
+	}
+	// Simulate a WINDOW_UPDATE frame's value (window size in bytes)
+	windowUpdateValue := 13
+	n, err = fwu.Conn.Write(b)
+
+	// Compare the number of bytes written with the WINDOW_UPDATE value
+	if n == windowUpdateValue {
```
purnesh42H, line 125: why do you have to compare the bytes again? The string comparison should always throw the error for frameWindowUpdate

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return &clientPrefaceLengthConn{Conn: conn}, nil
+}
+
+func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
+	if string(b) == framerWriteSettingsString {
+		return 0, errors.New("intended test case string detected: forced write error")
+	}
+	framerValue := 9
+	n, err = fws.Conn.Write(b)
```
purnesh42H, line 99: same here why to compare bytes again?

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,187 @@
...
+	net.Conn
+}
+
+type framerWindowUpdateConn struct {
+	net.Conn
+}
+
+func (cp *clientPrefaceConn) Write(b []byte) (n int, err error) {
+	if string(b) == string(ClientPreface) {
+		return 0, errors.New("preface write error")
```
purnesh42H, line 56: `preface write error` -> `force error for preface write`

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,187 @@
...
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &clientPrefaceLengthConn{Conn: conn}, nil
+}
+
+func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
+	if string(b) == framerWriteSettingsString {
+		return 0, errors.New("intended test case string detected: forced write error")
```
purnesh42H, line 96: `intended test case string detected: forced write error` -> `force error for framerWriteSettings`

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,187 @@
...
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &framerWriteSettingsConn{Conn: conn}, nil
+}
+
+func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {
+	if string(b) == framerWindowUpdateString {
+		return 0, errors.New("intended test case string detected: forced write error")
```
purnesh42H, line 118: `intended test case string detected: forced write error` -> `force error for framerWindowUpdate`

internal/transport/http2_client_test.go
hanut19: change done please check

internal/transport/http2_client_test.go
hanut19: please check done

internal/transport/http2_client_test.go
hanut19: changes done

internal/transport/http2_client_test.go
hanut19: if not compare the bytes. test case will not cover. so cannot remove this condition.

purnesh42H: but then there is no point of first comparison. You can't just assume that no other message will have same size as frameWriteSetting

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,187 @@
...
+			if err != nil {
+				t.Fatalf("Listen() = _, %v, want _, <nil>", err)
+			}
+			defer lis.Close()
+			ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+			defer cancel()
+
+			_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, test.opts, func(GoAwayReason) {})
+			if err == nil {
+				t.Errorf("Expected an error, but got nil")
```
purnesh42H, line 180: please follow got before want https://google.github.io/styleguide/go/decisions#got-before-want

t.Errorf("err = nil, want non-nil")

internal/transport/http2_client_test.go
purnesh42H: If you can't compare actual string, then you have to force the error based on order of writes. One way is to use atomic bool like this https://github.com/grpc/grpc-go/pull/7371/files#diff-c90a44d2d36c8cc9c90b97b290cce3bdf6a5086bdac1e51e42d0e7b8f415b8b8R2757-R2764

It will be initially false. You set this to true after client preface write is done and then in the conn check if this atomic bool is true and throw the error

internal/transport/http2_client_test.go
hanut19: done please check

internal/transport/http2_client_test.go
hanut19: done please check

internal/transport/http2_client_test.go
hanut19: done

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, err
+	}
+	return &clientPrefaceLengthConn{Conn: conn}, nil
+}
+
+func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
+	if string(b) == string(framerWriteSettings) {
+		return 0, errors.New("force error for Framer write setting")
+	}
+	n, err = fws.Conn.Write(b)
```
purnesh42H, line 98: nit: can just directly return `fws.Conn.Write(b)`

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, err
+	}
+	return &framerWriteSettingsConn{Conn: conn}, nil
+}
+
+func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {
+	if string(b) == string(framerWindowUpdate) {
+		return 0, errors.New("force error for windowupdate")
+	}
+	n, err = fwu.Conn.Write(b)
```
purnesh42H, line 114: nit: can just directly return `fwu.Conn.Write(b)`

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,174 @@
...
+	return &clientPrefaceConn{Conn: conn}, nil
+}
+
+func (cpl *clientPrefaceLengthConn) Write(b []byte) (n int, err error) {
+	if string(b) == string(ClientPreface) {
+		incorrectPreface := "INCORRECT PREFACE\r\n\r\n"
+		n, err = cpl.Conn.Write([]byte(incorrectPreface))
+		return n, err
+	}
+	// Normally write the bytes if they don't match the ClientPreface
```
purnesh42H, line 81: nit: no need of inline comment here. Its obvious

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,174 @@
...
+type framerWindowUpdateConn struct {
+	net.Conn
+}
+
+func (cp *clientPrefaceConn) Write(b []byte) (n int, err error) {
+	if string(b) == string(ClientPreface) {
+		return 0, errors.New("force error for preface write")
+	}
+
+	// Normally write the bytes if they don't match the ClientPreface
```
purnesh42H, line 59: nit: no need of inline comment here

purnesh42H (approved):
lgtm

internal/transport/http2_client_test.go
hanut19: done. please check

internal/transport/http2_client_test.go
hanut19: done. please check

internal/transport/http2_client_test.go
hanut19: done. please check

internal/transport/http2_client_test.go
hanut19: done. please check

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,169 @@
...
+func dialerFramerWriteSettings(_ context.Context, addr string) (net.Conn, error) {
+	conn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &framerWriteSettingsConn{Conn: conn}, nil
+}
+
+func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {
+	if string(b) == string(framerWindowUpdate) {
```
purnesh42H, line 107: nit: can use `bytes.equal(...)`

hanut19: done. please check

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"time"
+
+	"golang.org/x/net/http2"
+	"google.golang.org/grpc/resolver"
+)
+
+// ClientPreface is the HTTP/2 client preface string.
+var ClientPreface = []byte(http2.ClientPreface)
+
+var framerWindowUpdate = []byte{0, 0, 4, 8, 0, 0, 0, 0, 0, 0, 0, 56, 129}
```
purnesh42H, line 35: for better readability, we should change to something like `[]byte{[]http2Setting{...}}` instead of raw bytes

hanut19:
```
func SettingsToBytes(settings []http2.Setting) ([]byte, error) {
	buf := new(bytes.Buffer)
	for _, s := range settings {
		if err := binary.Write(buf, binary.BigEndian, s.ID); err != nil {
			return nil, err
		}
		if err := binary.Write(buf, binary.BigEndian, s.Val); err != nil {
			return nil, err
		}
	}
	return buf.Bytes(), nil
}
``` 
```
func (fws *framerWriteSettingsConn) Write(b []byte) (n int, err error) {
	framerWriteSettings, _ := SettingsToBytes([]http2.Setting{{ID: 0, Val: 4}, {ID: 0, Val: 0}})

	if bytes.Equal(b, framerWriteSettings) {
		fmt.Println(framerWriteSettings)
		return 0, errors.New("force error for Framer write setting")
	}
	return fws.Conn.Write(b)
}
```
while i am using this approach unable to cover test case and getting error find below:
 http2_client_test.go:202: TestNewHTTP2ClientTarget() = connection error: desc = "error reading server preface: read tcp 127.0.0.1:50163->127.0.0.1:50162: use of closed network connection", want connection error: desc = "transport: failed to write initial settings frame: force error for Framer write setting"

purnesh42H: you can do `json.Marshal([]http2.Setting{})` to convert to `[]byte`

purnesh42H: you need to make sure you are putting the correct values in `[]http2.Setting` object which is expected by the test case

hanut19: done. please check

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"golang.org/x/net/http2"
+	"google.golang.org/grpc/resolver"
+)
+
+// ClientPreface is the HTTP/2 client preface string.
+var ClientPreface = []byte(http2.ClientPreface)
+
+// framerWriteSettings represents the serialized byte sequence of an empty HTTP/2 SETTINGS frame.
+var framerWriteSettings = []byte{0, 0, 0, 4, 0, 0, 0, 0, 0}
```
purnesh42H, line 36: framerWriteSettings  is serialized byte sequence of `[]http2.Setting{}`

hanut19: done please check

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+// framerWindowUpdate represents the serialized byte sequence of an HTTP/2 WINDOW_UPDATE frame.
```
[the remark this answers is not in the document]
hanut19: done please check

purnesh42H:
@arvindbr8 adding you for one more review. Thanks!

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (s) TestNewHTTP2ClientTarget(t *testing.T) {
+	tests := []struct {
+		name     string
+		opts     ConnectOptions
+		expected string
+	}{
+		{
+			name:     "client-preface-write",
```
arvindbr8, line 131: Seems like the error path is already covered by other tests. https://app.codecov.io/gh/grpc/grpc-go/blob/master/internal%2Ftransport%2Fhttp2_client.go#L425

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			defer lis.Close()
+			ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+			defer cancel()
+
+			_, err = NewClientTransport(ctx, context.Background(), resolver.Address{Addr: lis.Addr().String()}, test.opts, func(GoAwayReason) {})
+			if err == nil {
+				t.Errorf("got nil, want an error")
+			}
+			if err.Error() != test.expected {
```
arvindbr8, line 167: Error string is implementation-specific and isn't part a package's contract, therefore comparing error strings is not advisable.  That being said, this is an internal package - so we are in control here (presumably).

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+		{
+			name:     "framer-write-settings",
+			opts:     ConnectOptions{Dialer: dialerFramerWriteSettings},
+			expected: "connection error: desc = \"transport: failed to write initial settings frame: force error for Framer write setting\"",
+		},
+		{
+			name:     "framer-write-windowUpdate",
+			opts:     ConnectOptions{Dialer: dialerFramerWriteWindowUpdate, InitialConnWindowSize: 80000},
+			expected: "connection error: desc = \"transport: failed to write window update: force error for windowupdate\"",
+		},
```
arvindbr8, line 149: should we stub a test framer for these tests instead?

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
+			// Create a server.
+			lis, err := net.Listen("tcp", "localhost:0")
+			if err != nil {
+				t.Fatalf("Listen() = _, %v, want _, <nil>", err)
+			}
+			defer lis.Close()
+			ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
```
arvindbr8, line 160:
```suggestion
			ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
```

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+		},
+		{
+			name:     "framer-write-windowUpdate",
+			opts:     ConnectOptions{Dialer: dialerFramerWriteWindowUpdate, InitialConnWindowSize: 80000},
+			expected: "connection error: desc = \"transport: failed to write window update: force error for windowupdate\"",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
```
arvindbr8, line 153: nix newline

arvindbr8 (changes requested):
Thanks for improving coverage with these tests!  A few thoughts here:

1. To maximize the value of these tests, let's explore how we can make them more closely reflect "real-world" scenarios. 
2. Also, while grouping tests by error paths in NewClientTranport can be helpful, consider if individual tests might provide greater clarity for each specific scenario.
3. Exact error string assertions are never ideal as we do not guarantee such contracts.

arvindbr8:
We should re-access this PR by looking at the coverage report for the file here: https://app.codecov.io/gh/grpc/grpc-go/blob/master/internal%2Ftransport%2Fhttp2_client.go

dfawley (changes requested):
Are we sure this is the right approach?

It seems like making a `conn` that disconnects frequently while just connecting and making RPCs might work better and cover other kinds of similar errors that don't have test cases covering them either.  It wouldn't be deterministic, but getting every error covered with unit tests is extremely difficult, fragile, and hard to maintain.

internal/transport/http2_client_test.go
hanut19: done please check

internal/transport/http2_client_test.go
hanut19: done please check

internal/transport/http2_client_test.go
hanut19: done please check

internal/transport/http2_client_test.go
hanut19: removed this test. as specified already covered.

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"golang.org/x/net/http2"
+	"google.golang.org/grpc/resolver"
+)
+
+// ClientPreface is the HTTP/2 client preface string.
+var ClientPreface = []byte(http2.ClientPreface)
+
+// framerWriteSettings is serialized byte sequence of []http2.Setting{}
+var framerWriteSettings = []byte{0, 0, 0, 4, 0, 0, 0, 0, 0}
```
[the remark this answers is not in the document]
hanut19: done please check

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"net"
+	"testing"
+	"time"
+
+	"golang.org/x/net/http2"
+	"google.golang.org/grpc/resolver"
+)
+
+// ClientPreface is the HTTP/2 client preface string.
+var ClientPreface = []byte(http2.ClientPreface)
```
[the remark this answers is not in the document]
hanut19: yes using in test case.

internal/transport/http2_client_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+// framerWriteSettings is serialized byte sequence of []http2.Setting{}
+var framerWriteSettings = []byte{0, 0, 0, 4, 0, 0, 0, 0, 0}
+
+// framerWindowUpdate is serialized byte sequence of http2.Setting{ID:  http2.SettingInitialWindowSize, Val: 14465}
+var framerWindowUpdate = []byte{0, 0, 4, 8, 0, 0, 0, 0, 0, 0, 0, 56, 129}
```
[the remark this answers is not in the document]
hanut19: another way of doing this
`func (fwu *framerWindowUpdateConn) Write(b []byte) (n int, err error) {

	var buf bytes.Buffer
	framer := http2.NewFramer(&buf, nil)
	framer.WriteWindowUpdate(0, 14465)
	
	if bytes.Equal(b, buf.Bytes()) {
		return 0, errors.New("force error for windowupdate")
	}
	return fwu.Conn.Write(b)
}`

PR #7469, opened 2024-08-01
doc: update keepalive ClientParameters doc about doubling the interval upon GOAWAY

menghanl:
Add to the godoc the behavior that the client doubles the keepalive ping interval when GOAWAY is received due to too-many-pings.

And link to the design doc.

RELEASE NOTES: N/A

### DIFF
--- a/keepalive/keepalive.go
+++ b/keepalive/keepalive.go
@@ -34,6 +34,16 @@ type ClientParameters struct {
 	// After a duration of this time if the client doesn't see any activity it
 	// pings the server to see if the transport is still alive.
 	// If set below 10s, a minimum value of 10s will be used instead.
+	//
+	// Note that server has a default EnforcementPolicy.MinTime of 5 minutes
+	// (which means the client shouldn't ping more frequently than every 5 minutes).
+	//
+	// But it's not a strong requirement to set this less than
+	// EnforcementPolicy.MinTime, because the server sends a GOAWAY with error code
+	// ENHANCE_YOUR_CALM and debug data equal to ASCII "too_many_pings", and the
+	// client will double this interval upon receiving that.
+	//
+	// For more details, see https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md
 	Time time.Duration // The current default value is infinity.
 	// After having pinged for keepalive check, the client waits for a duration
 	// of Timeout and if no activity is seen even after that the connection is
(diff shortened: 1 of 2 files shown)


easwars (commented):
While you are here, should we also try and fix the code where we check if `Time` is less than `10s` and if so, set it to `10s`. We should check for `0s` explicitly there and probably not set it to `10s` if the value is `0s`? This will be a behavior change though. @dfawley

dfawley (commented):
I'm undecided on the 0s vs. 10s. vs infinity question so far.

The question to me is: if the user actually sets keepalive, why would they want it to have an infinite `Time`?  Is there ever a reason to do that?  If not, then the default being 10s (_when configuring keepalive_) seems pretty reasonable to me.

If we keep the behavior the same, we should update the comment, though, to say "keepalives are disabled by default, but the default value for this field is 10s".

menghanl (commented):
Updated. PTAL

keepalive/keepalive.go
```diff
@@ -33,7 +33,19 @@ import (
...
+	// minutes (which means the client shouldn't ping more frequently than every
+	// 5 minutes).
+	//
+	// Though not ideal, it's not a strong requirement for Time to be less than
+	// EnforcementPolicy.MinTime.  Time will automatically double if the server
+	// disconnects due to its enforcement policy.
+	//
+	// For more details, see
+	// https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md
 	Time time.Duration // The current default value is infinity.
```
dfawley, line 49: Can you delete this comment, please?  Explaining the behavior can be done via the existing text above and in `grpc.WithKeepaliveParams` (which ideally would be updated as part of this cleanup, if you don't mind, to say something like "`Keepalive is disabled by default`" or "`Keepalive pings are disabled by default`").

keepalive/keepalive.go
```diff
@@ -33,7 +33,19 @@ import (
 type ClientParameters struct {
 	// After a duration of this time if the client doesn't see any activity it
 	// pings the server to see if the transport is still alive.
-	// If set below 10s, a minimum value of 10s will be used instead.
+	// If set below 10s (including 0), a minimum value of 10s will be used
```
dfawley, line 36: The "including 0" is probably not needed if we get rid of the  comment below as mentioned.

dfawley (approved):
Thanks for the PR!

keepalive/keepalive.go
menghanl: Done

keepalive/keepalive.go
menghanl: Done

dfawley (approved):

easwars (approved):

Issue #7444, opened 2024-07-26
Add godoc linter


dfawley:
Since removing golint, we've had numerous occasions of missing docstrings.  This seems like the consensus replacement for golint: https://github.com/mgechev/revive -- we should set it up to at least check the docstrings.

EDIT: 

- [x] add the `Revive` inter in quite mode: https://github.com/grpc/grpc-go/pull/7472**

- [x] Enforce Revive Linter: #7589

- [x] fix below existing lint issues and enable the disabled rule

#7575 fixes
- [x] 83 exported

#7574 fixes
- [x] 8  package-comments

#7552 fixes
- [x] 15  redefines-builtin-id

#7528  fixes
- [x] 6  var-declaration
- [x] 3  indent-error-flow
- [x] 1  increment-decrement  
- [x]  1  superfluous-else  

#7577 #7580 fixes
- [x] 546 unused-parameter
- [x] 28  var-naming (Ignore `reflection/test/grpc_testing_not_regenerate/`)
- [x] 1  context-as-argument (Ignore `reflection/test/grpc_testing_not_regenerate/`)

purnesh42H:
Fix the existing lint issues one by one as caught in https://github.com/grpc/grpc-go/pull/7472/files as part of vet check

janardhanvissa:
@purnesh42H Please assign me this issue.

janardhanvissa:
@purnesh42H Right now, In the tests (vet) I'm getting like don't use an underscore in package name https://revive.run/r#var-naming. 
It's not expecting underscore but by default as it's generating from the proto file like ("google.golang.org/grpc/reflection/test/grpc_testing_not_regenerate"). So, what need to do for further. Please suggest.

janardhanvissa:
@purnesh42H Fixed linter issues for superfluous-else, increment-decrement, indent-error-flow, var-declaration. Please review the changes.

https://github.com/janardhanvissa/grpc-go/pull/1

janardhanvissa:
@purnesh42H Fixed below godoc linter issues. 

   15  redefines-builtin-id  
    8  package-comments      
    8  empty-block           
    6  var-declaration       
    3  indent-error-flow     
    2  increment-decrement    
    1  superfluous-else   
    
    
    Please find the PR's below
    
    https://github.com/janardhanvissa/grpc-go/pull/1
    https://github.com/janardhanvissa/grpc-go/pull/2
    https://github.com/janardhanvissa/grpc-go/pull/3

[...]

PR #7472, opened 2024-08-01
tools: Add github.com/mgechev/revive

purnesh42H:
cc #7444

RELEASE NOTES: None


dfawley (approved):
Tactically speaking...

```
Warnings:
  546  unused-parameter      
   83  exported              
   28  var-naming            
   15  redefines-builtin-id  
    8  package-comments      
    8  empty-block           
    6  var-declaration       
    3  indent-error-flow     
    2  increment-decrement   
    1  context-as-argument   
    1  superfluous-else      
```

...we're going to want to fix the issues that have few instances of them ASAP.  But we will probably want to make this a blocking checker before fixing all the "unused-parameter" items, since there are so many of them.

scripts/vet.sh
```diff
@@ -167,4 +168,10 @@ XXXXX PleaseIgnoreUnused'
   popd
 done
 
+# - Collection of revive linter analysis checks
+REV_OUT="$(mktemp)"
+revive -formatter unix ./... >"${REV_OUT}" || true
+
+(noret_grep -v "[unused-parameter]" "${REV_OUT}" | not grep -v "\.pb\.go:") || true
```
dfawley, line 175: Now this doesn't show us the other errors.

Maybe we can just make revive skip that check?  I assume that's possible?

purnesh42H: After switching to `plain` formatter, it works

dfawley (approved):

PR #7474, opened 2024-08-01
examples/advancedtls: example code for different security configurations for grpc-go using `advancedtls`

gtcooke94:
This PR adds runnable examples of various pieces of advancedtls

RELEASE NOTES: N/A


examples/features/advancedtls/README.md
```diff
@@ -0,0 +1,26 @@
...
+* Port 8883: A server running using InsecureCredentials.
+
+The clients are designed to call these servers with varying configurations of credentials and revocation configurations.
+* mTLS with certificate providers and CRLs
+* mTLS with custom verification
+* mTLS with credentials from credentials.NewTLS (directly using the tls.Config)
+* Insecure Credentials
+
+## Generate the credentials used in the examples
+Run `./generate.sh` from `/path/to/grpc-go/examples/features/advancedtls` to generate the `creds` directory containing the certificates and CRLs needed for these examples.
```
dfawley, line 16: I assume this isn't actually necessary since there is a `creds` directory in this PR?  If so, maybe move this section to the bottom and update the text here saying this is how you can do this if you need to make changes or something.

examples/features/advancedtls/client/main.go
```diff
@@ -0,0 +1,304 @@
...
+
+	if err != nil {
+		fmt.Printf("Error %v\n", err)
+		os.Exit(1)
+	}
+	fullServerAddr := serverAddr + ":" + port
+	runWithCredentials(clientTLSCreds, fullServerAddr, !shouldFail)
+}
+
+func tlsWithCrlsToGoodServer(credsDirectory string) {
```
dfawley, line 112: `CRLs` please throughout.

examples/features/advancedtls/client/main.go
```diff
@@ -0,0 +1,304 @@
...
+	flag.Parse()
+
+	if *credsDirectory == "" {
+		fmt.Println("Must set credentials_directory argument to this repo's creds directory")
+		os.Exit(1)
+	}
+	tlsWithCrls(*credsDirectory)
+	customVerification(*credsDirectory)
+	credentialsNewTLSExample(*credsDirectory)
+	insecureCredentialsExample()
```
dfawley, line 303: Probably too late to change stuff now, but I think it would be good to split things up into multiple clients & servers instead.  This one binary does a lot of different things, and doing so makes the code pretty complicated and hard to understand as an example.  My 2c.

examples/features/advancedtls/creds/localhost-openssl.cnf
```diff
@@ -0,0 +1,24 @@
+[req]
```
dfawley, line 1: Should these `cnf` files have comments to explain anything that isn't obvious, or point out what's important / not boilerplate?

examples/features/advancedtls/generate.sh
```diff
@@ -0,0 +1,87 @@
+rm -rf creds
```
dfawley, line 1: A little scary to me when I see `rm -rf` in a script.  You could either test for existence of the directory and refuse to run or prompt for confirmation or something.

dfawley, line 1: `#!/bin/bash`

examples/features/advancedtls/generate.sh
```diff
@@ -0,0 +1,87 @@
...
+rm 01.pem
+rm 02.pem
+rm 03.pem
+rm 04.pem
+rm *csr*
+rm *.txt*
+
+
+
+popd
```
dfawley, line 87: Please add a newline here.

examples/features/advancedtls/server/main.go
```diff
@@ -0,0 +1,188 @@
...
+const insecurePort int = 50053
+
+func (s *server) UnaryEcho(ctx context.Context, req *pb.EchoRequest) (*pb.EchoResponse, error) {
+	return &pb.EchoResponse{Message: req.Message}, nil
+}
+
+func tlsServers(credentialsDirectory string) {
+	go func() {
+		createAndRunTLSServer(credentialsDirectory, false, goodServerWithCrlPort)
+	}()
```
dfawley, line 56: `go createAndRunTLSServer(...)` instead please.  And elsewhere.

But also in Go it's an antipattern to have a func spawn something in a goroutine vs. just putting `go` on the call to the func.  Probably delete these two functions and call `createAndRun*` from `main`.  This also ties with the comment at the bottom..........

examples/features/advancedtls/server/main.go
```diff
@@ -0,0 +1,188 @@
...
+	flag.Parse()
+	if *credentialsDirectory == "" {
+		fmt.Println("Must set credentials_directory argument")
+		os.Exit(1)
+	}
+	tlsServers(*credentialsDirectory)
+	insecureServer()
+	for {
+		time.Sleep(1 * time.Second)
+	}
```
dfawley, line 187: If you want to block forever you can do `<-make(chan struct{})`.

But also, if you are calling `createAndRun...` from here directly, just don't do the last one in a `go` call, and it will block by virtue of `Serve` blocking?

dfawley (approved):

examples/features/advancedtls/client/main.go
gtcooke94: Discussed, will leave this for now and potentially make changes in the future

examples/features/advancedtls/creds/localhost-openssl.cnf
gtcooke94: Discussed, no because they are out of scope for this as an example - they are a prerequisite for getting the credentials needed

examples/features/advancedtls/server/main.go
gtcooke94: Good call-out on the anti-pattern, fixed

examples/features/advancedtls/README.md
gtcooke94: Discussed, moved the bottom as a developer note

PR #7478, opened 2024-08-04
Use builtin min and max functions

jhalterman:
With the upgrade to Go 1.21 (#7282), grpc-go can use the new builtin [min](https://pkg.go.dev/builtin#min) and [max](https://pkg.go.dev/builtin#max) functions. This commit removes existing functions and uses the built-ins instead.

RELEASE NOTES: none

### DIFF
--- a/internal/transport/controlbuf.go
+++ b/internal/transport/controlbuf.go
@@ -30,7 +30,6 @@ import (
 
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
-
 	"google.golang.org/grpc/internal/grpclog"
 	"google.golang.org/grpc/internal/grpcutil"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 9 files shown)


purnesh42H (approved):
lgtm. Thanks for the contribution!

internal/transport/controlbuf.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
+
```
arvindbr8, line 33: revert unrelated changes, pls?

arvindbr8 (approved):
LGTM, do you mind reverting the unrelated changes please?

internal/transport/controlbuf.go
arvindbr8: and others?

jhalterman: Sure, done!

dfawley (approved):

PR #7484, opened 2024-08-06
balancer/rls: Add picker metrics

zasweq:
This PR adds the implementation for RLS metrics.

RELEASE NOTES:
* balancer/rls: Add picker Pick information metrics

### DIFF
--- a/balancer/rls/balancer.go
+++ b/balancer/rls/balancer.go
@@ -28,8 +28,10 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/google/uuid"
 	"google.golang.org/grpc/balancer"
 	"google.golang.org/grpc/connectivity"
+	estats "google.golang.org/grpc/experimental/stats"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/internal"
 	"google.golang.org/grpc/internal/backoff"
@@ -77,6 +79,42 @@ var (
 	clientConnUpdateHook = func() {}
 	dataCachePurgeHook   = func() {}
 	resetBackoffHook     = func() {}
+
+	cacheEntriesMetric = estats.RegisterInt64Gauge(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.cache_entries",
+		Description: "EXPERIMENTAL. Number of entries in the RLS cache.",
+		Unit:        "entry",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid"},
+		Default:     false,
+	})
+	cacheSizeMetric = estats.RegisterInt64Gauge(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.cache_size",
+		Description: "EXPERIMENTAL. The current size of the RLS cache.",
+		Unit:        "By",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid"},
+		Default:     false,
+	})
+	defaultTargetPicksMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.default_target_picks",
+		Description: "EXPERIMENTAL. Number of LB picks sent to the default target.",
+		Unit:        "pick",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid", "grpc.lb.pick_result"},
+		Default:     false,
+	})
+	targetPicksMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.target_picks",
+		Description: "EXPERIMENTAL. Number of LB picks sent to each RLS target. Note that if the default target is also returned by the RLS server, RPCs sent to that target from the cache will be counted in this metric, not in grpc.rls.default_target_picks.",
+		Unit:        "pick",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid", "grpc.lb.pick_result"},
+		Default:     false,
+	})
+	failedPicksMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.failed_picks",
+		Description: "EXPERIMENTAL. Number of LB picks failed due to either a failed RLS request or the RLS channel being throttled.",
+		Unit:        "pick",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target"},
+		Default:     false,
+	})
 )
 
 func init() {
@@ -95,6 +133,7 @@ func (rlsBB) Build(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.
 		done:               grpcsync.NewEvent(),
 		cc:                 cc,
 		bopts:              opts,
+		uuid:               uuid.New().String(),
 		purgeTicker:        dataCachePurgeTicker(),
 		dataCachePurgeHook: dataCachePurgeHook,
 		lbCfg:              &lbConfig{},
@@ -122,6 +161,7 @@ type rlsBalancer struct {
 	done               *grpcsync.Event // Fires when Close() is done.
 	cc                 balancer.ClientConn
 	bopts              balancer.BuildOptions
+	uuid               string
 	purgeTicker        *time.Ticker
 	dataCachePurgeHook func()
 	logger             *internalgrpclog.PrefixLogger
@@ -240,7 +280,16 @@ func (b *rlsBalancer) purgeDataCache(doneCh chan struct{}) {
 		case <-b.purgeTicker.C:
 			b.cacheMu.Lock()
 			updatePicker := b.dataCache.evictExpiredEntries()
+
+			b.stateMu.Lock()
+			rlsLookupService := b.lbCfg.lookupService
+			b.stateMu.Unlock()
+			cacheSize := b.dataCache.currentSize
+			cacheEntries := int64(len(b.dataCache.entries))
 			b.cacheMu.Unlock()
+			grpcTarget := b.bopts.Target.String()
+			cacheSizeMetric.Record(b.bopts.MetricsRecorder, cacheSize, grpcTarget, rlsLookupService, b.uuid)
+			cacheEntriesMetric.Record(b.bopts.MetricsRecorder, cacheEntries, grpcTarget, rlsLookupService, b.uuid)
 			if updatePicker {
 				b.sendNewPicker()
 			}
@@ -304,7 +353,12 @@ func (b *rlsBalancer) UpdateClientConnState(ccs balancer.ClientConnState) error 
 		// `stateMu` if we are to hold both locks at the same time.
 		b.cacheMu.Lock()
 		b.dataCache.resize(newCfg.cacheSizeBytes)
+		dataCacheSize := b.dataCache.currentSize
+		dataCacheEntries := int64(len(b.dataCache.entries))
 		b.cacheMu.Unlock()
+		grpcTarget := b.bopts.Target.String()
+		cacheSizeMetric.Record(b.bopts.MetricsRecorder, dataCacheSize, grpcTarget, newCfg.lookupService, b.uuid)
+		cacheEntriesMetric.Record(b.bopts.MetricsRecorder, dataCacheEntries, grpcTarget, newCfg.lookupService, b.uuid)
 	}
 	return nil
 }
@@ -490,15 +544,17 @@ func (b *rlsBalancer) sendNewPickerLocked() {
 	if b.defaultPolicy != nil {
 		b.defaultPolicy.acquireRef()
 	}
+
 	picker := &rlsPicker{
-		kbm:           b.lbCfg.kbMap,
-		origEndpoint:  b.bopts.Target.Endpoint(),
-		lb:            b,
-		defaultPolicy: b.defaultPolicy,
-		ctrlCh:        b.ctrlCh,
-		maxAge:        b.lbCfg.maxAge,
-		staleAge:      b.lbCfg.staleAge,
-		bg:            b.bg,
+		kbm:             b.lbCfg.kbMap,
+		origEndpoint:    b.bopts.Target.Endpoint(),
+		lb:              b,
+		defaultPolicy:   b.defaultPolicy,
+		ctrlCh:          b.ctrlCh,
+		maxAge:          b.lbCfg.maxAge,
+		staleAge:        b.lbCfg.staleAge,
+		bg:              b.bg,
+		rlsServerTarget: b.lbCfg.lookupService,
 	}
 	picker.logger = internalgrpclog.NewPrefixLogger(logger, fmt.Sprintf("[rls-picker %p] ", picker))
 	state := balancer.State{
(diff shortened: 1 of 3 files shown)


balancer/rls/cache.go
```diff
@@ -163,7 +163,8 @@ func (l *lru) getLeastRecentlyUsed() cacheKey {
 //
 // It is not safe for concurrent access.
 type dataCache struct {
-	maxSize     int64 // Maximum allowed size.
```
easwars, line 166: Is this a formatting change alone? My eyes are failing me. If so, can you please revert this.

balancer/rls/balancer.go
```diff
@@ -77,6 +79,42 @@ var (
...
+		Description: "EXPERIMENTAL. The current size of the RLS cache.",
+		Unit:        "By",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid"},
+		Default:     false,
+	})
+	defaultTargetPicksMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.default_target_picks",
+		Description: "EXPERIMENTAL. Number of LB picks sent to the default target.",
+		Unit:        "pick",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid", "grpc.lb.pick_result"},
```
easwars, line 101: The labels mentioned in the design for this metric are:
`grpc.target, grpc.lb.rls.server_target, grpc.lb.rls.data_plane_target, grpc.lb.pick_result`

balancer/rls/balancer.go
```diff
@@ -77,6 +79,42 @@ var (
...
+		Default:     false,
+	})
+	targetPicksMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.target_picks",
+		Description: "EXPERIMENTAL. Number of LB picks sent to each RLS target. Note that if the default target is also returned by the RLS server, RPCs sent to that target from the cache will be counted in this metric, not in grpc.rls.default_target_picks.",
+		Unit:        "pick",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid", "grpc.lb.pick_result"},
```
easwars, line 108: Same as above. We need a `grpc.lb.rls.data_plane_target` label on this.

balancer/rls/balancer.go
```diff
@@ -77,6 +79,42 @@ var (
...
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid"},
+		Default:     false,
+	})
+	cacheSizeMetric = estats.RegisterInt64Gauge(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.cache_size",
+		Description: "EXPERIMENTAL. The current size of the RLS cache.",
+		Unit:        "By",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.instance_uuid"},
+		Default:     false,
+	})
```
easwars, line 96: Did you consider passing the following to the `dataCache` at creation time:
- `cacheEntriesMetric`
- `cacheSizeMetric`
- Value of the `grpc.target` label
- Value of the `grpc.lb.rls.instance_uuid` label

`grpc.lb.rls.server_target` label is not available at `Build` time (which is when the dataCache is created). So, we would have to add a method on the `dataCache` to update the value of this label.

Once the above is done, the `dataCache` can take care of updating the cache related metrics.
- `cacheEntriesMetric` is straight-forward
  - The only two operations that change the number of entries in the `dataCache` are `addEntry` and `deleteAndCleanup`. You could update this metric everytime you return from these methods of the `dataCache`.
- `cacheSizeMetric` is also straight-forward
  - There are three operations that change the current size of the `dataCache`
    - `addEntry`
    - `updateEntrySize`
    - `deleteAndCleanup`
  - And you could update the value of this metric when returning from these methods of the `dataCache`  

Let me know what you think about this approach.

easwars (changes requested):

easwars:
Also, please add a more descriptive release note entry.

balancer/rls/balancer.go
zasweq: Yeah noticed this while unit testing and forgot to update this PR> Thanks.

balancer/rls/balancer.go
zasweq: Done.

balancer/rls/cache.go
zasweq: Oh whoops reverted.

balancer/rls/balancer.go
zasweq: Discussed offline; decided to make these metrics be asynchronous polling gauges. So a callback will be passed from here to OpenTelemetry.

I grab the cache mu in the callback to read the caches current properties, and also need to read the most recent rls server target. I could do this through the helper you mentioned above but decided to just grab RLS state mu and read it from cfg and use that to record.

balancer/rls/picker.go
```diff
@@ -85,7 +86,13 @@ func (p *rlsPicker) Pick(info balancer.PickInfo) (balancer.PickResult, error) {
...
 
 	p.lb.cacheMu.Lock()
-	defer p.lb.cacheMu.Unlock()
+	var pr balancer.PickResult
+	var err error
+	metricsCallback := func() {}
+	defer func() {
+		p.lb.cacheMu.Unlock()
+		metricsCallback()
+	}()
```
easwars, line 95: What is the rationale behind changing the methods, `delegateToChildPoliciesLocked` and `useDefaultPickIfPossible` to return a callback that records the metrics instead of recording the metrics inline in those methods?

balancer/rls/picker.go
```diff
@@ -164,29 +175,46 @@ func (p *rlsPicker) delegateToChildPoliciesLocked(dcEntry *cacheEntry, info bala
...
+// a function to be invoked to record metrics.
+func (p *rlsPicker) useDefaultPickIfPossible(info balancer.PickInfo, errOnNoDefault error) (balancer.PickResult, func(), error) {
 	if p.defaultPolicy != nil {
 		state := (*balancer.State)(atomic.LoadPointer(&p.defaultPolicy.state))
-		return state.Picker.Pick(info)
+		res, err := state.Picker.Pick(info)
+		pr := "complete"
+		if err != nil {
+			pr = "fail"
+		}
```
easwars, line 209: I don't think this is as simple. In C-core, the picker returns the chosen subchannel and a pick-result. The picker can return one of the following result types:

1. Complete: The pick is complete, and the picker returns a subchannel to send the RPC on.  It may also optionally return some headers to add to the RPC or request notification when the RPC ends.

2. Queue: The picker does not currently have a subchannel available to send the RPC on, but it is working on getting one (e.g., if there is a connection attempt currently in flight).  The channel will queue the RPC, and it will retry the pick when the LB policy returns a new picker.

3. Fail: The picker has failed to connect to all possible subchannels and has nowhere to send the RPC.  In the channel, if the RPC is wait_for_ready, it will be queued; otherwise, the RPC attempt will fail (although there may be additional attempts made by the retry code).

4. Drop: This is similar to "Fail", except that (a) even wait_for_ready RPCs will be failed and (b) no more retry attempts will be made.

In our metrics, we are not emitting anything for `Queue`, but we are for the other three. If you see the documentation for the `Pick` method in the `balancer` package, you will see:
```
	// If an error is returned:
	//
	// - If the error is ErrNoSubConnAvailable, gRPC will block until a new
	//   Picker is provided by the balancer (using ClientConn.UpdateState).
	//
	// - If the error is a status error (implemented by the grpc/status
	//   package), gRPC will terminate the RPC with the code and message
	//   provided.
	//
	// - For all other errors, wait for ready RPCs will wait, but non-wait for
	//   ready RPCs will be terminated with this error's Error() string and
	//   status code Unavailable.
```

This means that for us, the pick result can be deduced as follows:
- `err == nil`: pick_result is "complete"
- `err == ErrNoSubConnAvailable`: pick_result is "queue"
- `err` is not a `status` error: pick_result is "fail"
  - although, i believe we will not be able to differentiate here between `wait_for_ready` RPCs and the others. This needs to be clarified with someone to see if this is OK.
  - @dfawley : What do you think? Do you think it would be an OK idea to add a `FailFast` field to `PickInfo`? The `picker_wrapper` has this information, it just doesn't pass it to the `Pick` method.
- `err` is a `status` error: pick_result is "drop"

easwars (changes requested):

balancer/rls/picker.go
zasweq: We want to record this outside the cacheMu, since metrics recording can potentially be expensive. Doing it outside the cacheMu makes it not *block all other RPCs* which was Doug's main concern with doing inside cache mu.

balancer/rls/picker.go
```diff
@@ -61,12 +61,13 @@ type rlsPicker struct {
 
 	// The picker is given its own copy of the below fields from the RLS LB policy
 	// to avoid having to grab the mutex on the latter.
-	defaultPolicy *childPolicyWrapper // Child policy for the default target.
-	ctrlCh        *controlChannel     // Control channel to the RLS server.
-	maxAge        time.Duration       // Cache max age from LB config.
-	staleAge      time.Duration       // Cache stale age from LB config.
-	bg            exitIdler
-	logger        *internalgrpclog.PrefixLogger
+	rlsServerTarget string
```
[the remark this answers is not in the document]
zasweq: Yeah I initially had that but I figured since this already persists bOpts around just use that. I guess code cleanliness it worth it in this case because it's only an extra string and pointer persisted around.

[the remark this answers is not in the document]
zasweq: Similar to how you store endpoint as well rather than access through balancer's Build options.

balancer/rls/balancer.go
```diff
@@ -77,6 +78,28 @@ var (
 	clientConnUpdateHook = func() {}
 	dataCachePurgeHook   = func() {}
 	resetBackoffHook     = func() {}
+
+	defaultTargetPicksMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.lb.rls.default_target_picks",
+		Description: "EXPERIMENTAL. Number of LB picks sent to the default target.",
+		Unit:        "pick",
+		Labels:      []string{"grpc.target", "grpc.lb.rls.server_target", "grpc.lb.rls.data_plane_target", "grpc.lb.pick_result"},
+		Default:     false,
```
easwars, line 87: Nit: Ignore the `Default` field since `false` is the zero value for `bool`?

balancer/rls/picker.go
```diff
@@ -164,29 +178,53 @@ func (p *rlsPicker) delegateToChildPoliciesLocked(dcEntry *cacheEntry, info bala
...
 			res, err := state.Picker.Pick(info)
 			if err != nil {
-				return res, err
+				pr := "fail"
+				if _, ok := status.FromError(err); ok {
+					pr = "drop"
+				}
+				return res, func() {
+					targetPicksMetric.Record(p.metricsRecorder, 1, p.grpcTarget, p.rlsServerTarget, cpw.target, pr)
+				}, err
```
easwars, line 187: How about making a helper since this logic is used in two places now:

```golang
// errToPickResult is a helper function which converts the error value returned
// by Pick() to a string that represents the pick result.
func errToPickResult(err error) string {
	if err == nil {
		return "complete"
	}
	if errors.Is(err, balancer.ErrNoSubConnAvailable) {
		return "queue"
	}
	if _, ok := status.FromError(err); ok {
		return "drop"
	}
	return "fail"
}
```

And we need to handle the `balancer.ErrNoSubConnAvailable` case in your code, don't we? Otherwise, we will wrongly mark picks that a child policy wanted to be queued as "fail".

And if the decision is to not record a metric for queued picks, we need to make sure that we ignore that explicitly after we handle the `balancer.ErrNoSubConnAvailable` case, as in the helper I described above.

balancer/rls/picker.go
easwars: This is worth documenting with a comment.

easwars:
Also, how do you plan to test these changes?

balancer/rls/picker.go
zasweq: Oh for some reason I thought the balancer.ErrNoSubConnAvailable was only possible to be emitted from this RLS Layer. I forgot this could emit from the child as well in these error cases. I will try your helper out.

balancer/rls/balancer.go
zasweq: I'd prefer to keep this explicit since a user now knows they need to configure it on through the OTel API. All of these non per call metrics are experimental and default off for now.

balancer/rls/picker.go
zasweq: Done. Added comment "Record metrics without the cache mutex held, to prevent lock contention between concurrent RPC's and their Pick calls. Metrics Recording can potentially be expensive."

zasweq:
Thanks for the pass; great suggestions. I plan on testing this with unit tests for more specific sceanrios/counters, and e2e with RLS deployed as a top level balancer of channel, with a OpenTelemetry component configured. I can then verify OpenTelemetry eventually emits the expected metrics atoms for the RLS Balancer metrics.

balancer/rls/balancer.go
zasweq: Decided to do this and emit on synchronous gauges with cache mu held in cache operations. Just to note I think you only need to pass target at construction, the metrics variables are global to the balancer package, and since the cache is 1:1 with the balancer, I think I can just create a UUID in newDataCache (and record current cache size/entries is 0). I'll send this out in a new PR.

zasweq: Like your suggestion for picker, I also passed a metrics recorder.

balancer/rls/balancer.go
easwars: Ack.

balancer/rls/picker.go
```diff
@@ -164,29 +197,54 @@ func (p *rlsPicker) delegateToChildPoliciesLocked(dcEntry *cacheEntry, info bala
...
 			if err != nil {
-				return res, err
+				pr := errToPickResult(err)
+				rf := func() {
+					targetPicksMetric.Record(p.metricsRecorder, 1, p.grpcTarget, p.rlsServerTarget, cpw.target, pr)
+				}
+				if pr == "queue" {
+					// Don't record metrics for queued Picks.
+					rf = func() {}
+				}
```
easwars, line 207: Here and down below, I would pull in the check for queued picks inside the callback:

```
rf := func() {
	if pr == "queue" {
		// Don't record metrics for queued Picks.
		return
	}
	targetPicksMetric.Record(p.metricsRecorder, 1, p.grpcTarget, p.rlsServerTarget, cpw.target, pr)
}
```

The above approach would mean that for completed and failed picks, we would incur the extra cost of running a conditional `if pr == "queue" { ... }`, and branch statements can be expensive with CPU pipelining. I'm not too concerned about it though. I prefer the way the code looks when you pull in it. But if you are @dfawley thinks the cost is not worth the ergonomics, I'm fine with it.

easwars (approved):

balancer/rls/picker.go
zasweq: Doug said he doesn't mind either, so I switched to this.

PR #7486, opened 2024-08-06
benchmark: wire in new gzip compressor

PapaCharlie:
Without this change, the benchmarks did not actually include executions with the new gzip compressor from the `encoding/gzip` package. This adds the option to run the suite with said new compressor and adds it to the "all" suite as well.

RELEASE NOTES: none

### DIFF
--- a/benchmark/benchmain/main.go
+++ b/benchmark/benchmain/main.go
@@ -65,6 +65,7 @@ import (
 	"google.golang.org/grpc/benchmark/latency"
 	"google.golang.org/grpc/benchmark/stats"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/encoding/gzip"
 	"google.golang.org/grpc/experimental"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/internal"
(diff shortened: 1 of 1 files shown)


benchmark/benchmain/main.go
```diff
@@ -128,10 +129,11 @@ const (
 	workloadsUnconstrained = "unconstrained"
 	workloadsAll           = "all"
 	// Compression modes.
-	compModeOff  = "off"
-	compModeGzip = "gzip"
-	compModeNop  = "nop"
-	compModeAll  = "all"
+	compModeOff    = "off"
+	compModeGzip   = "gzip"
```
dfawley, line 133: Request: add a `compModeGzipV1 = "gzipV1"`, and make `"gzip"` an alias that uses the latest compressor API version we have (i.e. `"gzipV2"` for now).

PapaCharlie: Addressed offline, instead of adding a new flag just replace the behavior of the current flag since it was testing very very old deprecated code.

dfawley (approved):

easwars (approved):

PR #7499, opened 2024-08-09
balancer: Clarifying PickResult's Done behaviour in documentation

AnomalRoil:
The current documentation seems to imply the `Done` function has pointer parameters, which it does not.

A `Done func(DoneInfo)` doing a `nil` check on the `DoneInfo` would not be valid Go since `DoneInfo` is a struct and not a pointer.

RELEASE NOTES: None

### DIFF
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -329,7 +329,8 @@ type PickResult struct {
 	SubConn SubConn
 
 	// Done is called when the RPC is completed.  If the SubConn is not ready,
-	// this will be called with a nil parameter.  If the SubConn is not a valid
+	// this will be called with an empty DoneInfo: meaning that Pick did not
+	// error but no bytes were sent nor received.  If the SubConn is not a valid
 	// type, Done may not be called.  May be nil if the balancer does not wish
 	// to be notified when the RPC completes.
 	Done func(DoneInfo)


purnesh42H (approved):
lgtm

balancer/balancer.go
```diff
@@ -329,7 +329,8 @@ type PickResult struct {
 	SubConn SubConn
 
 	// Done is called when the RPC is completed.  If the SubConn is not ready,
-	// this will be called with a nil parameter.  If the SubConn is not a valid
+	// this will be called with an empty DoneInfo: meaning that Pick did not
+	// error but no bytes were sent nor received.  If the SubConn is not a valid
```
dfawley, line 333: Grammar nit: it should be "but no bytes were sent _or_ received" here, not "nor".

Adding an "either" or "neither" to the sentence, only "either" makes sense:

"but no bytes were _either_ sent _or_ received" vs
"but no bytes were _neither_ sent _nor_ received".

If you say "but bytes were neither sent nor received", that would be correct, too, but a bit awkward.

balancer/balancer.go
```diff
@@ -329,7 +329,8 @@ type PickResult struct {
 	SubConn SubConn
 
 	// Done is called when the RPC is completed.  If the SubConn is not ready,
-	// this will be called with a nil parameter.  If the SubConn is not a valid
+	// this will be called with an empty DoneInfo: meaning that Pick did not
```
dfawley, line 332: Please us a comma instead of a colon.

PR #7522, opened 2024-08-16
rbac: fix usage of AuthInfo

dfawley:
This could actually lead to a panic if the auth type reports "tls" but the `AuthInfo` is not a `credentials.TLSInfo`.  There is no evidence this is happening and has been this way for a long time.

RELEASE NOTES: none


ejona86 (approved):

arvindbr8 (approved):
you have my authoritative LGTM

PR #7525, opened 2024-08-17
stats/opentelemetry: Optimize slice allocations

ash2k:
First commit moves attribute construction under `if` to only construct them when they are used.

Second commit uses `WithAttributeSet()` instead of `WithAttributes()`. This avoids copying the slice, which is not necessary for this code. See https://github.com/open-telemetry/opentelemetry-go/blob/metric/v1.28.0/metric/instrument.go#L349-L368.

Also, a bit of allocations can be removed by constructing the vararg slice one time vs when calling `Record()` each time.

```go
var sink []string

func abc(vals ...string) {
	sink = vals
}

func BenchmarkFuncCall(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()
	arg := []string{"a", "b", "c"}
	for i := 0; i < b.N; i++ {
		abc(arg...)
	}
}
```

The above benchmark gives me these results:

```
// Pass "a", "b", "c"
BenchmarkFuncCall-10    	43171615	        27.14 ns/op	      48 B/op	       1 allocs/op

// Pass arg...
BenchmarkFuncCall-10    	1000000000	         0.9530 ns/op	       0 B/op	       0 allocs/op
```

See https://go.dev/ref/spec#Passing_arguments_to_..._parameters.

RELEASE NOTES: None

### DIFF
--- a/stats/opentelemetry/client_metrics.go
+++ b/stats/opentelemetry/client_metrics.go
@@ -128,9 +128,13 @@ func (h *clientStatsHandler) streamInterceptor(ctx context.Context, desc *grpc.S
 }
 
 func (h *clientStatsHandler) perCallMetrics(ctx context.Context, err error, startTime time.Time, ci *callInfo) {
-	s := status.Convert(err)
-	callLatency := float64(time.Since(startTime)) / float64(time.Second)
-	h.clientMetrics.callDuration.Record(ctx, callLatency, otelmetric.WithAttributes(otelattribute.String("grpc.method", ci.method), otelattribute.String("grpc.target", ci.target), otelattribute.String("grpc.status", canonicalString(s.Code()))))
+	callLatency := float64(time.Since(startTime)) / float64(time.Second) // calculate ASAP
+	attrs := otelmetric.WithAttributeSet(otelattribute.NewSet(
+		otelattribute.String("grpc.method", ci.method),
+		otelattribute.String("grpc.target", ci.target),
+		otelattribute.String("grpc.status", canonicalString(status.Code(err))),
+	))
+	h.clientMetrics.callDuration.Record(ctx, callLatency, attrs)
 }
 
 // TagConn exists to satisfy stats.Handler.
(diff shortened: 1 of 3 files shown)


arjan-bal (approved):
LGTM, thanks for the contribution!

Adding another reviewer for a second approval.

stats/opentelemetry/client_metrics.go
```diff
@@ -128,9 +128,13 @@ func (h *clientStatsHandler) streamInterceptor(ctx context.Context, desc *grpc.S
...
 
 func (h *clientStatsHandler) perCallMetrics(ctx context.Context, err error, startTime time.Time, ci *callInfo) {
-	s := status.Convert(err)
-	callLatency := float64(time.Since(startTime)) / float64(time.Second)
-	h.clientMetrics.callDuration.Record(ctx, callLatency, otelmetric.WithAttributes(otelattribute.String("grpc.method", ci.method), otelattribute.String("grpc.target", ci.target), otelattribute.String("grpc.status", canonicalString(s.Code()))))
+	callLatency := float64(time.Since(startTime)) / float64(time.Second) // calculate ASAP
+	attrs := otelmetric.WithAttributeSet(otelattribute.NewSet(
+		otelattribute.String("grpc.method", ci.method),
+		otelattribute.String("grpc.target", ci.target),
+		otelattribute.String("grpc.status", canonicalString(status.Convert(err).Code())),
```
dfawley, line 135: Can we use `status.Code(err)` here?

dfawley (commented):
Wow, I wish otel had renamed `WithAttributeSet` to `WithAttributes` and made some other function behave the way `WithAttributes` behaves.  That is unexpected.  I might even go so far as to mark `WithAttributes` deprecated if I was them, to make it more clear that it's probably not what you ever want to use.

stats/opentelemetry/client_metrics.go
ash2k: Done. Rebased and addressed this comment. No other changes.

dfawley: For future consideration, please don't force push during github PR reviews -- or if you want to do a rebase, please don't modify the contents of existing commits.  Github doesn't maintain review consistency well in these cases.  It's much easier to see what's changed over time if you only ever add new commits.  Thanks!

dfawley (approved):

stats/opentelemetry/client_metrics.go
ash2k: Ok, got it. Thank you. I mainly do it to have a clean history where each commit makes sense, but if in this repo PRs are squash-merged anyway, then there is no point. I'll keep this in mind. 👍

PR #7528, opened 2024-08-18
Fix revive identified linter issues: var-declaration, indent-error-flow, increment-decrement, superfluous-else

janardhanvissa:
Partially addresses https://github.com/grpc/grpc-go/issues/7444 

Fixed below linter issues
    - 6  var-declaration       
    - 3  indent-error-flow     
    - 2  increment-decrement   
    - 1  superfluous-else

RELEASE NOTES: None


purnesh42H (approved):
lgtm

arvindbr8:
Thanks for taking a stab at this @janardhanvissa -- However I would prefer if these fixes also have an effect in our CI. I have #7543 in review. 

Once we merge it, we can then re-enable the checks which are fixed in this PR

arvindbr8:
I spoke to @purnesh42H offline and seems like @janardhanvissa  is working on all the remaining fixes (expect for unused-parameters) at the moment.  Seems like it's not worth doing #7543. 

I would have preferred if there were some evidence that these checks are actually been fixed as this PR does not include any changes to the CI 

I'm okay to iterate over this

arvindbr8 (approved):

purnesh42H:
> I spoke to @purnesh42H offline and seems like @janardhanvissa is working on all the remaining fixes (expect for unused-parameters) at the moment. Seems like it's not worth doing #7543.
> 
> I would have preferred if there were some evidence that these checks are actually been fixed as this PR does not include any changes to the CI
> 
> I'm okay to iterate over this

Thanks Arvind. Yeah we can rely on `check annotations` in the PR `/files` view. Even though they are not enforced in the PR checks, they do appear there as warnings. So, If they are not appearing anymore in the changed files then we can conclude they are fixed.

PR #7535, opened 2024-08-19
credentials/tls: default GRPC_ENFORCE_ALPN_ENABLED to true

arjan-bal:
Related issue: https://github.com/grpc/grpc-go/issues/434

RELEASE NOTES:
* Clients and servers will reject TLS connections that don't support ALPN. This can be disabled by setting environment variable `GRPC_ENFORCE_ALPN_ENABLED` to `false` (case insensitive). Please file a bug if you encounter any issues with this behaviour. This will become the default behaviour in an upcoming release.

### DIFF
--- a/credentials/credentials_test.go
+++ b/credentials/credentials_test.go
@@ -311,7 +311,7 @@ func tlsServerHandshake(conn net.Conn) (AuthInfo, error) {
 
 func tlsClientHandshake(conn net.Conn, _ string) (AuthInfo, error) {
 	clientTLSConfig := &tls.Config{
-		InsecureSkipVerify: true,
+		InsecureSkipVerify: true, // NOLINT
 		NextProtos:         []string{"h2"},
 	}
 	clientConn := tls.Client(conn, clientTLSConfig)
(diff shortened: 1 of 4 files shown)


dfawley:
Let's wait a few weeks for this since we have a lot of other semi-risky things in flight right now.

dfawley (approved):

arjan-bal:
Waiting to fix internal usages before submitting this PR.

atollena:
Small note on the release note: 

> Clients and servers will reject TLS connections that don't support ALPN. This can be disabled by setting environment variable GRPC_ENFORCE_ALPN_ENABLED to false (case insensitive). Please file a bug if you encounter any issues with this behaviour. This will become the default behaviour in an upcoming release.

Is the last sentence correct? Shouldn't it just read "The environment variable to revert this behavior will be removed in an upcoming release."?

arjan-bal:
> Small note on the release note:
> 
> > Clients and servers will reject TLS connections that don't support ALPN. This can be disabled by setting environment variable GRPC_ENFORCE_ALPN_ENABLED to false (case insensitive). Please file a bug if you encounter any issues with this behaviour. This will become the default behaviour in an upcoming release.
> 
> Is the last sentence correct? Shouldn't it just read "The environment variable to revert this behavior will be removed in an upcoming release."?

You're correct. It was a copy/paste error. Updated the release note

PR #7540, opened 2024-08-20
cherry-pick: transport: add timeout for writing GOAWAY on http2Client.Close() #7371

aranjans:
fixes https://github.com/grpc/grpc-go/issues/7314

Problem
In gRPC 1.64, we released a https://github.com/grpc/grpc-go/pull/7015 "Send GOAWAY to server on Client Transport Shutdown", post which in case of network hang when we are not able to write GOAWAY frame from client, closing connections will take a very long time.

What does this PR fix
This PR fixes this issue by having a timeout for loopyWriter to write GOAWAY so that http2Client.Close() doesn't wait for too long.

RELEASE NOTES:

- transport/http2_client: fix waiting on http2Client.Close() in case of network hang for writing GOAWAY.

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -59,6 +59,8 @@ import (
 // atomically.
 var clientConnectionCounter uint64
 
+var goAwayLoopyWriterTimeout = 5 * time.Second
+
 var metadataFromOutgoingContextRaw = internal.FromOutgoingContextRaw.(func(context.Context) (metadata.MD, [][]string, bool))
 
 // http2Client implements the ClientTransport interface with HTTP2.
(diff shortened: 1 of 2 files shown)


arvindbr8:
The relnotes should talk more about the bug along with what the fix was.  Users would be interested to know if they might be impacted by the bug.

aranjans:
@arvindbr8 updated the description

arvindbr8 (approved):

arvindbr8:
Also feel free to update the relnotes for the release itself. Now when i think about it, the relnotes in the description here does not get picked up

dfawley:
No user will understand this release note, either; it's written more like it's for a gRPC developer.  The release notes should talk about the user-visible problem is being fixed.  "client: prevent hanging during ClientConn.Close() when the network is unreachable" or something like that.

arvindbr8:
cc: @aranjans

aranjans:
Acknowledged, updated the release note.

PR #7543, opened 2024-08-20
vet: dont revive from revive

arvindbr8:
I would prefer if we took this approach with revive + #7444.  This way we can re-enable each of the check as in when we fix them. This gives us a better way of testing changes like #7528 

Let me know how you think about this

RELEASE NOTES: none

### DIFF
--- a/scripts/revive.toml
+++ b/scripts/revive.toml
@@ -1,0 +1,21 @@
+# TODO(#7444): Enable these rules once the linter issues are fixed.
+[rule.context-as-argument]
+    Disabled = true
+[rule.empty-block]
+    Disabled = true
+[rule.exported]
+    Disabled = true
+[rule.increment-decrement]
+    Disabled = true
+[rule.indent-error-flow]
+    Disabled = true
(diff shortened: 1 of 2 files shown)


arvindbr8:
@purnesh42H -- I want to take a look at https://github.com/mgechev/revive#available-rules and see if there are some more rules we would love to include which are not part of the default set.

also staticcheck might already be doing some of this... ugh, why did golint have to go away

purnesh42H:
Just to confirm the rules mentioned in .toml file are disabled and rest of the rules are enabled by default? or we will only check for rules which are in .toml file? 

I think former is better to exclude the rules which we don't want and let everything else being checked. In my PR https://github.com/grpc/grpc-go/pull/7472/files#diff-68ca30ee8b2dba2db2f3c93ad8d56c67f4dab2bd4475848f24f70c41f4681f22R177 i had only excluded `unused-parameter` as there were so many. Initial plan was to enforce revive once everything except `unused-parameter` is fixed, which should be quick

Issue #7538, opened 2024-08-19
xDS load reporting is not populating `total_issued_requests` field in LRS reports


dfawley:
Looks like we did not ever populate this field, but the other gRPC languages are.  The fields we should be setting are not specifically mentioned in gRFC A27 at this time, so we should also make sure to improve that gRFC.

PR #7544, opened 2024-08-21
xdsclient: Populate total_issued_requests count in LRS load report

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/7538

This PR populates the [total issued request count](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/load_report.proto#envoy-v3-api-msg-config-endpoint-v3-clusterstats:~:text=in%20the%20locality.-,total_issued_requests,-(uint64)%20The) which is already being added by Java and c-core clients.

The `issued request` counter is incremented when a new request is started. The counter is reset when a load report is sent to the LRS server

RELEASE NOTES:
* LRS load reports contain the `total_issued_requests` field.

### DIFF
--- a/xds/internal/balancer/clusterimpl/balancer_test.go
+++ b/xds/internal/balancer/clusterimpl/balancer_test.go
@@ -142,7 +142,7 @@ func (s) TestDropByCategory(t *testing.T) {
 	sc1.UpdateState(balancer.SubConnState{ConnectivityState: connectivity.Ready})
 	// Test pick with one backend.
 
-	const rpcCount = 20
+	const rpcCount = 24
 	if err := cc.WaitForPicker(ctx, func(p balancer.Picker) error {
 		for i := 0; i < rpcCount; i++ {
 			gotSCSt, err := p.Pick(balancer.PickInfo{})
@@ -159,11 +159,11 @@ func (s) TestDropByCategory(t *testing.T) {
 			if gotSCSt.Done == nil {
 				continue
 			}
-			// Fail half of the requests that are not dropped.
-			if i%4 == 1 {
-				gotSCSt.Done(balancer.DoneInfo{})
-			} else {
+			// Fail 1/4th of the requests that are not dropped.
+			if i%8 == 1 {
 				gotSCSt.Done(balancer.DoneInfo{Err: fmt.Errorf("test error")})
+			} else {
+				gotSCSt.Done(balancer.DoneInfo{})
 			}
 		}
 		return nil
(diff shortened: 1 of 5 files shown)


xds/internal/balancer/clusterimpl/balancer_test.go
```diff
@@ -356,7 +362,10 @@ func (s) TestDropCircuitBreaking(t *testing.T) {
 		Service:    testServiceName,
 		TotalDrops: uint64(maxRequest),
 		LocalityStats: map[string]load.LocalityData{
-			assertString(xdsinternal.LocalityID{}.ToString): {RequestStats: load.RequestData{Succeeded: uint64(rpcCount - maxRequest + 50)}},
+			assertString(xdsinternal.LocalityID{}.ToString): {RequestStats: load.RequestData{
+				Succeeded: uint64(rpcCount - maxRequest + 50),
+				Issued:    uint64(rpcCount - maxRequest + 50),
```
dfawley, line 367: Are there no tests that include failed calls?  It would be nice to have at least one test case where succeeded != issued.

dfawley (commented):
LGTM, just one request for a test case.

xds/internal/balancer/clusterimpl/balancer_test.go
arjan-bal: Changed a couple of test cases to have both errors and successful RPCs.

xds/internal/balancer/clusterimpl/balancer_test.go
```diff
@@ -178,7 +184,8 @@ func (s) TestDropByCategory(t *testing.T) {
 		Drops:      map[string]uint64{dropReason: dropCount},
 		LocalityStats: map[string]load.LocalityData{
 			assertString(xdsinternal.LocalityID{}.ToString): {RequestStats: load.RequestData{
-				Succeeded: rpcCount - dropCount,
+				Succeeded: (rpcCount - dropCount) / 2,
+				Errored:   (rpcCount - dropCount) / 2,
```
dfawley, line 188: Can we make it so these aren't the same?  I.e. if we reversed the two in code then the test would still pass.  E.g. make it fail 1/4th and succeed in 3/4ths.

dfawley (approved):

xds/internal/balancer/clusterimpl/balancer_test.go
arjan-bal: Done.

PR #7545, opened 2024-08-21
stats/otel: upgrade grpc version that contains the experimental/stats package

arvindbr8:
Building stats/otel was broken on its own because this package depends on `grpc/experimental/stats` which is not released yet. Our tests passed because of the replace directive magic.

RELEASE NOTES: none

### DIFF
--- a/stats/opentelemetry/go.mod
+++ b/stats/opentelemetry/go.mod
@@ -12,7 +12,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.28.0
 	go.opentelemetry.io/otel/sdk v1.28.0
 	go.opentelemetry.io/otel/sdk/metric v1.28.0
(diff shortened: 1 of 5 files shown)


stats/opentelemetry/go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.28.0
 	go.opentelemetry.io/otel/sdk v1.28.0
 	go.opentelemetry.io/otel/sdk/metric v1.28.0
-	google.golang.org/grpc v1.65.0
```
dfawley, line 15: Shouldn't we use 1.66.0?

arvindbr8: Done, 1.66.0 was not available at the time.

dfawley (approved):

PR #7546, opened 2024-08-21
cherry-pick #7356 to v1.66.x branch

dfawley:
RELEASE NOTES:
* tbd


arvindbr8 (approved):

internal/transport/controlbuf.go
```diff
@@ -962,43 +971,50 @@ func (l *loopyWriter) processData() (bool, error) {
...
+			// Note that this is only supposed to be nil in tests. Otherwise, stream is
+			// always initialized with a BufferPool.
+			pool = mem.DefaultBufferPool()
+		}
+		buf = pool.Get(size)
+		defer pool.Put(buf)
 
-	size := hSize + dSize
+		copy((*buf)[:hSize], dataItem.h)
+		_, _ = dataItem.reader.Read((*buf)[hSize:])
```
printchard, line 996: Maybe to late to change, but I noticed something today. Doesn't this mean that we are copying the message even if there is no need to? i.e. `dataItem.h` is empty, so we only have `dataItem.d`, thus we could just call `WriteData(/* params */, dataItem.d)`, but here we are doing `dataItem.reader.Read` which is performing a copy.

PR #7574, opened 2024-08-30
.*: fix revive package-comments lint issues

purnesh42H:
Addresses: #7444 

RELEASE NOTES: None

### DIFF
--- a/examples/features/csm_observability/client/main.go
+++ b/examples/features/csm_observability/client/main.go
@@ -16,6 +16,7 @@
  *
  */
 
+// Binary client is an example client.
 package main
 
 import (
--- a/examples/features/csm_observability/client/main.go
+++ b/examples/features/csm_observability/client/main.go
@@ -16,7 +16,7 @@
  *
  */
 
-// Binary client is an example client for CSM Observability.
+// Binary client is a client for the CSM Observability example.
 package main
 
 import (
--- a/examples/features/csm_observability/client/main.go
+++ b/examples/features/csm_observability/client/main.go
@@ -16,7 +16,7 @@
  *
  */
 
-// Binary client is an example client.
+// Binary client is an example client for CSM Observability.
 package main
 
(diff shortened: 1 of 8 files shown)


examples/features/csm_observability/client/main.go
```diff
@@ -16,6 +16,7 @@
  *
  */
 
+// Binary client is an example client.
```
dfawley, line 19: Can you include the example's purpose/name here too?  E.g.

`// Binary client is an example client for CSM observability.`

purnesh42H: Added. Should we change other examples as well? I just followed the other examples and they didn't have names

dfawley: The goal is not to increase the scope of this work.  I'm fine with leaving existing things alone -- unless you _want_ to go change them, in which case I will approve.

In general, though, we should _never_ use "some other thing we have does X" as an excuse to do "X", if we know "X" is not optimal.

examples/features/customloadbalancer/client/main.go
```diff
@@ -16,6 +16,7 @@
  *
  */
 
+// Binary client is an example client for custom load balancer.
```
dfawley, line 19: This is not proper English.  "for _a_ custom load balancer" or "a client for the custom load balancer example", etc.

dfawley (approved):
One small nit, otherwise LGTM.  Thanks!

examples/features/customloadbalancer/client/main.go
purnesh42H: Thanks. Done

examples/features/csm_observability/client/main.go
purnesh42H: I agree. I was just not sure if it was intentional to leave out the names. I will open a small issue to improve the example documentation

PR #7581, opened 2024-08-31
Add full method names list to generated code for enhanced gRPC interceptor control

Elsharaky:
Add full method names list for each service to enhance custom gRPC interceptor control

- Implemented a feature that generates a list of full method names for each gRPC service in the generated code.
- This list is integrated into the code to facilitate service-wide control mechanisms in custom gRPC interceptors.
- By providing access to all method names, this feature allows developers to apply consistent logic, such as authentication, logging, or rate-limiting, across an entire service in a more streamlined and efficient manner.
- Updated code generation logic to include the method names list, ensuring better introspection and control over gRPC services.

### DIFF
--- a/balancer/grpclb/grpc_lb_v1/load_balancer_grpc.pb.go
+++ b/balancer/grpclb/grpc_lb_v1/load_balancer_grpc.pb.go
@@ -41,6 +41,12 @@ const (
 	LoadBalancer_BalanceLoad_FullMethodName = "/grpc.lb.v1.LoadBalancer/BalanceLoad"
 )
 
+var (
+	LoadBalancer_FullMethodNames = []string{
+		"/grpc.lb.v1.LoadBalancer/BalanceLoad",
+	}
+)
+
 // LoadBalancerClient is the client API for LoadBalancer service.
 //
 // For semantics around ctx use and closing/ending streaming RPCs, please refer to https://pkg.go.dev/google.golang.org/grpc/?tab=doc#ClientConn.NewStream.
--- a/channelz/grpc_channelz_v1/channelz_grpc.pb.go
+++ b/channelz/grpc_channelz_v1/channelz_grpc.pb.go
@@ -49,6 +49,18 @@ const (
 	Channelz_GetSocket_FullMethodName        = "/grpc.channelz.v1.Channelz/GetSocket"
 )
 
+var (
+	Channelz_FullMethodNames = []string{
+		"/grpc.channelz.v1.Channelz/GetTopChannels",
+		"/grpc.channelz.v1.Channelz/GetServers",
+		"/grpc.channelz.v1.Channelz/GetServer",
+		"/grpc.channelz.v1.Channelz/GetServerSockets",
+		"/grpc.channelz.v1.Channelz/GetChannel",
+		"/grpc.channelz.v1.Channelz/GetSubchannel",
+		"/grpc.channelz.v1.Channelz/GetSocket",
+	}
+)
+
 // ChannelzClient is the client API for Channelz service.
 //
 // For semantics around ctx use and closing/ending streaming RPCs, please refer to https://pkg.go.dev/google.golang.org/grpc/?tab=doc#ClientConn.NewStream.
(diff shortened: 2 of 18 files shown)


purnesh42H:
Thanks @Elsharaky. Could you provide more details about what is your use case to require below functionalities and what is preventing to achieve the same with existing functionalities?

- This list is integrated into the code to facilitate service-wide control mechanisms in custom gRPC interceptors.
- By providing access to all method names, this feature allows developers to apply consistent logic, such as authentication, logging, or rate-limiting, across an entire service in a more streamlined and efficient manner.

Elsharaky:
Thanks @purnesh42H  for the feedback!

My use case involves implementing custom gRPC interceptors that need to apply logic at the service level, rather than on a per-method basis. Specifically, I’m working with interceptors that handle tasks like authentication, logging, and rate-limiting. These interceptors need to uniformly apply the same logic across all methods within a service.

Currently, achieving this requires manually maintaining a list of method names for each service, which can be error-prone and difficult to manage, especially as services evolve. By generating a list of full method names, the code becomes more maintainable and reduces the risk of human error. It also simplifies the interceptor logic by allowing it to reference a single, generated source of truth.

Regarding your concerns about the potential downsides:

Code size: The additional code is relatively small, especially compared to the benefits of reducing manual intervention.
Complexity: The complexity is managed within the code generator, meaning it doesn’t impact the end user directly.
Maintenance overhead: The generated list is automatically updated with the service definition, reducing the maintenance burden compared to manually tracking method names.
I understand the need to weigh these benefits against the potential downsides, and I’m open to suggestions on how we could mitigate any concerns.

purnesh42H:
Thanks @Elsharaky. Actually, interceptors are well-suited for implementing logic that is not specific to a single RPC method. If you haven't already, please check out the [official gRPC documentation on interceptors](https://grpc.io/docs/guides/interceptors/).

Interceptors can be added when building a gRPC channel or server, and they will be called for every RPC on that channel or server. This allows for consistent application of logic, such as authentication, logging, or rate-limiting, across all methods within a service.

Regarding your requirement for having a list of full method names, have you considered using the `FullMethod` property provided by gRPC's `UnaryServerInfo` and `StreamServerInfo` structures? This property can be accessed within the interceptor context and dynamically provides the service and method being called. For more details, you can refer to the [gRPC Go documentation](https://pkg.go.dev/google.golang.org/grpc#UnaryServerInfo).

Elsharaky:
**Thanks @purnesh42H  for the follow-up!**

I understand how the `FullMethod` property can be useful in many scenarios, but my use case is a bit different. I’m working with a dynamic implementation gRPC server that handles multiple microservices, each containing several gRPC services. 

In this setup, I need to apply an interceptor selectively to specific services rather than across all methods. For example, if I want to enforce authentication only for a particular service, I need a way to efficiently identify and manage all methods within that service.

Currently, achieving this requires explicitly marking each method within the service as authenticated. This is where a generated list of method names per service would be particularly valuable. It allows for a clean and automated way to apply such logic without manually handling each method.

The `FullMethod` property is helpful for method-level control, but it doesn’t fully address the need for service-wide control in a dynamic server environment where services are managed centrally and may share common infrastructure.

I hope this provides more context for why I believe the generated list would be beneficial in this specific scenario. I’m open to any alternative suggestions you might have on how to achieve this with existing features.

purnesh42H:
@Elsharaky this sounds like a specific usecase. Generally it's a good idea to start with a feature request rather than jumping straight to a pull request. A feature request allows the community and maintainers to provide feedback on whether the feature aligns with the project’s goals and more importantly, if it’s something that would benefit other users as well

You can open a new feature request here https://github.com/grpc/grpc-go/issues/new/choose

Elsharaky:
**Thanks @purnesh42H**

I’ll go ahead and open a feature request to explain the use case in detail and see if there’s broader interest or alternative solutions that could work within the project’s goals.

Thanks again for your guidance, and I appreciate the feedback throughout this process!

purnesh42H:
Thanks @Elsharaky. I will close this PR for now and provide further updates on the issue.

PR #7584, opened 2024-09-02
endpointsharding: Skip nil child pickers while updating ClientConn state

arjan-bal:
The child pickers are initialized to `nil` when child LB policies are created
https://github.com/grpc/grpc-go/blob/ca4865d6dd6f3d8b77f1943ccfd6c9e78223912d/balancer/endpointsharding/endpointsharding.go#L118-L122

When pick is called on endpointsharding's picker, it tries to call `Pick()` on a nil picker here and panics
https://github.com/grpc/grpc-go/blob/ca4865d6dd6f3d8b77f1943ccfd6c9e78223912d/balancer/endpointsharding/endpointsharding.go#L254-L258

This change ignores nil child pickers during aggregation. If all pickers are nil, the channel state is set to CONNECTING with error `balancer.ErrNoSubConnAvailable`.

Test logs demonstrating the panic: [panic_logs.txt](https://github.com/user-attachments/files/17193245/panic_logs.txt)

RELEASE NOTES:
* endpointsharding: Skip `nil` child pickers while updating ClientConn state. This avoid panics when an RPC is made before the child policy generates a picker.

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -116,7 +116,13 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 			bal = child.(*balancerWrapper)
 		} else {
 			bal = &balancerWrapper{
-				childState: ChildState{Endpoint: endpoint},
+				childState: ChildState{
+					Endpoint: endpoint,
+					State: balancer.State{
+						ConnectivityState: connectivity.Connecting,
+						Picker:            base.NewErrPicker(balancer.ErrNoSubConnAvailable),
+					},
+				},
 				ClientConn: es.cc,
 				es:         es,
 			}
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -116,13 +116,7 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 			bal = child.(*balancerWrapper)
 		} else {
 			bal = &balancerWrapper{
-				childState: ChildState{
-					Endpoint: endpoint,
-					State: balancer.State{
-						ConnectivityState: connectivity.Connecting,
-						Picker:            base.NewErrPicker(balancer.ErrNoSubConnAvailable),
-					},
-				},
+				childState: ChildState{Endpoint: endpoint},
 				ClientConn: es.cc,
 				es:         es,
 			}
@@ -196,6 +190,7 @@ func (es *endpointSharding) updateState() {
 	defer es.mu.Unlock()
 
 	children := es.children.Load()
+	pickerCount := 0
 	childStates := make([]ChildState, 0, children.Len())
 
 	for _, child := range children.Values() {
@@ -203,6 +198,10 @@ func (es *endpointSharding) updateState() {
 		childState := bw.childState
 		childStates = append(childStates, childState)
 		childPicker := childState.State.Picker
+		if childPicker == nil {
+			continue
+		}
+		pickerCount++
 		switch childState.State.ConnectivityState {
 		case connectivity.Ready:
 			readyPickers = append(readyPickers, childPicker)
@@ -214,6 +213,15 @@ func (es *endpointSharding) updateState() {
 			transientFailurePickers = append(transientFailurePickers, childPicker)
 			// connectivity.Shutdown shouldn't appear.
 		}
+	}
+
+	// If there are no pickers available yet, return a queuing picker.
+	if pickerCount == 0 {
+		es.cc.UpdateState(balancer.State{
+			ConnectivityState: connectivity.Connecting,
+			Picker:            base.NewErrPicker(balancer.ErrNoSubConnAvailable),
+		})
+		return
 	}
 
 	// Construct the round robin picker based off the aggregated state. Whatever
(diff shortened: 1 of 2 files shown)


balancer/endpointsharding/endpointsharding_test.go
```diff
@@ -157,3 +162,54 @@ func (s) TestEndpointShardingBasic(t *testing.T) {
...
+	cc, err := grpc.Dial(mr.Scheme()+":///", grpc.WithResolvers(mr), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatalf("Failed to dial: %v", err)
+	}
+	defer cc.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), defaultShortTestTimeout)
+	defer cancel()
+	client := testgrpc.NewTestServiceClient(cc)
+
+	// Even though the child LB policy hasn't given an picker updates, it is
```
purnesh42H, line 210: > // Even though the child LB policy hasn't given an picker updates, it is assumed that its in connecting state

We should verify that there is no state change i.e. balancer was created with `connecting` state and balancer is still in `connecting` state. May be check `State.ConnectivityState`?

dfawley:
Is this just covering up for a child LB policy that isn't following the rules of the API?  Should the child policy be required to set its state?

Note that we will want some children to start IDLE, specifically we'll probably want to create a wrapper LB policy that creates a single child in IDLE and only starts it when it gets an RPC or an `ExitIdle` call.

dfawley:
> The endpointsharding policy was not setting the initial state for child policies and would panic if the child hasn't provided a balancer state update and an RPC is made.

What is the panic, exactly?  There's no explanation in the PR description and no linked issue here.

arjan-bal:
> What is the panic, exactly? There's no explanation in the PR description and no linked issue here.

Attached logs to the PR description. The problem is that the child pickers in the child state are not initialized here:
https://github.com/grpc/grpc-go/blob/ca4865d6dd6f3d8b77f1943ccfd6c9e78223912d/balancer/endpointsharding/endpointsharding.go#L118-L122

When pick is called on the endpointsharding balancer, it tries to call `Pick()` on a nil picker here:
https://github.com/grpc/grpc-go/blob/ca4865d6dd6f3d8b77f1943ccfd6c9e78223912d/balancer/endpointsharding/endpointsharding.go#L254-L258

arjan-bal:
> Is this just covering up for a child LB policy that isn't following the rules of the API? Should the child policy be required to set its state?
 
Are LB policies required to produce a picker when `UpdateClientConnState` is called?
I found this bug while working on `pickfirst` changes. From a discussion on the `pickfirst` PR, I found that `pickfirst` can assume that it starts in state `CONNECTING`. Reason being that the LB policy should not be built if it doesn't need to connect. I then copied an optimization from Java, which ignored sending duplicate `CONNECTING` pickers as the state is remains the same. I found that `endpointsharding` panics if the child policy doesn't send a picker on the first call to `UpdateClientConnState`.

dfawley:
> The problem is that the child pickers in the child state are not initialized here

But..there are no child pickers yet.  It seems like the _real_ problem is that the endpointsharding balancer shouldn't perform any `UpdateState` calls if no children have done so yet, either.

> Are LB policies required to produce a picker when UpdateClientConnState is called?

No, there is no such requirement today.

> From a discussion on the pickfirst PR, I found that pickfirst can assume that it starts in state CONNECTING.

I don't think that's quite correct.  Pickfirst should START in state CONNECTING.  It should do so by producing a picker that queues RPCs and sets the state to CONNECTING.

> I then copied an optimization from Java, which ignored sending duplicate CONNECTING pickers as the channel state and picker error remains the same.

This seems OK as long as a picker has been sent that says CONNECTING.  You shouldn't suppress the first update based on the assumption that you started CONNECTING.

dfawley:
I think this change is actually OK, but I'm concerned.

I would be OK with the assumptions here if it's documented in the balancer API _and_ if we audit _all_ parent LB policies that we own to make sure they conform to this behavior.

This is not currently documented, however, so it's questionable to propagate this assumption throughout the tree.

Also I don't see anything in the channel that does this, so I think it's the gracefulswitch balancer that makes the channel behave this way.  But I'm not exactly clear on where we would set the channel state to CONNECTING while we're still waiting for the initial name resolver update.  Maybe we aren't, and that's a bug?  Or maybe we're initializing more than I was expecting even while waiting for the initial update?

dfawley:
I filed #7686 to make sure we have a test case for the behavior in the last paragraph I wrote above.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -210,6 +215,15 @@ func (es *endpointSharding) updateState() {
 		}
 	}
 
+	// If there are no pickers available yet, return a queuing picker.
+	if pickerCount == 0 {
+		es.cc.UpdateState(balancer.State{
+			ConnectivityState: connectivity.Connecting,
+			Picker:            base.NewErrPicker(balancer.ErrNoSubConnAvailable),
+		})
```
dfawley, line 223: If no children have produced a picker, why are we doing so here?

I think the safest behavior here is to either do what you said before (and go through the exercise to make "initial state == connecting" an API guarantee), or to only have this endpointsharding LB policy produce pickers in sync with its children: i.e. whenever any child updates its picker, this policy updates its combined picker.

arjan-bal: If we don't send a picker here, the unit test added in the PR indicates that the channel remains in IDLE mode.

arjan-bal:
Closing this PR, will open it once https://github.com/grpc/grpc-go/issues/7686 is fixed.

PR #7575, opened 2024-08-30
.*: fix lint issues of not having comments for exported funcs and vars along with any remaining issues and enable remaining disabled rules

purnesh42H:
Addresses: #7444

RELEASE NOTES: None

PR #7589, opened 2024-09-04
vet: enforce revive linter

purnesh42H:
Addresses: #7444 

Enforce the revive linter with a configuration that enables only the rules that are already fixed, while keeping the rules that are yet to be fixed disabled. We will enable the remaining rules as we address and resolve them.

[Recommended lint rules](https://github.com/mgechev/revive/blob/master/README.md#recommended-configuration) by revive (cover default golint rules) except empty-block

RELEASE NOTES: None


dfawley (approved):

scripts/revive.toml
```diff
@@ -0,0 +1,27 @@
...
+[rule.indent-error-flow]
+[rule.errorf]
+[rule.superfluous-else]
+[rule.unused-parameter]
+[rule.unreachable-code]
+[rule.redefines-builtin-id]
+
+# Disabled rules
+[rule.empty-block]
+    Disabled = true
```
arvindbr8, line 27: Since we have enough effective usages of empty-block, imo we should not bother about it at all. 

```suggestion
```

scripts/revive.toml
arvindbr8, line 1: not trying to be pedantic here, but going back to first principles - I think we wanted to improve our OSS CI to make the internal linters pass more reliably. 

Looking at the list posted [here](https://github.com/mgechev/revive/blob/master/README.md#available-rules) - the ones that also apply for golint are a subset of the recommended list. Golint ignores:
- [rule.superfluous-else]
- [rule.unused-parameter]
- [rule.unreachable-code]
- [rule.redefines-builtin-id]

That being said, having them seems like a good idea to me!

Here are some more I think is worth adding:

- [rule.import-shadowing]
  - Spots identifiers that shadow an import 	
- [rule.use-any] 
  - Proposes to replace `interface{}` with its alias `any`
  - remove the existing check from [vet.sh](https://github.com/grpc/grpc-go/blob/master/scripts/vet.sh#L59)
- [rule.empty-lines]
  - Warns when there are heading or trailing newlines in a block
    - unclear what a block means here. needs investigation
<details>
<summary>Adding this here for posterity</summary>

Available in revive and not in golint
| Name                  | Config | Description                                                      | `golint` | Typed |
| --------------------- | :----: | :--------------------------------------------------------------- | :------: | :---: |
| [`if-return`]           |  n/a   | Redundant if when returning an error.                            |   no    |  no   |
| [`argument-limit`]      |  int (defaults to 8)  | Specifies the maximum number of arguments a function can receive |    no    |  no   |
| [`cyclomatic`]          |  int (defaults to 10)   | Sets restriction for maximum Cyclomatic complexity.              |    no    |  no   |
| [`max-public-structs`]  |  int (defaults to 5)  | The maximum number of public structs in a file.                  |    no    |  no   |
| [`file-header`]         | string (defaults to none)| Header which each file should have.                              |    no    |  no   |
| [`superfluous-else`]    |  []string   | Prevents redundant else statements (extends [`indent-error-flow`]) |    no    |  no   |
| [`confusing-naming`]    |  n/a   | Warns on methods with names that differ only by capitalization   |    no    |  no   |
| [`get-return`]          |  n/a   | Warns on getters that do not yield any result                    |    no    |  no   |
| [`modifies-parameter`]  |  n/a   | Warns on assignments to function parameters                      |    no    |  no   |
| [`confusing-results`]   |  n/a   | Suggests to name potentially confusing function results          |    no    |  no   |
| [`deep-exit`]           |  n/a   | Looks for program exits in funcs other than `main()` or `init()` |    no    |  no   |
| [`unused-parameter`]    |  n/a   | Suggests to rename or remove unused function parameters          |    no    |  no   |
| [`unreachable-code`]    |  n/a   | Warns on unreachable code                                        |    no    |  no   |
| [`add-constant`]        |  map   | Suggests using constant for magic numbers and string literals    |    no    |  no   |
| [`flag-parameter`]      |  n/a   | Warns on boolean parameters that create a control coupling       |    no    |  no   |
| [`unnecessary-stmt`]    |  n/a   | Suggests removing or simplifying unnecessary statements          |    no    |  no   |
| [`struct-tag`]          |  []string   | Checks common struct tags like `json`, `xml`, `yaml`               |    no    |  no   |
| [`constant-logical-expr`]   |  n/a   | Warns on constant logical expressions                        |    no    |  no   |
| [`bool-literal-in-expr`]|  n/a   | Suggests removing Boolean literals from logic expressions        |    no    |  no   |
| [`redefines-builtin-id`]|  n/a   | Warns on redefinitions of builtin identifiers                    |    no    |  no   |
| [`function-result-limit`] |  int (defaults to 3)| Specifies the maximum number of results a function can return    |    no    |  no   |
| [`imports-blocklist`]   | []string | Disallows importing the specified packages                     |    no    |  no   |
| [`range-val-in-closure`]|  n/a   | Warns if range value is used in a closure dispatched as goroutine|    no    |  no   |
| [`waitgroup-by-value`]  |  n/a   | Warns on functions taking sync.WaitGroup as a by-value parameter |    no    |  no   |
| [`atomic`]              |  n/a   | Check for common mistaken usages of the `sync/atomic` package    |    no    |  no   |
| [`empty-lines`]   | n/a | Warns when there are heading or trailing newlines in a block              |    no    |  no   |
| [`line-length-limit`]   | int (defaults to 80) | Specifies the maximum number of characters in a line             |    no    |  no   |
| [`call-to-gc`]   | n/a    | Warns on explicit call to the garbage collector    |    no    |  no   |
| [`duplicated-imports`] | n/a  | Looks for packages that are imported two or more times   |    no    |  no   |
| [`import-shadowing`]   | n/a    | Spots identifiers that shadow an import    |    no    |  no   |
| [`bare-return`] | n/a  | Warns on bare returns   |    no    |  no   |
| [`unused-receiver`]   | n/a    | Suggests to rename or remove unused method receivers    |    no    |  no   |
| [`cognitive-complexity`]          |  int (defaults to 7) | Sets restriction for maximum Cognitive complexity.              |    no    |  no   |
| [`string-format`]          |  map   | Warns on specific string literals that fail one or more user-configured regular expressions            |    no    |  no   |
| [`early-return`]          | []string   | Spots if-then-else statements where the predicate may be inverted to reduce nesting |    no    |  no   |
| [`unconditional-recursion`]          |  n/a   | Warns on function calls that will lead to (direct) infinite recursion |    no    |  no   |
| [`identical-branches`]          |  n/a   | Spots if-then-else statements with identical `then` and `else` branches       |    no    |  no   |
| [`defer`]          |  map   |  Warns on some [defer gotchas](https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1)       |    no    |  no   |
| [`unexported-naming`]          |  n/a   |  Warns on wrongly named un-exported symbols       |    no    |  no   |
| [`function-length`]          |  int, int (defaults to 50 statements, 75 lines)   |  Warns on functions exceeding the statements or lines max |    no    |  no   |
| [`nested-structs`]          |  n/a   |  Warns on structs within structs |    no    |  no   |
| [`useless-break`]          |  n/a   |  Warns on useless `break` statements in case clauses |    no    |  no   |
| [`banned-characters`]          |  []string (defaults to []string{})   |  Checks banned characters in identifiers |    no    |  no   |
| [`optimize-operands-order`]          |  n/a   |  Checks inefficient conditional expressions |    no    |  no   |
| [`use-any`]          |  n/a   |  Proposes to replace `interface{}` with its alias `any` |    no    |  no   |
| [`datarace`]          |  n/a   |  Spots potential dataraces |    no    |  no   |
| [`comment-spacings`]          |  []string   |  Warns on malformed comments |    no    |  no   |
| [`redundant-import-alias`]          |  n/a   |  Warns on import aliases matching the imported package name |    no    |  no   |
| [`import-alias-naming`]         | string or map[string]string (defaults to allow regex pattern ^[a-z][a-z0-9]{0,}$) | Conventions around the naming of import aliases.                              |    no    |  no   |
| [`enforce-map-style`] |  string (defaults to "any")  |  Enforces consistent usage of `make(map[type]type)` or `map[type]type{}` for map initialization. Does not affect `make(map[type]type, size)` constructions. |    no    |  no   |
| [`enforce-slice-style`] |  string (defaults to "any")  |  Enforces consistent usage of `make([]type, 0)` or `[]type{}` for slice initialization. Does not affect `make(map[type]type, non_zero_len, or_non_zero_cap)` constructions. |    no    |  no   |
| [`enforce-repeated-arg-type-style`] |  string (defaults to "any")  |  Enforces consistent style for repeated argument and/or return value types. |    no    |  no   |
| [`max-control-nesting`] |  int (defaults to 5)  | Sets restriction for maximum nesting of control structures. |    no    |  no   |
| [`comments-density`] |  int (defaults to 0)  | Enforces a minimum comment / code relation |    no    |  no   |
| [`time-equal`]         |  n/a   | Suggests to use `time.Time.Equal` instead of `==` and `!=` for equality check time.                 |   no    |  yes  |
| [`unchecked-type-assertions`]         |  n/a   | Disallows type assertions without checking the result.                 |   no    |  yes  |
| [`empty-block`]         |  n/a   | Warns on empty code blocks                                       |    no    |  yes   |
| [`modifies-value-receiver`] |  n/a   | Warns on assignments to value-passed method receivers        |    no    |  yes  |
| [`range-val-address`]|  n/a   | Warns if address of range value is used dangerously |    no    |  yes   |
| [`unhandled-error`]   | []string   | Warns on unhandled errors returned by function calls    |    no    |  yes   |
| [`string-of-int`]          |  n/a   | Warns on suspicious casts from int to string            |    no    |  yes   |

Available in revive and in golint

| Name                  | Config | Description                                                      | `golint` | Typed |
| --------------------- | :----: | :--------------------------------------------------------------- | :------: | :---: |
| [`context-keys-type`]   |  n/a   | Disallows the usage of basic types in `context.WithValue`.       |   yes    |  yes  |
| [`time-naming`]         |  n/a   | Conventions around the naming of time variables.                 |   yes    |  yes  |
| [`var-declaration`]     |  n/a   | Reduces redundancies around variable declaration.                |   yes    |  yes  |
| [`unexported-return`]   |  n/a   | Warns when a public return is from unexported type.              |   yes    |  yes  |
| [`errorf`]              |  n/a   | Should replace `errors.New(fmt.Sprintf())` with `fmt.Errorf()`   |   yes    |  yes  |
| [`blank-imports`]       |  n/a   | Disallows blank imports                                          |   yes    |  no   |
| [`context-as-argument`] |  n/a   | `context.Context` should be the first argument of a function.    |   yes    |  no   |
| [`dot-imports`]         |  n/a   | Forbids `.` imports.                                             |   yes    |  no   |
| [`error-return`]        |  n/a   | The error return parameter should be last.                       |   yes    |  no   |
| [`error-strings`]       |  []string   | Conventions around error strings.                                |   yes    |  no   |
| [`error-naming`]        |  n/a   | Naming of error variables.                                       |   yes    |  no   |
| [`exported`]            |  []string   | Naming and commenting conventions on exported symbols.           |   yes    |  no   |
| [`increment-decrement`] |  n/a   | Use `i++` and `i--` instead of `i += 1` and `i -= 1`.            |   yes    |  no   |
| [`var-naming`]          |  allowlist & blocklist of initialisms   | Naming rules.                                                    |   yes    |  no   |
| [`package-comments`]    |  n/a   | Package commenting conventions.                                  |   yes    |  no   |
| [`range`]               |  n/a   | Prevents redundant variables when iterating over a collection.   |   yes    |  no   |
| [`receiver-naming`]     |  n/a   | Conventions around the naming of receivers.                      |   yes    |  no   |
| [`indent-error-flow`]   |  []string   | Prevents redundant else statements.                              |   yes    |  no   |

</details>

scripts/revive.toml
arvindbr8: Instead can we add explicit disable for the fixes that are in flight. That way we can rebase this change, and remove that check as we fix them. Seems cleaner to me. That way we can avoid having to see all the linter error annotations on PRs

wdyt?

scripts/vet.sh
```diff
@@ -178,7 +178,7 @@ done
 
 # Collection of revive linter analysis checks
 REV_OUT="$(mktemp)"
-revive -set_exit_status=1 -exclude "reflection/test/grpc_testing_not_regenerate/" -formatter plain ./... >"${REV_OUT}" || true
+revive -set_exit_status=1 -exclude "reflection/test/grpc_testing_not_regenerate/" -formatter plain -config "$(dirname "$0")/revive.toml" ./... >"${REV_OUT}" || true
```
arvindbr8, line 181: I think we can safely combine lines 180-184 to be something like:

```suggestion
revive \
  -set_exit_status=1 \
  -exclude "reflection/test/grpc_testing_not_regenerate/" \
  -exclude "**/*.pb.go" \
  -config "$(dirname "$0")/revive.toml" \
  ./...
```

scripts/revive.toml
arvindbr8: cc: @dfawley

scripts/revive.toml
purnesh42H: I like the idea. i modified to disable the rules that are yet to merged. Let me know if that looks good

scripts/revive.toml
purnesh42H: I added use-any. empty-lines check though is weird. It is detecting even cases like `func main() {` which we don't want I think.

Btw, do you want to enable all rules? If yes, we can `enableAllRules = true` and disable only what we don't want currently.

scripts/vet.sh
purnesh42H: Done

examples/features/advancedtls/client/main.go
arvindbr8, line 1: Let's revert these changes from this pr and add it to https://github.com/grpc/grpc-go/pull/7575

scripts/revive.toml
arvindbr8: Sounds good to me. Please consider this comment non-blocking

arvindbr8 (approved):

rpc_util.go
arvindbr8, line 1: I missed to comment on this file in the last pass

scripts/revive.toml
```diff
@@ -0,0 +1,33 @@
...
+# Disabled rules
+# https://github.com/grpc/grpc-go/issues/7444 to track TODO(s)
+[rule.empty-block]
+    Disabled = true
+[rule.import-shadowing]
+    Disabled = true
+[rule.exported]
+    Disabled = true # TODO: Enable after existing issues are fixed
+[rule.redefines-builtin-id]
+    Disabled = true # TODO: Enable after existing issues are fixed
```
arvindbr8, line 33: newline please

examples/features/advancedtls/client/main.go
purnesh42H: Done

rpc_util.go
purnesh42H: Me too :)

arvindbr8 (approved):
sorry, i last comment

scripts/revive.toml
purnesh42H: Added

PR #7595, opened 2024-09-08
mem: replace flate.Reader reference

ash2k:
This interface has nothing to do with compression and there are no other references to that package in the codebase. If we remove the reference we same compiler and linker some work - no need to parse, compile, link that package just to use one interface from it (assuming no other references from the codebase).

RELEASE NOTES:
- mem: remove dependency on `compress/flate` package

### DIFF
--- a/mem/buffer_slice.go
+++ b/mem/buffer_slice.go
@@ -19,7 +19,6 @@
 package mem
 
 import (
-	"compress/flate"
 	"io"
 )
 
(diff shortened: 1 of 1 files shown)


dfawley (approved):
cc @PapaCharlie

PapaCharlie:
Yep makes sense to me!

easwars (approved):

purnesh42H (approved):

PR #7598, opened 2024-09-09
Change version to 1.67.0

purnesh42H:
RELEASE NOTES: None


dfawley:
FYI: this PR should not be merged until immediately before the release.

dfawley:
Also the target is wrong here.  It should be going into the v1.67.x branch.

PR #7614, opened 2024-09-10
balancer/rls: Add picker and cache unit tests for RLS Metrics

zasweq:
This PR adds picker and cache unit tests to test RLS Metrics.

RELEASE NOTES: N/A

### DIFF
--- a/balancer/rls/cache_test.go
+++ b/balancer/rls/cache_test.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	estats "google.golang.org/grpc/experimental/stats"
 	"google.golang.org/grpc/internal/backoff"
 	"google.golang.org/grpc/internal/testutils/stats"
 )
@@ -251,7 +252,7 @@ func (s) TestDataCache_Metrics(t *testing.T) {
 		{size: 4},
 		{size: 5},
 	}
-	tmr := stats.NewTestMetricsRecorder(t)
+	tmr := stats.NewTestMetricsRecorder()
 	dc := newDataCache(50, nil, tmr, "")
 
 	dc.updateRLSServerTarget("rls-server-target")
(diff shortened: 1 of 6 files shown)


internal/testutils/stats/test_metrics_recorder.go
```diff
@@ -75,6 +75,15 @@ func (r *TestMetricsRecorder) AssertDataForMetric(metricName string, wantVal flo
 	}
 }
 
+// AssertNoDataForMetric asserts no data is present for metric.
+func (r *TestMetricsRecorder) AssertNoDataForMetric(metricName string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if _, ok := r.data[estats.Metric(metricName)]; ok {
+		r.t.Fatalf("Data is present for metric %v", metricName)
```
dfawley, line 83: This is following an anti-pattern: https://google.github.io/styleguide/go/decisions#assert

Now that we have more time, let's address this.

balancer/rls/cache_test.go
```diff
@@ -242,3 +242,50 @@ func (s) TestDataCache_ResetBackoffState(t *testing.T) {
 		t.Fatalf("unexpected diff in backoffState for cache entry after dataCache.resetBackoffState(): %s", diff)
 	}
 }
+
+var cacheEntriesMetricsTests = []*cacheEntry{
```
dfawley, line 246: Please move inside test function so it's not global.

balancer/rls/cache_test.go
```diff
@@ -242,3 +242,50 @@ func (s) TestDataCache_ResetBackoffState(t *testing.T) {
...
+	// 5 total entries, size 3 each, so should record 5 entries and 15
+	// size.
+	tmr.AssertDataForMetric("grpc.lb.rls.cache_entries", 5)
+	tmr.AssertDataForMetric("grpc.lb.rls.cache_size", 15)
+
+	// Resize down the cache to 3 entries. This should scale down the cache to 3
+	// entries with 3 bytes each, so should record 3 entries and 9 size.
+	dc.resize(9)
+	tmr.AssertDataForMetric("grpc.lb.rls.cache_entries", 3)
+	tmr.AssertDataForMetric("grpc.lb.rls.cache_size", 9)
```
dfawley, line 272: This relies on having the same size for every entry? Or does order matter?

balancer/rls/cache_test.go
```diff
@@ -242,3 +242,50 @@ func (s) TestDataCache_ResetBackoffState(t *testing.T) {
 		t.Fatalf("unexpected diff in backoffState for cache entry after dataCache.resetBackoffState(): %s", diff)
 	}
 }
+
+var cacheEntriesMetricsTests = []*cacheEntry{
+	{size: 3},
+	{size: 3},
+	{size: 3},
+	{size: 3},
```
dfawley, line 250: It seems like these should vary to have better coverage.

balancer/rls/cache_test.go
zasweq: Ah good point. Done.

balancer/rls/cache_test.go
zasweq: Yeah it relies. I think operations are deterministic if I tweak the numbers, but I don't know if we'll have that guarantee in the future. Let me play around with this some.

balancer/rls/cache_test.go
zasweq: Switched to 12345, and scaled the assertions to follow this. Luckily all the cache operations are deterministic.

internal/testutils/stats/test_metrics_recorder.go
zasweq: Didn't pass t to this component anymore, did all the assertions inline.

balancer/weightedroundrobin/balancer_test.go
```diff
@@ -234,12 +235,28 @@ func (s) TestWRRMetricsBasic(t *testing.T) {
...
+	}
+	if got := mr.Data[estats.Metric("grpc.lb.wrr.endpoint_weight_not_yet_usable")]; got != 1 {
+		t.Fatalf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.wrr.endpoint_weight_not_yet_usable", got, 1)
+	}
+	// Unusable, so no endpoint weight. Due to only one SubConn, this will never
+	// update the weight. Thus, this will stay 0.
+	if got := mr.Data[estats.Metric("grpc.lb.wrr.endpoint_weight_stale")]; got != 0 {
+		t.Fatalf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.wrr.endpoint_weight_stale", got, 0)
+	}
+	/*mr.AssertDataForMetric("grpc.lb.wrr.rr_fallback", 1)           // Falls back because only one SubConn.
```
dfawley, line 254: Please fix this up.

balancer/rls/cache_test.go
```diff
@@ -242,3 +243,61 @@ func (s) TestDataCache_ResetBackoffState(t *testing.T) {
...
+	for i, k := range cacheKeys {
+		dc.addEntry(k, cacheEntriesMetricsTests[i])
+	}
+
+	const cacheEntriesKey = "grpc.lb.rls.cache_entries"
+	const cacheSizeKey = "grpc.lb.rls.cache_size"
+	// 5 total entries which add up to 15 size, so should record that.
+	if got := tmr.Data[estats.Metric(cacheEntriesKey)]; got != 5 {
+		t.Fatalf("Unexpected data for metric %v, got: %v, want: %v", cacheEntriesKey, got, 5)
+	}
```
dfawley, line 268: Consider the pattern: `if got, want := tmr.Data[], 5; got != want { t.Fatalf(..., got, want) }`

balancer/weightedroundrobin/balancer_test.go
```diff
@@ -87,7 +88,7 @@ var (
 		OOBReportingPeriod:      stringp("0.005s"),
 		BlackoutPeriod:          stringp("0s"),
 		WeightExpirationPeriod:  stringp("60s"),
-		WeightUpdatePeriod:      stringp(".050s"),
+		WeightUpdatePeriod:      stringp("30s"),
```
dfawley, line 91: Why is this changed?

internal/testutils/stats/test_metrics_recorder.go
```diff
@@ -44,57 +44,30 @@ type TestMetricsRecorder struct {
 	floatHistoCh *testutils.Channel
 	intGaugeCh   *testutils.Channel
 
-	// The most recent update for each metric name.
-	mu   sync.Mutex
-	data map[estats.Metric]float64
+	// Mu protects Data.
+	Mu sync.Mutex
+	// Data is the most recent update for each metric name.
+	Data map[estats.Metric]float64
```
dfawley, line 50: A getter may be better than exporting?

```go
func (...) Metric(name string) float64
```

balancer/weightedroundrobin/balancer_test.go
zasweq: Not strictly needed; but I figured I'd make it hard for the scheduler to emit during the test which would race with test (but I protect with mu so no issue there).

balancer/rls/cache_test.go
zasweq: Yeah I was doing everything not aliased and the gots were too long and figured I could do it with wants to, but decided just to do got. It looks nice though so I'l switch this.

balancer/weightedroundrobin/balancer_test.go
zasweq: Whoops deleted.

balancer/rls/cache_test.go
zasweq: Discussed offline; not possible due to returning two values from helper.

internal/testutils/stats/test_metrics_recorder.go
```diff
@@ -36,18 +35,16 @@ import (
...
-
 	intCountCh   *testutils.Channel
 	floatCountCh *testutils.Channel
 	intHistoCh   *testutils.Channel
 	floatHistoCh *testutils.Channel
 	intGaugeCh   *testutils.Channel
 
-	// Mu protects Data.
+	// Mu protects data.
 	Mu sync.Mutex
```
dfawley, line 45: The main goal of making a getter was so that `mu` didn't need to be exported and to make access simpler, since mutexes are generally internal state and complicate things.

Seems like you need something more complex if you need an atomic read of multiple values (though this wasn't supported previously?).  `func Metrics(name []string) (*float64)`???

zasweq: Hmmmmm I just deleted the grab of it in balancer_test. I was worried due to the low time for weight updates (when time.Ticker goes off and reupdates scheduler) but I made it 30 seconds so should never race (it'll fail much earlier than that). So mutex still grabbed sometimes in this component in exported helpers.

dfawley (approved):

Issue #7657, opened 2024-09-21
High memory usage when XDS is used


snewell:
### What version of gRPC are you using?
Verified with multiple versions from v1.61.0 through v1.66.2

### What version of Go are you using (`go version`)?
Tested with multiple go versions include 1.21.11 and 1.22.6

### What operating system (Linux, Windows, …) and version?
Docker containers based on two versions of Linux: ubuntu 22 and redhat 9

### What did you do?
Memory usage increases seemingly unchecked on a server when XDS is enabled and there are inbound requests.  This does not happen without XDS, on a completely idle server, or when grpc-go < v1.61.0 (verified with v1.60.0 and v1.60.1).  A full reproducible example with server, client, and protobufs (all stripped to what I believe are minimum) is available here: https://github.com/snewell/memleak-demo.

We're using an istio sidecar for XDS.  I don't have access to another XDS provider (e.g., traffic director).  Verified the memory usage is in my gRPC server, not the sidecar (pprof output included below).  Without XDS memory usage is flat, even with istio enabled.

Created a stripped down server and protobuf that does nothing other than provide an endpoint.  With a client connecting and sending requests every 5 seconds, memory increased.  This was also observed a more complex system as well, but the stripped down version is included here.  Relevant code:

#### Protos
```
syntax = "proto3";

option go_package = "github.com/snewell/memleak-demo/internal/pb";

package service;

service SS {
  rpc DoIt (Request) returns (stream Response) {}
}

message Request {
}

message Response {
}
```

#### Server
```
type ss struct {
	pb.UnimplementedSSServer
}

func (s ss) DoIt(*pb.Request, grpc.ServerStreamingServer[pb.Response]) error {
	return nil
}

func serverFn(cmd *cobra.Command, args []string) {
	go func() {
		log.Println(http.ListenAndServe(":6060", nil))
	}()

	listenAddr := fmt.Sprintf(":%v", listenPort)
	log.Printf("Listening on %v", listenAddr)
	lis, err := net.Listen("tcp", listenAddr)
	if err != nil {
		log.Fatalf("failed to listen: %v", err)

	}

	creds := insecure.NewCredentials()
	if creds, err = xdscreds.NewServerCredentials(xdscreds.ServerOptions{FallbackCreds: creds}); err != nil {
		log.Fatalf("failed to create server-side xDS credentials: %v", err)
	}

	xdsServer, err := xds.NewGRPCServer(grpc.Creds(creds))
	if err != nil {
[...]

PR #7664, opened 2024-09-23
xds/server: Fix xDS Server leak

zasweq:
Discovered in #7657.

My Dynamic RDS fix added a ref to the server transport in the wrapped connection: https://github.com/grpc/grpc-go/pull/6915/files#diff-dd56a1b7688625b5b70cd616b08c301d12f7f01edbd9be95c506743fd58a6155R140 to Drain correctly even right after Accepting the connection. It also added a ref to wrapped connection, which only gets cleared on the xDS Server's Serving State changing to Not Serving and filter chain updates: https://github.com/grpc/grpc-go/pull/6915/files#diff-e4706c72ae912399b7f8ee6f04cec2374ef7a7679b12358f201ddb0b45e34146R344. In the production environment/test case the connection closes but the listener continues to listen and the server continues to serve in a state that is not NOT_SERVING thus keeping a ref to the wrapped connection around. The solution is to clear the ref to the wrapped connection when the connection closes.

RELEASE NOTES:
* xds/server: Fix xDS Server memory leak

### DIFF
--- a/xds/internal/server/listener_wrapper.go
+++ b/xds/internal/server/listener_wrapper.go
@@ -364,7 +364,6 @@ func (l *listenerWrapper) Close() error {
 		l.cancelWatch()
 	}
 	l.rdsHandler.close()
-	l.conns = nil
 	return nil
 }
 
--- a/xds/internal/server/listener_wrapper.go
+++ b/xds/internal/server/listener_wrapper.go
@@ -136,7 +136,7 @@ type listenerWrapper struct {
 
 	// mu guards access to the current serving mode and the active filter chain
 	// manager.
-	mu sync.RWMutex
+	mu sync.Mutex
 	// Current serving mode.
 	mode connectivity.ServingMode
 	// Filter chain manager currently serving.
@@ -203,11 +203,8 @@ func (l *listenerWrapper) maybeUpdateFilterChains() {
 	// gracefully shut down with a grace period of 10 minutes for long-lived
 	// RPC's, such that clients will reconnect and have the updated
 	// configuration apply." - A36
-	var connsToClose map[*connWrapper]bool
-	if l.activeFilterChainManager != nil { // If there is a filter chain manager to clean up.
-		connsToClose = l.conns
-		l.conns = make(map[*connWrapper]bool)
-	}
+	connsToClose := l.conns
+	l.conns = make(map[*connWrapper]bool)
 	l.activeFilterChainManager = l.pendingFilterChainManager
 	l.pendingFilterChainManager = nil
 	l.instantiateFilterChainRoutingConfigurationsLocked()
@@ -348,7 +345,7 @@ func (l *listenerWrapper) Accept() (net.Conn, error) {
 	}
 }
 
-func (l *listenerWrapper) RemoveConn(conn *connWrapper) {
+func (l *listenerWrapper) removeConn(conn *connWrapper) {
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	delete(l.conns, conn)
--- a/xds/internal/server/listener_wrapper.go
+++ b/xds/internal/server/listener_wrapper.go
@@ -86,6 +86,7 @@ func NewListenerWrapper(params ListenerWrapperParams) net.Listener {
 		xdsC:              params.XDSClient,
 		modeCallback:      params.ModeCallback,
 		isUnspecifiedAddr: params.Listener.Addr().(*net.TCPAddr).IP.IsUnspecified(),
+		conns:             make(map[*connWrapper]bool),
 
 		mode:   connectivity.ServingModeNotServing,
 		closed: grpcsync.NewEvent(),
@@ -141,7 +142,7 @@ type listenerWrapper struct {
 	// Filter chain manager currently serving.
 	activeFilterChainManager *xdsresource.FilterChainManager
 	// conns accepted with configuration from activeFilterChainManager.
-	conns []*connWrapper
+	conns map[*connWrapper]bool
 
 	// These fields are read/written to in the context of xDS updates, which are
 	// guaranteed to be emitted synchronously from the xDS Client. Thus, they do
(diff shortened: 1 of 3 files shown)


xds/internal/server/listener_wrapper.go
```diff
@@ -341,12 +342,18 @@ func (l *listenerWrapper) Accept() (net.Conn, error) {
...
 		cw := &connWrapper{Conn: conn, filterChain: fc, parent: l, urc: fc.UsableRouteConfiguration}
-		l.conns = append(l.conns, cw)
-		l.mu.RUnlock()
+		l.conns[cw] = true
+		l.mu.Unlock()
 		return cw, nil
 	}
 }
 
+func (l *listenerWrapper) RemoveConn(conn *connWrapper) {
```
dfawley, line 351: Unexport?  I don't think this is here for an interface.

xds/internal/server/listener_wrapper_test.go
```diff
@@ -151,5 +157,109 @@ func (s) TestListenerWrapper(t *testing.T) {
...
+	case mode := <-modeCh:
+		if mode != connectivity.ServingModeServing {
+			t.Fatalf("mode change received: %v, want: %v", mode, connectivity.ServingModeServing)
+		}
+	}
+
+	server := grpc.NewServer(grpc.Creds(insecure.NewCredentials()))
+	testgrpc.RegisterTestServiceServer(server, &testService{})
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
dfawley, line 226: Is this necessary?

Does `Serve` block on anything after `Stop` (not `GracefulStop`) is called?

xds/internal/server/listener_wrapper_test.go
```diff
@@ -151,5 +157,109 @@ func (s) TestListenerWrapper(t *testing.T) {
...
+	}()
+
+	// Make 100 connections to the server, and make an RPC on each one.
+	for i := 0; i < 100; i++ {
+		cc, err := grpc.NewClient(lw.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+		if err != nil {
+			t.Fatalf("grpc.NewClient failed with err: %v", err)
+		}
+		client := testgrpc.NewTestServiceClient(cc)
+		if _, err := client.EmptyCall(ctx, &testpb.Empty{}, grpc.WaitForReady(true)); err != nil {
```
dfawley, line 241: WFR should basically never be needed unless you're expecting transient errors.

xds/internal/server/listener_wrapper_test.go
```diff
@@ -151,5 +157,109 @@ func (s) TestListenerWrapper(t *testing.T) {
...
+		lisWrapper.mu.Lock()
+		if lenConns = len(lisWrapper.conns); lenConns == 0 {
+			lisWrapper.mu.Unlock()
+			break
+		}
+		lisWrapper.mu.Unlock()
+	}
+	if ctx.Err() != nil {
+		t.Fatalf("timeout waiting for lis wrapper conns to clear, size: %v", lenConns)
+	}
```
dfawley, line 261: Maybe a simplification:

```suggestion
	lenConns := 1
	for ; ctx.Err() == nil && lenConns > 0; <-time.After(time.Millisecond) {
		lisWrapper.mu.Lock()
		lenConns = len(lisWrapper.conns)
		lisWrapper.mu.Unlock()
	}
	if lenConns > 0 {
		t.Fatalf("timeout waiting for lis wrapper conns to clear, size: %v", lenConns)
	}
```

xds/internal/server/listener_wrapper_test.go
zasweq: No, but I forget the testing requirement for the goroutines not leaking. The signal of Stop causes Serve to eventually exit, but then it's not guaranteed it'll exit before the goroutine returns. Although there really isn't a distinction because you can execute wg.Done() yield testing goroutine returns and goroutine hasn't finished, but I think you mentioned to me a while back that this is ok with respect to leak checker.

xds/internal/server/listener_wrapper.go
zasweq: Ah good point I forgot conn wrapper was in same package.

xds/internal/server/listener_wrapper.go
```diff
@@ -304,15 +305,15 @@ func (l *listenerWrapper) Accept() (net.Conn, error) {
 			return nil, fmt.Errorf("received connection with non-TCP address (local: %T, remote %T)", conn.LocalAddr(), conn.RemoteAddr())
 		}
 
-		l.mu.RLock()
```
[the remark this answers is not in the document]
zasweq: Ah ok will switch.

xds/internal/server/listener_wrapper.go
```diff
@@ -202,17 +203,17 @@ func (l *listenerWrapper) maybeUpdateFilterChains() {
 	// gracefully shut down with a grace period of 10 minutes for long-lived
 	// RPC's, such that clients will reconnect and have the updated
 	// configuration apply." - A36
-	var connsToClose []*connWrapper
+	var connsToClose map[*connWrapper]bool
 	if l.activeFilterChainManager != nil { // If there is a filter chain manager to clean up.
```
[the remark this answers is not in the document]
zasweq: I switched this to do it unconditionally. Let me know if you think it looks cleaner, if not I can switch back.

xds/internal/server/listener_wrapper_test.go
zasweq: Deleted this wait group.

xds/internal/server/listener_wrapper_test.go
zasweq: Deleted.

xds/internal/server/listener_wrapper_test.go
zasweq: Switched. Verified it works the same.

xds/internal/server/listener_wrapper_test.go
dfawley: Actually, all the functionality you're doing is already implemented in the channel itself for its own reasons:

`Serve` does the `wg.Done` when it exits:
https://github.com/grpc/grpc-go/blob/6f50403edb0c2db1bb557168cf4e6f87ea2efdb5/server.go#L851

And `Stop` does the `wg.Wait` (way) before it returns:
https://github.com/grpc/grpc-go/blob/6f50403edb0c2db1bb557168cf4e6f87ea2efdb5/server.go#L1911

(`Stop` needs `Serve` to end in order to be sure it can close _all_ the connections the channel created.)

dfawley (approved):

xds/internal/server/listener_wrapper_test.go
zasweq: Ah, ok, that's interesting. Noted for next time. I already deleted the wait group and the wait, so I'll go ahead and merge this.

PR #7665, opened 2024-09-24
Transport: Prevent DNS lookup if host not provided

antonyj7:
Currently, if the host is not specified, it defaults to localhost. This can lead to issues if the DNS server does not resolve localhost correctly, potentially resulting in errors such as `{"code":14,"message":"dns: A record lookup error: lookup localhost: i/o timeout"}`.

I believe that if the host is not provided, the system should skip DNS resolution entirely and directly use the host, which would help prevent these unintended behaviors.

RELEASE NOTES: 

-  Skip DNS resolution when host not specified

### DIFF
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -372,7 +372,7 @@ func formatIP(addr string) (addrIP string, ok bool) {
 // target: "www.google.com" defaultPort: "443" returns host: "www.google.com", port: "443"
 // target: "ipv4-host:80" defaultPort: "443" returns host: "ipv4-host", port: "80"
 // target: "[ipv6-host]" defaultPort: "443" returns host: "ipv6-host", port: "443"
-// target: ":80" defaultPort: "443" returns host: "localhost", port: "80"
+// target: ":80" defaultPort: "443" returns host: "127.0.0.1", port: "80"
 func parseTarget(target, defaultPort string) (host, port string, err error) {
 	if target == "" {
 		return "", "", internal.ErrMissingAddr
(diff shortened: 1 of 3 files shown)


purnesh42H:
@antonyj7 thanks for the change. Though i think hardcoding to only 127.0.0.1 could cause problems in IPv6-preferred or IPv6-only environments. To ensure broader compatibility and future-proof the code, it would be better to handle both IPv4 (127.0.0.1) and IPv6 (::1) loopback addresses when skipping DNS resolution and that would require some more custom logic. Using localhost and relying on the system's DNS resolution to handle the appropriate mapping between IPv4 (127.0.0.1) and IPv6 (::1) might be better in most cases.

Are you facing any specific issues with using localhost? If yes, could you provide more details or logs for that?

antonyj7:
> localhost

Thanks for the review. We have been having issues with some of our VPN servers whereby the DNS does not resolve localhost and spits out `{"code":14,"message":"dns: A record lookup error: lookup localhost: i/o timeout"}` error.

However, the broader point here was - do we even have to go down the route of resolving DNS in case of localhost ?
 
`grpc.NewClient(endpoint, opts...)` fails when we use `localhost:$port` or `:$port` (ignoring host) , we have this fixed by passing `127.0.0.1:$port` as endpoint
 
Its not a big deal for us, and we can live with this :-)   but if this can be fixed in the library, then we can remove 127.0.0.1 in our code :-)

antonyj7:
closing PR as it does not address both IPv4 and IPv6, will raise a new PR if required with support for both models...

antonyj7:
closing this, will raise another PR in case we have a generic solution to cater for IPv6 etc..

PR #7670, opened 2024-09-24
github/workflows: Fix the tag used to enable buffer leak checks

arjan-bal:
The build tag used to enable buffer leaks is defined here:
https://github.com/grpc/grpc-go/blob/e7a8097342e07dbfa093589eed67a8e0d6558423/internal/leakcheck/leakcheck_enabled.go#L1

The Github workflow seems to be using the wrong tag name.

RELEASE NOTES: N/A

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -57,7 +57,7 @@ jobs:
 
           - type: tests
             goversion: '1.22'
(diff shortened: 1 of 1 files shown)


arjan-bal:
Some fixes need to be made in the grpc code before this check can be enabled. Closing this PR in the meantime.

Issue #7429, opened 2024-07-21
Why does connection to `localhost`  fail on local ?


manuraj17:
I faced this issue when I was testing out a simple application; Not able to put a finger on what exactly am missing.

My client code is returning with this error
```sh
could not greet: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

I have my server running as 
```go
const (
	port = ":50051"
)

// server is used to implement helloworld.GreeterServer.
type server struct {
	helloworld.UnimplementedGreeterServer
}

// SayHello implements helloworld.GreeterServer
func (s *server) SayHello(ctx context.Context, in *helloworld.HelloRequest) (*helloworld.HelloReply, error) {
	log.Printf("Received: %v\n", in.GetName())
	return &helloworld.HelloReply{Message: "Hello " + in.Name}, nil
}

func main() {
	lis, err := net.Listen("tcp", port)
	if err != nil {
		log.Fatalf("failed to listen: %v", err)
	}
	s := grpc.NewServer()
	helloworld.RegisterGreeterServer(s, &server{})
	log.Printf("server starting at %v\n", lis.Addr())
	if err := s.Serve(lis); err != nil {
		log.Fatalf("failed to serve: %v", err)
	}
}
```

Client connecting to `localhost` as 
```go
const (
	address = "localhost:50051"
	// address     = "[::1]:50051"

	// address     = "127.0.0.1:50051"
	defaultName = "world"
)

func main() {
	// Set up a connection to the server.
	conn, err := grpc.NewClient(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
	if err != nil {
		log.Fatalf("did not connect: %v", err)
	}
	defer conn.Close()
	c := helloworld.NewGreeterClient(conn)

	// Contact the server and print out its response.
	name := defaultName
	if len(os.Args) > 1 {
		name = os.Args[1]
	}
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()
	r, err := c.SayHello(ctx, &helloworld.HelloRequest{Name: name})
	if err != nil {
		log.Fatalf("could not greet: %v", err)
	}
	log.Printf("Greeting: %s", r.GetMessage())
}
```

This code fails.

But when I connect with either of 
```
	address     = "[::1]:50051"
	address     = "127.0.0.1:50051"
```

It works. I am trying to understand what am missing here.  Can anyone help me understand? TIA!

[...]

PR #7672, opened 2024-09-26
handle dns resolver error for localhost

antonyj7:


purnesh42H:
@antonyj7 we usually don't do draft PRs on main repo. If you want us to validate your approach. I am assuming its a follow up PR for the issue of facing delay in dns lookup for localhost. You can provide a proposal or link a PR from your fork in the issue https://github.com/grpc/grpc-go/issues/7429

PR #7674, opened 2024-09-26
balancer/endpointsharding: Ignore empty endpoints

zasweq:
This PR ignores empty endpoints in endpoint sharding, and adds a validation helper in the resolver package that all petiole policies should call.

RELEASE NOTES:
- resolver: add ValidateEndpoints
- balancer/endpointsharding: ignore endpoints with zero addresses.


balancer/endpointsharding/endpointsharding.go
```diff
@@ -80,18 +80,27 @@ type endpointSharding struct {
...
+	addrs := resolver.NewAddressMap()
+	// Check/return early if any endpoints have no addresses, or there is a duplicate address map.
 	// TODO: make this configurable if needed.
 	for i, endpoint := range state.ResolverState.Endpoints {
 		if len(endpoint.Addresses) == 0 {
 			return fmt.Errorf("endpoint %d has empty addresses", i)
 		}
+		for _, addr := range endpoint.Addresses {
+			if _, ok := addrs.Get(addr); ok {
+				return fmt.Errorf("duplicate addr %v present in endpoints list", addr)
```
arjan-bal, line 100: Should this wrap `balancer.ErrBadResolverState` so that the re-resolution is triggered?

arjan-bal (commented):
Can you add a unit test for this behaviour to catch future regressions?

balancer/endpointsharding/endpointsharding.go
dfawley: I don't think so.  This is being fed back to the name resolver anyway.

I don't think this magic error is actually used anymore, and I don't see anything in the channel checking for it.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -80,18 +80,27 @@ type endpointSharding struct {
...
+// error without forwarding any updates. Otherwise returns first error found
+// from a child, but fully processes the new update.
 func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState) error {
 	if len(state.ResolverState.Endpoints) == 0 {
 		return errors.New("endpoints list is empty")
 	}
-	// Check/return early if any endpoints have no addresses.
+
+	addrs := resolver.NewAddressMap()
+	// Check/return early if any endpoints have no addresses, or there is a duplicate address map.
```
dfawley, line 92: s/map//.

dfawley (commented):
+1 to the request for a test case.

balancer/endpointsharding/endpointsharding.go
arjan-bal: Okay.

balancer/endpointsharding/endpointsharding.go
zasweq: Done.

zasweq:
Added a unit test, thanks.

balancer/endpointsharding/endpointsharding_test.go
```diff
@@ -157,3 +157,26 @@ func (s) TestEndpointShardingBasic(t *testing.T) {
...
 }
+
+// TestEndpointShardingBalancerAPI tests that when endpoint sharding gets a
+// Client Conn update with a duplicate address across endpoints the
+// UpdateClientConnState operation errors.
+func (s) TestEndpointShardingBalancerAPI(t *testing.T) {
+	es := NewBalancer(nil, balancer.BuildOptions{})
+	addr := resolver.Address{Addr: "addr1"}
+	err := es.UpdateClientConnState(balancer.ClientConnState{
+		ResolverState: resolver.State{
```
dfawley, line 168: This is pretty limited in what it covers.  E.g. it doesn't cover addresses across endpoints, or endpoints with multiple addresses.  Please add some more test cases to cover more possibilities -- probably make it table driven where `[]Endpoint` is the input and the result is probably just a `bool` to indicate whether there should be an error or not.

zasweq: Added more coverage for passing and failing tests.

balancer/endpointsharding/endpointsharding_test.go
```diff
@@ -158,25 +159,62 @@ func (s) TestEndpointShardingBasic(t *testing.T) {
...
+			wantErr: false,
+		},
+		{
+			name: "endpoint-with-no-addresses",
+			endpoints: []resolver.Endpoint{
+				{Addresses: []resolver.Address{addr1, addr2}},
+				{Addresses: []resolver.Address{}},
+			},
+			wantErr: true,
 		},
```
dfawley, line 206: Please add a couple more:

- no error with addr1+2 and addr3+4 in separate endpoints -- you should generally always have a "pass" test.
- error where endpoint1 has the same addr multiple times

zasweq: Discussed offline; had the first case. Added the second case.

zasweq:
Updated endpointsharding as per offline discussion:
* Validate errors if empty endpoints or no addresses present in endpoints
* UpdateClientConnState simply ignores endpoints with no addresses

balancer/endpointsharding/endpointsharding.go
```diff
@@ -76,24 +75,28 @@ type endpointSharding struct {
...
-// addresses, returns error without forwarding any updates. Otherwise returns
-// first error found from a child, but fully processes the new update.
-func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState) error {
-	if len(state.ResolverState.Endpoints) == 0 {
+// ValidateEndpoints returns an error if the endpoints list is empty, or no
+// addresses are present in endpoint list.
+func ValidateEndpoints(endpoints []resolver.Endpoint) error {
+	if len(endpoints) == 0 {
 		return errors.New("endpoints list is empty")
 	}
```
dfawley, line 83: I think you can just delete this?  It's a subset of the other validation?

balancer/endpointsharding/endpointsharding.go
```diff
@@ -76,24 +75,28 @@ type endpointSharding struct {
...
-// UpdateClientConnState creates a child for new endpoints and deletes children
-// for endpoints that are no longer present. It also updates all the children,
-// and sends a single synchronous update of the childrens' aggregated state at
-// the end of the UpdateClientConnState operation. If any endpoint has no
-// addresses, returns error without forwarding any updates. Otherwise returns
-// first error found from a child, but fully processes the new update.
-func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState) error {
-	if len(state.ResolverState.Endpoints) == 0 {
+// ValidateEndpoints returns an error if the endpoints list is empty, or no
+// addresses are present in endpoint list.
```
dfawley, line 79: This comment should indicate how/why/when to use this, not what it does. 

```go
// The EndpointSharding balancer assumes ValidateEndpoints has been used to validate all Endpoints
// passed to UpdateCLientConnState.
```

?

Or is this scenario actually not a problem for the endpointsharding balancer itself?  In which case maybe this function belongs in package `resolver` or `balancer`?

balancer/endpointsharding/endpointsharding.go
```diff
@@ -76,24 +75,28 @@ type endpointSharding struct {
...
 	}
+	return errors.New("endpoints list contains no addresses")
+}
 
+// UpdateClientConnState creates a child for new endpoints and deletes children
+// for endpoints that are no longer present. It also updates all the children,
+// and sends a single synchronous update of the childrens' aggregated state at
+// the end of the UpdateClientConnState operation. If any endpoint has no
+// addresses it will ignore that endpoint. Otherwise, returns first error found
+// from a child, but fully processes the new update.
```
dfawley, line 98: Nobody will see this godoc comment here since it's on an unexported type.

balancer/endpointsharding/endpointsharding.go
zasweq: I kept this here soley for the sake of the more specific error string. I think the two are distinct, which is why I left this clause for a different error message.

balancer/endpointsharding/endpointsharding.go
zasweq: Yeah I can move it to resolver or balancer, as this should trigger failure case in all petiole policies.

balancer/endpointsharding/endpointsharding.go
zasweq: Are you ok with that?

balancer/endpointsharding/endpointsharding.go
zasweq: But also since it's gatekeeped in all *petiole* policies maybe it does make sense to keep it here? Although looks like we won't use this in petiole like ring hash so I'll move it to resolver.

zasweq: Updated comment as well, let me know what you think.

resolver/resolver.go
```diff
@@ -330,3 +331,19 @@ type AuthorityOverrider interface {
 	// typically in line, and must keep it unchanged.
 	OverrideAuthority(Target) string
 }
+
+// ValidateEndpoints validates endpoints are valid from a petiole policies
```
dfawley, line 335: "Valid" now appears 3 times.  And spelling mistake:
`ValidateEndpoints validates endpoints from a petiole _policy's_ perspective.`

But now you should also explain what that actually means?

How about:

> ValidateEndpoints validates that the endpoints are valid to use in a petiole policy - i.e. they must be non-empty and each must contain at least one address.

And what's a petiole policy? Is this defined somewhere?  Maybe add: "See [A61] for details"?

balancer/endpointsharding/endpointsharding.go
dfawley: I'm fine with comments that aren't visible to anyone but someone reading the code.  But make sure they aren't supposed to be showing up in godoc.  Godoc should have enough info that the package is usable by someone who doesn't know what it does before viewing the godocs, and unexported comments should focus on how the implementation works and why it's the way it is.

resolver/resolver.go
zasweq: Hmmm ok. At first I thought I needed to add Outlier Detection to this, but that isn't a petiole policy (can't use it directly on top of pick first, that is invalid), and this is meant to trigger reresolution and reject the update. This happens from the petiole policy layer so the only thing important in OD is to ignore endpoints with no addresses and to have the undefined behavior in shared addresses, and leave petiole policies to trigger this error.

zasweq: Switched to

```
// ValidateEndpoints validates endpoints from a petiole policy's perspective.
// Petiole policies should call this before calling into their children. See
//[gRPC A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md)
// for details.
```

balancer/endpointsharding/endpointsharding.go
zasweq: I reread this comment and I think the language here appropriate for this not public facing operation (all relate to the UpdateClientConnState() operation), and the exported comments seem good enough to me. Do you want me to take some of this information and put it in exported go doc?

PR #7719, opened 2024-10-09
[google-c2p resolver] Introduce SetUniverseDomain API

apolcyn:
As title

### DIFF
--- a/xds/googledirectpath/googlec2p.go
+++ b/xds/googledirectpath/googlec2p.go
@@ -30,6 +30,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net/url"
+	"sync"
 	"time"
 
 	"google.golang.org/grpc/grpclog"
@@ -46,7 +47,7 @@ const (
 	c2pScheme    = "google-c2p"
 	c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
 
-	tdURL                   = "dns:///directpath-pa.googleapis.com"
+	defaultUniverseDomain   = "googleapis.com"
 	zoneURL                 = "http://metadata.google.internal/computeMetadata/v1/instance/zone"
 	ipv6URL                 = "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s"
 	ipv6CapableMetadataName = "TRAFFICDIRECTOR_DIRECTPATH_C2P_IPV6_CAPABLE"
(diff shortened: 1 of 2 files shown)


xds/googledirectpath/googlec2p.go
```diff
@@ -58,15 +59,68 @@ const (
...
 // For overriding in unittests.
 var (
-	onGCE   = googlecloud.OnGCE
-	randInt = rand.Int
-	logger  = internalgrpclog.NewPrefixLogger(grpclog.Component("directpath"), logPrefix)
+	onGCE            = googlecloud.OnGCE
+	randInt          = rand.Int
+	logger           = internalgrpclog.NewPrefixLogger(grpclog.Component("directpath"), logPrefix)
+	universeDomainMu sync.Mutex
+	universeDomain   = ""
```
dfawley, line 66: Can you move the comment above to `onGCE` / `randInt` instead of the whole block, please?  It no longer applies to everything.

xds/googledirectpath/googlec2p.go
```diff
@@ -58,15 +59,68 @@ const (
...
 )
 
 func init() {
 	resolver.Register(c2pResolverBuilder{})
 }
 
+// SetUniverseDomain informs the gRPC library of the TPC universe domain
+// in which the process is running. It is the caller's responsibility to
```
dfawley, line 74: Is there any documentation we can point to about this?

xds/googledirectpath/googlec2p.go
```diff
@@ -58,15 +59,68 @@ const (
...
+// configure its dependencies.
+//
+// If a gRPC channel is created with the "google-c2p" URI scheme and this
+// function has not been called, gRPC configures the universe domain as
+// "googleapis.com".
+//
+// Arguments:
+//
+//	d: The DNS domain for Google APIs in the current TPC universe
+//	   (for example, "googleapis.com" or "apis-s3ns.fr").
```
dfawley, line 86: Is this a new Go style thing?  I've not seen it before.  If not, please follow the normal style of just mentioning it inline.  Or I think it can just be removed and `d` renamed to `domain`.

xds/googledirectpath/googlec2p.go
```diff
@@ -58,15 +59,68 @@ const (
...
+//
+// Otherwise, returns an error.
+func SetUniverseDomain(d string) error {
+	universeDomainMu.Lock()
+	defer universeDomainMu.Unlock()
+	if d == "" {
+		return fmt.Errorf("universe domain cannot be empty")
+	}
+	if universeDomain == "" {
+		universeDomain = d
```
dfawley, line 101: Do we need to do any paranoid validation in case `SetUniverseDomain` is called using unsanitized data?

I'm not sure that anything could actually go wrong if we don't (aside from it just not working), though.

dfawley (commented):
LGTM, mostly just some nits about comments.

xds/googledirectpath/googlec2p.go
apolcyn: done

xds/googledirectpath/googlec2p.go
apolcyn: Asking around, I'll get back to this comment after I hear back

xds/googledirectpath/googlec2p.go
apolcyn: Sorry this is probably not golang style (this is something my editor suggested). Moved to a more inline comment.

xds/googledirectpath/googlec2p.go
apolcyn: We perhaps could, but I'm not sure how aggressive we want to be in this check.

Because this is a DNS domain that we're configuring, standard DNS rules should apply:

1) The entire Traffic Director DNS name cannot exceed 255 characters (https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.4). So we could perhaps do a simple check here that this domain does not exceed 255 characters. We could perhaps also do a more complicated check that each label does not exceed 63 characters.

2) Each character within a label should be a letter, number, or hyphen (https://datatracker.ietf.org/doc/html/rfc1034#section-3.5)

Asking around to see what other clients are doing here. I'll get back to this.

xds/googledirectpath/googlec2p.go
apolcyn: No documentation currently

xds/googledirectpath/googlec2p.go
apolcyn: Circling back on this.

In principle it seems very reasonable to add some input validation. And we can *probably* do it according to those standard DNS rules I mentioned.

However, client libraries are apparently not doing any such input validation here, and I've yet to hear of any actual specific validation that's been agreed upon.

So as for now, I think it's safest to not do any validation beyond the empty string check.

As you mentioned, things are expected to not work if the input is bad. Most likely DNS resolution of the Traffic Director endpoint will fail.

apolcyn:
@dfawley PTAL

dfawley (approved):

PR #7723, opened 2024-10-10
Add ALTS support to interop stress client

sagarsudo:
Tested using:
```
./client --server_addresses="google-c2p:///directpath-grpclb-with-pick-first-test-jz.googleapis.com" --test_cases="empty_unary:100" --use_alts=true --test_duration_secs=10
Total calls made: 54346384
```

### DIFF
--- a/interop/stress/client/main.go
+++ b/interop/stress/client/main.go
@@ -36,6 +36,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/google"
+	"google.golang.org/grpc/credentials/alts"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/interop"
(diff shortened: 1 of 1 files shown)


eshitachandwani:
Hi @sagarsudo, it seems you may want to include this flag in your application for testing purposes. Could you please clarify the reason for requesting this change in the gRPC-Go OSS repository?

sagarsudo:
Just noticed a discrepancy in between interop/client and interop/stress/client, where the former has ALTS support but the latter does not.

sagarsudo:
It would be good to have ALTS support in the stress client as it would make it easier to stress test against GCP services.

eshitachandwani:
Hey @sagarsudo ,
These test frameworks are for internal testing and after reviewing the test frameworks and checking other languages, it appears that ALTS support in the stress client isn't a common feature.  Therefore, we've decided not to prioritize adding it at this time.

If you believe ALTS support is critical or have a specific use case in mind, please don't hesitate to reopen this issue and provide more details.

PR #7725, opened 2024-10-10
pickfirst: Implement Happy Eyeballs

arjan-bal:
As part of the [Dualstack design](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#happy-eyeballs-in-the-pick_first-lb-policy), the pickfirst policy should implement the happy eyeballs algorithm while connecting to multiple backends.

The timeout for the happy eyeballs connection timer is NOT configurable as that's an optional requirement in the gRFC.

RELEASE NOTES:
* The new experimental `pickfirst` LB policy (disabled by default) supports Happy Eyeballs to attempt connections to multiple backends concurrently. The experimental `pickfirst` policy can be enabled by setting the environment variable `GRPC_EXPERIMENTAL_ENABLE_NEW_PICK_FIRST` to `true`.


easwars:
Should we mention the environment variables in the release note? Or at least in the PR description?

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -622,3 +682,57 @@ func equalAddressIgnoringBalAttributes(a, b *resolver.Address) bool {
...
+
+// callbackScheduleris used to schedule the execution of a callback after a
+// a specified delay. It is not safe for concurrent access.
+type callbackScheduler struct {
+	cancelScheduled func()
+	closed          bool
+	wg              sync.WaitGroup
+}
+
+// schedule schedules the execution of a callback. It cancels any previously
```
easwars, line 694: > It cancels any previously scheduled callbacks

We can only have one pending scheduled callback, right?

What is the real purpose/advantage of having this type over simply using `time.AfterFunc` and storing the returned timer in the LB policy and calling `Stop` on it when you want to cancel pending callback?

easwars (commented):
I couldn't complete a full pass, but some comment here to get satrted.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -622,3 +682,57 @@ func equalAddressIgnoringBalAttributes(a, b *resolver.Address) bool {
 		a.Attributes.Equal(b.Attributes) &&
 		a.Metadata == b.Metadata
 }
+
+// callbackScheduleris used to schedule the execution of a callback after a
```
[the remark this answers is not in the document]
arjan-bal: Removed the `callbackScheduler` type in favour of `time.TimeAfterFunc`.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -396,18 +419,44 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
...
+			b.requestConnectionLocked()
+		}
+	}, connectionDelayInterval)
 }
 
 func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.SubConnState) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	oldState := sd.state
+	// Record a connection attempt when existing CONNECTING.
```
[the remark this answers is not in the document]
arjan-bal: Typo fixed.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -384,8 +405,10 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
 		switch scd.state {
 		case connectivity.Idle:
 			scd.subConn.Connect()
+			b.scheduleNextConnectionLocked()
 		case connectivity.TransientFailure:
 			// Try the next address.
+			scd.connectionAttempted = true
```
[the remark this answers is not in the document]
arjan-bal: The connection attempted flag is reset when the first pass is started. A re-used subchannel may not have completed backoff when the attempt is made. This case is mentioned in [A62](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#happy-eyeballs-in-the-pick_first-lb-policy):

> Note that every time the LB policy receives a new address list, it will start an initial Happy Eyeballs pass over the new list, even if some of the subchannels are not actually new due to their addresses having been present on both the old and new lists. This means that on the initial pass through the address list for a subsequent address list update, when pick_first decides to start a connection attempt on a given subchannel (whether because it is the first subchannel in the list or because the timer fired before the previous address' connection attempt completed), that subchannel may not be in state IDLE, which is the only state in which a connection attempt may be requested.

and later:
> * If the subchannel is in state TRANSIENT_FAILURE, then we know that it is in backoff due to a recent connection attempt failure, so we treat it as if we have already made a connection attempt on this subchannel, and we will immediately move on to the next subchannel.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -384,8 +405,10 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
 		switch scd.state {
```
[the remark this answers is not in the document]
arjan-bal: Changed to call return in each case.

> Or maybe we can get rid of the for loop in this method and instead make a recursive call when the state is TF.

It was recursive initially. I changed it to an iterative version when there was a PR comment about the base case for the recursion. With iteration, it's easier to see that the function ends after checking all the subchannels once.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -288,9 +299,18 @@ func (b *pickfirstBalancer) ExitIdle() {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	if b.state == connectivity.Idle && b.addressList.currentAddress() == b.addressList.first() {
```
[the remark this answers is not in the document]
arjan-bal: I initially added it because I was (incorrectly) thinking that ExitIdle will start the iteration over the address list from the start. This would cause the first subchannel to be connected which isn't in TF. If it was called multiple times, subchannels that were in TF earlier could complete the backoff and exit TF.

However, that's not the case. The address list is reset while entering IDLE, not exiting IDLE. So repeated calls to `requestConnectionLocked()` will be a no-op.

Thanks for spotting this, I've removed the check and the `addressList.first()` method  I added to support this.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -104,8 +115,9 @@ type scData struct {
 	subConn balancer.SubConn
 	addr    resolver.Address
 
-	state   connectivity.State
-	lastErr error
+	state               connectivity.State
+	lastErr             error
+	connectionAttempted bool
```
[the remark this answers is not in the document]
arjan-bal: Renamed to `connectionFailed`.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -56,20 +58,29 @@ var (
 	// It is changed to "pick_first" in init() if this balancer is to be
 	// registered as the default pickfirst.
 	Name = "pick_first_leaf"
+	// timerFunc allows mocking the timer for testing connection delay related
+	// functionality.
+	timerFunc = time.After
```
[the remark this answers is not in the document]
arjan-bal: Renamed to `timeAfterFunc`.

arjan-bal:
> Should we mention the environment variables in the release note? Or at least in the PR description?

Updated the release notes.

.github/workflows/testing.yml
```diff
@@ -73,7 +73,9 @@ jobs:
           - type: tests
             goversion: '1.22'
             testflags: -race
-            grpcenv: 'GRPC_EXPERIMENTAL_ENABLE_NEW_PICK_FIRST=true'
+            grpcenv: |
+              'GRPC_EXPERIMENTAL_ENABLE_NEW_PICK_FIRST=true'
+              'GRPC_EXPERIMENTAL_ENABLE_PICK_FIRST_HAPPY_EYEBALLS=true'
```
zasweq, line 78: Optional: I feel like gating the new pick first policy with an environment variable is good enough. I don't feel the need for a second environment variable for this functionality, especially since both will eventually go default true. Or is the goal for a user to be able to just disable this algorithm even if using new pick first? I don't see many use cases for that, especially with the normal usage of endpoints. If that's the case, maybe a knob too (which I know the configs are hard defined in gRFC).

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
easwars: Thanks. Makes sense.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -384,30 +405,69 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	closeFn := timerAfterFunc(connectionDelayInterval, func() {
+		b.mu.Lock()
+		defer b.mu.Unlock()
+		// If the scheduled task is cancelled while acquiring the mutex, return.
+		if ctx.Err() != nil {
+			return
+		}
+		if b.logger.V(2) {
+			b.logger.Infof("Happy Eyeballs timer expired.")
```
easwars, line 451: Would it be possible to include some other data here, either the pointer to the subchannel or the address of the subchannel, so that the log statement is more meaningful.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -384,30 +405,69 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
...
+		}
+		if b.addressList.increment() {
+			b.requestConnectionLocked()
+		}
+	}).Stop
+	b.cancelScheduled = sync.OnceFunc(func() {
+		closeFn()
+		cancel()
```
easwars, line 459: Shouldn't the context be canceled first before stopping the timer?

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -56,20 +58,29 @@ var (
...
 )
 
-// TODO: change to pick-first when this becomes the default pick_first policy.
-const logPrefix = "[pick-first-leaf-lb %p] "
+const (
+	// TODO: change to pick-first when this becomes the default pick_first policy.
+	logPrefix = "[pick-first-leaf-lb %p] "
+	// connectionDelayInterval is the time to wait for during the happy eyeballs
+	// pass before starting the next connection attempt.
+	connectionDelayInterval = 250 * time.Millisecond
```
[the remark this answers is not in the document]
arjan-bal: I'm not sure. When I saw that Java didn't make the timeout configurable, I didn't spend too much time finding a solution. 
One idea I had was to use a dial option and pass the value from the channel to LB policies as a resolver attribute in `balancer.ClientConnState`.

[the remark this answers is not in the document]
easwars: Fair enough.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -56,20 +58,29 @@ var (
 	// It is changed to "pick_first" in init() if this balancer is to be
 	// registered as the default pickfirst.
 	Name = "pick_first_leaf"
+	// timerAfterFunc allows mocking the timer for testing connection delay
+	// related functionality.
+	timerAfterFunc = time.AfterFunc
```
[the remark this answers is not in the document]
easwars: Thanks.

easwars (commented):
Quoting this paragraph from A61:
---
Note that because we do not report TRANSIENT_FAILURE until after the Happy Eyeballs pass has completed and we start a new Happy Eyeballs pass whenever we receive a new address list, there is a potential failure mode where we may never report TRANSIENT_FAILURE if we are receiving new address lists faster than we are completing Happy Eyeballs passes. This is a pre-existing problem, and each gRPC implementation currently deals with it in its own way. This design does not propose any changes to those existing approaches, although a future gRFC may attempt to achieve further convergence here.
---

I guess we never had any logic/code to handle this prior to Happy Eyeballs. Do have an idea of how to handle this?

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +854,268 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	}
+	timerCh <- struct{}{}
+	if err := servers[2].awaitContacted(ctx); err != nil {
+		t.Fatalf("Server with address %q not contacted: %v", addrs[2], err)
+	}
+
+	// First SubConn Fails.
+	servers[0].closeConn()
+
+	// No TF should be reported until the first pass is complete.
```
easwars, line 939: Reporting TF is not strictly tied to completion of the first pass. It is tied to all connection attempts failing, right?

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -955,3 +1220,94 @@ type ccStateSubscriber struct {
 func (c *ccStateSubscriber) OnMessage(msg any) {
 	c.transitions = append(c.transitions, msg.(connectivity.State))
 }
+
+// hangingServer is a server that accept a TCP connection and remains idle until
+// asked to close or respond to the connection. They can be used to control
+// how long it takes for a subchannel to report a TRANSIENT_FAILURE in tests.
+type hangingServer struct {
```
easwars, line 1227: We have a bunch of utilities in the `testutils` package that can be used I feel, instead of implementing this new hangingServer.

There is a blocking dialer that lets you block, fail or succeed connection attempts: https://github.com/grpc/grpc-go/blob/ec10e73f02a38fcf1e4f6e3cdee5f73c89bfa4ac/internal/testutils/blocking_context_dialer.go#L33
There is a restartable listener: https://github.com/grpc/grpc-go/blob/master/internal/testutils/restartable_listener.go#L37
There is a wrapped listener that returns a wrapped conn: https://github.com/grpc/grpc-go/blob/master/internal/testutils/wrappers.go

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +854,268 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+
+	// No TF should be reported until the first pass is complete.
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	testutils.AwaitNotState(shortCtx, t, cc, connectivity.TransientFailure)
+
+	// Move off the end of the list, pickfirst should still be waiting for TFs
+	// to be reported.
+	timerCh <- struct{}{}
```
easwars, line 947: This is slightly off the gRFC. Not a major concern in terms of correctness. The gRFC does say that if we are starting a connection to the last address on the list, we shouldn't be starting the timer. But our code does start the timer.

easwars (commented):
I've made comments only on one of the tests, but some of them apply to the other two tests as well. And if you end up being able to use one or more of the existing utilities, then the tests are anyways going to change a bit. Thanks.

balancer/pickfirst/internal/internal.go
```diff
@@ -18,7 +18,15 @@
 // Package internal contains code internal to the pickfirst package.
 package internal
 
-import "math/rand"
+import (
+	"math/rand"
```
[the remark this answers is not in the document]
arjan-bal: Acknowledged.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
arjan-bal: `BlockingDialer` fits the requirement perfectly! I've replaced the `hangingServer` that I implemented with `BlockingDialer`.

arjan-bal:
> Quoting this paragraph from A61: " Note that because we do not report TRANSIENT_FAILURE until after the Happy Eyeballs pass has completed and we start a new Happy Eyeballs pass whenever we receive a new address list, there is a potential failure mode where we may never report TRANSIENT_FAILURE if we are receiving new address lists faster than we are completing Happy Eyeballs passes. This is a pre-existing problem, and each gRPC implementation currently deals with it in its own way. This design does not propose any changes to those existing approaches, although a future gRFC may attempt to achieve further convergence here. "
> 
> I guess we never had any logic/code to handle this prior to Happy Eyeballs. Do have an idea of how to handle this?

This failure mode exists in the existing pickfirst which calls `addrConn.updateAddrs`, to handle this scenario. I'm not sure if `grpc-go` does anything to prevent this. The DNS resolver has time based throttling, but that's not a universal fix. Java doesn't have anything to handle this in the `pickfirst` LB policy, though the channel may have a way of preventing this.

Let me start a discussion in the cross language group.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +854,268 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+		return ret
+	}
+
+	defer func() {
+		pfinternal.TimeAfterFunc = originalTimer
+	}()
+
+	defer func() {
+		pfinternal.TimeAfterFunc = originalTimer
+	}()
```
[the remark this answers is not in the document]
easwars: I'm still seeing two defers. Maybe you didn't push a commit?

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+
+	dialer := testutils.NewBlockingDialer()
+	cc, rb, bm := setupPickFirstLeaf(t, 3,
+		grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingConfig": [{"%s":{}}]}`, pickfirstleaf.Name)),
+		grpc.WithContextDialer(dialer.DialContext),
+	)
```
easwars, line 889: If possible, it would be nicer to not have to break this function call into multiple lines. See here for some ideas: go/go-style/decisions#func-formatting

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	// Verify that only the first server is contacted.
+	if got, want := holds[0].Wait(ctx), true; got != want {
+		t.Fatalf("hold[%q].Wait() = %t, want %t", addrs[0], got, want)
+	}
+
+	// Ensure no other servers are contacted.
+	if got, want := holds[1].IsStarted(), false; got != want {
+		t.Fatalf("holds[%q].IsStarted() = %t, want %t", addrs[1], got, want)
+	}
+
```
easwars, line 906: Nit: remove this newline, since this is essentially part of the same logical code block under the comment `	// Ensure no other servers are contacted.`

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +854,268 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	// Ensure no other servers are contacted.
+	if got, want := servers[1].isContacted(), false; got != want {
+		t.Fatalf("Server[%q].isContacted() = %t, want %t", addrs[1], got, want)
+	}
+	if got, want := servers[2].isContacted(), false; got != want {
+		t.Fatalf("Server[%q].isContacted() = %t, want %t", addrs[2], got, want)
+	}
+
+	// Make the happy eyeballs timer fire twice so that pickfirst reaches the
+	// last address in the list.
```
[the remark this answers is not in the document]
easwars: I get it now. Maybe logically group them differently then?

```golang
	// Make the happy eyeballs timer fire once and verify that the 
	// second server is contacted, but the third isn't.
	timerCh <- struct{}{}
	if got, want := holds[1].Wait(ctx), true; got != want {
		t.Fatalf("hold[%q].Wait() = %t, want %t", addrs[1], got, want)
	}
	if got, want := holds[2].IsStarted(), false; got != want {
		t.Fatalf("holds[%q].IsStarted() = %t, want %t", addrs[2], got, want)
	}

	// Make the happy eyeballs timer fire once more and verify that the 
	// third server is contacted.
	timerCh <- struct{}{}
	if got, want := holds[2].Wait(ctx), true; got != want {
		t.Fatalf("hold[%q].Wait() = %t, want %t", addrs[2], got, want)
	}
```

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+		t.Fatalf("hold[%q].Wait() = %t, want %t", addrs[1], got, want)
+	}
+
+	if got, want := holds[2].IsStarted(), false; got != want {
+		t.Fatalf("holds[%q].IsStarted() = %t, want %t", addrs[2], got, want)
+	}
+
+	timerCh <- struct{}{}
+	// Verify that the second server is contacted and 3rd isn't.
+	if got, want := holds[1].Wait(ctx), true; got != want {
```
easwars, line 926: This should be `holds[2]` right?

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+		t.Fatalf("holds[%q].IsStarted() = %t, want %t", addrs[2], got, want)
+	}
+
+	// Make the happy eyeballs timer fire twice so that pickfirst reaches the
+	// last address in the list.
+	timerCh <- struct{}{}
+
+	// Verify that the second server is contacted and 3rd isn't.
+	if got, want := holds[1].Wait(ctx), true; got != want {
+		t.Fatalf("hold[%q].Wait() = %t, want %t", addrs[1], got, want)
```
easwars, line 917: I understand the style guide recommendation to identify the function in the test error message, but this is a function on a test utility type and not the function under test. So, maybe something like:
```
	t.Fatalf("Timeout when waiting for second server with address %q to be contacted", addrs[1])
```

Similarly for other error strings where we are calling `hold.Wait` or `hold.IsStarted`.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	}
+
+	// First SubConn Fails.
+	holds[0].Fail(fmt.Errorf("test error"))
+
+	// No TF should be reported until the first pass is complete.
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
```
easwars, line 936: Similar to other comments. Remove this newline, since waiting to ensure that we dont reach TF is part of the same comment above.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+		return timer
+	}
+
+	defer func() {
+		pfinternal.TimeAfterFunc = originalTimer
+	}()
+
+	defer func() {
+		pfinternal.TimeAfterFunc = originalTimer
+	}()
```
easwars, line 978: Again same here. Still see two defers.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,253 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+		}()
+		return timer
+	}
+
+	defer func() {
+		pfinternal.TimeAfterFunc = originalTimer
+	}()
+
+	dialer := testutils.NewBlockingDialer()
+	cc, rb, bm := setupPickFirstLeaf(t, 3,
```
easwars, line 1043: Many of the comments on the first test, applies to the second and third one as well. Please do take care of all three tests. Thanks.

easwars (changes requested):

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
arjan-bal: Changed. Using the server numbers instead of the words "first, second etc." as its easier to write copy/paste and write vim macros to make all the changes in one go.

.github/workflows/testing.yml
arjan-bal: PR description updated.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	}
+	if holds[2].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 2, addrs[2])
+	}
+
+	// Make the happy eyeballs timer fire once and verify that the
+	// second server is contacted, but the third isn't.
+	timerCh <- struct{}{}
+
+	// Verify that the second server is contacted and 3rd isn't.
```
easwars, line 911: Nit: You could remove the newline and this comment since the comment on line 907 conveys the same thing and the two verifications that are below are covered as part of that comment. Thanks.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	// No TF should be reported until the first pass is complete.
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	testutils.AwaitNotState(shortCtx, t, cc, connectivity.TransientFailure)
+
+	// Third SubConn fails.
+	holds[2].Fail(fmt.Errorf("test error"))
+
+	testutils.AwaitNotState(shortCtx, t, cc, connectivity.TransientFailure)
```
easwars, line 938: The short context has to be recreated here since the deadline would have elapsed in the previous call to `testutils.AwaitNotState`.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	}
+
+	timerCh <- struct{}{}
+
+	// Second connection attempt is successful.
+	if holds[1].Wait(ctx) != true {
+		t.Fatalf("Timeout waiting for server %d with address %q to be contacted", 1, addrs[1])
+	}
+	holds[1].Resume()
+	testutils.AwaitState(ctx, t, cc, connectivity.Ready)
```
easwars, line 1000: Nit: Better use of newlines and comments would make this block more readable I feel.

```golang

	// Make the happy eyeballs timer fire once and verify that the
	// second server is contacted.
	timerCh <- struct{}{}
	if holds[1].Wait(ctx) != true {
		t.Fatalf("Timeout waiting for server %d with address %q to be contacted", 1, addrs[1])
	}
	
	// Get the connection attempt to the second server to succeed and verify
	// that the channel becomes READY.
	holds[1].Resume()
	testutils.AwaitState(ctx, t, cc, connectivity.Ready)
```

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+}
+
+// TestPickFirstLeaf_HappyEyeballs_TFThenTimerFires tests the pickfirst balancer
+// by causing a SubConn to fail and then jumping to the 3rd SubConn after the
+// happy eyeballs timer expires.
+func (s) TestPickFirstLeaf_HappyEyeballs_TFThenTimerFires(t *testing.T) {
```
easwars, line 1006: Here and in the first test added in this PR, could we add an `_` after the `TF`? So, this would be `TestPickFirstLeaf_HappyEyeballs_TF_ThenTimerFires`.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
 	}
 }
 
+// TestPickFirstLeaf_HappyEyeballs_TFAfterEndOfList verifies that pickfirst
```
easwars, line 853: Optional: These test names are super long and the comment could simply be `// Test verifies blah`. Here and in the other tests in this PR.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+		t.Fatalf("Timeout waiting for server %d with address %q to be contacted", 0, addrs[0])
+	}
+
+	// Ensure no other servers are contacted.
+	if holds[1].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 1, addrs[1])
+	}
+	if holds[2].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 2, addrs[2])
+	}
```
easwars, line 1058: Nit: similar to this block in the first test, this could be a single comment and single logical block.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+
+	// Verify that only the second server is contacted.
+	if holds[1].Wait(ctx) != true {
+		t.Fatalf("Timeout waiting for server %d with address %q to be contacted", 1, addrs[1])
+	}
+
+	// Ensure no other servers are contacted.
+	if holds[2].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 2, addrs[2])
+	}
```
easwars, line 1077: Similarly, this could be a single block with one comment and no newlines between them.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,245 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	// Ensure no other servers are contacted.
+	if holds[1].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 1, addrs[1])
+	}
+	if holds[2].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 2, addrs[2])
+	}
+
+	// Replace the timer channel so that the old timers don't attempt to read
+	// messages pushed next.
```
easwars, line 1061: Old timers should get canceled when subsequent subchannels are created, right? Why do we need to do this?

easwars (approved):
LGTM. Some minor nits in the tests.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
easwars: I see what you are saying. That seems better to me, unless it is too much work.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
easwars: This one still needs fixing.

arjan-bal: Created a new context.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
arjan-bal: Refactored to have the internal.TimeAfterFunc return a `cancelFunc()` instead of a timer. This allowed the test to stop the timer when `pickfirst` cancels the timer. I also created a helper function to return a timer function and a function to trigger the timer manually instead of having the tests write on channel.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,190 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	holds[0].Fail(fmt.Errorf("test error"))
+
+	// No TF should be reported until the first pass is complete.
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	testutils.AwaitNotState(shortCtx, t, cc, connectivity.TransientFailure)
+
+	// Third SubConn fails.
+	shortCancel()
```
easwars, line 918: Do we need this? Won't `testutils.AwaitNotState` fail the test if the specified state is reached before the context expires?

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -851,6 +850,190 @@ func (s) TestPickFirstLeaf_EmptyAddressList(t *testing.T) {
...
+	// Verify that only the second server is contacted.
+	if holds[1].Wait(ctx) != true {
+		t.Fatalf("Timeout waiting for server %d with address %q to be contacted", 1, addrs[1])
+	}
+	if holds[2].IsStarted() != false {
+		t.Fatalf("Server %d with address %q contacted unexpectedly", 2, addrs[2])
+	}
+
+	// The happy eyeballs timer expires, skipping server[1] and requesting the creation
+	// of a third SubConn.
```
easwars, line 1025: Why do you say we are skipping server[1] here? IIUC correctly:
- we first started a connection to server[0]
- connection to server[0] failed before the HE timer fired
- so, we started a connection to server[1]
- now, the HE timer has fired
- so, we would start a connection to server[2]

I don't see where we are skipping server[1].

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
arjan-bal: It's not required because of the way `testutils.AwaitNotState` works. When I tried to ignore the first cancel function as follows:
```go
shortCtx, _ := context.WithTimeout(ctx, defaultTestShortTimeout)
```
`govet` complains about a possible context leak because it can't ensure that the context will be cancelled at compile time. If we re-assign the `cancel` func later, `govet` doesn't complain but I still called `cancel` just to be consistent. Removed the call now.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
arjan-bal: The test doesn't skip the SubConn but it skips waiting for the SubConn to report a success or failure and moves on to the next SubConn. The comment was copied taken from [Java's test case](https://github.com/grpc/grpc-java/blob/36e29abf4168646b4e3866061c00b5cbea6fcead/core/src/test/java/io/grpc/internal/PickFirstLeafLoadBalancerTest.java#L2553). I've improved the wording now.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -580,9 +643,17 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
 	}
 }
 
-func (b *pickfirstBalancer) endFirstPassLocked(lastErr error) {
+func (b *pickfirstBalancer) endFirstPassIfPossibleLocked(lastErr error) {
+	if b.addressList.isValid() || b.subConns.Len() < b.addressList.size() {
```
dfawley, line 647: `isValid` feels a lot like `firstPass`.  And shouldn't `b.subConns.Len()` always be equal to the `addressList`'s index, so if `addressList.isValid` then the second case should always be true, too?  I'm sensing some duplication here that might be able to be removed & simplified.  Or, I'm not understanding something and we need some comments to explain why these checks are here.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -456,30 +473,73 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
...
 		case connectivity.Idle:
 			scd.subConn.Connect()
+			b.scheduleNextConnectionLocked()
+			return
 		case connectivity.TransientFailure:
-			// Try the next address.
+			// The SubConn is being re-used and failed during a previous pass
+			// over the addressList. It has not completed backoff yet.
+			// Mark it as having failed and try the next address.
+			scd.connectionFailed = true
```
dfawley, line 482: `connectionFailed` is a bit like `lastErr != nil`.  Do we need both?

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -456,30 +473,73 @@ func (b *pickfirstBalancer) requestConnectionLocked() {
...
+			return
 		case connectivity.Shutdown:
 			// Should never happen.
 			b.logger.Errorf("SubConn with state SHUTDOWN present in SubConns map")
+			return
 		case connectivity.Connecting:
-			// Wait for the SubConn to report success or failure.
+			// Wait for the connection attempt to complete or the timer to fire
+			// before attempting the next address.
+			b.scheduleNextConnectionLocked()
```
dfawley, line 496: It feels like either this or the one in the `Idle` case is redundant.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -580,9 +643,17 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
...
+		sd := v.(*scData)
+		if !sd.connectionFailed {
+			return
+		}
+	}
 	b.firstPass = false
-	b.numTF = 0
 	b.state = connectivity.TransientFailure
 
 	b.cc.UpdateState(balancer.State{
```
dfawley, line 659: Are we really supposed to attempt re-connection on every already-idle subchannel at the same instant?  And not apply the timer or anything?

internal/envconfig/envconfig.go
```diff
@@ -43,8 +43,7 @@ var (
 	// EnforceALPNEnabled is set if TLS connections to servers with ALPN disabled
 	// should be rejected. The HTTP/2 protocol requires ALPN to be enabled, this
 	// option is present for backward compatibility. This option may be overridden
-	// by setting the environment variable "GRPC_ENFORCE_ALPN_ENABLED" to "true"
-	// or "false".
+	// by setting the environment variable "GRPC_ENFORCE_ALPN_ENABLED" to "false".
```
dfawley, line 46: I suspect you didn't want these diffs.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
arjan-bal: > isValid feels a lot like firstPass.

For the first pass to be completed, all the SubConns need to report a TF. `!isValid()` will indicate that `SubConn.Connect()` has been called on all the SubConns, but it doesn't tell us if TFs have been reported. So the `isValid()` check is used as an optimization to return early and avoid iterating on the entire SubConns map.

> And shouldn't b.subConns.Len() always be equal to the addressList's index, so if addressList.isValid then the second case should always be true, too?

Yes, the check for `b.subConns.Len() < b.addressList.size()` seems redundant because if `!b.addressList.isValid()`, then a SubConn would have been created for every address while iterating over the list. Removed the check for `b.subConns.Len() < b.addressList.size()`.

Added some comments explaining the check and what the function does.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
arjan-bal: Yes, this is mentioned in [A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#happy-eyeballs-in-the-pick_first-lb-policy):

> If the first pass completes without a successful connection attempt, we will switch to a mode where we keep trying to connect to all addresses at all times, with no regard for the order of the addresses.

internal/envconfig/envconfig.go
arjan-bal: Reverted. I initially added an env var to toggle happy eyeballs, which was later removed it. I made a similar change to the doc comment for new env var based on reviewer feedback.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
dfawley: OK, I see.  I thought this was called from `updateSubConnState` -- and it is, but only in the case where the subchannel update was a TF update.  Thanks!

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
dfawley: Raising this concern elsewhere. :)

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
dfawley: OK, I see what's happening here, thanks for the explanation.

Maybe name it `connectionFailed(In/During)FirstPass`?

dfawley (approved):
LGTM modulo the one request to change `connectionFailed` to be a little more specific.

Thanks!!

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
arjan-bal: Renamed to `connectionFailedInFirstPass`.

PR #7726, opened 2024-10-10
xds: switching to stubserver in tests instead of testservice implementation

janardhanvissa:

RELEASE NOTES: None

### DIFF
--- a/xds/internal/httpfilter/fault/fault_test.go
+++ b/xds/internal/httpfilter/fault/fault_test.go
@@ -37,6 +37,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/internal/grpctest"
+	"google.golang.org/grpc/internal/stubserver"
 	"google.golang.org/grpc/internal/testutils"
 	"google.golang.org/grpc/internal/testutils/xds/e2e"
 	"google.golang.org/grpc/metadata"
@@ -67,25 +68,6 @@ func Test(t *testing.T) {
 	grpctest.RunSubTests(t, s{})
 }
 
-type testService struct {
-	testgrpc.TestServiceServer
-}
-
-func (*testService) EmptyCall(context.Context, *testpb.Empty) (*testpb.Empty, error) {
-	return &testpb.Empty{}, nil
-}
-
-func (*testService) FullDuplexCall(stream testgrpc.TestService_FullDuplexCallServer) error {
-	// End RPC after client does a CloseSend.
-	for {
-		if _, err := stream.Recv(); err == io.EOF {
-			return nil
-		} else if err != nil {
-			return err
-		}
-	}
-}
-
 // clientSetup performs a bunch of steps common to all xDS server tests here:
 // - spin up an xDS management server on a local port
 // - spin up a gRPC server and register the test service on it
@@ -106,9 +88,26 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	bootstrapContents := e2e.DefaultBootstrapContents(t, nodeID, managementServer.Address)
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
+	stub := &stubserver.StubServer{
+		EmptyCallF: func(context.Context, *testpb.Empty) (*testpb.Empty, error) {
+			return &testpb.Empty{}, nil
+		},
+		FullDuplexCallF: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+			for {
+				if _, err := stream.Recv(); err == io.EOF {
+					return nil
+				} else if err != nil {
+					return err
+				}
+			}
+		},
+	}
 	// Initialize a gRPC server and register the stubServer on it.
 	server := grpc.NewServer()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
+
+	// Set the server in the stub and start the test service.
+	stub.S = server
+	stubserver.StartTestService(t, stub)
 
 	// Create a local listener and pass it to Serve().
 	lis, err := testutils.LocalTCPListener()
--- a/xds/internal/httpfilter/fault/fault_test.go
+++ b/xds/internal/httpfilter/fault/fault_test.go
@@ -79,7 +79,7 @@ func Test(t *testing.T) {
 //     sent by the xdsClient for queries.
 //   - the port the server is listening on
 //   - cleanup function to be invoked by the tests when done
-func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
+func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32) {
 	// Spin up a xDS management server on a local port.
 	nodeID := uuid.New().String()
 	managementServer := e2e.StartManagementServer(t, e2e.ManagementServerOptions{})
@@ -88,9 +88,15 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	bootstrapContents := e2e.DefaultBootstrapContents(t, nodeID, managementServer.Address)
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
-	// Initialize a test gRPC server, assign it to the stub server, and start
-	// the test service.
+	// Create a local listener.
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatalf("testutils.LocalTCPListener() failed: %v", err)
+	}
+
+	// Initialize a test gRPC server, assign it to the stub server, and start the test service.
 	stub := &stubserver.StubServer{
+		Listener: lis,
 		EmptyCallF: func(context.Context, *testpb.Empty) (*testpb.Empty, error) {
 			return &testpb.Empty{}, nil
 		},
@@ -106,16 +112,11 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 		},
 	}
 
-	// Create a local listener and assign it to the stub server.
-	lis, err := testutils.LocalTCPListener()
-	if err != nil {
-		t.Fatalf("testutils.LocalTCPListener() failed: %v", err)
-	}
-	stub.Listener = lis
 	stubserver.StartTestService(t, stub)
-	return managementServer, nodeID, uint32(lis.Addr().(*net.TCPAddr).Port), func() {
+	t.Cleanup(func() {
 		stub.S.Stop()
-	}
+	})
+	return managementServer, nodeID, uint32(lis.Addr().(*net.TCPAddr).Port)
 }
 
 func (s) TestFaultInjection_Unary(t *testing.T) {
@@ -457,8 +458,7 @@ func (s) TestFaultInjection_Unary(t *testing.T) {
 		}},
 	}}
 
-	fs, nodeID, port, cleanup := clientSetup(t)
-	defer cleanup()
+	fs, nodeID, port := clientSetup(t)
 
 	for tcNum, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
@@ -540,8 +540,8 @@ func (s) TestFaultInjection_Unary(t *testing.T) {
 }
 
 func (s) TestFaultInjection_MaxActiveFaults(t *testing.T) {
-	fs, nodeID, port, cleanup := clientSetup(t)
-	defer cleanup()
+	fs, nodeID, port := clientSetup(t)
+
 	resources := e2e.DefaultClientResources(e2e.ResourceParams{
 		DialTarget: "myservice",
 		NodeID:     nodeID,
--- a/xds/internal/httpfilter/fault/fault_test.go
+++ b/xds/internal/httpfilter/fault/fault_test.go
@@ -88,6 +88,7 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	bootstrapContents := e2e.DefaultBootstrapContents(t, nodeID, managementServer.Address)
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
+	// Initialize an xDS-enabled gRPC server and use the helper to start the test service.
 	stub := &stubserver.StubServer{
 		EmptyCallF: func(context.Context, *testpb.Empty) (*testpb.Empty, error) {
 			return &testpb.Empty{}, nil
@@ -104,7 +105,6 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	}
 	server := grpc.NewServer()
 
-	// Set the server in the stub and start the test service.
 	stub.S = server
 	stubserver.StartTestService(t, stub)
 
--- a/xds/internal/httpfilter/fault/fault_test.go
+++ b/xds/internal/httpfilter/fault/fault_test.go
@@ -113,10 +113,9 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	}
 	stub.Listener = lis
 	stubserver.StartTestService(t, stub)
-	cleanup := func() {
+	return managementServer, nodeID, uint32(lis.Addr().(*net.TCPAddr).Port), func() {
 		stub.S.Stop()
 	}
-	return managementServer, nodeID, uint32(lis.Addr().(*net.TCPAddr).Port), cleanup
 }
 
 func (s) TestFaultInjection_Unary(t *testing.T) {
--- a/xds/internal/httpfilter/fault/fault_test.go
+++ b/xds/internal/httpfilter/fault/fault_test.go
@@ -88,7 +88,7 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	bootstrapContents := e2e.DefaultBootstrapContents(t, nodeID, managementServer.Address)
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
-	// Initialize an xDS-enabled gRPC server and use the helper to start the test service.
+	// Initialize a test gRPC server, assign it to the stub server, and start the test service.
 	stub := &stubserver.StubServer{
 		EmptyCallF: func(context.Context, *testpb.Empty) (*testpb.Empty, error) {
 			return &testpb.Empty{}, nil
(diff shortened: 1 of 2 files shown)


xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,9 +88,26 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
...
+			for {
+				if _, err := stream.Recv(); err == io.EOF {
+					return nil
+				} else if err != nil {
+					return err
+				}
+			}
+		},
+	}
 	// Initialize a gRPC server and register the stubServer on it.
```
purnesh42H, line 105: can we remove this comment as we are not doing it anymore?

xds/server_ext_test.go
```diff
@@ -144,13 +124,33 @@ func (s) TestServingModeChanges(t *testing.T) {
...
 	server, err := xds.NewGRPCServer(grpc.Creds(insecure.NewCredentials()), modeChangeOpt, xds.BootstrapContentsForTesting(bootstrapContents))
 	if err != nil {
 		t.Fatalf("Failed to create an xDS enabled gRPC server: %v", err)
 	}
+
+	// Set the server in the stub and start the test service.
+	stub.S = server
+	stubserver.StartTestService(t, stub)
+
 	defer server.Stop()
```
purnesh42H, line 152: defer server.Stop() should ideally be right after server be created. So, you can move the new code after that. It is because we need to execute it that even in case of errors. Right now, if something goes wrong in 149 and 150, server.Stop() won't execute.

Applicable everywhere

xds/server_ext_test.go
```diff
@@ -271,12 +271,34 @@ func (s) TestResourceNotFoundRDS(t *testing.T) {
...
```
purnesh42H, line 300: same here: move the new code after `defer server.Stop()`

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,9 +88,25 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	bootstrapContents := e2e.DefaultBootstrapContents(t, nodeID, managementServer.Address)
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
-	// Initialize a gRPC server and register the stubServer on it.
```
purnesh42H, line 109: similar comment as previous PR. Modify the comment to specify setting on stub server and remove the below extra line

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,9 +88,25 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
...
+	// Initialize an xDS-enabled gRPC server and use the helper to start the test service.
```
purnesh42H, line 91: this can be simply written like this

`// Initialize a test gRPC server, assign it to the stub server, and start the test service.`

janardhanvissa: Addressed the comments and pushed the changes. Please re-review once.

purnesh42H:
@easwars for second review

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,9 +88,26 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	bootstrapContents := e2e.DefaultBootstrapContents(t, nodeID, managementServer.Address)
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
-	// Initialize a gRPC server and register the stubServer on it.
+	// Initialize a test gRPC server, assign it to the stub server, and start the test service.
```
arvindbr8, line 91:
```suggestion
	// Initialize a test gRPC server, assign it to the stub server, and start
	// the test service.
```

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,9 +88,26 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
...
+					return err
+				}
+			}
+		},
+	}
+
 	server := grpc.NewServer()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
+
+	stub.S = server
```
arvindbr8, line 109: you can skip these lines.  This will be done for you if `stub.S == nil`

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,9 +88,26 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
...
 	testutils.CreateBootstrapFileForTesting(t, bootstrapContents)
 
-	// Initialize a gRPC server and register the stubServer on it.
+	// Initialize a test gRPC server, assign it to the stub server, and start the test service.
+	stub := &stubserver.StubServer{
+		EmptyCallF: func(context.Context, *testpb.Empty) (*testpb.Empty, error) {
+			return &testpb.Empty{}, nil
+		},
+		FullDuplexCallF: func(stream testgrpc.TestService_FullDuplexCallServer) error {
+			for {
```
arvindbr8, line 97: Re-add the comment from deleted lines?

xds/server_ext_test.go
```diff
@@ -145,12 +125,31 @@ func (s) TestServingModeChanges(t *testing.T) {
...
+	}
 	server, err := xds.NewGRPCServer(grpc.Creds(insecure.NewCredentials()), modeChangeOpt, xds.BootstrapContentsForTesting(bootstrapContents))
 	if err != nil {
 		t.Fatalf("Failed to create an xDS enabled gRPC server: %v", err)
 	}
 	defer server.Stop()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
+
+	stub.S = server
+	stubserver.StartTestService(t, stub)
```
arvindbr8, line 151: you combine all of this to one line by passing these sopts to StartTestService

xds/server_ext_test.go
```diff
@@ -145,12 +125,31 @@ func (s) TestServingModeChanges(t *testing.T) {
...
+
```
arvindbr8, line 152: nix this new line pls

xds/server_ext_test.go
```diff
@@ -145,12 +125,31 @@ func (s) TestServingModeChanges(t *testing.T) {
...
 	go func() {
 		if err := server.Serve(lis); err != nil {
```
arvindbr8, line 154: wait, StartTestService already does `server.Serve`. What is the expected behavior here?

xds/server_ext_test.go
```diff
@@ -271,12 +270,31 @@ func (s) TestResourceNotFoundRDS(t *testing.T) {
...
+	}
 	server, err := xds.NewGRPCServer(grpc.Creds(insecure.NewCredentials()), modeChangeOpt, xds.BootstrapContentsForTesting(bootstrapContents))
 	if err != nil {
 		t.Fatalf("Failed to create an xDS enabled gRPC server: %v", err)
 	}
 	defer server.Stop()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
+
+	stub.S = server
+	stubserver.StartTestService(t, stub)
```
arvindbr8, line 296: same comment as before

arvindbr8 (changes requested):

arvindbr8:
@janardhankrishna-sai -- Also please mention the PR description why does this diff "partially" fix the issue. The last few comments in https://github.com/grpc/grpc-go/issues/7291 is also not very clear.

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: If stub.S is nil, that initialization will be handled automatically. I'll go ahead and skip those lines as suggested.

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: re-adding the deleted comment lines

xds/server_ext_test.go
janardhanvissa: updated to stub.S.Server(lis)

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: Updated the comments as suggested

arvindbr8: this suggestion was to wrap comments at 80-col

xds/server_ext_test.go
arvindbr8: Could you comment on why we need to call Serve(lis) twice? (once in StartTestService, and then here?)

janardhanvissa: First Call: This initializes the server with the StubServer, enabling it to handle incoming RPCs. This setup is essential for the server to be ready to respond to any requests that may come in during the test.

Second Call : This call is necessary to start the server on the provided listener (lis). It allows the server to begin accepting connections. It enables specific configurations or behaviors that need to be applied after the StartTestService setup.

janardhanvissa: Please correct me if I'm not in the right direction.

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: yeah understood and done that.

xds/server_ext_test.go
janardhanvissa: combined server initialization and passed sopts to StartTestService.

xds/server_ext_test.go
janardhanvissa: combined server initialization and passed sopts to StartTestService.

xds/server_ext_test.go
arvindbr8: Sure. [stubserver.StartTestService](https://github.com/grpc/grpc-go/blob/master/internal/stubserver/stubserver.go#L277) calls [StartServer](https://github.com/grpc/grpc-go/blob/master/internal/stubserver/stubserver.go#L197) which calls [ss.S.serve(lis)](https://github.com/grpc/grpc-go/blob/master/internal/stubserver/stubserver.go#L203). This seems redundant in these tests. Could you check if this is really required?

janardhanvissa: **stubserver.StartTestService():**
This function sets up and registers the necessary gRPC service handlers (in our case, EmptyCall, UnaryCall, and FullDuplexCall) on the server (stub.S). Without this, the server won’t have the expected handlers for the incoming RPCs, resulting in unimplemented errors.

**Serve() in a goroutine:**
This actually starts the server and begins listening for incoming connections on the specified listener (lis).
Without calling Serve(), the server won’t accept any connections, leading to errors such as connection refused.

purnesh42H: @janardhanvissa Arvind is right. `StartTestService` already starts the server on a default listener (localhost:0) if no listener is provided to stub. What you should do here is assign the listener to stub i.e. `stub.Listener = lis` and then do stub.StartTestService(...). Then, you can delete the below go routine in the test which is starting the server.

@arvindbr8 great catch. I didn't notice it as that goroutine  was not visible in the diff.

xds/server_ext_test.go
```diff
@@ -145,14 +125,32 @@ func (s) TestServingModeChanges(t *testing.T) {
...
 		t.Fatalf("Failed to create an xDS enabled gRPC server: %v", err)
 	}
-	defer server.Stop()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
+	stubserver.StartTestService(t, stub, sopts...)
+	defer stub.S.Stop()
+
 	go func() {
-		if err := server.Serve(lis); err != nil {
+		if err := stub.S.Serve(lis); err != nil {
```
purnesh42H, line 153: same thing here. Do the same as suggested above

xds/server_ext_test.go
```diff
@@ -271,14 +269,32 @@ func (s) TestResourceNotFoundRDS(t *testing.T) {
...
```
purnesh42H, line 297: same. Do the same as suggested above.

xds/server_ext_test.go
janardhanvissa: Thank you, @purnesh42H  and @arvindbr8 , for catching that! I overlooked the fact that StartTestService defaults to using localhost:0 when no listener is assigned. Assigning stub.Listener = lis before stub.StartTestService(...) makes sense to avoid the extra goroutine in the test for starting the server. I'll update the tests accordingly.

janardhanvissa:
> @janardhankrishna-sai -- Also please mention the PR description why does this diff "partially" fix the issue. The last few comments in #7291 is also not very clear.

xds/server_ext_test.go
janardhanvissa: Done with the changes.

xds/server_ext_test.go
janardhanvissa: Done with the changes.

xds/server_ext_test.go
janardhanvissa: Done with the changes.

xds/server_ext_test.go
purnesh42H: @janardhanvissa remove this new line. New lines should only be there to isolate logical group of code

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -117,13 +115,13 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
 	}
 
 	go func() {
-		if err := server.Serve(lis); err != nil {
+		if err := stub.S.Serve(lis); err != nil {
```
purnesh42H, line 118: @janardhanvissa you haven't removed the redundant `s.serve()` here

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -117,13 +115,13 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
...
 			t.Errorf("Serve() failed: %v", err)
 		}
 	}()
 
 	return managementServer, nodeID, uint32(lis.Addr().(*net.TCPAddr).Port), func() {
-		server.Stop()
+		stub.S.Stop()
```
purnesh42H, line 124: this should be right after StartTestService with `defer`

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: missed to update it. Done now.

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: The defer stub.S.Stop() call in clientSetup is stopping the server too early, before the test completes. This causes subsequent RPC calls in the test to fail with a Canceled error, and leads to a nil pointer dereference when trying to access the stub server's resources.

purnesh42H: ignore my comment of defer. The function was being returned with stub.S.Stop so we should keep it that way

arvindbr8: Could we use `t.Cleanup` directly in this helper function? It would simplify the calling code by removing the need for an explicit defer statement.

arvindbr8 (changes requested):
Thanks for making the changes per conversations. I just had a few more comments, please take a look.

xds/server_ext_test.go
```diff
@@ -145,17 +125,29 @@ func (s) TestServingModeChanges(t *testing.T) {
...
 	}
-	defer server.Stop()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
-	go func() {
-		if err := server.Serve(lis); err != nil {
-			t.Errorf("Serve() failed: %v", err)
-		}
-	}()
+	stub.Listener = lis
+	stubserver.StartTestService(t, stub, sopts...)
```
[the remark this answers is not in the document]
janardhanvissa: moved Listener to stub initialization and removed sopts in StartTestService.

xds/server_ext_test.go
```diff
@@ -271,17 +263,29 @@ func (s) TestResourceNotFoundRDS(t *testing.T) {
...
```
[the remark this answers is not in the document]
janardhanvissa: removed sopts in StartTestService.

xds/server_ext_test.go
```diff
@@ -145,17 +125,29 @@ func (s) TestServingModeChanges(t *testing.T) {
...
 		t.Fatalf("Failed to create an xDS enabled gRPC server: %v", err)
 	}
-	defer server.Stop()
-	testgrpc.RegisterTestServiceServer(server, &testService{})
-	go func() {
-		if err := server.Serve(lis); err != nil {
-			t.Errorf("Serve() failed: %v", err)
-		}
-	}()
+	stub.Listener = lis
```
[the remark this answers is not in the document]
janardhanvissa: moved Listener to stub initialization.

xds/server_ext_test.go
```diff
@@ -271,17 +263,29 @@ func (s) TestResourceNotFoundRDS(t *testing.T) {
...
```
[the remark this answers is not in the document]
janardhanvissa: moved Listener to stub initialization.

xds/internal/httpfilter/fault/fault_test.go
janardhanvissa: Yes, using t.Cleanup directly in the helper function is the preferred approach for cleanup in tests. Updated the tests accordingly.

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -106,25 +88,35 @@ func clientSetup(t *testing.T) (*e2e.ManagementServer, string, uint32, func()) {
...
+					return err
+				}
+			}
+		},
 	}
+
+	stubserver.StartTestService(t, stub)
+	t.Cleanup(func() {
+		stub.S.Stop()
+	})
```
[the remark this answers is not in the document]
janardhanvissa: done with the change.

xds/internal/httpfilter/fault/fault_test.go
```diff
@@ -549,8 +540,8 @@ func (s) TestFaultInjection_Unary(t *testing.T) {
 }
 
 func (s) TestFaultInjection_MaxActiveFaults(t *testing.T) {
-	fs, nodeID, port, cleanup := clientSetup(t)
-	defer cleanup()
+	fs, nodeID, port := clientSetup(t)
+
```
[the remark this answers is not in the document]
janardhanvissa: removed empty line

PR #7727, opened 2024-10-10
orca: switching to stubserver in tests instead of testservice implementation

janardhanvissa:

RELEASE NOTES: None

### DIFF
--- a/orca/service_test.go
+++ b/orca/service_test.go
@@ -153,10 +153,6 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 		currentRequests := requests
 		mu.Unlock()
 
-		if currentRequests == numRequests {
-			break
-		}
-
 		wantProto := &v3orcapb.OrcaLoadReport{
 			CpuUtilization:         50.0,
 			MemUtilization:         0.9,
@@ -179,6 +175,10 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 	if _, err := testStub.EmptyCall(ctx, &testpb.Empty{}); err != nil {
 		t.Fatalf("EmptyCall failed: %v", err)
 	}
+
+	// A short delay to ensure the metrics deletion is processed.
+	time.Sleep(shortReportingInterval * 2)
+
 	// Wait for the server to push empty metrics which indicate the processing
 	// of the above EmptyCall RPC.
 	for {
--- a/orca/service_test.go
+++ b/orca/service_test.go
@@ -172,9 +172,6 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 		t.Fatalf("EmptyCall failed: %v", err)
 	}
 
-	// A short delay to ensure the metrics deletion is processed.
-	time.Sleep(shortReportingInterval * 2)
-
 	// Wait for the server to push empty metrics which indicate the processing
 	// of the above EmptyCall RPC.
 	for {
--- a/orca/service_test.go
+++ b/orca/service_test.go
@@ -65,20 +65,17 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 
 	var requests int
 	var mu sync.Mutex
-	var wg sync.WaitGroup
 
 	const numRequests = 20
-	wg.Add(numRequests)
 
 	stub := &stubserver.StubServer{
 		UnaryCallF: func(ctx context.Context, req *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
-			defer wg.Done()
-
 			mu.Lock()
 			requests++
+			currRequests := requests
 			mu.Unlock()
 
-			smr.SetNamedUtilization(requestsMetricKey, float64(requests)*0.01)
+			smr.SetNamedUtilization(requestsMetricKey, float64(currRequests)*0.01)
 			smr.SetCPUUtilization(50.0)
 			smr.SetMemoryUtilization(0.9)
 			smr.SetApplicationUtilization(1.2)
@@ -130,8 +127,6 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 		errCh <- nil
 	}()
 
-	wg.Wait()
-
 	// Start the server streaming RPC to receive custom backend metrics.
 	oobStub := v3orcaservicegrpc.NewOpenRcaServiceClient(cc)
 	stream, err := oobStub.StreamCoreMetrics(ctx, &v3orcaservicepb.OrcaLoadReportRequest{ReportInterval: durationpb.New(shortReportingInterval)})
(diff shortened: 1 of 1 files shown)


orca/service_test.go
```diff
@@ -112,11 +110,11 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
 	defer cc.Close()
 
-	// Spawn a goroutine which sends 20 unary RPCs to the test server. This
+	// Spawn a goroutine which sends 20 unary RPCs to the stub server. This
 	// will trigger the injection of custom backend metrics from the
-	// testServiceImpl.
+	// stubServer.
 	const numRequests = 20
-	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
purnesh42H, line 117: Why do we need to change timeout here?

orca/service_test.go
```diff
@@ -138,16 +136,17 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 		t.Fatalf("Failed to create a stream for out-of-band metrics")
 	}
 
+	// Wait for the goroutine to finish before processing metrics.
+	if err := <-errCh; err != nil {
```
purnesh42H, line 140: why is this out of the select-context block?

orca/service_test.go
janardhanvissa: The default is 5 secs, getting some context deadline issues. So, updated to 10 secs.

orca/service_test.go
```diff
@@ -131,6 +130,8 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 		errCh <- nil
 	}()
 
+	wg.Wait()
```
purnesh42H, line 133: We don't need this waitGroup, there is already a for loop which Wait for the server to push metrics which indicate the completion of all the unary RPCs made from the above goroutine.

orca/service_test.go
```diff
@@ -151,11 +147,15 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 		default:
 		}
 
+		mu.Lock()
```
purnesh42H, line 150: why did you add this? Please provide inline comments explaining such changes. The previous wantProto should work as is I think. If not, then please mention what is happening

orca/service_test.go
```diff
@@ -151,11 +147,15 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
+		expectedUtilization := float64(requests) * 0.01
```
purnesh42H, line 151: why did you change this value to float64(requests) * 0.01? it was `numRequests * 0.01`. The numRequests is a constant so that's we should have in want. requests are being calculated within stub server call and may not be synchronized. You shouldn't need mutex to assign a constant value

purnesh42H (approved):
lgtm

purnesh42H:
@easwars for second review

orca/service_test.go
```diff
@@ -93,14 +63,42 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
 		t.Fatalf("orca.EnableOutOfBandMetricsReportingForTesting() failed: %v", err)
 	}
 
-	// Register the test service implementation on the same grpc server, and start serving.
-	testgrpc.RegisterTestServiceServer(s, &testServiceImpl{smr: smr})
+	// Assign the gRPC server to the stub server and start serving.
+	stub.S = s
+	stubserver.StartTestService(t, stub)
 	go s.Serve(lis)
 	defer s.Stop()
```
arvindbr8, line 103: Similar to my comment from the other PR. We can just pass in sopts to `StartTestService`

orca/service_test.go
```diff
@@ -93,14 +63,42 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
 	if err := orca.Register(s, opts); err != nil {
 		t.Fatalf("orca.EnableOutOfBandMetricsReportingForTesting() failed: %v", err)
 	}
 
-	// Register the test service implementation on the same grpc server, and start serving.
-	testgrpc.RegisterTestServiceServer(s, &testServiceImpl{smr: smr})
+	// Assign the gRPC server to the stub server and start serving.
+	stub.S = s
+	stubserver.StartTestService(t, stub)
 	go s.Serve(lis)
```
arvindbr8, line 102: StartTestService already does s.Serve. What is the expected behavior here

arvindbr8 (changes requested):

purnesh42H:
@arvindbr8 for re-review

orca/service_test.go
```diff
@@ -93,16 +63,42 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 	opts := orca.ServiceOptions{MinReportingInterval: shortReportingInterval, ServerMetricsProvider: smr}
 	internal.AllowAnyMinReportingInterval.(func(*orca.ServiceOptions))(&opts)
 
+	var requests int64
+	var mu sync.Mutex
```
arvindbr8, line 67: can we consider using `atomics` here?

orca/service_test.go
```diff
@@ -93,16 +63,42 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
+
+	const numRequests = 20
```
arvindbr8, line 69: Let's move this back to where the usage is.

orca/service_test.go
```diff
@@ -93,16 +63,42 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
+			smr.DeleteNamedUtilization(requestsMetricKey)
+			smr.SetCPUUtilization(0)
+			smr.SetMemoryUtilization(0)
+			smr.DeleteApplicationUtilization()
+			return &testpb.Empty{}, nil
+		},
+	}
+
+	// Assign the gRPC server to the stub server and start serving.
+	stub.Listener = lis
```
arvindbr8, line 94: Let's specify this where `stub` is initialized

orca/service_test.go
```diff
@@ -93,16 +63,42 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
...
+	var mu sync.Mutex
+
+	const numRequests = 20
+
+	stub := &stubserver.StubServer{
+		UnaryCallF: func(ctx context.Context, req *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+			mu.Lock()
+			requests++
+
+			smr.SetNamedUtilization(requestsMetricKey, float64(requests)*0.01)
```
arvindbr8, line 76: seems like we have changed how the lock is handled here, but this seems to be the right way of doing it. 

But this would go away if we use an atominc.Int64 here.

arvindbr8 (changes requested):

orca/service_test.go
```diff
@@ -93,16 +63,37 @@ func (s) TestE2E_CustomBackendMetrics_OutOfBand(t *testing.T) {
 	opts := orca.ServiceOptions{MinReportingInterval: shortReportingInterval, ServerMetricsProvider: smr}
 	internal.AllowAnyMinReportingInterval.(func(*orca.ServiceOptions))(&opts)
 
+	var requests atomic.Int64
+
+	stub := &stubserver.StubServer{
+		Listener: lis,
+		UnaryCallF: func(ctx context.Context, req *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+
```
arvindbr8, line 71:
```suggestion
```

arvindbr8 (approved):
LGTM

PR #7733, opened 2024-10-11
loopyWriter: fix dataItem.h

luxcgo:
```go
	hSize := min(maxSize, len(dataItem.h))
        ...
	if hSize != 0 && dSize == 0 {
		buf = &dataItem.h
        }
```

If `hSize` is less than `len(dataItem.h)`, we still send the whole header.
So afterward, we need to set it nil, instead of using the wrong `hSize`

### DIFF
--- a/internal/transport/controlbuf.go
+++ b/internal/transport/controlbuf.go
@@ -958,17 +958,17 @@ func (l *loopyWriter) processData() (bool, error) {
 		return false, nil
 	}
 
-	// Figure out the maximum size we can send
-	maxSize := http2MaxFrameLen
-	if strQuota := int(l.oiws) - str.bytesOutStanding; strQuota <= 0 { // stream-level flow control.
+	strQuota := int(l.oiws) - str.bytesOutStanding
+	if strQuota <= 0 { // stream-level flow control.
 		str.state = waitingOnStreamQuota
 		return false, nil
-	} else if maxSize > strQuota {
-		maxSize = strQuota
 	}
-	if maxSize > int(l.sendQuota) { // connection-level flow control.
-		maxSize = int(l.sendQuota)
-	}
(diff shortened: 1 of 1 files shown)


eshitachandwani:
Hey @luxcgo , we are reviewing your PR, but in the mean time would you write unit test for this change. It looks like a logic change it should always be accompanied by unit tests that fail before the change and pass after the change.

internal/transport/controlbuf.go
```diff
@@ -958,17 +958,17 @@ func (l *loopyWriter) processData() (bool, error) {
...
 	}
+	// Figure out the maximum size we can send (guaranteed to be bigger than 0)
+	// Shall be the smallest one among the following values:
+	// - http2MaxFrameLen
+	// - strQuota(stream-level flow control)
+	// - l.sendQuota(connection-level flow control)
+	maxSize := min(http2MaxFrameLen, strQuota, int(l.sendQuota))
 	// Compute how much of the header and data we can send within quota and max frame length
 	hSize := min(maxSize, len(dataItem.h))
 	dSize := min(maxSize-hSize, dataItem.reader.Remaining())
```
arjan-bal, line 974: Will we also need to adjust the calculation of `remainingBytes` and `size` accordingly since it's used to close the reader and adjusting flow control below?
Practically, I believe `remainingBytes` calculation will not cause problem since there would always be data left to write if we're writing headers.

arjan-bal: I think fixing `remainingBytes` and `size` may make the flow control windows negative. So maybe writing the complete header bytes while not having enough quota is a bug?

arjan-bal:
Closing the PR as the current behaviour looks correct based on https://github.com/grpc/grpc-go/pull/7733#discussion_r1806694975. Please re-open if you think this is not the case.

Issue #7049, opened 2024-03-19
Update docs and examples and tests to use `NewClient` instead of `Dial`


dfawley:
After #7010 we need to make sure all our documentation recommends `NewClient`.  We should still retain some tests for `Dial` but the majority of them should probably be converted to use `NewClient` as well.

silaselisha:
@dfawley I would like to work on this issue

dfawley:
Sounds good @silaselisha, that would be great -- we need to start only inside the repo for now, because we don't want to update any other documentation until our next release which includes the new function.

Note that we can't just do a trivial `sed` to replace `grpc.Dial` with `grpc.NewClient` (though maybe that's a decent way to start), because there are minor behavior differences that some of our tests probably (hopefully) will rely on.

silaselisha:
@dfawley  I'll start working on the `example repo`, and as mentioned I'll strictly traverse through the files just to manually replacing `grpc.Dial` with `grpc.NewClient`.

SamYuan1990:
To be specific, how to set timeout makes me confusing as https://github.com/grpc/grpc-go/issues/7093

dfawley:
Timeouts are not supported in `NewClient`. See https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#dialing-in-grpc for more info.

If you need timeouts, feel free to keep using `Dial` or `DialContext`; they are not going to be deleted in 1.x, they are just not a good idea to use per the information above.

SamYuan1990:
> Timeouts are not supported in `NewClient`. See https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#dialing-in-grpc for more info.
> 
> If you need timeouts, feel free to keep using `Dial` or `DialContext`; they are not going to be deleted in 1.x, they are just not a good idea to use per the information above.

can we remove Deprecated on `Dial` or `DialContext` to avoid lint error?

SamYuan1990:
https://github.com/Hyperledger-TWGC/tape/actions/runs/8551020475/job/23429132147?pr=384#step:5:25
```
Error: SA1019: grpc.DialContext is deprecated: use NewClient instead.  Will be supported throughout 1.x.  (staticcheck)
```

dfawley:
https://www.merriam-webster.com/dictionary/deprecate

> to withdraw official support for **_or discourage the use of (something, such as a software product) in favor of a newer or better alternative_**

[...]

PR #7737, opened 2024-10-14
cleanup: replace grpc.Dial with grpc.NewClient in tests

hanut19:
replace grpc.Dial with grpc.NewClient in test files

Addresses: #7049 

RELEASE NOTES: None

### DIFF
--- a/balancer/grpclb/grpclb_test.go
+++ b/balancer/grpclb/grpclb_test.go
@@ -832,11 +832,12 @@ func (s) TestGRPCLB_Fallback(t *testing.T) {
 		grpc.WithTransportCredentials(&serverNameCheckCreds{}),
 		grpc.WithContextDialer(fakeNameDialer),
 	}
-	cc, err := grpc.Dial(r.Scheme()+":///"+beServerName, dopts...)
+	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create new client to the backend %v", err)
 	}
 	defer cc.Close()
+	cc.Connect()
 	testC := testgrpc.NewTestServiceClient(cc)
 
 	// Push an update to the resolver with fallback backend address stored in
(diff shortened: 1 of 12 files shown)


internal/idle/idle_e2e_test.go
```diff
@@ -131,12 +131,12 @@ func (s) TestChannelIdleness_Disabled_NoActivity(t *testing.T) {
 		grpc.WithIdleTimeout(0), // Disable idleness.
 		grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`),
 	}
-	cc, err := grpc.Dial(r.Scheme()+":///test.server", dopts...)
+	cc, err := grpc.NewClient(r.Scheme()+":///test.server", dopts...)
 	if err != nil {
-		t.Fatalf("grpc.Dial() failed: %v", err)
+		t.Fatalf("Failed to create new client: %v", err)
```
purnesh42H, line 136: i see that in some places you replaced with grpc.NewClient() and some places you have "Failed to create new client". Please be consistent. grpc.Dial() failed can be changed to "grpc.NewClient() failed" and "Failed to dial" can be changed to "Failed to create new client"

hanut19: done please check

resolver_balancer_ext_test.go
```diff
@@ -109,11 +110,12 @@ func (s) TestResolverBuildFailure(t *testing.T) {
 	resolver.Register(&resolverBuilderWithErr{errCh: resErrCh, scheme: name})
 
 	resErrCh <- nil
-	cc, err := grpc.Dial(name+":///", grpc.WithTransportCredentials(insecure.NewCredentials()))
+	cc, err := grpc.NewClient(name+":///", grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		t.Fatalf("grpc.Dial error: %v", err)
+		t.Fatalf("grpc.NewClient error: %v", err)
```
purnesh42H, line 115: nit: `grpc.NewClient()`

purnesh42H (approved):
lgtm

resolver_balancer_ext_test.go
hanut19: done please check

balancer/grpclb/grpclb_test.go
```diff
@@ -832,11 +832,12 @@ func (s) TestGRPCLB_Fallback(t *testing.T) {
...
 		grpc.WithContextDialer(fakeNameDialer),
 	}
-	cc, err := grpc.Dial(r.Scheme()+":///"+beServerName, dopts...)
+	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create new client to the backend %v", err)
 	}
 	defer cc.Close()
+	cc.Connect()
```
dfawley, line 840: This should be unnecessary, throughout.

If it is necessary, there should be a comment explaining exactly why.

I suspect the reason for this one is that we're updating the resolver state before the channel has started?  If it is needed here, you can use `r.InitialState` before creating the client instead of `r.UpdateState` afterwards.  But it's possible it works without the `Connect` call, even without that change.

hanut19: cc.Connect() to ensure that the resolver is built. The clientconn is passed to the resolver though the resolver.Build() method. grpc.NewClient is lazy, i.e. it doesn't start creating subchannels until an RPC is made or cc.Connect() is called.

purnesh42H:
@hanut19 let's close this and do a shorter PR with one/two change at a time. We shouldn't be using `cc.connect` if rpc is being made in the test.

PR #7740, opened 2024-10-14
stats/opentelemetry/csm: Get mesh_id local label from "CSM_MESH_ID" environment variable, rather than parsing from bootstrap file

zasweq:
This PR switches the mesh_id local label to get set from the CSM_MESH_ID environment variable, with the same logic as other labels set from the environment. It no longer comes from the bootstrap, or needs to be parsed from the node_id of the bootstrap.

RELEASE NOTES:
* stats/opentelemetry/csm: Get mesh_id local label from "CSM_MESH_ID" environment variable, rather than parsing from bootstrap file

### DIFF
--- a/stats/opentelemetry/csm/observability_test.go
+++ b/stats/opentelemetry/csm/observability_test.go
@@ -32,8 +32,6 @@ import (
 	"google.golang.org/grpc/encoding/gzip"
 	istats "google.golang.org/grpc/internal/stats"
 	"google.golang.org/grpc/internal/stubserver"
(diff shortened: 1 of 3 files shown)


purnesh42H:
> stats/opentelemetry/csm: Get mesh_id local label from "CSM_MESH_ID" environment variable, rather than parsing from bootstrap file

Not fully familiar with the background on this but for now should we keep both for backward compatibility?

zasweq:
No not in this case, as we are getting rid of the old way entirely.

zasweq:
Interop tests passed.

PR #7751, opened 2024-10-17
github: add Go 1.23 testing and make staticcheck work locally with go1.23

arjan-bal:
RELEASE NOTES: None


.github/workflows/testing.yml
```diff
@@ -43,35 +43,35 @@ jobs:
       matrix:
         include:
           - type: vet
-            goversion: '1.22'
+            goversion: '1.23'
```
dfawley, line 46: I thought we agreed to keep vet running at 1.22?

arjan-bal: You're right, reverted to 1.22 for vet. There is a problem in vet.sh where it sets `PATH` using `GOROOT` even when `GOROOT` is unset. Because of that, vet was using go1.23. I've made the fix in this PR.

arjan-bal:
@dfawley, I've removed the `-go` flag being passed to staticcheck and that seems to have fixed the failures while using go1.23 locally.

dfawley (approved):

PR #7756, opened 2024-10-17
Make balancer registration threadsafe

ryanpbrewster:
Although it's not a well-supported pattern, I have use-cases where I can't easily register all of my balancer implementations in `init` functions.

Today, any call to `balancer.Register` can trigger a fatal error via an unsynchronized concurrent map access:
```
fatal error: concurrent map read and map write
goroutine 452 [running]:
google.golang.org/grpc/balancer.Get({0x2407e87, 0xa})
	external/org_golang_google_grpc/balancer/balancer.go:88 +0xea
```

This PR is intended to be a very lightweight change that makes it a bit less dangerous to do just-in-time balancer registration. Any existing use-cases should be unaffected, and the performance difference should be negligible.

RELEASE NOTES:
- Made `balancer.Register` threadsafe

### DIFF
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"net"
 	"strings"
+	"sync"
 
 	"google.golang.org/grpc/channelz"
 	"google.golang.org/grpc/connectivity"
@@ -40,7 +41,7 @@ import (
 
 var (
 	// m is a map from name to balancer builder.
-	m = make(map[string]Builder)
+	m sync.Map
 
 	logger = grpclog.Component("balancer")
 )
@@ -62,7 +63,7 @@ func Register(b Builder) {
 		// remove this warning and update the docstrings for Register and Get.
 		logger.Warningf("Balancer registered with name %q. grpc-go will be switching to case sensitive balancer registries soon", b.Name())
 	}
-	m[name] = b
+	m.Store(name, b)
 }
 
 // unregisterForTesting deletes the balancer with the given name from the
@@ -70,7 +71,7 @@ func Register(b Builder) {
 //
 // This function is not thread-safe.
 func unregisterForTesting(name string) {
-	delete(m, name)
+	m.Delete(name)
 }
 
 // connectedAddress returns the connected address for a SubConnState. The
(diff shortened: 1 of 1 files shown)


eshitachandwani (approved):

eshitachandwani:
Adding @purnesh42H for a second review.

eshitachandwani (approved):

eshitachandwani (commented):
We might not want to allow balancers to be registered after init time.

ryanpbrewster:
> We might not want to allow balancers to be registered after init time. 

I have a load balancer and I want to emit metrics from it. The approach I'm currently taking is to initialize a metrics client and then plumb it into the `balancer.Builder` implementation so that instances of my `balancer.Balancer` have access to it.

This approach has worked fairly well, but it does not play nicely with `init` functions. Is there a downside I'm missing to this approach?

eshitachandwani:
Hey @ryanpbrewster ,

The map is designed to be immutable after initialization to prevent potential issues, such as duplicate balancer registrations or inconsistencies. This ensures predictable behavior and avoids conflicts that could arise from concurrent modifications.

For scenarios where dynamic registration is needed, lazy initialization could be a suitable workaround. 
Please let me know if you have any further questions or concerns.

bobby-stripe:
@eshitachandwani from reading the code, today duplicate balancer registrations are _not_ detected or prevented.  Additionally this PR makes the map thread safe, preventing data races and basic consistency problems.

Can you share more of the reasoning here?  This PR solves an actual problem we have, and I don't understand from the comments why this simple change would make things worse or why it was rejected.  Thanks!

eshitachandwani:
@bobby-stripe The balancer map is intentionally not thread-safe, and we strongly recommend registering all balancers during init time.  This is because registering a balancer while RPCs are in flight can create a race condition. An RPC might attempt to use a balancer before the registration is fully complete, leading to the RPC failing to find the balancer and causing errors.

While this PR makes the map itself thread-safe, it doesn't eliminate this underlying race condition between balancer registration and active RPCs.

ryanpbrewster:
If the map is thread-safe then it is _possible_ to do runtime registration safely (as long as the caller ensures that the channels that ask for the new balancer are always created _after_ the balancer has been registered).

If the map is not thread-safe then it is flatly impossible to do any kind of runtime registration. It is not just a race-condition that might cause a few RPCs to fail, it is a fatal error that will kill the whole process.

purnesh42H:
@ryanpbrewster design philosophy for not making it thread-safe and not allowing runtime registration here is to emphasize stability and predictable behavior. 

The library maintains a global map to store the registered resolvers. This map is accessed whenever a new gRPC connection is established. Modifying this map concurrently during runtime can lead to race conditions and unpredictable behavior. Another reason I can think of is that while establishing connections, library uses the resolver map during its execution. Allowing runtime registration could mean that the resolver might not be available during dial sometimes, leading to inconsistencies.

Issue #7676, opened 2024-09-27
Buffer leak tests are not enabled


arjan-bal:
There is a GitHub workflow configured to run buffer leak detection: https://github.com/grpc/grpc-go/blob/e7a8097342e07dbfa093589eed67a8e0d6558423/.github/workflows/testing.yml#L58-L61

Here is the build tag that enabled the buffer leak detection: https://github.com/grpc/grpc-go/blob/e7a8097342e07dbfa093589eed67a8e0d6558423/internal/leakcheck/leakcheck_enabled.go#L1

The workflow is using the tag `buffer_pooling` while the build tag is `checkbuffers`.

Errors are reported after using the correct tag: https://github.com/grpc/grpc-go/actions/runs/11020737778/job/30606203907?pr=7670

From maintainers' group:
> We know there are some leaks in error situations. Paul had disabled it for some tests, but I think that stuff to disable it got torn out?
note that these leaks aren't actual memory leaks but just failure to re-use memory in some rare situations

arjan-bal:
@PapaCharlie would you be open to taking look at this?

PR #7762, opened 2024-10-21
Disable buffer_pooling tests

aranjans:
Related to: https://github.com/grpc/grpc-go/issues/7676 

RELEASE NOTES: n/a


arjan-bal (approved):

purnesh42H (commented):
can we comment this with a TODO instead of removing so that we don't forget about it? Or we are never going to fix?

aranjans:
We can comment this out with TODO but this would be dependent on papacharlie change. In case we are going to do this, then adding TODO sounds good to me. WDYT?

purnesh42H (approved):
lgtm. @dfawley to confirm if never plan to include or not? If we want to include this in future, we can keep the issue open or comment out this as TODO

dfawley (approved):

dfawley:
> lgtm. @dfawley to confirm if never plan to include or not? If we want to include this in future, we can keep the issue open or comment out this as TODO

No, this doesn't make sense, and was just left in the original PR by mistake.  Buffer pooling is always enabled.

PR #7764, opened 2024-10-21
balancer/roundrobin: fix Pick() when subConnsLen is 0

aranjans:
RELEASE NOTES: 

- balancer/roundrobin: fix Pick() when subConnsLen is 0.


arjan-bal (changes requested):
The base balancer returns an error when the resolver produces an empty address list: https://github.com/grpc/grpc-go/blob/98959d9a4904e98bbf8b423ce6a3cb5d36f90ee1/balancer/base/balancer.go#L139-L142

Have you observed the roundrobin picker getting an empty `SubConn` list? Could you add a unit test to demonstrate such a scenario?

aranjans:
@arjan-bal Ah, yeah. Sounds good, and hence this PR would be redundant. Hence, closing this PR

PR #7768, opened 2024-10-22
test: Add unit test for channel state waiting for first resolver update

janardhanvissa:
Addresses:  #7686 

RELEASE NOTES: None

### DIFF
--- a/test/clientconn_state_transition_test.go
+++ b/test/clientconn_state_transition_test.go
@@ -622,7 +622,7 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 // TestChannelStateWaitingForFirstResolverUpdate verifies the initial
 // state of the channel when a manual name resolver doesn't provide any updates.
 func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
-	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as Connect() is called soon as Connect() is called. See issue #7686.")
+	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as Connect() is called. See issue #7686.")
 
 	backend := stubserver.StartTestService(t, nil)
 	defer backend.Stop()
@@ -636,10 +636,12 @@ func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
 	}
 	defer cc.Close()
 
+	if state := cc.GetState(); state != connectivity.Idle {
+		t.Fatalf("Expected initial state to be IDLE, got %v", state)
+	}
+
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
-
-	testutils.AwaitState(ctx, t, cc, connectivity.Idle)
 
 	// The channel should transition to CONNECTING automatically when Connect()
 	// is called.
@@ -653,7 +655,7 @@ func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
 }
 
 func (s) TestChannelStateTransitionWithRPC(t *testing.T) {
-	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as an RPC call is made soon as Connect() is called. See issue #7686.")
+	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as an RPC call is made. See issue #7686.")
 
 	backend := stubserver.StartTestService(t, nil)
 	defer backend.Stop()
@@ -667,10 +669,12 @@ func (s) TestChannelStateTransitionWithRPC(t *testing.T) {
 	}
 	defer cc.Close()
 
+	if state := cc.GetState(); state != connectivity.Idle {
+		t.Fatalf("Expected initial state to be IDLE, got %v", state)
+	}
+
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
-
-	testutils.AwaitState(ctx, t, cc, connectivity.Idle)
 
 	// Make an RPC call to transition the channel to CONNECTING.
 	go func() {
--- a/test/clientconn_state_transition_test.go
+++ b/test/clientconn_state_transition_test.go
@@ -38,7 +38,6 @@ import (
 	"google.golang.org/grpc/internal/grpcsync"
 	"google.golang.org/grpc/internal/stubserver"
 	"google.golang.org/grpc/internal/testutils"
-	testgrpc "google.golang.org/grpc/interop/grpc_testing"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/resolver/manual"
 )
@@ -622,54 +621,40 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 // TestChannelStateWaitingForFirstResolverUpdate verifies the initial
 // state of the channel when a manual name resolver doesn't provide any updates.
 func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping test in short mode.")
+	}
+
 	backend := stubserver.StartTestService(t, nil)
 	defer backend.Stop()
 
 	mr := manual.NewBuilderWithScheme("e2e-test")
 	defer mr.Close()
 
-	cc, err := grpc.NewClient(mr.Scheme()+":///", grpc.WithResolvers(mr), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	cc, err := grpc.Dial(mr.Scheme()+":///", grpc.WithResolvers(mr), grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		t.Fatalf("Failed to create new client: %v", err)
 	}
 	defer cc.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
-	defer cancel()
-
-	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
-	defer shortCancel()
-
-	cc.Connect()
-	testutils.AwaitState(shortCtx, t, cc, connectivity.Idle)
+	testutils.AwaitState(context.Background(), t, cc, connectivity.Idle)
 
 	mr.UpdateState(resolver.State{
 		Addresses: []resolver.Address{{Addr: backend.Address}},
 	})
 
-	// Check that the channel transitions to CONNECTING then READY
-	testutils.AwaitState(shortCtx, t, cc, connectivity.Connecting)
-	t.Skipf("After connecting, before READY state: current state = %v\n", cc.GetState())
+	cc.Connect()
 
-	testutils.AwaitState(shortCtx, t, cc, connectivity.Ready)
-	t.Skipf("After entering READY state: current state = %v\n", cc.GetState())
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
 
-	mr.UpdateState(resolver.State{Addresses: nil})
+	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
 
-	done := make(chan struct{})
-	go func() {
-		_, err := testgrpc.NewTestServiceClient(cc).EmptyCall(shortCtx, &testgrpc.Empty{})
-		if err == nil {
-			t.Errorf("Expected RPC to fail, but it succeeded")
-		}
-		close(done)
-	}()
+	time.Sleep(100 * time.Millisecond)
 
-	// Wait for the goroutine to finish
-	select {
-	case <-done:
-		// RPC call completed as expected
-	case <-shortCtx.Done():
-		t.Error("Timed out waiting for RPC call to complete")
-	}
+	mr.UpdateState(resolver.State{
+		Addresses: []resolver.Address{{Addr: backend.Address}},
+	})
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Ready)
 }
--- a/test/clientconn_state_transition_test.go
+++ b/test/clientconn_state_transition_test.go
@@ -640,15 +640,15 @@ func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
 	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
 	defer shortCancel()
 
-	// TODO: Change this assertion to testutils.AwaitState(shortCtx, t, cc,
-	// connectivity.Connecting) when the channel correctly transitions to
-	// CONNECTING while waiting for the first resolver update.
 	testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Idle)
 
 	internal.EnterIdleModeForTesting.(func(*grpc.ClientConn))(cc)
 
 	cc.Connect()
 
+	// TODO: Change this assertion to testutils.AwaitState(shortCtx, t, cc,
+	// connectivity.Connecting) when the channel correctly transitions to
+	// CONNECTING while waiting for the first resolver update.
 	testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Idle)
 
 	shortCancel()
--- a/test/clientconn_state_transition_test.go
+++ b/test/clientconn_state_transition_test.go
@@ -644,7 +644,7 @@ func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
 	testutils.AwaitState(shortCtx, t, cc, connectivity.Idle)
 
 	mr.UpdateState(resolver.State{
-		Addresses: []resolver.Address{{Addr: backend.Address}}, // Delivering the address
+		Addresses: []resolver.Address{{Addr: backend.Address}},
 	})
 
 	// Check that the channel transitions to CONNECTING then READY
--- a/test/clientconn_state_transition_test.go
+++ b/test/clientconn_state_transition_test.go
@@ -621,9 +621,8 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 // TestChannelStateWaitingForFirstResolverUpdate verifies the initial
 // state of the channel when a manual name resolver doesn't provide any updates.
 func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping test.")
-	}
+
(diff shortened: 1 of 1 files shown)


test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,52 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	// TODO: Change this assertion to testutils.AwaitState(shortCtx, t, cc,
+	// connectivity.Connecting) when the channel correctly transitions to
+	// CONNECTING while waiting for the first resolver update.
```
arjan-bal, line 645: This comment should be on top of the next `testutils.AwaitNoStateChange` call (line 652 presently) which is made after calling `cc.Connect()`.

arjan-bal:
Please merge the latest master. There was a fix for the failing vet test added recently.

test/clientconn_state_transition_test.go
janardhanvissa: Addressed the above comments  and pushed the changes. All the checks are passing now. Please re-review once.

arjan-bal (approved):

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,52 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Idle)
+
+	internal.EnterIdleModeForTesting.(func(*grpc.ClientConn))(cc)
+
+	cc.Connect()
+
+	// TODO: Change this assertion to testutils.AwaitState(shortCtx, t, cc,
```
dfawley, line 649: Please instead assert the proper behavior and use `t.Skip` instead with a message that refers to the issue.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,52 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	// connectivity.Connecting) when the channel correctly transitions to
+	// CONNECTING while waiting for the first resolver update.
+	testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Idle)
+
+	shortCancel()
+	shortCtx, shortCancel = context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	go func() {
+		_, err := testgrpc.NewTestServiceClient(cc).EmptyCall(shortCtx, &testgrpc.Empty{})
```
dfawley, line 659: Is it necessary to do this RPC? Why?

If so, you need to block until this goroutine completes before returning from the outer Test...() function.  But I don't think it should be needed.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,52 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+		if err == nil {
+			t.Errorf("Expected RPC to fail, but it succeeded")
+		}
+	}()
+
+	// TODO: Change this assertion to testutils.AwaitState(shortCtx, t, cc,
```
dfawley, line 665: I'm not sure what this part is attempting to test.  It should already be CONNECTING per the above assertion.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,52 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 		}
 	}
 }
+
+// TestChannelStateWaitingForFirstResolverUpdate verifies the initial
+// state of the channel when a manual name resolver doesn't provide any updates.
+func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
+	backend := stubserver.StartTestService(t, nil)
+	defer backend.Stop()
```
dfawley, line 626: Nothing uses backend -- delete it or eventually deliver the address and expect READY.

janardhanvissa:
Addressed above mentioned comments by using t.Skip instead with a message and channel and goroutine

test/clientconn_state_transition_test.go
janardhanvissa: Used `t.Skip` with a message.

test/clientconn_state_transition_test.go
janardhanvissa: Delivered the address as READY

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,58 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	t.Skipf("After connecting, before READY state: current state = %v\n", cc.GetState())
+
+	testutils.AwaitState(shortCtx, t, cc, connectivity.Ready)
+	t.Skipf("After entering READY state: current state = %v\n", cc.GetState())
+
+	mr.UpdateState(resolver.State{Addresses: nil})
+
+	done := make(chan struct{})
+	go func() {
+		_, err := testgrpc.NewTestServiceClient(cc).EmptyCall(shortCtx, &testgrpc.Empty{})
```
dfawley, line 661: This shouldn't use the `shortCtx` - we only use that for operations we expect to not happen.  We want this one to succeed.

Actually...literally everywhere you use `shortCtx` is wrong.  Use `ctx` instead.

But, we should use `shortCtx` in one check: before we produce the proper address from the resolver, and after we've asserted that we entered CONNECTING, we should do `testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Connecting)`

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,58 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	mr.UpdateState(resolver.State{
+		Addresses: []resolver.Address{{Addr: backend.Address}},
+	})
+
+	// Check that the channel transitions to CONNECTING then READY
+	testutils.AwaitState(shortCtx, t, cc, connectivity.Connecting)
+	t.Skipf("After connecting, before READY state: current state = %v\n", cc.GetState())
+
+	testutils.AwaitState(shortCtx, t, cc, connectivity.Ready)
+	t.Skipf("After entering READY state: current state = %v\n", cc.GetState())
```
dfawley, line 655: `Skip` skips the rest of the test, so just put one at the top of the test function and assert the proper behavior here instead.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,58 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	defer cc.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+
+	cc.Connect()
+	testutils.AwaitState(shortCtx, t, cc, connectivity.Idle)
```
dfawley, line 644: Why are you waiting for IDLE _after_ connecting?  And...ideally upon returning from `cc.Connect()`, we should _already_ be CONNECTING, so this should never pass.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,58 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+
+	mr.UpdateState(resolver.State{
+		Addresses: []resolver.Address{{Addr: backend.Address}},
+	})
+
+	// Check that the channel transitions to CONNECTING then READY
+	testutils.AwaitState(shortCtx, t, cc, connectivity.Connecting)
```
dfawley, line 651: This needs to be _before_ the `UpdateState`.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,58 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	t.Skipf("After connecting, before READY state: current state = %v\n", cc.GetState())
+
+	testutils.AwaitState(shortCtx, t, cc, connectivity.Ready)
+	t.Skipf("After entering READY state: current state = %v\n", cc.GetState())
+
+	mr.UpdateState(resolver.State{Addresses: nil})
```
dfawley, line 657: What's this about?

test/clientconn_state_transition_test.go
janardhanvissa: Moved to before the `UpdateState`

test/clientconn_state_transition_test.go
janardhanvissa: cc.Connect() moved to after the connectivity.Idle

test/clientconn_state_transition_test.go
janardhanvissa: used ctx instead of shortCtx

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,38 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 		}
 	}
 }
+
+// TestChannelStateWaitingForFirstResolverUpdate verifies the initial
+// state of the channel when a manual name resolver doesn't provide any updates.
+func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
+	if testing.Short() {
```
dfawley, line 624: What is this and why?

I think we have a bug, which this test doesn't actually appear to be testing, but which is described in the issue that asked for this test.  IF there is a bug, we need to always skip the test.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,38 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+		t.Fatalf("Failed to create new client: %v", err)
+	}
+	defer cc.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	cc.Connect()
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Idle)
```
dfawley, line 645: This has to be before Connect or else it will race..

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,38 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+
+	mr.UpdateState(resolver.State{
+		Addresses: []resolver.Address{{Addr: backend.Address}},
+	})
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
```
dfawley, line 651: This is supposed to be immediately after Connect(), and if we have the bug I am thinking we have, then you should see this fail when you run the test.

test/clientconn_state_transition_test.go
dfawley: I don't see the AwaitNoStateChange.

test/clientconn_state_transition_test.go
janardhanvissa: Yes, we have a bug that's why we added this test to skip.

test/clientconn_state_transition_test.go
janardhanvissa: `connectivity.Idle` moved up before `cc.Connect()`

purnesh42H:
@janardhanvissa let's re-review it in BLR once since there were many changes after first commit. Closing it for now

test/clientconn_state_transition_test.go
arjan-bal: @janardhanvissa you can skip the test unconditionally, i.e. remote the `if testing.Short()`. Also add more details in the skip message.
```go
t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as Connect() is called.")
```

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,32 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 		}
 	}
 }
+
+// TestChannelStateWaitingForFirstResolverUpdate verifies the initial
+// state of the channel when a manual name resolver doesn't provide any updates.
+func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
+
```
arjan-bal, line 624: nit: Please remove blank line at the beginning of the method.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,32 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Idle)
+
+	cc.Connect()
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
+
```
arjan-bal, line 647: nit: Remove blank line at the end of the method.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,32 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create new client: %v", err)
+	}
+	defer cc.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Idle)
+
```
arjan-bal, line 643: You can add a comment saying that "The channel should transition to CONNECTING automatically when Connect() is called". Then remove the blank like after `cc.Connect()` since both the lines are part of the same logical code block.
```go
// The channel should transition to CONNECTING automatically when Connect() is called.
cc.Connect()
testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
```

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +617,32 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	cc.Connect()
+
+	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
```
arjan-bal, line 646: Based on Doug's earlier comment, we can also add a check that the channel remains in CONNECTING.
```go
testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Connecting)
```

arjan-bal (commented):
Can you add another test case similar to the existing one, but that makes an RPC instead of calling `cc.Connect()`?

test/clientconn_state_transition_test.go
janardhanvissa: removed blank line

test/clientconn_state_transition_test.go
janardhanvissa: removed blank line

test/clientconn_state_transition_test.go
janardhanvissa: added comment

test/clientconn_state_transition_test.go
janardhanvissa: Added a check of AwaitNoStateChange

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,74 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 		}
 	}
 }
+
+// TestChannelStateWaitingForFirstResolverUpdate verifies the initial
+// state of the channel when a manual name resolver doesn't provide any updates.
+func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
+	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as Connect() is called.")
```
arjan-bal, line 625: Doug mentioned that we should  reference the Github issue in the skip message. Can you please add the line "See issue #7686." as the end?

```go
t.Skip("... soon as Connect() is called. See issue #7686.")
```
Same change for the other test case also.

arjan-bal (approved):
Couple of minor comments, otherwise LGTM!

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,74 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+
+	// Make an RPC call to transition the channel to CONNECTING.
+	go func() {
+		_, err := testgrpc.NewTestServiceClient(cc).EmptyCall(ctx, &testgrpc.Empty{})
+		if err == nil {
+			t.Errorf("Expected RPC to fail, but it succeeded")
+		}
+	}()
+
+	// The channel should transition to CONNECTING automatically when Connect()
```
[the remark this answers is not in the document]
janardhanvissa: done with the change

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,74 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
...
+	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
+
+	// Verify that the channel remains in CONNECTING state for a short time.
+	shortCtx, shortCancel := context.WithTimeout(ctx, defaultTestShortTimeout)
+	defer shortCancel()
+	testutils.AwaitNoStateChange(shortCtx, t, cc, connectivity.Connecting)
+}
+
+func (s) TestChannelStateTransitionWithRPC(t *testing.T) {
+	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as an RPC call is made soon as Connect() is called. See issue #7686.")
```
dfawley, line 656: Copy/paste issues here too

test/clientconn_state_transition_test.go
dfawley: This should go above the context creation and just be an assertion that `State() == IDLE`.  No need to `Await` since it _should_ always be in this state upon creation.

dfawley (commented):
Thanks!  Looks better, just a couple minor things.

test/clientconn_state_transition_test.go
```diff
@@ -616,3 +618,74 @@ func (s) TestConnectivityStateSubscriber(t *testing.T) {
 		}
 	}
 }
+
+// TestChannelStateWaitingForFirstResolverUpdate verifies the initial
+// state of the channel when a manual name resolver doesn't provide any updates.
+func (s) TestChannelStateWaitingForFirstResolverUpdate(t *testing.T) {
+	t.Skip("The channel remains in IDLE until the LB policy updates the state to CONNECTING. This is a bug and the channel should transition to CONNECTING as soon as Connect() is called soon as Connect() is called. See issue #7686.")
```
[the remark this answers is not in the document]
janardhanvissa: Updated the comment.

test/clientconn_state_transition_test.go
janardhanvissa: Updated the comment here as well.

janardhanvissa:
Replaced the initial AwaitState() call with a direct GetState() check.

dfawley (approved):
Thanks!

PR #7778, opened 2024-10-24
proposal: add cause to canceled context when Client sending a RST_STREAM due to an error on their end

### LINKED ISSUE #7541: Add cancel cause when client sends RST_STREAM frame

When a client resets a connection, the http2_server cancels the current context to interrupt all "server internal processing". Relevant code is [`t.closeStream(s, false, 0, false)`](https://github.com/grpc/grpc-go/blob/6a5a283b699dd16c8caa240d209d7302988ff753/internal/transport/http2_server.go#L832) and [`s.cancel()`](https://github.com/grpc/grpc-go/blob/6a5a283b699dd16c8caa240d209d7302988ff753/internal/transport/http2_server.go#L1315). 

When there is a big context stack within the server's logic, it can be quite hard to find out why the context was canceled. 
Was it due to some internal error? Timeout? Or is it due to a client sending the RST frame? 

A real-world problem analysis is described in https://github.com/grafana/tempo/issues/3957. 

### Proposed Solution

I am proposing to replace [`context.CancelFunc`](https://github.com/grpc/grpc-go/blob/6a5a283b699dd16c8caa240d209d7302988ff753/internal/transport/transport.go#L246C2-L246C33) with `context.CancelCauseFunc` and adjust all context cancellations to  report the cause. 

For the example described above, it could be something like this

```go
func (t *http2Server) handleRSTStream(f *http2.RSTStreamFrame) {
	// If the stream is not deleted from the transport's active streams map, then do a regular close stream.
	if s, ok := t.getStream(f); ok {
		t.closeStream(s, false, 0, false, errors.New("RST_STREAM frame received"))
		return
	}

// ... 

func (t *http2Server) closeStream(s *Stream, rst bool, rstCode http2.ErrCode, eosReceived bool, cause error) {
	// In case stream sending and receiving are invoked in separate
	// goroutines (e.g., bi-directional streaming), cancel needs to be
	// called to interrupt the potential blocking on other goroutines.
	s.cancel(cause)
```

### Alternatives Considered

`nil`

### Additional Context

When a pattern is established to communicate the cause of context cancellation, there are a few other cases, in addition to the one described above, where it could be useful:

1. **Client closes connection**:
   When the client closes a connection before the server can process all requests (asynchronously), the server's context gets canceled. It is up to the server implementor to decide how to handle that case. Should it rollback a transaction? Continue execution until done? To achieve that, the connection closed (by client) should be set as the cancel cause in [http2_server.go#L1272](https://github.com/dastrobu/grpc-go/blob/cancel-cause/internal/transport/http2_server.go#L1272).

2. **Client sets timeout via the `grpc-timeout` header**:
   When the client sets a timeout, the server should be able to distinguish between a gRPC timeout and other timeouts set elsewhere. To achieve that, `context.WithTimeout(ctx, timeout)` should be replaced by something like `context.WithTimeoutCause(ctx, timeout, GrpcTimeoutErr)` in [context.go#L29](https://github.com/dastrobu/grpc-go/blob/cancel-cause/internal/transport/context.go#L29).

3. **Other situations**:
   Even for situations that should rarely happen, like `Received a HEADERS frame with :method %q which should be POST` in [http2_server.go#L564](https://github.com/dastrobu/grpc-go/blob/cancel-cause/internal/transport/http2_server.go#L564), it could be beneficial to communicate a cause.

In essence, each case that leads to a `s.cancel()` call informs the calling client about the reason (e.g., via an HTTP status) but does not inform the server's goroutines working on the stream about the reason. In the long run, the goroutines should get as much information as possible about the actions the calling client took (e.g., closing the connection, setting a timeout, sending an RST frame) to react accordingly or if gRPC canceled the context (e.g., receiving bad headers).

The purpose is to allow the server's goroutines to implement logic to react to the client's actions appropriately and, as an additional benefit, make debugging easier, and allow the implementation of metrics and alerting which can distinguish a badly configured server (timeouts, etc.) from a badly configured client (closing connections, etc.).

dastrobu:
**Related Issue:** [#7541](https://github.com/grpc/grpc-go/issues/7541)

**Summary:**
This PR introduces a feature to add a cause to the context cancellation, for example when a client sends an RST_STREAM frame. This enhancement aims to provide more transparency and ease debugging by distinguishing between different reasons for context cancellation.

**Use Case:**
When a client resets a connection, the `http2_server` cancels the current context to interrupt all server internal processing. However, it is challenging to determine the exact reason for the context cancellation, whether it was due to an internal error, a timeout, or a client sending an RST frame. This feature will help in identifying the cause of the context cancellation, making it easier to debug and handle errors appropriately.

**Proposed Solution:**
- Replace `context.CancelFunc` with `context.CancelCauseFunc`.
- Adjust all context cancellations to report the cause.
- For example, when handling an RST_STREAM frame, the context will be canceled with a specific error indicating the receipt of the RST_STREAM frame.

**Additional Context:**
This change is backward-compatible and does not affect existing application code that checks for `ctx.Err()`. The new `context.Cause(ctx)` method will provide additional information when available.

**Example:**

The PR adds an example which shows how to implement a server with asynchronous processing using goroutines. 
The example illustrates some client behaviors (RST frame, close connection, timeout) which are propagated to the worker goroutine through the context. 

Here are the logs of the example server (with comments):

```
starting work on message: "world"
starting work on message: "hello"
'context canceled' with cause 'NO_ERROR', message worker for message "world" stopping. 👈🏻 NO_ERROR indicates the context was caused by HTTP status 0
server: error receiving from stream: rpc error: code = Canceled desc = context canceled 
'context canceled' with cause 'NO_ERROR', message worker for message "hello" stopping. 👈🏻 see above
starting work on message: "world"
starting work on message: "hello"
'context canceled' with cause 'context canceled', message worker for message "hello" stopping. 👈🏻 this case could be made more transparent when setting a cancel cause for a closed connection as well. 
'context canceled' with cause 'context canceled', message worker for message "world" stopping. 👈🏻 see above
server: error receiving from stream: rpc error: code = Canceled desc = context canceled
starting work on message: "world"
starting work on message: "hello"
'context deadline exceeded' with cause 'context deadline exceeded', message worker for message "world" stopping. 👈🏻 this case could be made more transparent by setting a custom cause for the context deadline in `ctx, timoutCancel = context.WithTimeout(ctx, timeout)`
server: error receiving from stream: rpc error: code = DeadlineExceeded desc = context deadline exceeded
'context deadline exceeded' with cause 'context deadline exceeded', message worker for message "hello" stopping. 👈🏻 see above
```

**Open Issues**

 * This PR is still a draft and almost no unit tests have been added to production code. This can be done when the reviewers agree that the feature of this PR will be considered to be added to the library. 
 * Some uses cases (client closes connection, timeout) do not receive a proper cause in this PR, as it was explicitly requested to set the cause only for the RST frame case yet. 
 * So far one error type (`Http2CodeError`) has been added to illustrate how server implementations could do error matching. The design for error types could be discussed in more detail. I'd see more error cases in the future that could be propagated through the context (e.g. `ConnectionClosed` by client, `GrpcTimeout` to distinguish from other timeouts).

### DIFF
--- a/examples/features/context/README.md
+++ b/examples/features/context/README.md
@@ -1,0 +1,12 @@
+# Context
+
+This example shows how servers can process requests in separate goroutines and
+handle context cancellation.
+
+```
+go run server/main.go
+```
+
+```
+go run client/main.go
+```
--- a/examples/features/context/client/main.go
+++ b/examples/features/context/client/main.go
@@ -1,0 +1,161 @@
+/*
+ *
+ * Copyright 2024 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Binary client demonstrates how to cancel in-flight RPCs by canceling the
+// context passed to the RPC.
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"time"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
+	pb "google.golang.org/grpc/examples/features/proto/echo"
+	"google.golang.org/grpc/status"
+)
+
+var addr = flag.String("addr", "localhost:50051", "the address to connect to")
+
+func sendMessage(stream pb.Echo_BidirectionalStreamingEchoClient, msg string) error {
+	fmt.Printf("sending message %q\n", msg)
+	return stream.Send(&pb.EchoRequest{Message: msg})
+}
+
+func recvMessage(stream pb.Echo_BidirectionalStreamingEchoClient, wantErrCode codes.Code) {
+	res, err := stream.Recv()
+	if status.Code(err) != wantErrCode {
+		log.Fatalf("stream.Recv() = %v, %v; want _, status.Code(err)=%v", res, err, wantErrCode)
+	}
+	if err != nil {
+		fmt.Printf("stream.Recv() returned expected error %v\n", err)
+		return
+	}
+	fmt.Printf("received message %q\n", res.GetMessage())
+}
+
+func cancelStream() {
+	fmt.Println("sending two messages and then canceling")
+	conn, err := grpc.NewClient(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatalf("did not connect: %v", err)
+	}
+	defer conn.Close()
+
+	c := pb.NewEchoClient(conn)
+
+	// Initiate the stream with a context that supports cancellation.
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	stream, err := c.BidirectionalStreamingEcho(ctx)
+	if err != nil {
+		log.Fatalf("error creating stream: %v", err)
+	}
+
+	// Send some test messages.
+	if err := sendMessage(stream, "hello"); err != nil {
+		log.Fatalf("error sending on stream: %v", err)
+	}
+	if err := sendMessage(stream, "world"); err != nil {
+		log.Fatalf("error sending on stream: %v", err)
+	}
+
+	// Ensure the RPC is working.
+	recvMessage(stream, codes.OK)
+	recvMessage(stream, codes.OK)
+
+	fmt.Println("canceling context")
+	cancel()
+}
+
+func closeConnection() {
+	fmt.Println("sending two messages and then closing connection")
+	conn, err := grpc.NewClient(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatalf("did not connect: %v", err)
+	}
+	defer func() {
+		fmt.Println("closing connection")
+		conn.Close()
+	}()
+
+	c := pb.NewEchoClient(conn)
+
+	stream, err := c.BidirectionalStreamingEcho(context.Background())
+	if err != nil {
+		log.Fatalf("error creating stream: %v", err)
+	}
+
+	// Send some test messages.
+	if err := sendMessage(stream, "hello"); err != nil {
+		log.Fatalf("error sending on stream: %v", err)
+	}
+	if err := sendMessage(stream, "world"); err != nil {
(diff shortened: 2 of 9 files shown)


purnesh42H:
@dastrobu thanks for the PR. Apologies for getting late to it. These are too many errors to verify at once. Could you just narrow down this PR to one case? Perhaps only to RstStream code from server? Also, please describe in the PR (only for RST case) what is the limitation and how this change improves that. 

Also, not sure if we should have a separate function to create context. We need to have this change easy to just do natively. Like adding the cancelCause only when we need to and it should be left on the author of code to decide. Let me know what you think

purnesh42H:
@dastrobu let's narrow down this PR to only "Client sending a RST_STREAM due to an error on their end". We want to see how much change it is and if it has enough benefit to reason about. Thanks

purnesh42H:
@dastrobu ping on this if you have bandwidth for making the the change. Thanks

dastrobu:
@purnesh42H I will update the PR draft later this week.

dastrobu:
> Could you just narrow down this PR to one case? Perhaps only to RstStream code from server?

I narrowed it down and added an example. 

>  Also, please describe in the PR (only for RST case) what is the limitation and how this change improves that.

PR description was updated. 

> Also, not sure if we should have a separate function to create context. We need to have this change easy to just do natively. Like adding the cancelCause only when we need to and it should be left on the author of code to decide. Let me know what you think.

I extracted the function here to dry out code and make the logic it testable through unit tests (see `context_test.go`). 
The reason we cannot use any context in the two cases is that the cancel function is part of the `ServerStream` type. 

Hence, the cancel function must always be of the same type. 
In most go functions the context is cancelled at the end of the function that creates it (through a `defer` call). This library, however, needs to pass the cancel function around to call it in appropriate places.  

We could inline the `createContextWithTimeout` function, but for the reasons explained above this would lead to code duplication and make it harder to test. I am open for alternative approaches, if you have good suggestion.

> If you want, in the description of the related issue we can link the error cases you have identified

I am not sure I understand what error cases you mean here?

dfawley:
Can I request that we first enumerate all the reasons the grpc server might cancel the context?

I can think of:

1. Unexpected connection loss or a more graceful connection closure (preceded by a GOAWAY)
2. Client-side cancelation (which is transmitted via a RST_STREAM)
3. Deadline exceeded

I'm not sure about "internal error" mentioned in the first comment. Can you give an example of this?

After that, what would knowing the difference between these cases affect?  Is this purely for monitoring/alerting/metrics?

dastrobu:
> Can I request that we first enumerate all the reasons the grpc server might cancel the context?
> 
> I can think of:
> 
> 1. Unexpected connection loss or a more graceful connection closure (preceded by a GOAWAY)
> 2. Client-side cancelation (which is transmitted via a RST_STREAM)
> 3. Deadline exceeded
> 
> I'm not sure about "internal error" mentioned in the first comment. Can you give an example of this?
> 
> After that, what would knowing the difference between these cases affect? Is this purely for monitoring/alerting/metrics?

@dfawley thanks for the prompt reply. I will pick that up tomorrow and try to write up the cases in #7541

PR #7781, opened 2024-10-25
server: Only call FromIncomingContext with stats handlers

evanj:
Skip calling metadata.FromIncomingContext if there are no stats hnadlers. This function is relatively expensive because it makes a copy of the map of gRPC headers. For our production services that use tracing libraries, this means there are typically about ~10 headers on every call. For services with cheap gRPC calls, this can use nearly 1% of CPU.

There are unit tests that cover this code path in the stats package but sadly I did not find any existing microbenchmarks that we could easily use to verify this improvement.

RELEASE NOTES: none

### DIFF
--- a/server.go
+++ b/server.go
@@ -1783,17 +1783,20 @@ func (s *Server) handleStream(t transport.ServerTransport, stream *transport.Str
 	service := sm[:pos]
 	method := sm[pos+1:]
 
-	md, _ := metadata.FromIncomingContext(ctx)
-	for _, sh := range s.opts.statsHandlers {
-		ctx = sh.TagRPC(ctx, &stats.RPCTagInfo{FullMethodName: stream.Method()})
-		sh.HandleRPC(ctx, &stats.InHeader{
-			FullMethod:  stream.Method(),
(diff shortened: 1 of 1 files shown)


evanj:
For some additional details, here is a screenshot from Datadog's continuous profiler of a fairly high RPS production server with no stats handlers, showing that this call to FromIncomingMetadata takes about 0.8% of all CPU. It also is responsible for about 1.2% of all allocated memory.

Screenshot 2024-10-25 at 3 38 18 PM

aranjans:
Thanks @evanj for this change with detailed analysis. Overall LGTM, adding @easwars as a second reviewer.

easwars:
@zasweq : Could you please take a quick look as well. Looks safe to me. But since you authored the code being changed, it makes sense for you to have a look. Thanks.

zasweq (approved):

zasweq:
You might need to rebase onto the most recent upstream master and then force push in order to trigger 1.23.

zasweq:
Merged, thank you for the contribution.

PR #7782, opened 2024-10-25
balancer/endpointsharding: Call ExitIdle() on child if child reports IDLE

zasweq:
This PR adds the functionality of calling ExitIdle on any child that reports IDLE, to automatically retrigger the child to try and start connecting.

In the future, we could make this configurable through a config (which we will also need to add), but for all the use cases of this endpoint sharding policy the desired behavior is to trigger ExitIdle on the child so we can leave default for now.

This is because petiole policies used to be able to call Connect() on SubConns, but now that pick_first is the universal leaf, that will be the only balancer that will contain the logic for SubConn management, so petioles have no way of triggering this reconnection, as previously present in the system.

RELEASE NOTES:
* balancer/endpointsharding: Call ExitIdle() on child if child reports IDLE

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -65,6 +65,7 @@ type endpointSharding struct {
 	cc    balancer.ClientConn
 	bOpts balancer.BuildOptions
 
+	childMu  sync.Mutex // syncs balancer.Balancer calls into children
 	children atomic.Pointer[resolver.EndpointMap]
 
 	// inhibitChildUpdates is set during UpdateClientConnState/ResolverError
@@ -114,6 +115,7 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 			bal.Balancer = gracefulswitch.NewBalancer(bal, es.bOpts)
 		}
 		newChildren.Set(endpoint, bal)
+		es.childMu.Lock()
 		if err := bal.UpdateClientConnState(balancer.ClientConnState{
 			BalancerConfig: state.BalancerConfig,
 			ResolverState: resolver.State{
@@ -127,13 +129,16 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 			// validations, this is a current limitation for simplicity sake.
 			ret = err
 		}
+		es.childMu.Unlock()
 	}
 	// Delete old children that are no longer present.
 	for _, e := range children.Keys() {
 		child, _ := children.Get(e)
 		bal := child.(balancer.Balancer)
 		if _, ok := newChildren.Get(e); !ok {
+			es.childMu.Lock()
 			bal.Close()
+			es.childMu.Unlock()
 		}
 	}
 	es.children.Store(newChildren)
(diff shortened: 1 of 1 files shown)


balancer/endpointsharding/endpointsharding.go
```diff
@@ -123,6 +124,7 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 			bal.Balancer = gracefulswitch.NewBalancer(bal, es.bOpts)
 		}
 		newChildren.Set(endpoint, bal)
+		es.childMu.Lock()
```
dfawley, line 127: I think we can safely take this once at the stop and defer the unlock.  The only things that happen concurrently are:

This & Close (guaranteed to be non-concurrent) & the ExitIdle goroutine, which can be concurrent but it should be safe for that to block all of UpdateCCS.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -282,6 +291,15 @@ func (bw *balancerWrapper) UpdateState(state balancer.State) {
 	bw.es.mu.Lock()
 	bw.childState.State = state
 	bw.es.mu.Unlock()
+	// When a child balancer says it's IDLE, ping it to exit idle and reconnect.
+	// TODO: In the future, perhaps make this a knob in configuration.
+	if ei, ok := bw.Balancer.(balancer.ExitIdler); state.ConnectivityState == connectivity.Idle && ok {
+		go func() {
+			bw.es.childMu.Lock()
+			ei.ExitIdle()
```
dfawley, line 299: Don't we need to avoid this racing with Close?  Not worried about a deadlock or concurrency, but this sequence:

- Child reports IDLE
- This goroutine gets scheduled
- Parent calls `Close()`
- This goroutine runs and tells the child to `ExitIdle()` after `Close()`

balancer/endpointsharding/endpointsharding.go
dfawley: Oh, same thing can apply to `Close`, too, I'm pretty sure.

zasweq: ResolverError can also race with ExitIdle (all three balancer.Balancer operations downward can race). Grabbed mutex for the whole operation and added close guard.

balancer/endpointsharding/endpointsharding.go
zasweq: Added close guard.

balancer/endpointsharding/endpointsharding.go
zasweq: And I do defer an updateState() call which will now update upward with mutex held but since you mentioned can assume can't call back inline I'll go ahead and add it.

balancer/endpointsharding/endpointsharding.go
zasweq: Good catch. I didn't even think of this constraint of the balancer.Balancer API of nothing called after Close(). So I guess ExitIdle even though a separate interface is the part of same API.

I did an grpcsync.Event, but let me know if you would rather have a bool.

dfawley (approved):

PR #7792, opened 2024-10-29
Cherry-pick #7702 to 1.68.x

purnesh42H:
Original PR (#7702)

RELEASE NOTES:

- mem: use slice capacity instead of length, to determine whether to pool buffers or directly allocate them


arvindbr8:
vet failed (non-blocking, but worth checking) - 

```
/opt/hostedtoolcache/go/1.23.2/x64/src/runtime/time.go:204:17: cannot range over 3 (untyped int constant): requires go1.22 or later (compile)
```

Weird that we don't see this in every run

arvindbr8 (approved):

Issue #7794, opened 2024-10-29
Access to Raw Serialized Data (Raw Data) in Latest gRPC Version Without Marshaling - Previous Workaround No Longer Available


mohdjishin:
In prior gRPC versions, the stats handler provided direct access to raw serialized data through payload.Data in `*stats.InPayload`. This functionality enabled certain applications to retrieve unmodified byte data directly. While a workaround was previously available — leveraging `TagRPC` to add values to the context and `HandleRPC` to access raw bytes for `*stats.InPayload `— this approach is no longer possible in the latest version.

Issue: With this removal, it is challenging to obtain raw request bytes directly without additional marshaling. The previous workaround, as discussed in @dfawley  comment on May 17, 2023 [Workaround](https://github.com/grpc/grpc-go/issues/6251#issuecomment-1550432045), no longer works in the latest gRPC version.

The removal of payload.Data occurred in this [commit](https://github.com/grpc/grpc-go/commit/9ab8b62505c85a5e03be1ae97949e960e24b6d4f#diff-366e46a40f6f60b4f7614eb0976bb51820364bf5ca6ccc4787eb49d7bdbef3e6L7)
![Screenshot from 2024-10-30 02-26-02](https://github.com/user-attachments/assets/44b1ea6a-4a89-43b1-a40f-e6f98e9b65b9)

dfawley:
Apologies for the breakage!  This field needed to be removed for performance reasons as we no longer create a full contiguous copy of the serialized data (except within the default proto codec, unfortunately, but if they ever improve the proto library to take advantage of split input buffers, we would start using it _ASAP_).

It's possible we could add the data back as a `mem.BufferSlice` instead of a flat buffer.  However, another reason we thought it was a good idea to remove it was because it passes potentially sensitive user data to something whose purpose was supposed to be basic telemetry, and that can make it easier to create privacy problems.

Let me talk with the other gRPC leads (for Java/C++) and see what kind of options they might offer for something like this.  It could be the case that it's just not possible without doing things more manually -- i.e. not using the generated code to register your method handlers, and have a custom codec that just can pass the bytes through instead.

mohdjishin:
Thank you, @dfawley , for the insights and explanation regarding the changes. The removal of payload.Data does impact applications like digital signature verification is essential. In our scenario, we sign the serialized raw data, and upon receiving it, we verify the signature against the exact bytes to ensure the data’s authenticity and integrity before processing it.

[...]

PR #7798, opened 2024-11-01
server: add embedding requirement to ServerTransportStream

dfawley:
`grpc.ServerTransportStream` is currently not extensible, because it does not enforce that a delegate implementation is embedded.  This change creates that requirement.  I noticed this when I was investigating the possibility of using this kind of intercepting (i.e. extending it to support messages) to suggest a way to accomplish #7794.  I'm not sure if that's appropriate, but something like this generally seems like a good idea to do regardless.

cc @jhump - this is potentially a breaking change to an (experimental) API you created. Does this seem reasonable to you, or actually break anything of yours?

RELEASE NOTES:
- server: add embedding requirement to ServerTransportStream (experimental)

### DIFF
--- a/internal/transport/handler_server.go
+++ b/internal/transport/handler_server.go
@@ -333,7 +333,7 @@ func (ht *serverHandlerTransport) writeCustomHeaders(s *Stream) {
 	s.hdrMu.Unlock()
 }
 
-func (ht *serverHandlerTransport) Write(s *Stream, hdr []byte, data mem.BufferSlice, _ *WriteOptions) error {
+func (ht *serverHandlerTransport) Write(s *Stream, hdr []byte, data mem.BufferSlice, _ *Options) error {
 	// Always take a reference because otherwise there is no guarantee the data will
 	// be available after this function returns. This is what callers to Write
 	// expect.
--- a/internal/transport/handler_server.go
+++ b/internal/transport/handler_server.go
@@ -333,7 +333,7 @@ func (ht *serverHandlerTransport) writeCustomHeaders(s *Stream) {
 	s.hdrMu.Unlock()
 }
 
-func (ht *serverHandlerTransport) Write(s *Stream, hdr []byte, data mem.BufferSlice, _ *Options) error {
+func (ht *serverHandlerTransport) Write(s *Stream, hdr []byte, data mem.BufferSlice, _ *WriteOptions) error {
 	// Always take a reference because otherwise there is no guarantee the data will
 	// be available after this function returns. This is what callers to Write
 	// expect.
--- a/internal/transport/handler_server_test.go
+++ b/internal/transport/handler_server_test.go
@@ -440,7 +440,7 @@ func (s) TestHandlerTransport_HandleStreams_WriteStatusWrite(t *testing.T) {
 		st.bodyw.Close() // no body
 
 		st.ht.WriteStatus(s, status.New(codes.OK, ""))
-		st.ht.Write(s, []byte("hdr"), newBufferSlice([]byte("data")), &Options{})
+		st.ht.Write(s, []byte("hdr"), newBufferSlice([]byte("data")), &WriteOptions{})
 	})
 }
 
--- a/internal/transport/handler_server_test.go
+++ b/internal/transport/handler_server_test.go
@@ -440,7 +440,7 @@ func (s) TestHandlerTransport_HandleStreams_WriteStatusWrite(t *testing.T) {
 		st.bodyw.Close() // no body
 
 		st.ht.WriteStatus(s, status.New(codes.OK, ""))
-		st.ht.Write(s, []byte("hdr"), newBufferSlice([]byte("data")), &WriteOptions{})
+		st.ht.Write(s, []byte("hdr"), newBufferSlice([]byte("data")), &Options{})
 	})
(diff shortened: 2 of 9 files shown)


jhump:
@dfawley, I'm having trouble understanding this. This experimental API was used to provide a _separate_ implementation of a transport, of `grpc.ClientConnInterface`. To that effect, it will not have any `*grpc.Stream` instance that it can use as a delegate. I suspect this totally breaks that use case (but I can checkout [our impls](https://github.com/fullstorydev/grpchan) and try out this branch). Could custom transports embed a nil `*Stream` or would that guarantee a nil-derference panic? Is the idea here to make newer methods available on `ServerTransportStream` via embedding the `Stream`?

jhump:
Yeah, this seems to totally defeat the point of the experimental API and definitely breaks those transports. I see that it's not even possible to embed a nil `Stream` since the type that implements the necessary interface is not `grpc.Stream` but rather a type inside of an internal package 😦

dfawley:
@jhump thanks for looking.  For this kind of testing, I believe you could just embed the interface itself to satisfy the requirement:

```go
type myTestTransportStream struct {
	grpc.ServerTransportStream
	// etc
}

I thought the _primary_ purpose of this interface was to intercept the `SetHeader`/`SendHeader`/etc operations in unary server interceptors, which otherwise can't be intercepted, because our API requires method handlers to call `grpc.SetHeader(ctx)`/etc.

**Comment by jhump** [comment]
@dfawley, thanks for the reply. I did think of that and got things working:
https://github.com/fullstorydev/grpchan/pull/71

I'm still not clear on what the purpose of embedding is. This seems like a nasty nil-dereference risk in this library now, if there's anything else inside of grpc-go that might get a hold of our implementation and try to type-assert it.

**Comment by dfawley** [comment]
Per the description, the intent was to enable us to extend this interface in a backward compatible way.  Otherwise, adding new operations will break anyone just trying to intercept the existing set of operations.  So the expected use case for that "normal"(?) usage would be:

```go
func myInterceptor(ctx, ...) ... {
	curSTS := grpc.ServerTransportStreamFromContext(ctx)
	mySTS := &mySTSImpl{ServerTransportStream: curSTS, ...} // Overrides the functions it wants to intercept.
	ctx = grpc.NewContextWithServerTransportStream(ctx, mySTS)
	methodHandler(ctx, ...)
}
```

But since you have things that are implementing it from scratch, you're right that if we were to do that and you were to embed this interface and leave it at nil, then you would be broken at runtime instead of at compile time, which would be worse.

We _could_ instead create an optional interface that can be implemented if you want to intercept new things.  But I think that's also wrong.  E.g. `SetSendCompressor` and `ClientSupportedCompressors` are both actually broken right now in your current use case -- they fail at runtime, because the `ServerTransportStream` you put in the context is not a `*transport.Stream`.  That's not great, either.

Maybe we should do this instead:

```
type UnstableServerTransportStream interface {
	// All interceptable methods go here.
	// We reserve the right to add more and break you at compile time forever.
}

type ServerTrasnportStream interface {
	UnstableServerTransportStream
	mustEmbedDelegate()
}

func NewContextWithUnstableServerTransportStream(context.Context, UnstableServerTransportStream) ctx {}
```

You'd use the "unstable" symbols since you're implementing from scratch, and you'll learn when you upgrade that you are missing some methods.  That's probably better than your users finding out at runtime?

Thoughts?

dfawley:
@jhump - See the latest commit for a PoC of the above idea - it's basically exactly the same as what I wrote, just actually implemented with tests passing (I think?).

jhump:
I suppose I could make this work, but it still feels a little icky being broken each time a method is added.

What about this:
1. Go back to just a single `ServerTransportStream`, add the new methods there as well as the requirement to embed, so that new methods can safely be added to the interface in the future.
2. Provide a `NopServerTransportStream` concrete type that implements `ServerTransportStream`, where all methods have  a dummy/no-op body and return zero values. External implementations can embed this type to meet the new embed requirement, and will risk neither nil-dereference panic nor compile error when a new method is added in the future.

dfawley:
Yes, that sounds good with one extension: the way I noticed this and got here was that I was looking for some way to intercept _data_ operations at the transport level, too, as an attempt to make #7794 more feasible, and possibly good enough to consider #1805 done.  I really don't want to add a third (fourth?) interceptor API, and instead I think we could extend/replace this one to accomplish what we need.  But I still need to look a bit more closely at things in order to see how possible that would be.

jhump:
@dfawley, question about the minimum Go version being 1.22. I made it in a comment on linked PR, but I suspect you might not get a notification from a comment in that other repo. Here's the link to the thread: https://github.com/fullstorydev/grpchan/pull/71#discussion_r1828228310

dfawley:
I'm going to close this for now..it will take some extra consideration.  But I think the current state is problematic since there are some things that just don't work correctly when this is used (i.e. `SetSendCompressor` and `ClientSupportedCompressors`).

jhump:
@dfawley, please tag me in any future PRs when you get to an alternate solution for this, so I can prep necessary changes in downstream projects. And thank you so much for your consideration of those projects!

dfawley:
I most certainly will.  Thanks for your input @jhump!

PR #7816, opened 2024-11-06
xds: Plumb EDS endpoints through xDS Balancer Tree

zasweq:
This PR plumbs EDS endpoints through the xDS Balancer tree. This is done through the hierarchy package, which is how addresses were split to their respective children.

RELEASE NOTES: N/A

### DIFF
--- a/internal/hierarchy/hierarchy.go
+++ b/internal/hierarchy/hierarchy.go
@@ -48,18 +48,14 @@ func (p pathValue) Equal(o any) bool {
 	return true
 }
 
-// GetEndpoint returns the hierarchical path of endpoint.
-func GetEndpoint(endpoint resolver.Endpoint) []string {
-	attrs := endpoint.Attributes
-	if attrs == nil {
-		return nil
-	}
-	path, _ := attrs.Value(pathKey).(pathValue)
-	return ([]string)(path)
+// FromEndpoint returns the hierarchical path of endpoint.
+func FromEndpoint(endpoint resolver.Endpoint) []string {
+	path, _ := endpoint.Attributes.Value(pathKey).(pathValue)
+	return path
 }
 
-// SetEndpoint overrides the hierarchical path in endpoint with path.
-func SetEndpoint(endpoint resolver.Endpoint, path []string) resolver.Endpoint {
+// SetInEndpoint overrides the hierarchical path in endpoint with path.
+func SetInEndpoint(endpoint resolver.Endpoint, path []string) resolver.Endpoint {
 	endpoint.Attributes = endpoint.Attributes.WithValue(pathKey, pathValue(path))
 	return endpoint
 }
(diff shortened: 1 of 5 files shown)


internal/hierarchy/hierarchy.go
```diff
@@ -48,6 +48,22 @@ func (p pathValue) Equal(o any) bool {
 	return true
 }
 
+// GetEndpoint returns the hierarchical path of endpoint.
+func GetEndpoint(endpoint resolver.Endpoint) []string {
+	attrs := endpoint.Attributes
+	if attrs == nil {
```
dfawley, line 54: I think this function can just be:

```go
path, _ := endpoint.Attributes.Value(pathKey).(pathValue)
return path
```

internal/hierarchy/hierarchy.go
```diff
@@ -110,3 +126,19 @@ func Group(addrs []resolver.Address) map[string][]resolver.Address {
 	}
 	return ret
 }
+
+// GroupEndpoints groups endpoints as per Group above.
```
dfawley, line 130: Pet peeve: please don't use "as per" in our docs. :)  Just "as" or "per" is probably what you want instead.

Except in this case, we want to just copy/paste the whole thing and update as appropriate, because we'll be deleting the other one.

internal/hierarchy/hierarchy.go
```diff
@@ -48,6 +48,22 @@ func (p pathValue) Equal(o any) bool {
 	return true
 }
 
+// GetEndpoint returns the hierarchical path of endpoint.
+func GetEndpoint(endpoint resolver.Endpoint) []string {
```
dfawley, line 52: This name is incredibly misleading.  It's not getting an endpoint.  How about `FromEndpoint`?

internal/hierarchy/hierarchy.go
```diff
@@ -48,6 +48,22 @@ func (p pathValue) Equal(o any) bool {
...
+	attrs := endpoint.Attributes
+	if attrs == nil {
+		return nil
+	}
+	path, _ := attrs.Value(pathKey).(pathValue)
+	return ([]string)(path)
+}
+
+// SetEndpoint overrides the hierarchical path in endpoint with path.
+func SetEndpoint(endpoint resolver.Endpoint, path []string) resolver.Endpoint {
```
dfawley, line 62: `SetInEndpoint`?  (If you can think of something better, happy to hear it...)

xds/internal/balancer/clusterresolver/clusterresolver.go
```diff
@@ -252,7 +252,6 @@ func (b *clusterResolverBalancer) updateChildConfig() {
 	for i, a := range addrs {
 		endpoints[i].Attributes = a.BalancerAttributes
 		endpoints[i].Addresses = []resolver.Address{a}
-		endpoints[i].Addresses[0].BalancerAttributes = nil
```
dfawley, line 255: What was this here for before, and why is it removed?

zasweq: (I didn't write this code), but this was here before to clear the balancer attributes because we assumed this to be working.

This is needed for the way we do load reporting (temporary measure). A load report test was failing because the load reporting attribute wasn't stuck in the address. The old pf policy flattens the addresses from all the endpoints if endpoints are set and uses that (which it wasn't receiving prior to this change, so just uses addresses), but the issue is this was clearing the load reporting. Thus, the cluster impl policy wasn't looking at the top level endpoint load reporting, but this. This is all temporary code that will be phased out but that is the technical reason.

internal/hierarchy/hierarchy.go
zasweq: What's wrong with as per lol?

Rewrote comment to be like the one above except for discussing endpoints.

internal/hierarchy/hierarchy.go
zasweq: So the reason is because we're guaranteed this to be set?

So programmer error vs. simplicity? Changed.

internal/hierarchy/hierarchy.go
zasweq: SetInEndpoint sounds good to me.

hierarchy.SetInEndpoint(...)

internal/hierarchy/hierarchy.go
dfawley: No, because they are actually equivalent.  :)

internal/hierarchy/hierarchy.go
dfawley: > What's wrong with as per lol?

It's always redundant.  Just "as" or "per" is sufficient.  https://english.stackexchange.com/a/54880

But also personal preference, I guess.  Some people like saying "ATM machine" and "PIN number" too.

xds/internal/balancer/clusterresolver/clusterresolver.go
dfawley: OK that makes sense, thanks.

dfawley (approved):

internal/hierarchy/hierarchy.go
zasweq: Oh lol so it's redundant.

PR #7826, opened 2024-11-11
balancer/weightedroundrobin: Switch Weighted Round Robin to use pick first instead of SubConns

zasweq:
This PR switches weighted round robin to defer to pick first instead of creating and handling SubConns itself. This is part of DualStack, and makes this a petiole.

RELEASE NOTES:
* balancer/weightedroundrobin: Switch Weighted Round Robin to use pick first instead of SubConns

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -28,18 +28,29 @@ package endpointsharding
 import (
 	"encoding/json"
 	"errors"
+	rand "math/rand/v2"
 	"sync"
 	"sync/atomic"
-
-	rand "math/rand/v2"
 
 	"google.golang.org/grpc/balancer"
 	"google.golang.org/grpc/balancer/base"
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/internal/balancer/gracefulswitch"
+	"google.golang.org/grpc/internal/envconfig"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/serviceconfig"
 )
+
+// PickFirstConfig is a pick first config without shuffling enabled.
+var PickFirstConfig string
+
+func init() {
+	if envconfig.NewPickFirstEnabled {
+		PickFirstConfig = "[{\"pick_first\": {}}]"
+		return
+	}
+	PickFirstConfig = "[{\"pick_first_leaf\": {}}]"
+}
 
 // ChildState is the balancer state of a child along with the endpoint which
 // identifies the child balancer.
@@ -100,9 +111,6 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 
 	// Update/Create new children.
 	for _, endpoint := range state.ResolverState.Endpoints {
-		if len(endpoint.Addresses) == 0 {
-			continue
-		}
 		if _, ok := newChildren.Get(endpoint); ok {
 			// Endpoint child was already created, continue to avoid duplicate
 			// update.
@@ -143,6 +151,9 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 		}
 	}
 	es.children.Store(newChildren)
+	if newChildren.Len() == 0 {
+		return balancer.ErrBadResolverState
+	}
 	return ret
 }
 
@@ -306,6 +317,3 @@ func (bw *balancerWrapper) UpdateState(state balancer.State) {
 func ParseConfig(cfg json.RawMessage) (serviceconfig.LoadBalancingConfig, error) {
 	return gracefulswitch.ParseConfig(cfg)
 }
-
-// PickFirstConfig is a pick first config without shuffling enabled.
-const PickFirstConfig = "[{\"pick_first\": {}}]"
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -28,12 +28,15 @@ package endpointsharding
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	rand "math/rand/v2"
 	"sync"
 	"sync/atomic"
 
 	"google.golang.org/grpc/balancer"
 	"google.golang.org/grpc/balancer/base"
+	"google.golang.org/grpc/balancer/pickfirst"
+	"google.golang.org/grpc/balancer/pickfirst/pickfirstleaf"
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/internal/balancer/gracefulswitch"
 	"google.golang.org/grpc/internal/envconfig"
@@ -45,11 +48,11 @@ import (
 var PickFirstConfig string
 
 func init() {
-	if envconfig.NewPickFirstEnabled {
-		PickFirstConfig = "[{\"pick_first\": {}}]"
-		return
+	name := pickfirst.Name
+	if !envconfig.NewPickFirstEnabled {
+		name = pickfirstleaf.Name
 	}
-	PickFirstConfig = "[{\"pick_first_leaf\": {}}]"
+	PickFirstConfig = fmt.Sprintf("[{%q: {}}]", name)
 }
 
 // ChildState is the balancer state of a child along with the endpoint which
--- a/balancer/weightedroundrobin/balancer.go
+++ b/balancer/weightedroundrobin/balancer.go
@@ -19,7 +19,6 @@
 package weightedroundrobin
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	rand "math/rand/v2"
@@ -35,6 +34,7 @@ import (
 	"google.golang.org/grpc/connectivity"
 	estats "google.golang.org/grpc/experimental/stats"
 	"google.golang.org/grpc/internal/grpclog"
+	"google.golang.org/grpc/internal/grpcsync"
 	iserviceconfig "google.golang.org/grpc/internal/serviceconfig"
 	"google.golang.org/grpc/orca"
 	"google.golang.org/grpc/resolver"
@@ -211,7 +211,7 @@ type wrrBalancer struct {
 	mu               sync.Mutex
 	cfg              *lbConfig // active config
 	locality         string
-	stopPicker       func()
+	stopPicker       *grpcsync.Event
 	addressWeights   *resolver.AddressMap  // addr -> endpointWeight
 	endpointToWeight *resolver.EndpointMap // endpoint -> endpointWeight
 	scToWeight       map[balancer.SubConn]*endpointWeight
@@ -245,7 +245,7 @@ func (b *wrrBalancer) UpdateState(state balancer.State) {
 	defer b.mu.Unlock()
 
 	if b.stopPicker != nil {
-		b.stopPicker()
+		b.stopPicker.Fire()
 		b.stopPicker = nil
 	}
 
@@ -288,9 +288,8 @@ func (b *wrrBalancer) UpdateState(state balancer.State) {
 		target:          b.target,
 	}
 
-	var ctx context.Context
-	ctx, b.stopPicker = context.WithCancel(context.Background())
-	p.start(ctx)
+	b.stopPicker = grpcsync.NewEvent()
+	p.start(b.stopPicker)
 
 	b.ClientConn.UpdateState(balancer.State{
 		ConnectivityState: state.ConnectivityState,
@@ -388,7 +387,7 @@ func (b *wrrBalancer) updateSubConnState(sc balancer.SubConn, state balancer.Sub
 func (b *wrrBalancer) Close() {
 	b.mu.Lock()
 	if b.stopPicker != nil {
-		b.stopPicker()
+		b.stopPicker.Fire()
 		b.stopPicker = nil
 	}
 	b.mu.Unlock()
@@ -468,7 +467,7 @@ func (p *picker) regenerateScheduler() {
 	atomic.StorePointer(&p.scheduler, unsafe.Pointer(&s))
 }
 
-func (p *picker) start(ctx context.Context) {
+func (p *picker) start(done *grpcsync.Event) {
 	p.regenerateScheduler()
 	if len(p.weightedPickers) == 1 {
 		// No need to regenerate weights with only one backend.
@@ -480,7 +479,7 @@ func (p *picker) start(ctx context.Context) {
 		defer ticker.Stop()
 		for {
 			select {
-			case <-ctx.Done():
+			case <-done.Done():
 				return
 			case <-ticker.C:
 				p.regenerateScheduler()
--- a/balancer/weightedroundrobin/balancer.go
+++ b/balancer/weightedroundrobin/balancer.go
@@ -21,7 +21,6 @@ package weightedroundrobin
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	rand "math/rand/v2"
 	"sync"
@@ -30,7 +29,7 @@ import (
 	"unsafe"
 
 	"google.golang.org/grpc/balancer"
-	"google.golang.org/grpc/balancer/base"
+	"google.golang.org/grpc/balancer/endpointsharding"
 	"google.golang.org/grpc/balancer/weightedroundrobin/internal"
 	"google.golang.org/grpc/balancer/weightedtarget"
 	"google.golang.org/grpc/connectivity"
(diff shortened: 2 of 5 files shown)


balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
-		b.ResolverError(errors.New("resolver produced zero addresses")) // will call regeneratePicker
-		return balancer.ErrBadResolverState
-	}
+	b.updateEndpointsLocked(ccs.ResolverState.Endpoints)
+	b.mu.Unlock()
+
+	// Note: if this call ever starts erroring (as of writing this won't happen
+	// unless programmer error), will need to rethink this operation, as once it
+	// gets here it has already updated all the data structures tracking
+	// endpoints.
```
dfawley, line 232: Delete this comment?  I think it's fine now.  We give the update to the child, which is all we need to do.  It should never "reject" the update.  It always applies it as best as possible, and PF should never fail to act on any update.

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+	logger              *grpclog.PrefixLogger
+	target              string
+	metricsRecorder     estats.MetricsRecorder
+
+	mu               sync.Mutex
+	cfg              *lbConfig // active config
+	locality         string
+	stopPicker       func()
+	addressWeights   *resolver.AddressMap  // addr -> endpointWeight
+	endpointToWeight *resolver.EndpointMap // endpoint -> endpointWeight
```
dfawley, line 210: We should turn these into generics :)

(Obviously not in this PR)

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+		v:               rand.Uint32(), // start the scheduler at a random point
+		cfg:             b.cfg,
+		weightedPickers: readyPickersWeight,
+		metricsRecorder: b.metricsRecorder,
+		locality:        b.locality,
+		target:          b.target,
 	}
+
+	var ctx context.Context
+	ctx, b.stopPicker = context.WithCancel(context.Background())
```
dfawley, line 288: Oh...did I use a context here?  :thinking: I think this should be replace with an `Event` (or a signaling channel).

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+		b.updateSubConnState(sc, state)
+		oldListener(state)
 	}
-	if b.connectivityState != connectivity.TransientFailure {
-		// No need to update the picker since no error is being returned.
-		return
+
+	sc, err := b.ClientConn.NewSubConn([]resolver.Address{addr}, opts)
+	if err != nil {
+		return sc, err
```
dfawley, line 314: Nit/style: I believe it's more traditional to return `nil, err` here.

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+	}
+
+	b.mu.Lock()
+	ewv, ok := b.addressWeights.Get(addr)
+	if !ok {
+		b.mu.Unlock()
```
dfawley, line 320: `defer` instead?

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+		// SubConn state updates can come in for a no longer relevant endpoint
+		// weight (from the old system after a new config update is applied).
+		// Will eventually get cleared from scMap once receives Shutdown signal.
+		return sc, err
```
dfawley, line 324: As above, but moreso..return `nil, err` -- your `sc` is actually valid here.

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
-		// No need to update the picker since no error is being returned.
-		return
+
+	sc, err := b.ClientConn.NewSubConn([]resolver.Address{addr}, opts)
+	if err != nil {
+		return sc, err
+	}
+
+	b.mu.Lock()
+	ewv, ok := b.addressWeights.Get(addr)
```
dfawley, line 318: Please be consistent.  Above you use `ewi` and here you use `ewv`.

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+	if !ok {
+		b.mu.Unlock()
+		// SubConn state updates can come in for a no longer relevant endpoint
+		// weight (from the old system after a new config update is applied).
+		// Will eventually get cleared from scMap once receives Shutdown signal.
+		return sc, err
 	}
-	b.regeneratePicker()
+	ew := ewv.(*endpointWeight)
```
dfawley, line 326: Delete the local?

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+	// Read the scheduler atomically.  All scheduler operations are threadsafe,
+	// and if the scheduler is replaced during this usage, we want to use the
+	// scheduler that was live when the pick started.
+	sched := *(*scheduler)(atomic.LoadPointer(&p.scheduler))
+
+	pickedPicker := p.weightedPickers[sched.nextIndex()]
+	pr, err := pickedPicker.picker.Pick(info)
+	if err != nil {
+		return balancer.PickResult{}, err
+	}
```
dfawley, line 446: It says this isn't covered by tests.  Is it actually impossible? I think it might be...We never call child pickers that aren't `Ready`, and `Ready` pickers should never error from PF.  This could `logger.Errorf` if we don't think it should happen.

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+	if !p.cfg.EnableOOBLoadReport {
+		pr.Done = func(info balancer.DoneInfo) {
```
dfawley, line 448: This needs to call the old `pr.Done` instead of just overwriting.  Or are we asserting that PF will never set a done callback?  If that's the case, add a `logger.Errorf` if it's set, please.

balancer/weightedroundrobin/balancer_test.go
```diff
@@ -732,7 +791,7 @@ func ensureReached(ctx context.Context, t *testing.T, c testgrpc.TestServiceClie
 	reached := make(map[string]struct{})
 	for len(reached) != n {
 		var peer peer.Peer
-		if _, err := c.EmptyCall(ctx, &testpb.Empty{}, grpc.Peer(&peer)); err != nil {
+		if _, err := c.EmptyCall(ctx, &testpb.Empty{}, grpc.Peer(&peer), grpc.WaitForReady(true)); err != nil {
```
dfawley, line 794: Why?  Are we expecting connection failures?

balancer/weightedroundrobin/balancer_test.go
```diff
@@ -460,6 +460,65 @@ func (s) TestBalancer_TwoAddresses_OOBThenPerCall(t *testing.T) {
...
+	}
+
+	endpointsSharedAddress := []resolver.Endpoint{{Addresses: []resolver.Address{{Addr: "same-address"}}}, {Addresses: []resolver.Address{{Addr: "same-address"}}}}
+	srv.R.UpdateState(resolver.State{
+		Endpoints: endpointsSharedAddress,
+	})
+
+	// Make some RPC's and make sure doesn't crash.
+	for i := 0; i < 10; i++ {
+		srv.Client.EmptyCall(ctx, &testpb.Empty{}, grpc.WaitForReady(true))
```
dfawley, line 483: Why wait for ready?  It will never be ready as is, right?  So this takes 10 seconds to run?

balancer/weightedroundrobin/balancer_test.go
```diff
@@ -460,6 +460,65 @@ func (s) TestBalancer_TwoAddresses_OOBThenPerCall(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	srv := startServer(t, reportCall)
+	sc := svcConfig(t, perCallConfig)
+	if err := srv.StartClient(grpc.WithDefaultServiceConfig(sc)); err != nil {
+		t.Fatalf("Error starting client: %v", err)
+	}
+
+	endpointsSharedAddress := []resolver.Endpoint{{Addresses: []resolver.Address{{Addr: "same-address"}}}, {Addresses: []resolver.Address{{Addr: "same-address"}}}}
```
dfawley, line 476: Can we use an address that actually works?

balancer/weightedroundrobin/balancer.go
zasweq: No I forgot to call old pr.Done, I'll add it (as I call state listener downward and just intercept the call).

balancer/weightedroundrobin/balancer.go
zasweq: Can we? There's lots of operations on these that do a bunch of operations dependent on endpoint vs. address. Idk how that would be encoded in same operation.

balancer/weightedroundrobin/balancer.go
zasweq: Ok that's fine if we stick with that philosophy/we can make the assumption from the lower layers.

I guess all our error cases were either explicit TF from no endpoints which then we would want this update or it would apply the update and it would be undefined (which I used behavior to explicitly error).

balancer/weightedroundrobin/balancer.go
dfawley: I think we'd parameterize it just like https://pkg.go.dev/sync/atomic#Pointer ... Instead of storing an `any`, we could store the one type that we want.  So...  `addressWeights *resolver.AddressMap[endpointWeight]`

zasweq: Oh the value as an atomic I see. I thought you meant an atomic on address/endpoint map type structure.

balancer/weightedroundrobin/balancer.go
```diff
@@ -141,273 +147,311 @@ func (bb) Name() string {
...
+// starting and clearing any endpoint weights needed.
+//
+// Caller must hold b.mu.
+func (b *wrrBalancer) updateEndpointsLocked(endpoints []resolver.Endpoint) {
+	endpointSet := resolver.NewEndpointMap()
+	for _, endpoint := range endpoints {
+		endpointSet.Set(endpoint, nil)
+		var ew *endpointWeight
+		ewi, ok := b.endpointToWeight.Get(endpoint)
+		if ok {
```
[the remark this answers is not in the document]
zasweq: Done

balancer/weightedroundrobin/balancer.go
```diff
@@ -84,23 +83,30 @@ var (
 	})
 )
 
+var gracefulSwitchPickFirst serviceconfig.LoadBalancingConfig
+
 func init() {
 	balancer.Register(bb{})
+	var err error
+	gracefulSwitchPickFirst, err = endpointsharding.ParseConfig(json.RawMessage(endpointsharding.PickFirstConfig))
```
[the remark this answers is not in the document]
zasweq: Oh whoops I called this graceful switch config in the endpoint sharding package, but from this packages perspective yeah this is endpoint sharding config thanks.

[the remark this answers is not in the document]
zasweq: Called it endpointShardingLBConfig.

balancer/weightedroundrobin/balancer.go
zasweq: Done. Is this wrong? Or is it go style to convert errors received -> nil error and stop the flow there?

balancer/weightedroundrobin/balancer.go
zasweq: Ah yeah good point. Switched.

balancer/weightedroundrobin/balancer.go
zasweq: Ah ok. I deleted the last part of comment and moved it to updateSubConnState since that actually applies to updateSubConnState.

zasweq: Don't I still want to return the created SubConn to the child though? Even if it's from an old system/irrelevant. I guess it doesn't matter since that child will eventually be cleared anyway and this layer never tracks it.

balancer/weightedroundrobin/balancer.go
zasweq: Whoops I think this is a typo switched to ewi.

balancer/weightedroundrobin/balancer.go
zasweq: Done.

balancer/weightedroundrobin/balancer_test.go
zasweq: I don't remember why I had it, I think copied from the call option I plumbed in above. Deleted, it's weird though it didn't take 10 seconds to run or fail.

balancer/weightedroundrobin/balancer.go
zasweq: So I induced a nil panic from this change to return nil and I realized why, this caught a bug.

So in updateEndpoints say I got a update with endpoint with address 1. Then I got an endpoint with address 1 and 2. In the cleanup step, it would cleanup the old endpoints addresses even if present in the new one, leading to hitting this block and inducing a nil panic. Fixing that by tracking a current address map fixed it.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -28,19 +28,30 @@ package endpointsharding
...
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/serviceconfig"
 )
 
+// PickFirstConfig is a pick first config without shuffling enabled.
+var PickFirstConfig string
+
+func init() {
+	if envconfig.NewPickFirstEnabled {
+		PickFirstConfig = "[{\"pick_first\": {}}]"
```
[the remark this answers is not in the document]
zasweq: Alright sounds good about back-ticks noted for future. Switched.

balancer/weightedroundrobin/balancer_test.go
zasweq: Whoops I had this left over from when I was trying to fix what I solved in the other PR with the endpoint hierarchy. Removed.

balancer/weightedroundrobin/balancer_test.go
zasweq: Done, good point. Just used the servers address and added a comment saying it should just work even though undefined it should connect to one.

balancer/weightedroundrobin/balancer.go
zasweq: Yeah you did use a context here. If the main reason you want to switch this is goa conformance tests I think this part won't get searched. It graph searches down the exact symbols it calls so NewClient/Server, which is why it hits the xDS Client, but this happens underneath the hood and won't get searched.

So if this is the way you did it it seems like the right way to do it? Give it a context that you can close from here and close all the resources. This relates to Easwar's xDS Client context scoping we might want to refactor with Purneshes changes, what's the best practices for context scoping within a tree and within a tree that then needs to be able to deallocate resources it's spawned.

Let me know if you feel strongly about this and I can change it tomorrow. I got to all other comments.

dfawley: > It graph searches down 

I'm definitely concerned about this part.  If it ever needs to use WRR (it doesn't today AFAIK) then I think it will scan and see this.  But we can fix separately.

balancer/weightedroundrobin/balancer.go
dfawley: What you were doing is not wrong, because we'd expect `sc` _should_ be `nil` here anyway.  But it feels a little weird since errors should almost always be returned mutually exclusively to the data result.

balancer/weightedroundrobin/balancer.go
dfawley: Wait, isn't this bad?  We need to do the lookup before calling `NewSubConn` or else you're creating a subchannel you don't actually need.  In which cases it would need to be `Shutdown()` first.

dfawley:
LGTM modulo the one issue mentioned above.

balancer/weightedroundrobin/balancer.go
zasweq: Don't we want this to still be created, and eventually shutdown from the old child balancer going away? Why gate here? Isn't this an implicit part of the child balancers lifecycle? (When it goes away and Close() since for an irrelevant endpoint), it sends a Shutdown signal for it's created SC's? How will this process occur if we gate this creation here?

balancer/weightedroundrobin/balancer.go
zasweq: I can switch this to a grpcsync.Event and receive off that channel instead of ctx.Done() if you want me too.

zasweq:
Got to the two points we discussed in our 1:1.

balancer/weightedroundrobin/balancer.go
```diff
@@ -309,24 +312,19 @@ func (b *wrrBalancer) NewSubConn(addrs []resolver.Address, opts balancer.NewSubC
...
-	ew := ewv.(*endpointWeight)
-	b.scToWeight[sc] = ew
-	b.mu.Unlock()
-
+	sc, err := b.ClientConn.NewSubConn([]resolver.Address{addr}, opts)
+	if err != nil {
+		return nil, err
+	}
+	b.scToWeight[sc] = ewi.(*endpointWeight)
 	return sc, err
```
dfawley, line 328: `nil` explicitly, please, since you asserted `err == nil` already.

balancer/weightedroundrobin/balancer.go
```diff
@@ -463,7 +467,7 @@ func (p *picker) regenerateScheduler() {
 	atomic.StorePointer(&p.scheduler, unsafe.Pointer(&s))
 }
 
-func (p *picker) start(ctx context.Context) {
+func (p *picker) start(done *grpcsync.Event) {
```
dfawley, line 470: `stopPicker` so the name is consistent?

dfawley (approved):
Two minor requests, otherwise LGTM.  Thanks!

balancer/weightedroundrobin/balancer.go
zasweq: Ah, done.

balancer/weightedroundrobin/balancer.go
zasweq: Done.

PR #7829, opened 2024-11-12
grpc: prevent nil-dereference in clientStream.Trailer

naya451:
While calling this method from clientStream.finish cs.attempt may be nil. Check it before dereference.

Found by Linux Verification Center (linuxtesting.org) with SVACE.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -857,7 +857,7 @@ func (cs *clientStream) Trailer() metadata.MD {
 	// directions -- it will prevent races and should not meaningfully impact
 	// performance.
 	cs.commitAttempt()
-	if cs.attempt.s == nil {
+	if cs.attempt != nil || cs.attempt.s == nil {
 		return nil
 	}
 	return cs.attempt.s.Trailer()


arjan-bal:
@naya451, the constructor of clientStream ensures that cs.attempt is not nil in the following code: https://github.com/grpc/grpc-go/blob/274830d67a236ec65c1e705c4f4b92be6e326a82/stream.go#L323-L367

In cs.withRetry(), cs.Attempt is initialized for the first attempt: https://github.com/grpc/grpc-go/blob/274830d67a236ec65c1e705c4f4b92be6e326a82/stream.go#L777-L788

The static analyzer indicates the possibility of nil pointer dereference, but if it is actually possible depends on the conditional branches executed at runtime. If my understanding is correct, and cs.attempt is not expected to be nil once the constructor returns. So we should not handle the `nil` case in other parts of the code. If `cs.attempt` somehow becomes nil, it would indicate a logical bug in the logic and crashing may be the best option.

I'm closing this PR as this nil-derefference doesn't seem possible in practice. Feel free to re-open the PR if you have a reproducible test case or scenario in which this is possible.

PR #7831, opened 2024-11-12
deps: Remove Go minor version from go.mods

arjan-bal:
The Go minor versions were added when bumping the go major version from 1.21 to 1.22 by running `go get go@1.22 && go mod tidy`. Specifying the minor version is not necessary unless a dependency specifies a go minor version. This PR removes the minor version from all but one go.mod file which is used for testing.

## Why
As mentioned in [this thread](https://github.com/grpc/grpc-go/pull/7624#discussion_r1757159186), grpc-go being a library should not specify a patch version.

RELEASE NOTES:
* Relax minimum Go version requirement from go1.22.7 to go1.22.

### DIFF
--- a/cmd/protoc-gen-go-grpc/go.mod
+++ b/cmd/protoc-gen-go-grpc/go.mod
@@ -1,6 +1,6 @@
 module google.golang.org/grpc/cmd/protoc-gen-go-grpc
 
-go 1.22.7
+go 1.22
 
 require (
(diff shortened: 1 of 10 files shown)


dfawley (approved):
Thanks for cleaning this up.

matthewhartstonge:
Adding this to the conversation over at: https://github.com/golang/go/issues/65847

Tracking issues related to the toolchain directive

PR #7838, opened 2024-11-13
delegating resolver only

eshitachandwani:


internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -0,0 +1,196 @@
...
+	return targetBuilder.Build(target, &innerClientConn{resolver, "target"}, opts)
+}
+
+func proxyURIResolver(proxyURL *url.URL, opts resolver.BuildOptions, presolver *delegatingResolver) (resolver.Resolver, error) {
+	scheme := "dns"
+	proxyBuilder := resolver.Get(scheme)
+	if proxyBuilder == nil {
+		return nil, fmt.Errorf("resolver for proxy not found")
+	}
+	host := proxyURL.Hostname()
```
eshitachandwani, line 120: I earlier had `proxyTarget:=resolver.Target{URL: *proxyURL}` but that resulted in empty `url.Path` which in turn resulted in dns returning error saying missing endpoint. I am not sure why this is happening, form what I can see proxyURL is of the form `http://localhost:8000` like this and it gets stored in `URL.Host = localhost:8000` and `URL.path` remains empty. This way seems to solve the problem as we use `URL.parse` on the host which saves it in path also returning desired values. I have also confirmed it using the route_guide example.

PR #7839, opened 2024-11-13
balancer/pickfirst: Add pick first metrics

zasweq:
This PR adds pick first metrics according to [A78](https://github.com/grpc/proposal/blob/master/A78-grpc-metrics-wrr-pf-xds.md), and tests as well.

RELEASE NOTES:
* balancer/pickfirst: Emit Metrics from pick_first load balancing policy


dfawley:
Would you mind reviewing this @arjan-bal?

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -548,7 +575,12 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
 		return
 	}
 
+	if oldState == connectivity.Connecting && newState.ConnectivityState == connectivity.TransientFailure {
+		pickFirstConnectionAttemptsFailed.Record(b.metricsRecorder, 1, b.target)
```
easwars, line 579: Does this work if the passed in metrics recorder is `nil`? I'm guessing not, since I see changes in tests setting this to nop metrics recorder. Can we instead handle that in `Build`?

zasweq:
Thanks for the pass Easwar, I did this work with Doug who knows a lot more about the state transitions so I want to hear your input on Easwar's concerns @dfawley.

dfawley:
I think it would be better for you to reply to them first @zasweq.

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+	"go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest"
+)
+
+var pfConfig string
+
+func init() {
+	name := pickfirst.Name
+	if !envconfig.NewPickFirstEnabled {
+		name = pickfirstleaf.Name
+	}
```
arjan-bal, line 54: This is not required, the `init()` function in pickfirstleaf ensures that `pickfirstleaf.Name` is the name under which the balancer is registered:
https://github.com/grpc/grpc-go/blob/0775031253f395d1008f80da5f0582b8871f143e/balancer/pickfirst/pickfirstleaf/pickfirstleaf.go#L47-L53

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+func (s) TestPickFirstMetrics(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	ss := &stubserver.StubServer{
+		EmptyCallF: func(ctx context.Context, in *testpb.Empty) (*testpb.Empty, error) {
+			return &testpb.Empty{}, nil
+		},
+	}
+	ss.StartServer()
```
arjan-bal, line 77: nit: This can be shortened with `ss := stubserver.StartTestService(t, nil)`.

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+	defer ss.Stop()
+
+	sc := internal.ParseServiceConfig.(func(string) *serviceconfig.ParseResult)(pfConfig)
```
arjan-bal, line 80: nit: This can be evaluated where it's used to get rid of the temp variable.
```go
r.InitialState(resolver.State{
	ServiceConfig: internal.ParseServiceConfig.(func(string) *serviceconfig.ParseResult)(pfConfig),
	Addresses:     []resolver.Address{{Addr: ss.Address}}},
)
```

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	r := manual.NewBuilderWithScheme("whatever")
+	r.InitialState(resolver.State{
+		ServiceConfig: sc,
+		Addresses:     []resolver.Address{{Addr: ss.Address}}},
+	)
+
+	grpcTarget := r.Scheme() + ":///"
```
arjan-bal, line 88: nit: This can be inlined as well.

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+	tmr := stats.NewTestMetricsRecorder()
+	cc, err := grpc.NewClient(grpcTarget, grpc.WithStatsHandler(tmr), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithResolvers(r))
```
arjan-bal, line 90: nit: The opts can be refactored into a temp variable to shorten the line:
```go
opts := []grpc.DialOption{
   grpc.WithStatsHandler(tmr),
   grpc.WithTransportCredentials(insecure.NewCredentials()),
   grpc.WithResolvers(r)
}
```

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	defer cc.Close()
+
+	tsc := testgrpc.NewTestServiceClient(cc)
+	if _, err := tsc.EmptyCall(ctx, &testpb.Empty{}); err != nil {
+		t.Fatalf("EmptyCall() failed: %v", err)
+	}
+
+	if got, _ := tmr.Metric("grpc.lb.pick_first.connection_attempts_succeeded"); got != 1 {
+		t.Fatalf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.connection_attempts_succeeded", got, 1)
```
arjan-bal, line 102: nit: Does it make sense to allow the test to continue by using `t.Errorf()` instead?

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+		t.Fatal(err)
+	}
+}
+
+func pollForDisconnectedMetrics(ctx context.Context, tmr *stats.TestMetricsRecorder) error {
+	for ; ctx.Err() == nil; <-time.After(time.Millisecond) {
+		if got, _ := tmr.Metric("grpc.lb.pick_first.disconnections"); got == 1 {
+			return nil
+		}
+	}
```
arjan-bal, line 122: Instead of polling, can we instead wait for the cc to enter IDLE using `testutils.AwaitState`? That avoid extra CPU time spent due to polling at a high frequency.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -57,7 +58,28 @@ var (
 	// Name is the name of the pick_first_leaf balancer.
 	// It is changed to "pick_first" in init() if this balancer is to be
 	// registered as the default pickfirst.
-	Name = "pick_first_leaf"
+	Name                          = "pick_first_leaf"
+	pickFirstDisconnectionsMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
```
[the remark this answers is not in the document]
zasweq: Good point. Dropped pick first prefix.

balancer/pickfirst/pickfirstleaf/metrics_test.go
easwars: +1: I think we should keep going here to get as many errors as possible.

go/go-style/decisions#keep-going

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+	if got, _ := tmr.Metric("grpc.lb.pick_first.connection_attempts_failed"); got != 0 {
+		t.Fatalf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.connection_attempts_failed", got, 0)
+	}
+	if got, _ := tmr.Metric("grpc.lb.pick_first.disconnections"); got != 0 {
+		t.Fatalf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.disconnections", got, 0)
+	}
+
+	ss.Stop()
+	if err = pollForDisconnectedMetrics(ctx, tmr); err != nil {
```
easwars, line 112: There is a single caller for this. Prefer getting rid of the function and inlining it here.

balancer/pickfirst/pickfirstleaf/metrics_test.go
easwars: Agree.

balancer/pickfirst/pickfirstleaf/metrics_test.go
easwars, line 1: I think we should have tests for happy eyeballs cases. Did you consider enhancing existing tests to check for metric values at the end of the test? Going down that path would cover a lot of scenarios instead of very simple ones being tested here.

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+	defer cc.Close()
+
+	tsc := testgrpc.NewTestServiceClient(cc)
+	if _, err := tsc.EmptyCall(ctx, &testpb.Empty{}); err == nil {
+		t.Fatalf("EmptyCall() passed when expected to fail")
+	}
+
+	r.UpdateState(resolver.State{
+		ServiceConfig: sc,
+		Addresses:     []resolver.Address{{Addr: ss.Address}}}) // Will trigger successful connection metric.
```
easwars, line 207: Please see: go/go-style/decisions#literal-matching-braces

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
easwars: > the top level channel/server creates a metrics recorder list and sends it down, so it should never be nil

If that is the case, maybe just a comment somewhere up top (maybe along with the field that it is guaranteed to be non-nil) should be good enough

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Hmmm good point. Switched to that.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -575,6 +607,7 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
 		// the first address when the picker is used.
 		b.shutdownRemainingLocked(sd)
 		b.state = connectivity.Idle
+		pickFirstDisconnectionsMetric.Record(b.metricsRecorder, 1, b.target)
```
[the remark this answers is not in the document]
zasweq: "The second condition above is considered a successful connection (again, as specified in the comment)." - I discussed this with Doug before I wrote this, and even though there's a interspliced successful connection, it can be logically be treated as simply a disconnection.

@dfawley

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -548,7 +575,12 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
 		return
 	}
 
+	if oldState == connectivity.Connecting && newState.ConnectivityState == connectivity.TransientFailure {
+		pickFirstConnectionAttemptsFailed.Record(b.metricsRecorder, 1, b.target)
+	}
```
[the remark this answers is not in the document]
zasweq: I don't understand your suggestion of moving this to a "select arms". I see two connectivity.TransientFailure conditions: one if it's in first pass, and one if it's in past the first pass.

Oh, so you want me to emit this from those two cases rather than if tf { record metric } at this point.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -548,7 +575,12 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
 		return
 	}
 
+	if oldState == connectivity.Connecting && newState.ConnectivityState == connectivity.TransientFailure {
```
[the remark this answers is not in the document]
zasweq: I don't know, you and Easwar have much more context than me. All I know is Connecting -> TF means a connection attempt failed. Is there any harm in keeping the old state check?

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
zasweq: Also, as I wrote to Arjan above, is there a harm in keeping the oldState Connecting check, even if there as an invariant that TF can only come after connecting? That makes it clearer to me.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -36,6 +36,7 @@ import (
 	"google.golang.org/grpc/balancer"
 	"google.golang.org/grpc/balancer/pickfirst/internal"
 	"google.golang.org/grpc/connectivity"
+	estats "google.golang.org/grpc/experimental/stats"
```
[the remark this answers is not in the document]
zasweq: This is how I always import experimental/stats. So one, for consistency.

The reason I prefixed this imports in other places other than the name collision is the point of this experimental package. The point of me doing a lot of that non per call observability stuff in this package so it's explicit to users it's experimental. Prefixing experimental imports with e makes this clearer to me.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
zasweq: Removed old state check.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
zasweq: For the second case, accounted for the interspliced READY signal by recording on connection attempts succeeded as well as disconnections.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
zasweq: Done. Thanks.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Ah, noted. Switched.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Done for the assertions on metrics values across the two unit tests.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Good point, switched.

The WaitForStateChange API doesn't wait for a state to something, it waits for a state to go out of something. So I read state before calling ss.Stop, and then waiting for state to leave that state. If the state transitions in between then and the channel going IDLE the test would break, but it shouldn't, so the test is robust. Thanks.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: The reason I had this in a separate function is I would have to do a bool local.

Switched to that. Let me know what you think.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Prefer to keep it like this since it's two lines here and only needs EmptyCallF.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Prefer to keep as a local since it's so long. Let me know if you feel strongly though.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Agreed, good point. Done.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Prefer inline here.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: I never considered that. Would you rather me write my own happy eyeballs or scale up the existing? I know there was lots of previous contention about adding too many assertions to tests.

zasweq:
Thanks for the comments y'all. Got to all comments (split out implementation and testing changes in separate commits).

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -575,6 +607,12 @@ func (b *pickfirstBalancer) updateSubConnState(sd *scData, newState balancer.Sub
 		// the first address when the picker is used.
 		b.shutdownRemainingLocked(sd)
 		b.state = connectivity.Idle
+		// READY SubConn interspliced in between CONNECTING and IDLE, need to
+		// account for that.
+		if oldState == connectivity.Connecting && newState.ConnectivityState == connectivity.Idle {
```
dfawley, line 612: +1 to something like:

```go
if oldState == Connecting {
  // A known issue causes a race that prevents the READY state change notification.  This works around it.
```

Also it would be great if we could create an issue for that problem and link it here so we know to come back and delete this one it's resolved.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -80,9 +102,12 @@ const (
 
 type pickfirstBuilder struct{}
 
-func (pickfirstBuilder) Build(cc balancer.ClientConn, _ balancer.BuildOptions) balancer.Balancer {
+func (pickfirstBuilder) Build(cc balancer.ClientConn, bo balancer.BuildOptions) balancer.Balancer {
 	b := &pickfirstBalancer{
-		cc:                    cc,
+		cc:              cc,
+		target:          bo.Target.String(),
+		metricsRecorder: bo.MetricsRecorder, // ClientConn will always create a Metrics Recorder so guaranteed to be non nil.
```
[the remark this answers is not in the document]
zasweq: This was Easwar's suggestion after our back and forth: https://github.com/grpc/grpc-go/pull/7839#discussion_r1849161049. What do you think about this. I'll drop "so guaranteed to be non nil" from this one because it's documented on the field.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
```diff
@@ -57,7 +58,28 @@ var (
 	// Name is the name of the pick_first_leaf balancer.
 	// It is changed to "pick_first" in init() if this balancer is to be
 	// registered as the default pickfirst.
-	Name = "pick_first_leaf"
+	Name                 = "pick_first_leaf"
+	disconnectionsMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
```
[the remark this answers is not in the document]
zasweq: Good point. Done. The RLS/WRR metrics are also suffixed with Metric, just checked.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
zasweq: Removed second conditional. Created a Github issue and linked to it.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Oh haha I created a race using WaitForStateChange, I see what you were going for with testutils.AwaitState. Much better haha, thanks.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
arjan-bal: You can move a similar check for TF from above here and combine them.
```
// Record a connection attempt when exiting CONNECTING.
if newState.ConnectivityState == connectivity.TransientFailure {
	sd.connectionFailedInFirstPass = true
}
```

balancer/pickfirst/pickfirstleaf/metrics_test.go
arjan-bal: That's a good idea. I would suggest adding assertions to the same test because asserting both behaviours in the same test would give a better indication that the features work together. If this severely hampers readability, we can have separate tests.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
easwars: Personally, `estats` doesn't convey the fact that the package is experimental to me.

> This is how I always import experimental/stats. So one, for consistency.

Prior art is not always a good justification for future art.

Other name name collisions, I don't see a convincing reason for renaming imports. If your package always has to be renamed when used (to clarify the meaning of the package), why not name is appropriately in the first place.

I don't want to block this PR on this, so maybe we can discuss this offline. Thanks.

balancer/pickfirst/pickfirstleaf/metrics_test.go
easwars: Agree with @arjan-bal.

I'm OK with it being a separate PR as well.

balancer/pickfirst/pickfirstleaf/metrics_test.go
easwars: You don't need a boolean. If you change the body of the `for` to `return` when you see `grpc.lb.pick_first.disconnections` (instead of breaking), then the only case where you will execute code below the `for` would be when the context expires. And therefore you can check for `ctx.Err() != nil` instead of checking for the boolean.

easwars (approved):

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: If I return early I don't even need the ctx.Err check right? The only reason it would hit that would be for the ctx to expire without having found disconnection/returned right?

zasweq: So my failure mode is an uncondtional t.Fatalf("timeout waiting for grpc.lb.pick_first.disconnections metric").

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
zasweq: Discussed offline; decided to switch to expstats, and further discussion is warranted as to what to do here (i.e. stats/experimental, leave as is, etc.).

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
zasweq: I agree with you, but I think this might be a more accurate metrics it lies after the conditional check of "previously relevant subconns". So if you think this metric is semantically accurate above that check, I can move it, but I think this is more accurate metric (ignores "obsolete SubConns").

balancer/pickfirst/pickfirstleaf/metrics_test.go
```diff
@@ -0,0 +1,280 @@
...
+	var foundDisconnections bool
+	for ; ctx.Err() == nil; <-time.After(time.Millisecond) {
+		if got, _ := tmr.Metric("grpc.lb.pick_first.disconnections"); got == 1 {
+			foundDisconnections = true
+			break
+		}
+	}
+	if !foundDisconnections {
+		t.Fatalf("error waiting for grpc.lb.pick_first.disconnections metric: %v", ctx.Err())
+	}
```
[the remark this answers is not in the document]
zasweq: Hmm good point. I left the check above with the 110 before server stops since I think that's an important state to check, but then this check can be immediate after it goes idle. Thanks.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Actually, with Arjan's point about await state I can use that to poll/sync and then the metrics test can be deterministic one time check. Thanks Arjan for suggestion.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: Scaled up two basic happy eyeballs tests (in a separate commit).

Note that there is a limitation in our test metrics recorder that it only persist the most recently emitted metric value for a given metric name. The reason for this is vast non determinism in other unit tests that use this component, where the most recent was deterministic but not the total metric. It's too far down the rabbit hole now to change I think, but let me know if y'all feel strongly about trying to add a summation assertion/if you want me to add metrics assertions to more happy eyeballs tests/what you think about the additional assertions in general.

I'll be happy to add more, remove the commit, leave as is, etc.

zasweq:
Thanks for the passes. Got to all comments and scaled up two very basic happy eyeballs tests. Let me know what you think.

balancer/pickfirst/pickfirstleaf/metrics_test.go
zasweq: I added a test to the other test that test with a ClientConn, so I think that should be good enough outside the summation thing, which I don't think should be addressed in this PR.

I saw other tests test interleaving addresses by sending TF, but figured this coverage should be good enough.

balancer/pickfirst/pickfirstleaf/pickfirstleaf.go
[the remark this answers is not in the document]
arjan-bal: Yes, I'm suggesting moving the following block down, instead of moving the new code up.
```go
if newState.ConnectivityState == connectivity.TransientFailure {
	sd.connectionFailedInFirstPass = true
}
```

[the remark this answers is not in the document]
zasweq: Ah, I see. Done.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -921,6 +924,17 @@ func (s) TestPickFirstLeaf_HappyEyeballs_TF_AfterEndOfList(t *testing.T) {
 	// Last SubConn fails, this should result in a TF update.
 	holds[1].Fail(fmt.Errorf("test error"))
 	testutils.AwaitState(ctx, t, cc, connectivity.TransientFailure)
+
+	// Only connection attempt fails in this test.
+	if got, _ := tmr.Metric("grpc.lb.pick_first.connection_attempts_succeeded"); got != 0 {
+		t.Errorf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.connection_attempts_succeeded", got, 3)
```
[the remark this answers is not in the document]
zasweq: Oh whoops, done.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -921,6 +924,17 @@ func (s) TestPickFirstLeaf_HappyEyeballs_TF_AfterEndOfList(t *testing.T) {
...
+	}
+	if got, _ := tmr.Metric("grpc.lb.pick_first.connection_attempts_failed"); got != 1 {
+		t.Errorf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.connection_attempts_failed", got, 3)
+	}
+	if got, _ := tmr.Metric("grpc.lb.pick_first.disconnections"); got != 0 {
+		t.Errorf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.disconnections", got, 3)
```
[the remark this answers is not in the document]
zasweq: Done.

balancer/pickfirst/pickfirstleaf/pickfirstleaf_ext_test.go
```diff
@@ -921,6 +924,17 @@ func (s) TestPickFirstLeaf_HappyEyeballs_TF_AfterEndOfList(t *testing.T) {
 	// Last SubConn fails, this should result in a TF update.
 	holds[1].Fail(fmt.Errorf("test error"))
 	testutils.AwaitState(ctx, t, cc, connectivity.TransientFailure)
+
+	// Only connection attempt fails in this test.
+	if got, _ := tmr.Metric("grpc.lb.pick_first.connection_attempts_succeeded"); got != 0 {
+		t.Errorf("Unexpected data for metric %v, got: %v, want: %v", "grpc.lb.pick_first.connection_attempts_succeeded", got, 3)
+	}
+	if got, _ := tmr.Metric("grpc.lb.pick_first.connection_attempts_failed"); got != 1 {
```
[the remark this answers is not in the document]
zasweq: Great idea. I had to add another method that just compares the incr since I had no access to the metric descriptor (when I wrote this assertion I was making the metrics in the tests itself). So added those 3 but kept these three here to test the expected state at the end.

zasweq:
Thanks for the comments. Great suggestions.

arjan-bal (approved):
LGTM

PR #7840, opened 2024-11-13
balancer: fix SubConn embedding requirement to not recommend a nil panic hazard

dfawley:
I deleted the release notes from #7758 and updated them here.

RELEASE NOTES:
* balancer: An internal method is added to the `balancer.SubConn` interface to force implementors to embed a delegate implementation. This requirement is present in the interface documentation, but wasn't enforced earlier.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -267,3 +267,9 @@ const (
 // It currently has an experimental suffix which would be removed once
 // end-to-end testing of the policy is completed.
 const RLSLoadBalancingPolicyName = "rls_experimental"
+
+// EnforceSubConnEmbedding is used to enforce proper SubConn implementation
+// embedding.
+type EnforceSubConnEmbedding interface {
+	enforceSubConnEmbedding()
+}
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -73,17 +73,6 @@ func unregisterForTesting(name string) {
 	delete(m, name)
 }
 
-// connectedAddress returns the connected address for a SubConnState. The
-// address is only valid if the state is READY.
-func connectedAddress(scs SubConnState) resolver.Address {
-	return scs.connectedAddress
-}
-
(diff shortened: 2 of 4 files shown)


internal/internal.go
```diff
@@ -267,3 +267,9 @@ const (
 // It currently has an experimental suffix which would be removed once
 // end-to-end testing of the policy is completed.
 const RLSLoadBalancingPolicyName = "rls_experimental"
+
+// EnforceSubConnEmbedding is used to enforce proper SubConn implementation
+// embedding.
+type EnforceSubConnEmbedding interface {
+	enforceSubConnEmbedding()
+}
```
arjan-bal, line 275: I see, extracting the private method in a new interface ensures compilation fails if `acBalancerWrapper` and other SubConn implementations are missing a SubConn method. 

Should we also embed either the internal interface or the NoOpSubConn in [TestSubConn](https://github.com/grpc/grpc-go/blob/a365199cb21f2bcf8fc46a4f3edb87f9adfab637/internal/testutils/balancer.go#L35) and [testSubConn](https://github.com/grpc/grpc-go/blob/a365199cb21f2bcf8fc46a4f3edb87f9adfab637/balancer/base/balancer_test.go#L44).

arjan-bal (approved):
vet is requesting adding godoc for the NoOpSubConn methods. One small comment about embedding in test SubConns, otherwise LGTM.

internal/internal.go
dfawley: This is where I paused, actually.  Maybe we don't want `NopSubConn`..  The question is whether it's better to provide something that silently does the wrong thing because something is not implemented correctly, but is needed, or recommend embedding the interface and not setting it, making it panic instead.

I kind of think it should be the latter.  Iterating.

easwars (approved):

PR #7861, opened 2024-11-20
credentials/alts/internal/handshaker: avoid looking up SRV and TXT records for handshaker service

townba:
When dialing the handshake service, disable the service config to ignore any provided by the resolver.

When doing a DNS resolver lookup, if we've disabled the service config, don't do an SRV record lookup.

RELEASE NOTES: none


dfawley (commented):
Please add comments in both places explaining why we're doing these things.  Otherwise LGTM, thanks.

townba:
PTAL

dfawley (approved):

Issue #7843, opened 2024-11-14
Example code doesn't gracefully shutdown


bkane-msft:
### What version of gRPC are you using?

https://github.com/grpc/grpc-go/blob/89737ae09d06cf745ebeb3c414a83c94b58709e5/examples/helloworld/greeter_server/main.go

### What version of Go are you using (`go version`)?

n/a

### What operating system (Linux, Windows, …) and version?

n/a

### What did you do?

I looked to see a good example for graceful shutdown

### What did you expect to see?

The example ( https://github.com/grpc/grpc-go/blob/89737ae09d06cf745ebeb3c414a83c94b58709e5/examples/helloworld/greeter_server/main.go ) to show graceful shutdown on a signal

### What did you see instead?

The following:

```go
	s := grpc.NewServer()
	pb.RegisterGreeterServer(s, &server{})
	log.Printf("server listening at %v", lis.Addr())
	if err := s.Serve(lis); err != nil {
		log.Fatalf("failed to serve: %v", err)
	}
```

There is also nothing in the Go docs ( https://grpc.io/docs/languages/go/basics/#starting-the-server ) about draining requests / etc on shutdown or signal.

I think this should be in the docs/example as every app needs this. Would you accept a PR to add it?

bkane-msft:
I think the correct code is:

```go
	// graceful stop
	ch := make(chan os.Signal, 1)
	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
	go func() {
		<-ch
		server.GracefulStop()
	}()

	// startup
	lis, err := net.Listen("tcp", ":27166")
	if err != nil {
		panic(fmt.Sprintf("failed to listen: %v", err))
	}
	slog.Info("server listening", "addr", lis.Addr())
	if err := server.Serve(lis); err != nil {
		panic(fmt.Sprintf("failed to serve: %v", err))
	}
```

but I'm new to gRPC and would like an authoritative example

purnesh42H:
@bkane-msft yes, `GracefulStop()` https://github.com/grpc/grpc-go/blob/master/server.go#L1901 is the correct way of stopping the gRPC server gracefully. We don't have any official documentation mentioning it though. I will ask the other maintainers if there is any and if not, should we add it.

purnesh42H:
@bkane-msft we have decided to add a separate example to demonstrate graceful stop for server. We will

PR #7865, opened 2024-11-22
examples/features/gracefulstop: add example to demonstrate server graceful stop

purnesh42H:
Fixes: #7843

RELEASE NOTES: None

### DIFF
--- a/examples/examples_test.sh
+++ b/examples/examples_test.sh
@@ -71,6 +71,7 @@ EXAMPLES=(
     "features/orca"
     "features/retry"
     "features/unix_abstract"
+    "features/gracefulstop"
 )
 
 declare -A SERVER_ARGS=(
--- a/examples/examples_test.sh
+++ b/examples/examples_test.sh
@@ -130,6 +130,7 @@ declare -A EXPECTED_SERVER_OUTPUT=(
     ["features/retry"]="request succeeded count: 4"
     ["features/unix_abstract"]="serving on @abstract-unix-socket"
     ["features/advancedtls"]=""
+    ["features/gracefulstop"]="Server stopped gracefully."
 )
 
 declare -A EXPECTED_CLIENT_OUTPUT=(
@@ -155,6 +156,7 @@ declare -A EXPECTED_CLIENT_OUTPUT=(
     ["features/retry"]="UnaryEcho reply: message:\"Try and Success\""
     ["features/unix_abstract"]="calling echo.Echo/UnaryEcho to unix-abstract:abstract-unix-socket"
     ["features/advancedtls"]=""
+    ["features/gracefulstop"]="Successful unary requests processed by server and made by client are same."
 )
 
 cd ./examples
--- a/examples/features/gracefulstop/README.md
+++ b/examples/features/gracefulstop/README.md
@@ -1,0 +1,31 @@
+# Graceful Stop
+
+This example demonstrates how to gracefully stop a gRPC server using
+Server.GracefulStop().
+
+## How to run
+
+Start the server which will listen to incoming gRPC requests as well as OS
+interrupt signals (SIGINT/SIGTERM). After receiving interrupt signal, it calls
+`s.GracefulStop()` to gracefully shut down the server. If graceful shutdown
+doesn't happen in time, server is stopped forcefully.
+
+```sh
+$ go run server/main.go
+```
+
+In a separate terminal, start the client which will send multiple requests to
+the server with some delay between each request.
+
+```sh
+$ go run client/main.go
+```
+
+Use Ctrl+C or SIGTERM to signal the server to shut down.
+
+The server begins a graceful stop:
+- It finishes ongoing requests.
+- Rejects new incoming requests.
+
+The client will notice the server's shutdown when a request fails.
+
--- a/examples/features/gracefulstop/README.md
+++ b/examples/features/gracefulstop/README.md
@@ -5,29 +5,25 @@ This example demonstrates how to gracefully stop a gRPC server using
 
 ## How to run
 
-Start the server which will serve `ServerStreaming` gRPC requests and listen to
-an OS interrupt signal (SIGINT/SIGTERM). `ServerStreaming` handler returns an
-error after sending 5 messages to the client or if client's context is
-canceled. If an interrupt signal is received, it calls `s.GracefulStop()` to
-gracefully shut down the server to make sure in-flight RPC is finished before
-shutting down. If graceful shutdown doesn't happen in time, the server is
-stopped forcefully.
+Start the server which will serve `ServerStreaming` gRPC requests. It spawns
+a go routine before starting the server that waits for signal from handler that
+server stream has started and initiates `Server.GracefulStop()`. Therefore,
+server will stop only after the in-flight rpc is finished.
 
 ```sh
 $ go run server/main.go
 ```
 
 In a separate terminal, start the client which will start a server stream to
-the server and keep receiving messages until an error is received. Once an
-error is received, it closes the stream.
+the server and wait to receive 5 messages before closing the stream.
+`ServerStreaming` handler signals the server to initiate graceful shutdown and
+start sending stream of messages to client indefinitely until client closes the
+stream.
 
 ```sh
 $ go run client/main.go
 ```
 
-Once the client starts receiving messages from server, use Ctrl+C or SIGTERM to
-signal the server to shut down.
-
-The server begins a graceful stop. It finish the in-flight request. In this
-case, client will receive 5 messages followed by the stream failure from the
-server, allowing client to close the stream gracefully, before shutting down.
+Once the client client finish receiving 5 messages from server, it sends
+`stream.CloseSend()` to close the stream which finishes the in-flight request.
+The server then gracefully stop.
--- a/examples/features/gracefulstop/README.md
+++ b/examples/features/gracefulstop/README.md
@@ -1,31 +1,33 @@
 # Graceful Stop
 
 This example demonstrates how to gracefully stop a gRPC server using
-Server.GracefulStop().
+`Server.GracefulStop()`.
 
 ## How to run
 
-Start the server which will listen to incoming gRPC requests as well as OS
-interrupt signals (SIGINT/SIGTERM). After receiving interrupt signal, it calls
-`s.GracefulStop()` to gracefully shut down the server. If graceful shutdown
-doesn't happen in time, server is stopped forcefully.
+Start the server which will serve `ServerStreaming` gRPC requests and listen to
+an OS interrupt signal (SIGINT/SIGTERM). `ServerStreaming` handler returns an
+error after sending 5 messages to the client or if client's context is
+canceled. If an interrupt signal is received, it calls `s.GracefulStop()` to
+gracefully shut down the server to make sure in-flight RPC is finished before
+shutting down. If graceful shutdown doesn't happen in time, the server is
+stopped forcefully.
 
 ```sh
 $ go run server/main.go
 ```
 
-In a separate terminal, start the client which will send multiple requests to
-the server with some delay between each request.
+In a separate terminal, start the client which will start a server stream to
+the server and keep receiving messages until an error is received. Once an
+error is received, it closes the stream.
 
 ```sh
 $ go run client/main.go
 ```
 
-Use Ctrl+C or SIGTERM to signal the server to shut down.
+Once the client starts receiving messages from server, use Ctrl+C or SIGTERM to
+signal the server to shut down.
 
-The server begins a graceful stop:
-- It finishes ongoing requests.
-- Rejects new incoming requests.
-
-The client will notice the server's shutdown when a request fails.
-
+The server begins a graceful stop. It finish the in-flight request. In this
+case, client will receive 5 messages followed by the stream failure from the
+server, allowing client to close the stream gracefully, before shutting down.
--- a/examples/features/gracefulstop/README.md
+++ b/examples/features/gracefulstop/README.md
@@ -1,34 +1,45 @@
 # Graceful Stop
 
 This example demonstrates how to gracefully stop a gRPC server using
-`Server.GracefulStop()`.
+`Server.GracefulStop()`. The graceful shutdown process involves two key steps:
 
-## How to run
+- Initiate `Server.GracefulStop()`. This function blocks until all currently
+  running RPCs have completed.  This ensures that in-flight requests are
+  allowed to finish processing.
 
-Start the server with a client streaming and unary request handler. When client
-streaming is started, client streaming handler signals the server to initiate
-graceful stop and waits for the stream to be closed or aborted. Until the
-`Server.GracefulStop()` is initiated, server will continue to accept unary
+- It's crucial to call `Server.Stop()` with a timeout before calling
+  `GracefulStop()`. This acts as a safety net, ensuring that the server
+  eventually shuts down even if some in-flight RPCs don't complete within a
+  reasonable timeframe.  This prevents indefinite blocking.
+
+## Try it
+
+```
+go run server/main.go
+```
+
+```
+go run client/main.go
+```
+
+## Explanation
+
+The server starts with a client streaming and unary request handler. When
+client streaming is started, client streaming handler signals the server to
+initiate graceful stop and waits for the stream to be closed or aborted. Until
+the`Server.GracefulStop()` is initiated, server will continue to accept unary
 requests. Once `Server.GracefulStop()` is initiated, server will not accept
 new unary requests.
 
-```sh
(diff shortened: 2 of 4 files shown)


bkane-msft:
This looks great, thank you! Would you ok linking to this example from https://grpc.io/docs/languages/go/basics/#starting-the-server so folks following the "Basics" tutorial can easily find it?

examples/features/gracefulstop/README.md
```diff
@@ -0,0 +1,31 @@
+# Graceful Stop
+
+This example demonstrates how to gracefully stop a gRPC server using
+Server.GracefulStop().
```
dfawley, line 4: Backticks?

examples/features/gracefulstop/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## How to run
+
+Start the server which will listen to incoming gRPC requests as well as OS
+interrupt signals (SIGINT/SIGTERM). After receiving interrupt signal, it calls
```
dfawley, line 9: Grammar: _an_ interrupt signal

examples/features/gracefulstop/README.md
```diff
@@ -0,0 +1,31 @@
...
+`s.GracefulStop()` to gracefully shut down the server. If graceful shutdown
+doesn't happen in time, server is stopped forcefully.
```
dfawley, line 11: Grammar: _the_ server

examples/features/gracefulstop/README.md
```diff
@@ -0,0 +1,31 @@
...
+the server with some delay between each request.
+
+```sh
+$ go run client/main.go
+```
+
+Use Ctrl+C or SIGTERM to signal the server to shut down.
+
+The server begins a graceful stop:
+- It finishes ongoing requests.
```
dfawley, line 27: Nit: ongoing -> in flight / in progress

("Ongoing" means more like "continuous".)

examples/features/gracefulstop/README.md
```diff
@@ -0,0 +1,31 @@
...
+- Rejects new incoming requests.
```
dfawley, line 28: I'm not sure how precise we want to be here, but this isn't _exactly_ what happens, FYI.

The server sends a GOAWAY on the connection, which the client sees, and then the client will stop sending requests on the connection.  If the client ignores the GOAWAY, then the server will reject new streams, but that shouldn't actually happen in any properly functioning HTTP/2 system.

What you wrote might be good enough for a basics explanation though.

examples/features/gracefulstop/client/main.go
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Binary client demonstrates sending multiple requests to server and observe
+// graceful stop.
```
dfawley, line 19: Is there some way this can actually prove that the stop was graceful?  E.g. by starting a stream that runs until requests start failing, and then closing the stream from the client and having it exit successfully by getting a response from the server at that time?

E.g. the server could respond with a message indicating how many requests it processed.

examples/features/gracefulstop/client/main.go
```diff
@@ -0,0 +1,67 @@
...
+var addr = flag.String("addr", "localhost:50052", "the address to connect to")
+
+func main() {
+	flag.Parse()
+
+	// Set up a connection to the server.
+	conn, err := grpc.NewClient(*addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatalf("Failed to connect: %v", err)
+	}
```
dfawley, line 44: Nit: `NewClient` doesn't connect.  The only errors you'd get here would be providing illegal options to the client.

examples/features/gracefulstop/client/main.go
```diff
@@ -0,0 +1,67 @@
...
+	defer conn.Close()
+	c := pb.NewGreeterClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	for i := 1; i <= 10; i++ {
```
dfawley, line 51: Shouldn't the client do these until they fail with `Unavailable`?

examples/features/gracefulstop/server/main.go
```diff
@@ -0,0 +1,86 @@
...
+
+	s := grpc.NewServer()
+	pb.RegisterGreeterServer(s, &server{})
+
+	go func() {
+		// Wait for an OS signal for graceful shutdown.
+		<-stop
+		fmt.Println("Shutting down server...")
+		s.GracefulStop()
+		close(stop)
```
dfawley, line 69: Is this OK? Closing the channel given to `signal.Notify`?  What if ^C is pressed again after the channel is closed?  Wouldn't it panic?

Better to just do a `timer := time.AfterFunc(10*time.Second, s.Stop); defer timer.Stop(); s.GracefulStop()` here?

examples/features/gracefulstop/README.md
bkane-msft:
```suggestion
`Server.GracefulStop()`.
```

bkane-msft: GitHub has a somewhat hidden feature that lets you offer suggestions by editing the text selected in the comment box . Then the PR author can click "accept" instead of checking out the repo and editing it.

bkane-msft: https://haacked.com/archive/2019/06/03/suggested-changes/

purnesh42H:
> This looks great, thank you! Would you ok linking to this example from https://grpc.io/docs/languages/go/basics/#starting-the-server so folks following the "Basics" tutorial can easily find it?

@bkane-msft the basic tutorial might not be appropriate for a graceful stop example as it is slightly advanced in nature. Right now we are propose adding it as a cross-language "guide" https://grpc.io/docs/guides/. Will update here.

examples/features/gracefulstop/README.md
purnesh42H: Added backticks. Thanks

examples/features/gracefulstop/README.md
purnesh42H: Done

examples/features/gracefulstop/README.md
purnesh42H: Done

examples/features/gracefulstop/README.md
purnesh42H: Re-wrote the description as the logic is changed now. Using in-flight though.

examples/features/gracefulstop/README.md
purnesh42H: Thanks. Now i am just restricting the description to completing the in-flight requests because we are not demonstrating sending new request while server is shutting down.

examples/features/gracefulstop/client/main.go
purnesh42H: So, here is what i did now

I wrote an example server which has a server streaming handler that sends 5 messages to client with 1 second delay in-between. After 5 messages, it sends an error to client. If we interrupt server during those 5 messages, server will make sure to complete those 5 messages followed by the error which demonstrates graceful shutdown.

examples/features/gracefulstop/client/main.go
purnesh42H: Done

examples/features/gracefulstop/client/main.go
purnesh42H: yes, client now waits until receiving an error from server. Explained above.

examples/features/gracefulstop/server/main.go
purnesh42H: Done

examples/features/gracefulstop/server/main.go
```diff
@@ -0,0 +1,106 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			atomic.AddInt32(&streamMessages, 1)
+
+			mu.Lock()
+			if streamMessages > 5 {
+				return fmt.Errorf("request failed")
+			}
```
dfawley, line 65: Why return an error?  Exiting successfully feels more graceful.

Also you should _always_ return status errors from your method handlers.  This will get converted to UNKNOWN by gRPC, which "is fine", but we should never show an example that does anything but send statuses back.

examples/features/gracefulstop/client/main.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		log.Fatalf("Error starting stream: %v", err)
+	}
+
+	for {
+		r, err := stream.Recv()
+		if err != nil {
+			// Handle the error and close the stream gracefully
+			log.Printf("Error sending request: %v\n", err)
+			err := stream.CloseSend()
```
dfawley, line 58: This doesn't make sense.  The stream ended (by virtue of receiving an error), so there is no reason to `CloseSend`.  Probably `CloseSend` should also return an error in this scenario, but it's also probably not important.

examples/features/gracefulstop/client/main.go
dfawley: ^C to cancel makes this hard to test.  We _should_ be covering this (and all examples) in https://github.com/grpc/grpc-go/blob/master/examples/examples_test.sh.  Please find a way to demonstrate this but to be testable and ideally not require more than a couple seconds to test.  I think having the client do `CloseSend` after unary requests start failing seems pretty nice, and counting the RPCs is something that can easily be done on both sides to show that things are still working on this connection even after the server has gracefully stopped.  If you have other ideas, that's fine too, but the example should be tested.

bkane-msft:
@purnesh42H and @dfawley , as someone new to gRPC, I'm really grateful to you two for your hard work in this PR providing a correct production-ready shutdown example! Thank you!

examples/features/gracefulstop/client/main.go
purnesh42H: removed `ctrl+c` requirement. Here is what i am doing now.
- Server will spawn a go routine that waits for start stream signal before initiating graceful stop. 
- When the client starts server streaming, handler will signal the server to initiate graceful stop and starts sending messages to client
- Client will receives 5 messages and then close the stream. This will trigger handler to finish the rpc.
- Graceful stop completes then

examples/features/gracefulstop/server/main.go
purnesh42H: Not applicable now. Explained the flow above

examples/features/gracefulstop/client/main.go
purnesh42H: Not applicable now

examples/features/gracefulstop/client/main.go
purnesh42H: As discussed offline, i have updated the approach to the combination of client streaming and unary rpc. So, client will start client streaming which will initiates graceful shutdown after 1 sec. During this 1 sec, client will send as many unary requests it can and will close the stream once graceful shutdown is initiated.

examples/features/gracefulstop/server/main.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var (
+	port = flag.Int("port", 50052, "port number")
+)
+
+type server struct {
+	pb.UnimplementedEchoServer
+
+	unaryRequests int32         // to track number of unary RPCs processed
```
dfawley, line 43: `atomic.Int32` is much nicer and safer.

examples/features/gracefulstop/server/main.go
```diff
@@ -0,0 +1,100 @@
...
+
+	s := grpc.NewServer()
+	ss := &server{streamStart: make(chan struct{})}
+	pb.RegisterEchoServer(s, ss)
+
+	go func() {
+		<-ss.streamStart // wait until server streaming starts
+		time.Sleep(1 * time.Second)
+		log.Println("Initiating graceful shutdown...")
+		s.GracefulStop() // gracefully stop server after in-flight server streaming rpc finishes
```
dfawley, line 93: This still should do a hard Stop after N seconds (10?), probably via a `time.AfterFunc`.  This is what users will need to do, so we need to show it.

examples/features/gracefulstop/client/main.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Fatalf("Error starting stream: %v", err)
+	}
+
+	for {
+		r, err := c.UnaryEcho(ctx, &pb.EchoRequest{Message: "Hello"})
+		if err != nil {
+			log.Printf("Error calling `UnaryEcho`. Server graceful stop initiated: %v", err)
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
```
dfawley, line 62: Nit: 200ms?  Otherwise you're printing 100 lines to the console under normal operation which is excessive.  5 seems reasonable.

examples/features/gracefulstop/README.md
```diff
@@ -0,0 +1,34 @@
+# Graceful Stop
+
+This example demonstrates how to gracefully stop a gRPC server using
+`Server.GracefulStop()`.
+
+## How to run
```
dfawley, line 6: Let's have a section up front that describes the process for doing a graceful shutdown: Call `GracefulStop`, which blocks until all currently in flight RPCs have completed.  Ensure `Stop` is called with a timeout so that even in flight RPCs are forcefully stopped eventually.  Then this section can describe the example's behavior.

examples/features/gracefulstop/client/main.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Printf(r.Message)
+	}
+
+	r, err := stream.CloseAndRecv()
+	if err != nil {
+		log.Fatalf("Error closing stream: %v", err)
+	}
+	log.Printf(r.Message)
```
dfawley, line 70: The client should keep its own count of requests and confirm that the value returned in this message matches its count.  Then it should output a "happy" message, which is how we can test it in `examples_test.sh`

examples/features/gracefulstop/server/main.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return nil
+}
+
+// UnaryEcho implements the EchoService.UnaryEcho method. It increments
+// `s.unaryRequests` on every call and returns it as part of `EchoResponse`.
+func (s *server) UnaryEcho(_ context.Context, _ *pb.EchoRequest) (*pb.EchoResponse, error) {
+	atomic.AddInt32(&s.unaryRequests, 1)
+	return &pb.EchoResponse{Message: fmt.Sprintf("Request Processed: %d", s.unaryRequests)}, nil
```
dfawley, line 73: I think it would be a little simpler to just echo the request back as an echo RPC should do, and not confuse things by showing the count on every unary call.  But it's not a big deal.

examples/features/gracefulstop/server/main.go
purnesh42H: yeah, that's right. Added

examples/features/gracefulstop/client/main.go
purnesh42H: Done

examples/features/gracefulstop/README.md
purnesh42H: Added in the first para and moved explanation down

examples/features/gracefulstop/server/main.go
purnesh42H: yeah switched to just echo

examples/features/gracefulstop/client/main.go
purnesh42H: Done. At the end we compare the values and if not equal, raise error

purnesh42H:
> The example LGTM. Please add coverage for it in `examples_test.sh`

Added

examples/examples_test.sh
```diff
@@ -71,6 +71,7 @@ EXAMPLES=(
     "features/orca"
     "features/retry"
     "features/unix_abstract"
+    "features/gracefulstop"
```
dfawley, line 74: This needs expectations for server and/or client output to confirm it is working correctly.

purnesh42H: Added. PTAL

dfawley (approved):

dfawley:
LGTM. Should the PR comment 0 say "Fixes #7843" instead so it's closed?

purnesh42H:
> LGTM. Should the PR comment 0 say "Fixes #7843" instead so it's closed?

yeah we can close this issue but I think we should link it through grpc.io docs with similar guidelines for server graceful stop.

PR #7867, opened 2024-11-22
xds client interface poc

purnesh42H:


xds/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.Stream[clients.StreamRequest, any], error) {
+	switch method {
+	case v3adsgrpc.AggregatedDiscoveryService_StreamAggregatedResources_FullMethodName:
+		g.newADSStream(ctx)
+	case v3lrsgrpc.LoadReportingService_StreamLoadStats_FullMethodName:
+		return g.newLRSStream(ctx)
+	}
+	return nil, fmt.Errorf("unsupported method: %v", method)
```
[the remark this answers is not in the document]
purnesh42H: @dfawley are you suggesting something like this `stream, err := g.cc.NewStream(ctx, &grpc.StreamDesc{}, methodName)` but then user still needs to pass the method name. So, in case of ads user need to pass "/envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources"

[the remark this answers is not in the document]
purnesh42H: oh ok. Looks like grpc.StreamDesc{} need a value for both ads and lrs which is defined in their respective client interface so to differentiate the client we have to enumerate. However, any other implementation can choose to directly create the stream from the passed ctx and methodName

xds/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+	typedRes, ok := any(res).(Res)
+	if !ok {
+		return nil, fmt.Errorf("response type mismatch")
+	}
+	return &typedRes, nil
+}
+
+func (g *grpcTransport) newADSStream(ctx context.Context) (clients.Stream[clients.StreamRequest, any], error) {
```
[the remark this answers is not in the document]
purnesh42H: We need client for ads and lrs interface. That's why do this but any other implementation doesn't need to do this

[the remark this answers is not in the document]
dfawley: The xdsclient should pass the method name of the streaming RPC call to the transport.  The user shouldn't be doing anything here besides setting the grpctransport implementation in the xdsclient's config.

xds/xds_client_test.go
```diff
@@ -0,0 +1,120 @@
...
+	serverConfig := clients.ServerConfig{ServerURI: mgmtServer.Address}
+	grpcServerConfig := grpctransport.ServerConfig{Credentials: insecure.NewBundle()}
+	serverConfig.Extensions = grpcServerConfig
+	grpcTransportBuilder := &grpctransport.Builder{}
+
+	listenerResourceType := listenerResourceType{}
+	resourceTypes := map[string]xdsclient.ResourceType{}
+	resourceTypes[listenerResourceType.TypeURL()] = listenerResourceType
+
+	config := xdsclient.NewConfig(defaultAthourity, authorities, node, grpcTransportBuilder, resourceTypes)
```
markdroth, line 96: Why is `NewConfig()` taking `defaultAuthority` as a parameter?  There should not be a default authority as a first-class field in the config, only a default list of xDS servers -- I don't think we want to support any extensions of `clients.Authority` at the top level.

dfawley: `Authority` here _is_ apparently just a list of servers:

https://github.com/grpc/grpc-go/pull/7867/files/c81bb55cc5c3110790985aad0a091a5d07a1a3c9#diff-cc980a8a61329b02bc44b7437d28b5fa6741fb5758a1567ba626326d6a18324eR41

purnesh42H: I think its more clear if we just keep it as a []ServerConfig or a separate struct `ServerConfigs` which holds the slice

purnesh42H: Also, may be authority doesn't need Extensions since it doesn't get passed around to implementation plugins

markdroth: The authority extensions don't get passed through XdsClient, but they are used outside of it.  The listener resource name template fields are used in both the client and server code that calls XdsClient.

purnesh42H: Kept the config in DecodeOptions

dfawley:
Closing this as it's not an actionable PR.  In the future, it's usually preferred to open a PR in your fork instead for "preview" type changes.

PR #7869, opened 2024-11-24
stream: add jitter to retry backoff in accordance with gRFC A6

isgj:
Update the delay calculation to reflect the updated gRFC

Address #7514 

RELEASE NOTES:

* client: update Retry backoff to apply jitter per updates to [gRFC A6](https://github.com/grpc/proposal/blob/master/A6-client-retries.md).

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -218,7 +218,7 @@ func newClientStream(ctx context.Context, desc *StreamDesc, cc *ClientConn, meth
 
 	var mc serviceconfig.MethodConfig
 	var onCommit func()
-	var newStream = func(ctx context.Context, done func()) (iresolver.ClientStream, error) {
+	newStream := func(ctx context.Context, done func()) (iresolver.ClientStream, error) {
 		return newClientStreamWithParams(ctx, desc, cc, method, mc, onCommit, done, opts...)
 	}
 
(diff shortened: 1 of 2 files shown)


aranjans:
@isgj Thanks for raising the PR. Can you add an end to end test to verify the same?

stream.go
```diff
@@ -708,11 +708,10 @@ func (a *csAttempt) shouldRetry(err error) (bool, error) {
 		cs.numRetriesSincePushback = 0
 	} else {
 		fact := math.Pow(rp.BackoffMultiplier, float64(cs.numRetriesSincePushback))
-		cur := float64(rp.InitialBackoff) * fact
-		if max := float64(rp.MaxBackoff); cur > max {
-			cur = max
-		}
-		dur = time.Duration(rand.Int64N(int64(cur)))
+		cur := min(float64(rp.InitialBackoff)*fact, float64(rp.MaxBackoff))
+		// apply a jitter of plus or minus 0.2
```
aranjans, line 712: nits: Apply jitter by multiplying with a random factor between 0.8 and 1.2.

aranjans:
Would be good if we could update the README of examples for retry feature.

arjan-bal:
> Would be good if we could update the README of examples for retry feature.

@aranjans can you point to the README and the example file that needs updating?

stream.go
isgj: Updated

isgj:
> Would be good if we could update the README of examples for retry feature.

The README already points to the right gRFC https://github.com/grpc/grpc-go/blob/master/examples/features/retry/README.md?plain=1#L7 

---

> Can you add an end to end test to verify the same?

I see the tests at https://github.com/grpc/grpc-go/blob/master/test/retry_test.go. 
Are you thinking about any specific test? I can try to add a test that checks the time when the retry happens, but that kind of test might be flaky

aranjans:
@isgj Yes that's the correct reference to README. 
> Are you thinking about any specific test? I can try to add a test that checks the time when the retry happens, but that kind of test might be flaky
Ah yeah, that might be flaky. We could skip writing the test for that.

Overall this change LGTM, adding the second reviewer to take a look.

dfawley (approved):
Thank you for the PR!

PR #7874, opened 2024-11-25
stats/opentelemetry: Cleanup OpenTelemetry API's before stabilization

zasweq:
This PR cleans up some OpenTelemetry API's before stabilization.

RELEASE NOTES:
* stats/opentelemetry: Introduce new APIs to enable OpenTelemetry instrumentation for metrics on servers and clients

### DIFF
--- a/stats/opentelemetry/opentelemetry.go
+++ b/stats/opentelemetry/opentelemetry.go
@@ -16,6 +16,10 @@
 
 // Package opentelemetry implements opentelemetry instrumentation code for
 // gRPC-Go clients and servers.
+//
+// For details on configuring opentelemetry and various instruments that this
+// package creates, see
+// [gRPC OpenTelemetry Metrics](https://grpc.io/docs/guides/opentelemetry-metrics/).
 package opentelemetry
 
 import (
@@ -28,6 +32,7 @@ import (
 	estats "google.golang.org/grpc/experimental/stats"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/internal"
+	"google.golang.org/grpc/stats"
 	otelinternal "google.golang.org/grpc/stats/opentelemetry/internal"
 
 	otelattribute "go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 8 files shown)


stats/opentelemetry/opentelemetry.go
```diff
@@ -57,27 +62,28 @@ type Options struct {
...
+
+	// MethodAttributeFilter is a function that determines whether to record the
+	// method name of RPCs as an attribute, or to bucket into "other". Take care
+	// to limit the values allowed, as allowing too many will increase
+	// cardinality and could cause severe memory or performance problems. On
+	// Client Side, to record the method name along with metric instead of
+	// bucketing into "other pass a grpc.StaticMethodCallOption as a call option
+	// into Invoke or NewStream.
+	//
 	// This only applies for server side metrics.
```
dfawley, line 82:
```suggestion
	// cardinality and could cause severe memory or performance problems.
	//
	// This only applies for server-side metrics.  For clients, to record
	// the method name in the attributes, pass grpc.StaticMethodCallOption
	// to Invoke or NewStream. Note that when using protobuf generated
	// clients, this CallOption is included automatically.
```

stats/opentelemetry/opentelemetry.go
```diff
@@ -57,27 +62,28 @@ type Options struct {
...
 	MethodAttributeFilter func(string) bool
 
-	// OptionalLabels are labels received from LB Policies that this component
-	// should add to metrics that record after receiving incoming metadata.
+	// OptionalLabels specifies a list of optional labels that will be included
+	// on any metrics that have corresponding optional labels configured.
```
dfawley, line 86:
```suggestion
	// OptionalLabels specifies a list of optional labels to enable on any
	// metrics that support them.
```

dfawley (commented):
LGTM, just some minor wording suggestions.  Please check out the edits-as-commits and suggested changes.

stats/opentelemetry/opentelemetry.go
zasweq: Thanks, done.

stats/opentelemetry/opentelemetry.go
zasweq: Yeah I couldn't figure out a way to tersely convey it's just the subset that are "supported" that are enabled. (i.e. it's not always the full set added optionally, but the ones that are actually also configured. Changed to your suggestion.

dfawley (approved):

PR #7881, opened 2024-11-29
*: Correct gRFC A1 implementation

eshitachandwani:
Fixes: #7556 
- Remove environment variable detection from the transport layer. Instead, introduce a per-address attribute to specify the proxy CONNECT string. This change aligns with the xDS design and provides more control over proxy behavior.
- Create a new resolver that handles proxy configuration and delegates to child resolvers for target and proxy address resolution. This resolver does operations stated above
- Introduce a new DialOption to preserve the current behavior of grpc.Dial, where the resolved target address is used in the proxy CONNECT request. This option ensures backward compatibility for users relying on the existing behavior.

RELEASE NOTES: None


internal/testutils/proxy.go
```diff
@@ -0,0 +1,106 @@
...
+	if p.in != nil {
+		p.in.Close()
+	}
+	if p.out != nil {
+		p.out.Close()
+	}
+}
+
+// NewProxyServer create and starts a proxy server.
+func NewProxyServer(lis net.Listener, requestCheck func(*http.Request) error, errCh chan error, doneCh chan struct{}, backendAddr string, resolutionOnClient bool, proxyServerStarted func()) *ProxyServer {
```
purnesh42H, line 54: may be we can do something like this fake server in its own package https://github.com/grpc/grpc-go/blob/17d08f746b51fa70911754d36a799c247a2c4b79/internal/testutils/xds/fakeserver/server.go#L19 ? wdyt?

New should only create the object and StartServer/Run should start the go routine to accept requests

internal/transport/http2_client.go
```diff
@@ -153,8 +154,13 @@ type http2Client struct {
 	logger       *grpclog.PrefixLogger
 }
 
-func dial(ctx context.Context, fn func(context.Context, string) (net.Conn, error), addr resolver.Address, useProxy bool, grpcUA string) (net.Conn, error) {
+func dial(ctx context.Context, fn func(context.Context, string) (net.Conn, error), addr resolver.Address, grpcUA string) (net.Conn, error) {
 	address := addr.Addr
+
+	//if the ProxyConnectAddr is set in the aattribute, do a proxy dial.
```
purnesh42H, line 160: typo: attribute

internal/transport/proxy.go
```diff
@@ -72,21 +55,22 @@ func basicAuth(username, password string) string {
...
 		Method: http.MethodConnect,
 		URL:    &url.URL{Host: backendAddr},
 		Header: map[string][]string{"User-Agent": {grpcUA}},
 	}
-	if t := proxyURL.User; t != nil {
-		u := t.Username()
-		p, _ := t.Password()
+	if user := attributes.User(address); user != nil {
+		u := user.Username()
+		p, _ := user.Password()
```
purnesh42H, line 73: hmmm...we should not ignore the bool. What happens if password is not set?

internal/transport/proxy.go
```diff
@@ -117,28 +101,14 @@ func doHTTPConnectHandshake(ctx context.Context, conn net.Conn, backendAddr stri
 	return conn, nil
 }
 
-// proxyDial dials, connecting to a proxy first if necessary. Checks if a proxy
-// is necessary, dials, does the HTTP CONNECT handshake, and returns the
-// connection.
-func proxyDial(ctx context.Context, addr string, grpcUA string) (net.Conn, error) {
-	newAddr := addr
-	proxyURL, err := mapAddress(addr)
+// proxyDial dials, connecting to a proxy first if necessary. Dials, does the
```
purnesh42H, line 104: we can remove the necessary part now? It will always dial proxy if called?

test/proxy_test.go
```diff
@@ -0,0 +1,666 @@
...
+	})
+
+	proxyServer := testutils.NewProxyServer(proxyListener, reqCheck, errCh, doneCh, backendAddr, resolutionOnClient, func() { close(proxyStartedCh) })
+	t.Cleanup(proxyServer.Stop)
+
+	return proxyListener, errCh, doneCh, proxyStartedCh
+}
+
+// TestGrpcDialWithProxy tests grpc.Dial using a proxy and default
+// resolver in the target URI.and verifies that it connects to the proxy server
```
purnesh42H, line 103: typo space and remove period after target URI

test/proxy_test.go
```diff
@@ -0,0 +1,666 @@
...
+			}, nil
+		}
+		return nil, nil
+	}
+	defer overwriteAndRestore(hpfe)()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	conn, err := grpc.Dial(unresolvedTargetURI, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
purnesh42H, line 132: we can't use `grpc.NewClient` here?

test/proxy_test.go
```diff
@@ -0,0 +1,666 @@
...
+	// Update the proxy resolver state with the proxy's address.
+	mrProxy.InitialState(resolver.State{
+		Addresses: []resolver.Address{
+			{Addr: proxyLis.Addr().String()},
+		},
+	})
+	// Dial to the proxy server.
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	conn, err := grpc.Dial(unresolvedTargetURI, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
purnesh42H, line 192: same comment about using `grpc.NewClient`

eshitachandwani: Same : The test is specifically to test the behaviour of `grpc.Dial` , to check that it doesn't change. The other tests use `grpc.NewClient`

test/proxy_test.go
eshitachandwani: Done

internal/transport/proxy.go
eshitachandwani: Done

internal/transport/proxy.go
eshitachandwani: It will append an empty string with the username. I have referenced it from the earlier code itself.

internal/transport/http2_client.go
eshitachandwani: Done.

test/proxy_test.go
```diff
@@ -0,0 +1,606 @@
...
+	proxyLis, errCh, doneCh, _ := setupProxy(t, backendAddr, false, func(req *http.Request) error {
+		if req.Method != http.MethodConnect {
+			return fmt.Errorf("unexpected Method %q, want %q", req.Method, http.MethodConnect)
+		}
+		if req.URL.Host != unresolvedTargetURI {
+			return fmt.Errorf("unexpected URL.Host %q, want %q", req.URL.Host, unresolvedTargetURI)
+		}
+		wantProxyAuthStr := "Basic " + base64.StdEncoding.EncodeToString([]byte(user+":"+password))
+		if got := req.Header.Get("Proxy-Authorization"); got != wantProxyAuthStr {
+			gotDecoded, _ := base64.StdEncoding.DecodeString(got)
```
purnesh42H, line 548: we should not ignore decoding errors. it should return error if decoding fails

internal/transport/proxy.go
purnesh42H: then user.Password() should just return empty string? it looks weird to ignore or atleast we should log a warning. So, if password is empty will the connection still succeed or we should do early failure?

test/proxy_test.go
eshitachandwani: Done.

test/proxy_test.go
```diff
@@ -0,0 +1,606 @@
...
+	case <-resolutionCh:
+		t.Logf("target resolution happened on client")
+	default:
+		t.Fatalf("Client-side resolution should be called but wasn't")
+	}
+}
+
+// Tests grpc.NewClient with grpc.WithNoProxy() set and verifies that it does
+// not dail to proxy, but directly to backend.
+func (s) TestGrpcNewClientWithNoProxy(t *testing.T) {
```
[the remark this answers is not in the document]
purnesh42H: yeah so right now if HTTPSProxyFromEnviornment is not present, it doesn't matter if `WithNoProxy` or `WithContextDialer` is set? The resolution will happen at client anyways? So having it present and not being called in these two cases is what we want to verify.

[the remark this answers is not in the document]
eshitachandwani: Done.

internal/proxyattributes/proxyattributes_test.go
```diff
@@ -0,0 +1,123 @@
...
+			// Validate ConnectAddr returns the correct connect address in the attribute.
+			if got := ConnectAddr(tt.addr); got != tt.want {
+				t.Errorf("ConnetAddr(%v) = %v, want %v ", tt.addr, got, tt.want)
+			}
+		})
+	}
+}
+
+// Tests that User returns the correct user in the attribute.
+func TestUser(t *testing.T) {
```
purnesh42H, line 73: `func (s)` ?

clientconn.go
```diff
@@ -225,7 +225,7 @@ func Dial(target string, opts ...DialOption) (*ClientConn, error) {
 func DialContext(ctx context.Context, target string, opts ...DialOption) (conn *ClientConn, err error) {
 	// At the end of this method, we kick the channel out of idle, rather than
 	// waiting for the first rpc.
-	opts = append([]DialOption{withDefaultScheme("passthrough")}, opts...)
+	opts = append([]DialOption{withDefaultScheme("passthrough"), WithTargetResolutionEnabled()}, opts...)
```
arjan-bal, line 228: Why is `WithTargetResolutionEnabled` added by default? IIUC, we want to keep the default behaviour the same as before, i.e. don't resolve the target address on the client.

dialoptions.go
```diff
@@ -377,7 +382,15 @@ func WithInsecure() DialOption {
...
 func WithNoProxy() DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
-		o.copts.UseProxy = false
+		o.UseProxy = false
+	})
+}
+
+// WithTargetResolutionEnabled returns a DialOption which enables target
+// resolution on client. This is ignored if WithNoProxy is used.
+func WithTargetResolutionEnabled() DialOption {
```
arjan-bal, line 391: Please mark this function as experimental. We keep new APIs experimental initially so they can be changed based on user feedback.
```
//
// # Experimental
//
// later release.
```

dialoptions.go
```diff
@@ -377,7 +382,15 @@ func WithInsecure() DialOption {
 // later release.
 func WithNoProxy() DialOption {
 	return newFuncDialOption(func(o *dialOptions) {
-		o.copts.UseProxy = false
```
arjan-bal, line 380: The godoc says the following:
```
This is ignored if WithDialer or WithContextDialer are used.
```
Since the proxy handling is now in the resolver, will the dialer always see the proxy's address? If so, this seems to be a behaviour change for users of custom dialers. Maybe we should skip proxy handling in the resolver if a custom dialer is being used to maintain backward compatability.

internal/internal.go
```diff
@@ -238,6 +238,11 @@ var (
 	// SetBufferPoolingThresholdForTesting updates the buffer pooling threshold, for
 	// testing purposes.
 	SetBufferPoolingThresholdForTesting any // func(int)
+
+	// HTTPSProxyFromEnvironmentForTesting returns the URL of the proxy to use
+	// for testing purposes. It is used to override the `http.ProxyFromEnvironment`
+	// function for testing purposes.
+	HTTPSProxyFromEnvironmentForTesting any // func(*http.Request) (*url.URL, error)
```
arjan-bal, line 245: You can use the correct type here instead of declaring it as `any`. I believe some functions are declared as `any` to avoid circular dependencies among gRPC packages. Since this function doesn't use gRPC types, it should be fine to use the specific type.

internal/proxyattributes/proxyattributes.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/url"
+
+	"google.golang.org/grpc/resolver"
+)
+
+type keyType string
+
+const userAndConnectAddrKey = keyType("grpc.resolver.delegatingresolver.userAndConnectAddr")
+
+type userAndConnectAddr struct {
```
arjan-bal, line 33: You can export this type. In populate instead of accepting individual vales, accept `UserAndConnectAddr` as a parameter. Create a single function to get the `UserAndConnectAddr` from a `resolver.address`.

arjan-bal, line 33: Use a more generic name like `proxyOptions`. By using specific names like `userAndConnectAddr`, you would need to keep updating the name every time you want to add/remove a field.

internal/proxyattributes/proxyattributes_test.go
```diff
@@ -0,0 +1,123 @@
+/*
```
arjan-bal, line 1: optional: I haven't completed the entire review. If you can test the setting of proxy attributes in an e2e style test, you can get rid of these tests.

dialoptions.go
```diff
@@ -94,6 +94,11 @@ type dialOptions struct {
 	idleTimeout                 time.Duration
 	defaultScheme               string
 	maxCallAttempts             int
+	// TargetResolutionEnabled indicates whether the client should perform
+	// target resolution even when a proxy is enabled.
+	TargetResolutionEnabled bool
+	// UseProxy specifies if a proxy should be used.
+	UseProxy bool
```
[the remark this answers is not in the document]
eshitachandwani: Done.

dialoptions.go
eshitachandwani: Done.

internal/proxyattributes/proxyattributes.go
eshitachandwani: Done.

eshitachandwani: Changed it to Options so that it can be used as proxyattributes.Options

internal/proxyattributes/proxyattributes_test.go
eshitachandwani: Done.

internal/internal.go
```diff
@@ -238,6 +240,11 @@ var (
 	// SetBufferPoolingThresholdForTesting updates the buffer pooling threshold, for
 	// testing purposes.
 	SetBufferPoolingThresholdForTesting any // func(int)
+
+	// HTTPSProxyFromEnvironmentForTesting returns the URL of the proxy to use
+	// for testing purposes. It is used to override the `http.ProxyFromEnvironment`
+	// function for testing purposes.
+	HTTPSProxyFromEnvironmentForTesting func(*http.Request) (*url.URL, error)
```
arjan-bal, line 247: You can move these to a new `resolver/delegatingresolver/internal` module since only delegating resolver tests need to use it. See the [pickfirst/internal](https://github.com/grpc/grpc-go/blob/master/balancer/pickfirst/internal/internal.go) package as an example.

arjan-bal: As discussed offline, revert this back to `any` and remove the `"net/http", "net/url"` imports and the dependency tests should pass again.

clientconn.go
arjan-bal: Here is my understanding:
1 `grpc.Dial()` uses `passthrough` by default.
2. `passthrough` doesn't resolve target hostnames on the client.
3. `WithTargetResolutionEnabled` causes target hostnames to be resolved on the client.

From 1 & 2, `grpc.Dial()` doesn't resolve target hostnames on the client by default. Adding `WithTargetResolutionEnabled` seems to break this behaviour.

eshitachandwani: Actually, using the `WithTargetResolutionEnabled` will just use the target resolver and not by pass resolution. So if someone uses `WithTargetResolutionEnabled` with `grpc.Dial` and default resolver, it will use passthrough and not do the resoltion, but if they use "dns" with `grpc.Dial` it will use dns resolver and do the resolution. This will preserve the behaviour.

internal/internal.go
arjan-bal: This is change is from the other PR. Resolving.

arjan-bal (commented):
I haven't reviewed the tests yet, will take a look at them next.

internal/transport/proxy_test.go
arjan-bal, line 1: Is there a reason for moving proxy tests out of `internal/transport`? It's preferable to keep tests in separate packages as it allows tests to run in parallel.

internal/testutils/proxy.go
arjan-bal, line 1: Is there a reason to move the proxy server into testutils? If there is only one consumer of this, it should remain with in the same package. When the need arises, we can move this into a common place.

internal/transport/proxy_test.go
arjan-bal: It looks like these tests are deleted, not moved. We need these tests.

eshitachandwani: Whatever was being tested in the tests is being tested in the E2E tests, basic auth and connect requests and connecting to the proxy server. The mapAddress function was moved to delegating resolver and that is being tested in delegating resolver. Do we need to test is again in unit tests?

clientconn.go
```diff
@@ -225,7 +225,14 @@ func Dial(target string, opts ...DialOption) (*ClientConn, error) {
 func DialContext(ctx context.Context, target string, opts ...DialOption) (conn *ClientConn, err error) {
 	// At the end of this method, we kick the channel out of idle, rather than
 	// waiting for the first rpc.
-	opts = append([]DialOption{withDefaultScheme("passthrough")}, opts...)
+	//
+	// WithTargetResolutionEnabled in `grpc.Dial` ensures that it preserves
+	// behavior: when default scheme passthrough is used, skip hostname
+	// resolution, when any other scheme like "dns" is used for resolution,
```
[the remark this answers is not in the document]
eshitachandwani: Done.

clientconn.go
```diff
@@ -225,7 +225,14 @@ func Dial(target string, opts ...DialOption) (*ClientConn, error) {
...
+	// perform resolution on the client as expected.
```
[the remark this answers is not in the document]
eshitachandwani: Done.

clientconn.go
```diff
@@ -225,7 +225,14 @@ func Dial(target string, opts ...DialOption) (*ClientConn, error) {
...
+	opts = append([]DialOption{withDefaultScheme("passthrough"), WithTargetResolutionEnabled()}, opts...)
+
+	opts = append([]DialOption{withDefaultScheme("passthrough"), WithTargetResolutionEnabled()}, opts...)
```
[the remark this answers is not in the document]
eshitachandwani: Done.

dialoptions.go
```diff
@@ -94,6 +94,8 @@ type dialOptions struct {
 	idleTimeout                 time.Duration
 	defaultScheme               string
 	maxCallAttempts             int
+	targetResolutionEnabled     bool // Specifies if client should perform target resolution when proxy is enabled.
```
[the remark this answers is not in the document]
eshitachandwani: Done.

internal/testutils/proxy.go
eshitachandwani: Done.

internal/transport/http2_client.go
```diff
@@ -153,8 +154,12 @@ type http2Client struct {
 	logger       *grpclog.PrefixLogger
 }
 
-func dial(ctx context.Context, fn func(context.Context, string) (net.Conn, error), addr resolver.Address, useProxy bool, grpcUA string) (net.Conn, error) {
+func dial(ctx context.Context, fn func(context.Context, string) (net.Conn, error), addr resolver.Address, grpcUA string) (net.Conn, error) {
 	address := addr.Addr
+
+	if _, present := proxyattributes.Get(addr); present {
+		return proxyDial(ctx, addr, grpcUA)
```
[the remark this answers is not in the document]
eshitachandwani: Done.

dialoptions.go
```diff
@@ -384,7 +384,8 @@ func WithNoProxy() DialOption {
 }
 
 // WithTargetResolutionEnabled returns a DialOption which enables target
-// resolution on client. This is ignored if WithNoProxy is used.
+// resolution on client even when "dns" scheme is used. This is ignored if
```
arjan-bal, line 387: I think we should mention when `a proxy is used along with the the "dns" scheme`.

internal/transport/proxy_test.go
arjan-bal: [TestHTTPConnectWithServerHello](https://github.com/grpc/grpc-go/blob/e912015fd3f4aabdff6d6cf835e321c19a204afb/internal/transport/proxy_test.go#L215-L229) is added to verify that the optimization added in https://github.com/grpc/grpc-go/pull/7424 doesn't effect correctness. It is difficult to test it in en e2e fashion because it needs the server the server to send the first message after creation of the TCP connection and the proxy to buffer this message along with the CONNECT response. I think the other tests can be replaced.

test/proxy_test.go
```diff
@@ -0,0 +1,716 @@
...
+	origHTTPSProxyFromEnvironment := delegatingresolver.HTTPSProxyFromEnvironment
+	delegatingresolver.HTTPSProxyFromEnvironment = func(req *http.Request) (*url.URL, error) {
+		return httpproxy.FromEnvironment().ProxyFunc()(req.URL)
+	}
+	defer func() {
+		delegatingresolver.HTTPSProxyFromEnvironment = origHTTPSProxyFromEnvironment
+	}()
+
+	// Configure manual resolvers for both proxy and target backends
+	targetResolver := setupDNS(t)
```
arjan-bal, line 265: We can use the real DNS resolver to resolve "localhost". Why use a fake hostname and manual resolver?

eshitachandwani: I cannot use a local host because the `http.FromProxyEnvirnoment` as well as `httpproxy.FromEnviornment` returns a nil proxy URL if the target URI is local host. The documentation says :

[`http.FromProxyEnvirnoment`](https://pkg.go.dev/net/http#ProxyFromEnvironment)
As a special case, if req.URL.Host is "localhost" (with or without a port number), then a nil URL and nil error will be returned.

[`http.FromProxyEnvirnoment`](https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config.ProxyFunc) 
As a special case, if req.URL.Host is "localhost" or a loopback address (with or without a port number), then a nil URL and nil error will be returned.

test/proxy_test.go
```diff
@@ -0,0 +1,716 @@
...
+// established to the proxy server, sends the unresolved target URI in the HTTP
+// CONNECT request, and is successfully connected to the backend server.
+func (s) TestGRPCDialWithProxy(t *testing.T) {
+	backendAddr := createAndStartBackendServer(t)
+	pLis, errCh, doneCh, _ := setupProxy(t, backendAddr, false, requestCheck(unresolvedTargetURI))
+
+	// Overwrite the proxy environment and restore it after the test.
+	proxyEnv := os.Getenv("HTTPS_PROXY")
+	os.Setenv("HTTPS_PROXY", pLis.Addr().String())
+	defer func() { os.Setenv("HTTPS_PROXY", proxyEnv) }()
```
arjan-bal, line 203: Can use `t.SetEnv` that automatically resets the value as part of test cleanup.

internal/transport/proxy_utils.go
```diff
@@ -0,0 +1,186 @@
...
+	"google.golang.org/grpc/internal/testutils"
+)
+
+const defaultTestTimeout = 10 * time.Second
+
+// RequestCheck returns a function that validates the CONNECT method and
+// determines whether the address is resolved or unresolved. It only checks if
+// the address is an IP address or not, relying on RPC failures to handle
+// incorrect resolutions.
+func RequestCheck(isResolved bool) func(*http.Request) error {
```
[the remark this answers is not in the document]
arjan-bal: To clarify, this can be changed to something like:
```go
func isIPAddr(addr string) bool {
     // omitted
}
```

Then in the callers, you can use this utility function as follows:
```go
reqCheck := func(req *http.Request) {
    if got, want := isIPAddr(req.URL.Host); got != want {
        t.Errorf("isIPAddr(%q) = %t, want = false", req.URL.Host,   got, want)
    }
}

_ := SetupProxy(t, reqCheck, false)
```

internal/transport/proxy_test.go
eshitachandwani: Revived the test.

internal/transport/proxy_ext_test.go
```diff
@@ -0,0 +1,496 @@
...
+		if req.URL.Host == unresolvedTargetURI {
+			return &url.URL{
+				Scheme: "https",
+				User:   url.UserPassword(user, password),
+				Host:   pServer.Addr,
+			}, nil
+		}
+		t.Errorf("Unexpected request host to proxy: %s want %s", req.URL.Host, unresolvedTargetURI)
+		return nil, nil
+	}
```
arjan-bal, line 478: To test the integration with `http.ProxyFromEnvironment`, you can set environment variable and avoid mocking this. Set the hostname like `test` which according to [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html) is always unresolvable. In `onRequest` do the validation, set a variable indicating the proxy was called and then cancel the request context to make the RPC fail. You can assert that the proxy is called at the end of the test.

This way we can avoid any DNS mocking code in the proxy server while still using the real function that reads env vars.

arjan-bal (commented):
It almost looks good to me. I would like to get another review, especially for the external API (the dialoption) and it's godocs.

internal/transport/proxy_ext_test.go
eshitachandwani: Done.

internal/transport/testutils/testutils.go
```diff
@@ -126,25 +93,23 @@ func (p *ProxyServer) handleRequest(t *testing.T, in net.Conn, proxyStarted func
...
+
+	go io.Copy(p.in, p.out)
+	go io.Copy(p.out, p.in)
 }
 
-// Creates and starts a proxy server.
-func newProxyServer(t *testing.T, lis net.Listener, reqCheck func(*http.Request) error, proxyStarted func(), waitForServerHello bool) *ProxyServer {
+// SetupProxy initializes and starts a proxy server, registers a cleanup to
+// stop it, and returns the proxy's listener and helper channels.
+func SetupProxy(t *testing.T, reqCheck func(*http.Request), waitForServerHello bool) *ProxyServer {
```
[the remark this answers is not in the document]
arjan-bal: Okay.

internal/transport/testutils/testutils.go
```diff
@@ -126,25 +93,23 @@ func (p *ProxyServer) handleRequest(t *testing.T, in net.Conn, proxyStarted func
...
 	t.Helper()
+	pLis, err := net.Listen("tcp", "localhost:0")
```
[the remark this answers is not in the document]
arjan-bal: I know, it's probably created so people don't have guess the correct `host:port` combination that works in all envs.

internal/transport/proxy_ext_test.go
```diff
@@ -0,0 +1,496 @@
...
+		if req.Method != http.MethodConnect {
+			t.Errorf("unexpected Method %q, want %q", req.Method, http.MethodConnect)
+		}
+		host, _, err := net.SplitHostPort(req.URL.Host)
+		if err != nil {
+			t.Error(err)
+		}
+		if got, want := isIPAddr(host), false; got != want {
+			t.Errorf("isIPAddr(%q) = %t, want = %t", host, got, want)
+		}
```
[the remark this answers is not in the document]
arjan-bal: I believe the only possible value for `unresolvedTargetURI` when you expect a hostname (not IP) is `localhost`. Either way is fine, for tests the easier it is to see the exact values being asserted, the better. It increases the chances of spotting problems while reading the code.

internal/transport/proxy_ext_test.go
```diff
@@ -0,0 +1,496 @@
...
+		t.Errorf("Unexpected request host to proxy: %s want %s", req.URL.Host, unresolvedTargetURI)
+		return nil, nil
+	}
+	orighpfe := delegatingresolver.HTTPSProxyFromEnvironment
+	delegatingresolver.HTTPSProxyFromEnvironment = hpfe
+	defer func() { delegatingresolver.HTTPSProxyFromEnvironment = orighpfe }()
+
+	// Create and update a custom resolver for target URI.
+	resolvedTargetURI := fmt.Sprintf("[::0]:%d", testutils.ParsePort(t, unresolvedTargetURI))
```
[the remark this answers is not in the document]
arjan-bal: I don't see the benefit of hiding the stubserver. On the flipside, it makes the function harder to re-use for other tests that want to start the server, but also do other stuff with it.

internal/transport/proxy_ext_test.go
arjan-bal: I was thinking of removing the mocking from this test instead of adding a new one. A major part of this test is the logic to read the poxy config from the env, which is presently being mocked.

internal/transport/testutils/testutils.go
```diff
@@ -60,7 +60,9 @@ func (p *ProxyServer) handleRequest(t *testing.T, in net.Conn, waitForServerHell
 		t.Errorf("failed to read CONNECT req: %v", err)
 		return
 	}
-
+	if req.Method != http.MethodConnect {
+		t.Fatalf("unexpected Method %q, want %q", req.Method, http.MethodConnect)
```
arjan-bal, line 64: We should not call `t.Fatalf` in a separate go routine, only on the main goroutine running the test.

internal/transport/proxy_ext_test.go
```diff
@@ -0,0 +1,540 @@
...
+	_, err := netip.ParseAddr(addr)
+	return err == nil
+}
+
+// Tests the scenario where a gRPC client retrieves the proxy address from the
+// environment variable `HTTPS_PROXY`. The test sets up an HTTP proxy server
+// and verifies that the client connects to this proxy, sends the unresolved
+// target URI in the HTTP CONNECT request.
+func (s) TestHTTPProxyFromEnvironment(t *testing.T) {
+	unresolvedTargetURI := "example.test.com"
```
arjan-bal, line 77: nit: Remove the `com` suffix.

eshitachandwani: Done.

internal/transport/proxy_ext_test.go
```diff
@@ -61,33 +61,64 @@ func createAndStartBackendServer(t *testing.T) string {
...
+
 // Tests the scenario where grpc.Dial is performed using a proxy with the
 // default resolver in the target URI. The test verifies that the connection is
 // established to the proxy server, sends the unresolved target URI in the HTTP
 // CONNECT request and is successfully connected to the backend server.
 func (s) TestGRPCDialWithProxy(t *testing.T) {
-	unresolvedTargetURI := createAndStartBackendServer(t)
+	backendAddr := startBackendServer(t)
+	unresolvedTargetURI := fmt.Sprintf("localhost:%d", testutils.ParsePort(t, backendAddr))
+	proxyStarted := false
```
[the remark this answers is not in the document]
eshitachandwani: Yeah, done!

internal/transport/proxy_ext_test.go
eshitachandwani: Right.Done. But used `httpproxy.FromEnvironment` function because the actual `http.FromProxyEnvironment` function just reads the environment once and then uses that for all the tests, so when the test is run multiple times, it will fail.

internal/transport/testutils/testutils.go
eshitachandwani: Noted. Changed.

internal/transport/proxy_ext_test.go
```diff
@@ -504,21 +477,15 @@ func (s) TestBasicAuthInNewClientWithProxy(t *testing.T) {
...
-				Host:   pServer.Addr,
-			}, nil
-		}
-		t.Errorf("Unexpected request host to proxy: %s want %s", req.URL.Host, unresolvedTargetURI)
-		return nil, nil
+	t.Setenv("HTTPS_PROXY", user+":"+password+"@"+pServer.Addr)
+
+	origHTTPSProxyFromEnvironment := delegatingresolver.HTTPSProxyFromEnvironment
+	delegatingresolver.HTTPSProxyFromEnvironment = func(req *http.Request) (*url.URL, error) {
+		return httpproxy.FromEnvironment().ProxyFunc()(req.URL)
```
arjan-bal, line 484: nit: For the benefit of future readers, please add a comment mentioning this done to read the env var in every test run.

arjan-bal (approved):

dialoptions.go
```diff
@@ -377,7 +379,21 @@ func WithInsecure() DialOption {
...
+
+// WithTargetResolutionEnabled returns a DialOption which enables target
+// resolution on client when a proxy is used along with the the "dns" scheme.
+// This is ignored if WithNoProxy is used.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a
+// later release.
+func WithTargetResolutionEnabled() DialOption {
```
dfawley, line 394: Hmm, how about `WithLocalDNSResolution()`?  So the documentation says something like `WithLocalDNSResolution forces local DNS name resolution even when a proxy is specified in the environment.  By default, the server name is provided directly to the proxy as part of the CONNECT handshake.`?

Otherwise I'm worried this name might be a little too vague for users to understand the behavior of it.

internal/transport/http2_client.go
```diff
@@ -153,8 +154,12 @@ type http2Client struct {
 	logger       *grpclog.PrefixLogger
 }
 
-func dial(ctx context.Context, fn func(context.Context, string) (net.Conn, error), addr resolver.Address, useProxy bool, grpcUA string) (net.Conn, error) {
+func dial(ctx context.Context, fn func(context.Context, string) (net.Conn, error), addr resolver.Address, grpcUA string) (net.Conn, error) {
 	address := addr.Addr
+
+	if opts, present := proxyattributes.Get(addr); present {
+		return proxyDial(ctx, addr, grpcUA, opts)
+	}
```
dfawley, line 162: Why do we want this to take higher precedence than a custom dialer?

I see that we later make it so a custom dialer disables the delegating resolver that would set the proxy attribute.  But what about if xDS is used and the proxy is set that way, and the user has a custom dialer?  I would expect the custom dialer to have higher precedence, probably, since that one is set directly by the application.

It is admittedly a weird case, however.

internal/transport/testutils/testutils.go
dfawley, line 1: I don't think we want a separate package for this unless it's something that needs to be shared with other (non-transport) packages, in which case it should usually go into `grpc/internal/...`.  You can put the contents in `proxy_test.go` or for similar things if they're too big you can put them in other `_test.go` files to make sure they're only compiled for tests.

resolver_wrapper.go
```diff
@@ -78,7 +79,16 @@ func (ccr *ccResolverWrapper) start() error {
 			Authority:            ccr.cc.authority,
 		}
 		var err error
-		ccr.resolver, err = ccr.cc.resolverBuilder.Build(ccr.cc.parsedTarget, ccr, opts)
+		// The delegating resolver is used unless:
+		//   - A custom dialer is provided via WithContextDialer dialoption.
+		//   - Proxy usage is disabled through WithNoProxy dialoption.
```
dfawley, line 84: Nit: this list should ideally say `, or` after the first bullet, since it's not clear from the text here whether these conditions are ORs or ANDs.

dialoptions.go
eshitachandwani: Done.

internal/transport/testutils/testutils.go
eshitachandwani: Done.

internal/testutils/proxy.go
```diff
@@ -16,7 +16,6 @@
  *
  */
 
-// Package testutils contains helpers for transport tests.
 package testutils
```
dfawley, line 19: Could you please make this a separate package instead (subdirectory of this one)?  Generally we should split things when possible instead of building one gigantic package named "utils".  See https://google.github.io/styleguide/go/best-practices#util-packages and also https://google.github.io/styleguide/go/best-practices#package-size.  (Re: package-size: of course this isn't exported, so it's not critical, but our testing infrastructure grows over time and we want to make sure it's easy to use for ourselves.  This package is already probably too big.)

internal/testutils/proxy.go
```diff
@@ -0,0 +1,130 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+)
+
+// ProxyServer represents a test proxy server.
```
dfawley, line 32: It seems this implementation has a lot of limitations.  These should be documented.

For example, it only supports one connection (ever? or just concurrently?).

internal/testutils/proxy.go
```diff
@@ -0,0 +1,130 @@
...
+	p.in = in
+	p.in.Write(buf.Bytes())
+	p.out = out
+
+	go io.Copy(p.in, p.out)
+	go io.Copy(p.out, p.in)
+}
+
+// HTTPProxy initializes and starts a proxy server, registers a cleanup to
+// stop it, and returns the proxy's listener and helper channels.
```
dfawley, line 101: It's not returning those things, it's returning a `ProxyServer`.  And that only has one thing that's readable, the `Addr` field.

internal/testutils/proxy.go
```diff
@@ -0,0 +1,130 @@
...
+				return
+			}
+			// p.handleRequest is not invoked in a goroutine because the test
+			// proxy currently supports handling only one connection at a time.
+			p.handleRequest(t, in, waitForServerHello)
+		}
+	}()
+	t.Logf("Started proxy at: %q", pLis.Addr())
+	t.Cleanup(p.stop)
+	p.Addr = fmt.Sprintf("localhost:%d", ParsePort(t, pLis.Addr().String()))
```
dfawley, line 128: Can this go into the struct initialization instead of happening separately?

dfawley, line 128: Also why can't this just be `pLis.Addr().String()`?  There should be a comment explaining why it's doing all this extra stuff to seemingly just convert `localhost:123` into `localhost:123`.

internal/transport/proxy.go
```diff
@@ -81,15 +63,15 @@ func doHTTPConnectHandshake(ctx context.Context, conn net.Conn, backendAddr stri
...
 		Method: http.MethodConnect,
-		URL:    &url.URL{Host: backendAddr},
+		URL:    &url.URL{Host: opts.ConnectAddr},
 		Header: map[string][]string{"User-Agent": {grpcUA}},
 	}
-	if t := proxyURL.User; t != nil {
-		u := t.Username()
-		p, _ := t.Password()
+
+	if user := opts.User; user != (url.Userinfo{}) {
```
dfawley, line 70: This comparison with a zero-initialized Userinfo looks a little off to me.  Should this field in `opts` be a pointer instead to determine whether it was set intentionally?

Or should it check if either `Username()` or `Password()` are non-empty?

I'm not sure exactly what the right behavior is, but this seems questionable, as it's including any unexported fields in the comparison.

Related but potential existing problem to investigate: what if username is set but password is not?  Is that valid?  If so, is the proper behavior to encode it as `username:`, as it does currently, or should the `:` be omitted?

eshitachandwani: I have added a `UserSet` option as changing it to a pointer can allow the caller to modify the use info.

internal/testutils/proxy.go
eshitachandwani: Added.

internal/transport/proxy.go
eshitachandwani: Got it ! Done.

dfawley (approved):
One last thing otherwise LGTM.  Thanks!

internal/testutils/proxyserver/proxyserver.go
```diff
@@ -102,9 +102,9 @@ func (p *ProxyServer) handleRequest(t *testing.T, in net.Conn, waitForServerHell
 	go io.Copy(p.out, p.in)
 }
 
-// HTTPProxy initializes and starts a proxy server, registers a cleanup to
+// NewHTTPProxy initializes and starts a proxy server, registers a cleanup to
 // stop it, and returns a ProxyServer.
-func HTTPProxy(t *testing.T, reqCheck func(*http.Request), waitForServerHello bool) *ProxyServer {
+func NewHTTPProxy(t *testing.T, reqCheck func(*http.Request), waitForServerHello bool) *ProxyServer {
```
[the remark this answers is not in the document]
eshitachandwani: Right! Done!

PR #7891, opened 2024-12-03
test: Split import paths for message and service generated code

arjan-bal:
This is required to make bazel/blaze build work.

RELEASE NOTES: N/A


purnesh42H (approved):
lgtm

purnesh42H:
May be we should add the vet rule?

arjan-bal:
> May be we should add the vet rule?

We can discuss this offline, with other maintainers.

Issue #2886, opened 2019-06-27
Server should send RST_STREAM when deadline is exceeded


ejona86:
The server code relies on the application to return when the deadline is exceeded. But when it does return, it doesn't cancel the RPC; it sends a graceful close:
https://github.com/grpc/grpc-go/blob/v1.21.1/server.go#L1010

If the deadline is exceeded when the handler returns, the RPC should be cancelled and RST_STREAM should be sent (if one hasn't already been received from the client).

ejona86:
Note that this may have interactions with ServerInHandle, if ServerInHandle changes the deadline. It is only appropriate to use RST_STREAM when the client specified its deadline, since in that case the client will ignore what the server sends anyway. If the ServerInHandler reduced the deadline, then a normal close is appropriate.

infovivek2020:
@purnesh42H Please assign to me

janardhanvissa:
@purnesh42H please assign me this issue

arjan-bal:
I was going through the code to understand the parts responsible for sending RST_STREAM frames on context timeouts. It seems that only gRPC Go clients send RST_STREAMs on context expiration though two different paths.

### First path
To read data from a stream, both server and client streams call the read method on the shared `Stream` struct.
https://github.com/grpc/grpc-go/blob/ee3e8d90c4da9d7131433135304b77f2385121e1/internal/transport/client_stream.go#L54-L60
https://github.com/grpc/grpc-go/blob/ee3e8d90c4da9d7131433135304b77f2385121e1/internal/transport/server_stream.go#L53-L59

Stream.read() eventually calls the [recvBufferReader.Read](https://github.com/grpc/grpc-go/blob/ee3e8d90c4da9d7131433135304b77f2385121e1/internal/transport/transport.go#L154C11-L154C32). This method has different paths for server and client streams:
https://github.com/grpc/grpc-go/blob/ee3e8d90c4da9d7131433135304b77f2385121e1/internal/transport/transport.go#L167-L171
 
`readClient` calls `r.closeStream()` if it detects a context timeout.
https://github.com/grpc/grpc-go/blob/ee3e8d90c4da9d7131433135304b77f2385121e1/internal/transport/transport.go#L220-L241

`r.closeStream` is set only for grpc clients and calls the `ClientStream.Close` method.
https://github.com/grpc/grpc-go/blob/ee3e8d90c4da9d7131433135304b77f2385121e1/internal/transport/client_stream.go#L62-L73

## Second path
[...]

PR #7892, opened 2024-12-03
grpc: server should send RST_STREAM when deadline is exceeded

misvivek:
Fixes: #2886

### DIFF
--- a/internal/transport/handler_server_test.go
+++ b/internal/transport/handler_server_test.go
@@ -394,7 +394,18 @@ func (s) TestHandlerTransport_HandleStreams_Timeout(t *testing.T) {
 			return
 		}
 		s.WriteStatus(status.New(codes.DeadlineExceeded, "too slow"))
+		// rst flag setting to verify the noop function: signalDeadlineExceeded
+		ch := make(chan struct{}, 1)
+		origSignalDeadlineExceeded := signalDeadlineExceeded
+		signalDeadlineExceeded = func() {
+			ch <- struct{}{}
+		}
+		defer func() {
+			signalDeadlineExceeded = origSignalDeadlineExceeded
+		}()
+
 	}
+
 	ht.HandleStreams(
 		context.Background(), func(s *ServerStream) { go runStream(s) },
 	)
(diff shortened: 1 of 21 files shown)


internal/transport/handler_server_test.go
```diff
@@ -392,7 +393,22 @@ func (s) TestHandlerTransport_HandleStreams_Timeout(t *testing.T) {
...
+		ch := make(chan struct{}, 1)
+		origSignalDeadlineExceeded := signalDeadlineExceeded
+		signalDeadlineExceeded = func() {
+			ch <- struct{}{}
+		}
+		defer func() {
+			signalDeadlineExceeded = origSignalDeadlineExceeded
+		}()
 		s.WriteStatus(status.New(codes.DeadlineExceeded, "too slow"))
+		select {
```
purnesh42H, line 406: you have to wait on channel as well otherwise you are not verifying the if your function was triggered

```
select {
case <-ch:  // Signal received, continue with the test
case <-s.ctx.Done():
case <-time.After(5 * time.Second):
	t.Errorf("timeout waiting for ctx.Done")
	return
}
```

misvivek: done

purnesh42H: @misvivek i just saw you are doing this inside goroutine which won't affect the execution of test whether pass or fail.

You need to move both no op function setting and channel waiting code outside of go routine and check at the end when go runStream(s) is done, perhaps at the end after checking the trailers. So I would structure it like this 

```
// rst flag setting to verify the noop function: signalDeadlineExceeded
ch := make(chan struct{}, 1)
origSignalDeadlineExceeded := signalDeadlineExceeded
signalDeadlineExceeded = func() {
ch <- struct{}{}
}
defer func() {
signalDeadlineExceeded = origSignalDeadlineExceeded
}()

runStream := func(s *ServerStream) {....}

...
...

checkHeaderAndTrailer(t, rw, wantHeader, wantTrailer)
select {
case <-ch: // Signal received, continue with the test
case <-time.After(5 * time.Second):
t.Errorf("timeout waiting for ctx.Done")
return
}
```

purnesh42H: And i think the test will fail then because you haven't yet fixed the bug.

purnesh42H:
@misvivek you need to write the as i mentioned above and then verify if the test is failing. If it is, then we have to fix the bug first.

purnesh42H (commented):
Need to fix the bug as well along with test

purnesh42H:
Closing this because of no activity from author

PR #7909, opened 2024-12-06
xds/internal: provide xDS node ID in control plane errors

townba:
To fix internal b/364405720 ("Provide xDS node id in control plane errors (Golang)").

RELEASE NOTES: none


dfawley:
Closing until a strategy for logging (etc) node ID is decided.

Issue #7856, opened 2024-11-19
passthrough scheme not documented


rittneje:
As far as I can tell, the `passthrough:///` scheme does not appear to be documented anywhere. It is needed in order to properly use the bufconn package.

It is mentioned in this comment: https://github.com/grpc/grpc-go/issues/7091#issuecomment-2037663115

It is not mentioned in the language agnostic gRPC docs. https://github.com/grpc/grpc/blob/master/doc/naming.md

purnesh42H:
@rittneje yes, passthrough is only applicable in grpc-go and is already deprecated https://github.com/grpc/grpc-go/blob/master/resolver/passthrough/passthrough.go#L22. It only exist for backward compatibility right now.

rittneje:
@purnesh42H First, the file you linked only mentions the _package_ is deprecated and should not be imported. It doesn't mention anything about the scheme itself. Second, it is the only way to get bufconn to work, so you can't just deprecate it without a proper replacement. Please re-open this issue.

dfinkel:
FWIW, my team switched our bufconn uses to `localhost` as the address, since that's guaranteed to resolve with the DNS resolver.

purnesh42H:
yeah, its not explicitly documented that passthrough is deprecated apart from source code but that's because it is internal to grpc-go. In release 1.64, we deprecated `grpc.Dial` to use `grpc.NewClient` https://github.com/grpc/grpc-go/releases/tag/v1.64.0. With `grpc.Dial` the default resolution scheme is passthrough which uses passthrough resolver that i linked above. With `grpc.NewClient` the default scheme is `dns`. If you want to use passthrough with `grpc.NewClient`, you need to pass the adderss prefixed with `passthrough:///`.

Could you describe what issue you are facing with bufcon though?

rittneje:
@purnesh42H With bufconn, we need to use `grpc.WithContextDialer` to wire it up correctly. In this case, the target passed to `grpc.NewClient` is immaterial, but we need to prefix it with `passthrough:///` or else it will try to do a DNS resolution anyway and fail.

As `passthrough:///` is not documented (and apparently deprecated with no replacement) the only way to figure this out is to stumble across the GitHub comment I originally mentioned. Instead:
1. the package documentation for google.golang.org/grpc/test/bufconn should directly tell you what to do
2. either `passthrough:///` needs to be un-deprecated, or a proper replacement needs to be added and mentioned in the documentation

PR #7916, opened 2024-12-10
grpcs: update `WithContextDialer` documentation to include using passthrough resolver

purnesh42H:
Fixes: #7856

RELEASE NOTES: None

### DIFF
--- a/dialoptions.go
+++ b/dialoptions.go
@@ -428,10 +428,10 @@ func WithTimeout(d time.Duration) DialOption {
 // returned by f, gRPC checks the error's Temporary() method to decide if it
 // should try to reconnect to the network address.
 //
-// If you want to bypass DNS resolution and connect directly to a specified
-// address, prefix your target address with passthrough:/// when using
-// WithContextDialer. This will cause the passthrough resolver to be used,
-// which avoids any DNS lookups. For example: passthrough:///localhost:50051.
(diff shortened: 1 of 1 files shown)


purnesh42H:
@dfawley could you ptal?

dialoptions.go
```diff
@@ -428,6 +428,11 @@ func WithTimeout(d time.Duration) DialOption {
 // returned by f, gRPC checks the error's Temporary() method to decide if it
 // should try to reconnect to the network address.
 //
+// If you want to bypass DNS resolution and connect directly to a specified
+// address, prefix your target address with passthrough:/// when using
+// WithContextDialer. This will cause the passthrough resolver to be used,
+// which avoids any DNS lookups. For example: passthrough:///localhost:50051.
```
dfawley, line 434: "you" is too informal for docstrings.  Proposed change:

```suggestion
// Note that gRPC by default performs name resolution on the target passed to NewClient.
// To bypass name resolution and cause the target string to be passed directly to the dialer
// here instead, use the "passthrough" resolver by specifying it in the target string,
// e.g. "passthrough:target".
```

(Let's not suggest `localhost:port` since that should work fine with DNS.  This would be if DNS _doesn't work with_ a target string.)

purnesh42H: Done

dfawley (approved):

dfawley:
Thanks!

PR #7917, opened 2024-12-10
examples/features/gracefulstop: fix printf to use format specifier

eshitachandwani:
RELEASE NOTES : None

### DIFF
--- a/examples/features/gracefulstop/client/main.go
+++ b/examples/features/gracefulstop/client/main.go
@@ -65,7 +65,7 @@ func main() {
 		}
 		unaryRequests++
 		time.Sleep(200 * time.Millisecond)
-		log.Printf(r.Message)
+		log.Printf("%s",r.Message)
 	}
 	log.Printf("Successful unary requests made by client: %d", unaryRequests)
 


examples/features/gracefulstop/client/main.go
```diff
@@ -65,7 +65,7 @@ func main() {
 		}
 		unaryRequests++
 		time.Sleep(200 * time.Millisecond)
-		log.Printf(r.Message)
+		log.Printf("%s", r.Message)
```
arjan-bal, line 68: What does the linter say about `log.Printf(r.Message)`? 

Can we instead change this to `log.Print(r.Message)` which is more direct?

eshitachandwani: I cannot get the error msg again, but this the page it linked to : https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/printf
for this part  : 

Conversely, it also reports calls to Printf-like functions with a non-constant format string and no other arguments:
`fmt.Printf(message) // non-constant format string in call to fmt.Printf`
Such calls may have been intended for the function's Print-like counterpart: if the value of message happens to contain "%", misformatting will occur. In this case, the checker additionally suggests a fix to turn the call into:
`fmt.Printf("%s", message)`

eshitachandwani: > Can we instead change this to `log.Print(r.Message)` which is more direct?

The fix mentioned with the error, as well in the doc is to use `fmt.Printf()` with format specifier.

PR #7918, opened 2024-12-10
grpc: fix message length checks when compression is enabled and maxReceiveMessageSize is MaxInt

vinothkumarr227:
Fixes https://github.com/grpc/grpc-go/issues/4552

RELEASE NOTES:

- grpc: fix receiving empty messages when compression is enabled and maxReceiveMessageSize is maxInt


rpc_util.go
```diff
@@ -874,14 +874,35 @@ func decompress(compressor encoding.Compressor, d mem.BufferSlice, maxReceiveMes
...
 	}
 
-	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)+1), pool)
+	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
 	if err != nil {
 		out.Free()
 		return nil, 0, err
 	}
+	if err = checkReceiveMessageOverflow(int64(out.Len()), int64(maxReceiveMessageSize), dcReader); err != nil {
+		return nil, out.Len() + 1, err
```
arjan-bal, line 883: Adding 1 to `out.Len()` can result in an overflow on 32 bit systems. The  following part of the function seems strange to me: 
> Optionally, if data will be over maxReceiveMessageSize, just return the size

I suggest making the following change to avoid this:
1. Declare a global error for indicating that the max receive size is exceeded:
```go
var	errMaxMessageSizeExceeded = errors.New("max message size exceeded")
```  
2. When the check here fails, `nil, 0, errMaxMessageSizeExceeded`. Update the godoc to mention the same.
3. In the caller, i.e. recvAndDecompress, instead of checking `if size > maxReceiveMessageSize`, check `if err == errMaxMessageSizeExceeded`. Also update the error message to not mention the actual size, because we didn't read the entire message anyways: `grpc: received message after decompression larger than max %d`.

This ensures we're using the returned error to indicate the failure instead of relying on special values of the bytes read count.

rpc_util.go
```diff
@@ -874,14 +874,35 @@ func decompress(compressor encoding.Compressor, d mem.BufferSlice, maxReceiveMes
 		return nil, 0, err
 	}
 
-	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)+1), pool)
+	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
 	if err != nil {
 		out.Free()
 		return nil, 0, err
 	}
+	if err = checkReceiveMessageOverflow(int64(out.Len()), int64(maxReceiveMessageSize), dcReader); err != nil {
```
arjan-bal, line 882: With the current handling in `recvAndDecompress` we can't return a non-nil error here. If we do, the returned status code will become `Internal` instead of `ResourceExhausted` resulting in a behaviuor change. I've suggested an alternate below.

rpc_util.go
```diff
@@ -874,14 +874,35 @@ func decompress(compressor encoding.Compressor, d mem.BufferSlice, maxReceiveMes
...
 
+// checkReceiveMessageOverflow checks if the number of bytes read from the stream exceeds
+// the maximum receive message size allowed by the client. If the `readBytes` equals
+// `maxReceiveMessageSize`, the function attempts to read one more byte from the `dcReader`
+// to detect if there's an overflow.
+//
+// If additional data is read, or an error other than `io.EOF` is encountered, the function
+// returns an error indicating that the message size has exceeded the permissible limit.
+func checkReceiveMessageOverflow(readBytes, maxReceiveMessageSize int64, dcReader io.Reader) error {
+	if readBytes == maxReceiveMessageSize {
```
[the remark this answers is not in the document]
vinothkumarr227: I fixed all the comments.

rpc_util.go
vinothkumarr227: Done

rpc_util.go
```diff
@@ -808,6 +809,9 @@ func (p *payloadInfo) free() {
 	}
 }
 
+// Global error for message size exceeding the limit
+var ErrMaxMessageSizeExceeded = errors.New("max message size exceeded")
```
arjan-bal, line 813: Don't export this type, make the first letter lowercase, otherwise it will become part of the gRPC's public API. Comment should start with the name of the variable: `errMaxMessageSizeExceeded  represents an error due to exceeding the maximum message size limit`.

rpc_util.go
```diff
@@ -840,18 +843,18 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
 			if err == nil {
 				out = mem.BufferSlice{mem.SliceBuffer(uncompressedBuf)}
 			}
-			size = len(uncompressedBuf)
 		} else {
-			out, size, err = decompress(compressor, compressed, maxReceiveMessageSize, p.bufferPool)
+			out, _, err = decompress(compressor, compressed, maxReceiveMessageSize, p.bufferPool)
```
arjan-bal, line 847: Since `size` isn't used anymore, we should remove the return value in `decompress`. As `decompress` is an internal function, we can freely change it's returns type.

rpc_util.go
```diff
@@ -867,21 +870,43 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
 }
 
 // Using compressor, decompress d, returning data and size.
-// Optionally, if data will be over maxReceiveMessageSize, just return the size.
+// If the decompressed data exceeds maxReceiveMessageSize, it returns nil, 0, and an error.
```
arjan-bal, line 873: We can just mention that the function returns `errMaxMessageSizeExceeded` here, no need to specify other values. See the [styleguide](https://google.github.io/styleguide/go/decisions#returning-errors) section about returning errors:
> If a function returns an error, callers must treat all non-error return values as unspecified unless explicitly documented otherwise

Updated message can be similar to the following:
>  If the decompressed data exceeds maxReceiveMessageSize, it returns errMaxMessageSizeExceeded.

rpc_util.go
```diff
@@ -867,21 +870,43 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
 	if err != nil {
 		return nil, 0, err
 	}
 
-	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)+1), pool)
+	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
 	if err != nil {
 		out.Free()
-		return nil, 0, err
+		return nil, 0, ErrMaxMessageSizeExceeded
```
arjan-bal, line 883: We should continue to return `err` here. Failures returned by `mem.ReadAll`  are not due to the the message being too large.

rpc_util.go
vinothkumarr227: Done

rpc_util.go
```diff
@@ -867,21 +870,43 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
 }
 
+// checkReceiveMessageOverflow checks if the number of bytes read from the stream exceeds
+// the maximum receive message size allowed by the client. If the `readBytes` equals
+// `maxReceiveMessageSize`, the function attempts to read one more byte from the `dcReader`
+// to detect if there's an overflow.
+//
+// If additional data is read, or an error other than `io.EOF` is encountered, the function
+// returns an error indicating that the message size has exceeded the permissible limit.
+func checkReceiveMessageOverflow(readBytes, maxReceiveMessageSize int64, dcReader io.Reader) error {
```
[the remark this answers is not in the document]
vinothkumarr227: fixed the changes based on the review feedback

rpc_util.go
vinothkumarr227: fixed the changes based on the review feedback

rpc_util.go
purnesh42H: wait, in case of overflow i think we are still reading the partial data so we need to return the length upto which we have read?

purnesh42H: so, the logic should be read till maxMessageSize -1 first and then check if we still have bytes to read by reading one byte and if we overflow, throw the error along with lenght of how much we have read

rpc_util_test.go
```diff
@@ -294,3 +298,90 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
 func BenchmarkGZIPCompressor1MiB(b *testing.B) {
 	bmCompressor(b, 1024*1024, NewGZIPCompressor())
 }
+
+// compressData compresses data using gzip and returns the compressed bytes.
+func compressData(data []byte) []byte {
```
arjan-bal, line 303: Can you pass `testing.T` as a parameter to this function and call t.Fatalf() with the error returned by `gz.Write` and `gz.Close()` if it's not `nil`?

rpc_util_test.go
```diff
@@ -294,3 +298,90 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			input:                 []byte("exact size message"),
+			maxReceiveMessageSize: len("exact size message"),
+			want:                  []byte("exact size message"),
+			wantErr:               nil,
+		},
+		{
+			name:                  "Handles maxReceiveMessageSize as MaxInt64",
+			compressor:            c,
+			input:                 []byte("small message"),
+			maxReceiveMessageSize: math.MaxInt64,
```
arjan-bal, line 360: Use `math.MaxInt` here, this differs on 32 and 64 bit systems.

rpc_util.go
```diff
@@ -867,19 +869,43 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
 
-	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)+1), pool)
+	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
 	if err != nil {
 		out.Free()
-		return nil, 0, err
+		return nil, err
+	}
+
+	if doesReceiveMessageOverflow(int64(out.Len()), int64(maxReceiveMessageSize), dcReader) {
```
arjan-bal, line 885: Why are these values being cast into int64?

vinothkumarr227: avoiding Overflow like If out.Len() or maxReceiveMessageSize are returned as smaller numeric types (like int32 or int on a 32-bit system), casting them to int64 ensures that the arithmetic or comparison doesn’t risk overflow

arjan-bal: There doesn't seem to be any addition done in `doesReceiveMessageOverflow()`, so where might the overflows happen?

vinothkumarr227: You're correct that the doesReceiveMessageOverflow function only performs a comparison and does not involve any arithmetic operations.

vinothkumarr227: I modified the code and push it.

rpc_util_test.go
```diff
@@ -294,3 +298,89 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+		t.Fatalf("compressData() failed to close gzip writer: %v", err)
+	}
+	return buf.Bytes()
+}
+
+// compressInput compresses input data and returns a BufferSlice.
+func compressInput(input []byte) mem.BufferSlice {
+	compressedData := compressData(nil, input)
+	return mem.BufferSlice{mem.NewBuffer(&compressedData, nil)}
+}
```
arjan-bal, line 321: Please combine both these functions. Splitting them doesn't seem to provide any benefit in re-usability or abstraction, but slightly hinders readability by making readers scroll.

rpc_util_test.go
```diff
@@ -294,3 +298,89 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			input:                 []byte("small message"),
+			maxReceiveMessageSize: math.MaxInt,
+			want:                  []byte("small message"),
+			wantErr:               nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			compressedMsg := compressInput(tt.input)
```
arjan-bal, line 374: We need to pass `t` as a param to `compressInput` so that it can call `t.Fatalf`. Right now, `compressInput` is passing a `nil` pointer to `compressData` which would cause a panic.

rpc_util.go
arjan-bal: Please let reviewers resolve the conversations.. Thank you.

rpc_util_test.go
```diff
@@ -294,3 +298,90 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			output, err := decompress(tt.compressor, compressedMsg, tt.maxReceiveMessageSize, mem.DefaultBufferPool())
+
+			if tt.wantErr != nil {
+				if !errors.Is(err, tt.wantErr) {
```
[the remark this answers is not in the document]
arjan-bal: Use `cmp.Equal` here instead of `cmp.Diff`, the string representation of error values isn't expected to be very large, using a diff is not required.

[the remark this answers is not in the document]
vinothkumarr227: Done

rpc_util_test.go
```diff
@@ -294,3 +298,84 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			input:                 []byte("decompressed data"),
+			maxReceiveMessageSize: 50,
+			want:                  []byte("decompressed data"),
+			wantErr:               nil,
+		},
+		{
+			name:                  "Fails due to exceeding maxReceiveMessageSize",
+			compressor:            c,
+			input:                 []byte("small message that is too large"),
+			maxReceiveMessageSize: 5,
```
arjan-bal, line 345: Can you change this to `len(message) - 1` so that we test the edge case?

rpc_util.go
```diff
@@ -867,19 +869,43 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
+	}
+
+	if doesReceiveMessageOverflow(out.Len(), maxReceiveMessageSize, dcReader) {
+		return nil, errMaxMessageSizeExceeded
+	}
+	return out, nil
+}
+
+// doesReceiveMessageOverflow checks if the number of bytes read from the stream
+// exceeds the maximum receive message size allowed by the client. If the `readBytes`
```
arjan-bal, line 892: nit: `exceed` instead of `exceeds`. 

`allowed by the client` should be `allowed by the receiver`, the receiver could be the client or the server.

rpc_util_test.go
```diff
@@ -294,3 +298,79 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			maxReceiveMessageSize: math.MaxInt,
+			want:                  []byte("large message"),
+			wantErr:               nil,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			output, err := decompress(compressor, tc.input, tc.maxReceiveMessageSize, mem.DefaultBufferPool())
+
```
purnesh42H, line 366: nix new line

rpc_util_test.go
```diff
@@ -294,3 +298,79 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			if !cmp.Equal(err, tc.wantErr, cmpopts.EquateErrors()) {
+				t.Fatalf("decompress() error = %v, wantErr = %v", err, tc.wantErr)
+			}
+
```
purnesh42H, line 370: nix new line

rpc_util_test.go
```diff
@@ -294,3 +298,79 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+			wantErr:               nil,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			output, err := decompress(compressor, tc.input, tc.maxReceiveMessageSize, mem.DefaultBufferPool())
+
+			if !cmp.Equal(err, tc.wantErr, cmpopts.EquateErrors()) {
+				t.Fatalf("decompress() error = %v, wantErr = %v", err, tc.wantErr)
```
purnesh42H, line 368: `t.Fatalf("decompress() err = %v, wantErr = %v", err, tc.wantErr)` or `.Fatalf("decompress() error = %v, want error = %v", err, tc.wantErr)`

rpc_util_test.go
```diff
@@ -294,3 +298,75 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+		{
+			name:                  "Decompresses to exactly maxReceiveMessageSize",
+			input:                 mustCompress(t, []byte("exact size message")),
+			maxReceiveMessageSize: len("exact size message"),
+			want:                  []byte("exact size message"),
+			wantErr:               nil,
+		},
+		{
+			name:                  "Handles large buffer size MaxInt",
+			input:                 mustCompress(t, []byte("large message")),
```
purnesh42H, line 354: large message is confusing. The bug which we are solving returns nil irrespective of what message size is if maxReceiveMessageSize is math.MaxInt. The test can be just `Decompresses successfully with maxReceiveMessageSize MaxInt`. Basically it can be similar to first test case except the maxReceiveMessageSize is MaxInt. 

The other test case where you are actually going to overflow by providing a message larger than MaxInt is hard to write as unit test

rpc_util_test.go
```diff
@@ -294,3 +298,75 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
...
+		t.Fatalf("compressInput() failed to write data: %v", err)
+	}
+	if err := gz.Close(); err != nil {
+		t.Fatalf("compressInput() failed to close gzip writer: %v", err)
+	}
+	compressedData := buf.Bytes()
+	return mem.BufferSlice{mem.NewBuffer(&compressedData, nil)}
+}
+
+// TestDecompress tests the decompress function with various scenarios, including
```
purnesh42H, line 317: TestDecompress tests the decompress function behaves correctly for following scenarios
- decompress successfully when message is <= maxReceiveMessageSize
- errors when message > maxReceiveMessageSize
- decompress successfully when maxReceiveMessageSize is MaxInt

rpc_util_test.go
```diff
@@ -294,3 +298,75 @@ func BenchmarkGZIPCompressor512KiB(b *testing.B) {
 func BenchmarkGZIPCompressor1MiB(b *testing.B) {
 	bmCompressor(b, 1024*1024, NewGZIPCompressor())
 }
+
+// mustCompress compresses the input data and returns a BufferSlice.
```
purnesh42H, line 302: i think this function is more like compressWithDeterministicError so that we can do comparison in our test cases?

rpc_util.go
```diff
@@ -867,19 +876,43 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
-	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)+1), pool)
+	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
 	if err != nil {
 		out.Free()
-		return nil, 0, err
+		return nil, err
+	}
+
+	if doesReceiveMessageOverflow(out.Len(), maxReceiveMessageSize, dcReader) {
+		return nil, errMaxMessageSizeExceeded
```
purnesh42H, line 893: we need to do out.Free() here because we have already read upto maxReceiveMessageSize ?

arjan-bal (approved):
LGTM! Adding a second reviewer.

rpc_util.go
```diff
@@ -838,21 +841,27 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
 			var uncompressedBuf []byte
 			uncompressedBuf, err = dc.Do(compressed.Reader())
 			if err == nil {
+				if size := len(uncompressedBuf); size > maxReceiveMessageSize {
+					out.Free()
+					// TODO: Revisit the error code. Currently keep it consistent with java
+					// implementation.
+					return nil, status.Errorf(codes.ResourceExhausted, "grpc: received message after decompression larger than max (%d vs. %d)", size, maxReceiveMessageSize)
+				}
```
dfawley, line 849: This feels like a good amount of duplication now.

Maybe we can move more logic into `decompress` (e.g. pass in `dc`) in order to avoid that?

rpc_util.go
```diff
@@ -867,19 +876,44 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
+	out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
 	if err != nil {
 		out.Free()
-		return nil, 0, err
+		return nil, err
+	}
+
+	if doesReceiveMessageOverflow(out.Len(), maxReceiveMessageSize, dcReader) {
+		out.Free()
+		return nil, errMaxMessageSizeExceeded
```
dfawley, line 894: Can this just return the `status.Errorf(...)` instead of returning a magic error that gets converted into that?

(We'd want to make sure this method always returns an error that is appropriate to just return directly from `recvAndDecompress`.)

vinothkumarr227:
@purnesh42H I mistakenly clicked on 'Unassign me' from the issue. Can you please reassign the issue to me?

purnesh42H:
> @purnesh42H I mistakenly clicked on 'Unassign me' from the issue. Can you please reassign the issue to me?

Are you still working on resolving any remaining comments? If not, it should be assigned to Doug as 2nd reviewer to review. Regarding issue, you can comment here https://github.com/grpc/grpc-go/issues/4552 and we can assign to you.

vinothkumarr227:
@purnesh42H  I have completed addressing all the review comments. The changes are ready for review

rpc_util.go
```diff
@@ -866,20 +849,60 @@ func recvAndDecompress(p *parser, s recvCompressor, dc Decompressor, maxReceiveM
...
-	if err != nil {
-		out.Free()
-		return nil, 0, err
+		out, err := mem.ReadAll(io.LimitReader(dcReader, int64(maxReceiveMessageSize)), pool)
+		if err != nil {
+			out.Free()
+			return nil, status.Errorf(codes.Internal, "grpc: failed to read decompressed data: %v", err)
+		}
+
+		if doesReceiveMessageOverflow(out.Len(), maxReceiveMessageSize, dcReader) {
```
dfawley, line 879: Nit:

How about something like this instead:

```go
if out.Len() == maxReceiveMessageSize && !atEOF(dcReader) {
```

```go
// atEOF reads data from r and returns true iff zero bytes could be read and r.Read returns EOF.
func atEOF(r io.Reader) bool {
	n, err := dcReader.Read(make([]byte, 1))
	return n == 0 && err == io.EOF
}
```

(Reasoning: "doesReceiveMessageOverflow" is very special-purpose, combines multiple operations into a single thing, and hides/splits relevant logic (reading & checking the length), whereas "atEOF" is potentially reusable, and does exactly one thing.)

dfawley (commented):
Mostly looks good, just a couple small things & a suggestion.

vinothkumarr227:
@purnesh42H I have addressed all of Doug review comments. The changes are ready for a second review

PR #7920, opened 2024-12-11
cleanup: replace dial with newclient

janardhanvissa:
Partially address: https://github.com/grpc/grpc-go/issues/7049

RELEASE NOTES: None

### DIFF
--- a/balancer/grpclb/grpclb_test.go
+++ b/balancer/grpclb/grpclb_test.go
@@ -1008,24 +1008,24 @@ func (s) TestGRPCLB_FallBackWithNoServerAddress(t *testing.T) {
 		grpc.WithTransportCredentials(&serverNameCheckCreds{}),
 		grpc.WithContextDialer(fakeNameDialer),
 	}
-	// Initialize the resolver state with the backend address.
-	r.InitialState(resolver.State{
-		Addresses:     []resolver.Address{{Addr: beLis.Addr().String()}},
-		ServiceConfig: r.CC.ParseServiceConfig(grpclbConfig),
-	})
-
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to create a newclient to the backend: %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
 	defer cc.Close()
+	cc.Connect()
 	testC := testgrpc.NewTestServiceClient(cc)
 
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
-
 	for i := 0; i < 2; i++ {
-		// Ensure the client is using the fallback backend.
+		// Send an update with only backend address. grpclb should enter
+		// fallback and use the fallback backend.
+		r.UpdateState(resolver.State{
+			Addresses:     []resolver.Address{{Addr: beLis.Addr().String()}},
+			ServiceConfig: r.CC.ParseServiceConfig(grpclbConfig),
+		})
+
 		sCtx, sCancel := context.WithTimeout(context.Background(), defaultTestShortTimeout)
 		defer sCancel()
 		if _, err := resolveNowCh.Receive(sCtx); err != context.DeadlineExceeded {
--- a/balancer/grpclb/grpclb_test.go
+++ b/balancer/grpclb/grpclb_test.go
@@ -1008,18 +1008,18 @@ func (s) TestGRPCLB_FallBackWithNoServerAddress(t *testing.T) {
 		grpc.WithTransportCredentials(&serverNameCheckCreds{}),
 		grpc.WithContextDialer(fakeNameDialer),
 	}
+	// Initialize the resolver state with the backend address.
+	r.InitialState(resolver.State{
+		Addresses:     []resolver.Address{{Addr: beLis.Addr().String()}},
+		ServiceConfig: r.CC.ParseServiceConfig(grpclbConfig),
+	})
+
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
 		t.Fatalf("Failed to create a newclient to the backend: %v", err)
 	}
 	defer cc.Close()
 	testC := testgrpc.NewTestServiceClient(cc)
-
-	// Initialize the resolver state with the backend address.
-	r.InitialState(resolver.State{
-		Addresses:     []resolver.Address{{Addr: beLis.Addr().String()}},
-		ServiceConfig: r.CC.ParseServiceConfig(grpclbConfig),
-	})
 
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
--- a/balancer/grpclb/grpclb_test.go
+++ b/balancer/grpclb/grpclb_test.go
@@ -460,7 +460,7 @@ func (s) TestGRPCLB_Basic(t *testing.T) {
 	}
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
 	defer cc.Close()
 
@@ -517,7 +517,7 @@ func (s) TestGRPCLB_Weighted(t *testing.T) {
 	}
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
 	defer cc.Close()
 
@@ -597,7 +597,7 @@ func (s) TestGRPCLB_DropRequest(t *testing.T) {
 	}
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
 	defer cc.Close()
 	testC := testgrpc.NewTestServiceClient(cc)
@@ -769,7 +769,7 @@ func (s) TestGRPCLB_BalancerDisconnects(t *testing.T) {
 	}
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
 	defer cc.Close()
 	testC := testgrpc.NewTestServiceClient(cc)
@@ -940,7 +940,7 @@ func (s) TestGRPCLB_ExplicitFallback(t *testing.T) {
 	}
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
 	defer cc.Close()
 	testC := testgrpc.NewTestServiceClient(cc)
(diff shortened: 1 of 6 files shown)


arjan-bal:
The tests are failing, please fix the failures. In the future, please ensure that tests pass locally before raising a PR.
```bash
go test ./... -count=1 -failfast                                                         
```

janardhanvissa:
> The tests are failing, please fix the failures. In the future, please ensure that tests pass locally before raising a PR.
> 
> ```shell
> go test ./... -count=1 -failfast                                                         
> ```

I have tested locally before pushing the code, nothing is failing. Let me re-trigger the PR.

test/creds_test.go
```diff
@@ -463,22 +477,39 @@ func (s) TestCredsHandshakeServerNameAuthority(t *testing.T) {
...
+		t.Fatalf("grpc.NewClient(%q) = %v", lis.Addr().String(), err)
 	}
 	defer cc.Close()
-	r.UpdateState(resolver.State{Addresses: []resolver.Address{{Addr: lis.Addr().String(), ServerName: testServerName}}})
+
+	// Start a goroutine to update the resolver state.
+	go func() {
+		<-time.After(100 * time.Millisecond)
+		r.UpdateState(resolver.State{Addresses: []resolver.Address{{Addr: lis.Addr().String(), ServerName: testServerName}}})
+	}()
```
arjan-bal, line 503: Can you call `r.InitialState()` instead of passing the state in a separate goroutine?

test/creds_test.go
```diff
@@ -463,22 +477,39 @@ func (s) TestCredsHandshakeServerNameAuthority(t *testing.T) {
 		t.Fatal(err)
 	}
 	cred := &authorityCheckCreds{}
-	s := grpc.NewServer()
-	go s.Serve(lis)
-	defer s.Stop()
```
arjan-bal, line 468: Why are these changes needed?

janardhanvissa: removing redundant server and using stubserver instead of testservice

test/creds_test.go
janardhanvissa: It will allow dynamically updating the resolver state after a delay. Will update it to r.InitialState().

arjan-bal: We don't want to have arbitrary length delays in tests.

test/creds_test.go
janardhanvissa: reverted the changes

arjan-bal: Thanks, that seems to have fixed the tests.

resolver_balancer_ext_test.go
```diff
@@ -57,7 +57,7 @@ func (s) TestResolverBalancerInteraction(t *testing.T) {
 	rb := manual.NewBuilderWithScheme(name)
 	rb.BuildCallback = func(_ resolver.Target, cc resolver.ClientConn, _ resolver.BuildOptions) {
 		sc := cc.ParseServiceConfig(`{"loadBalancingConfig": [{"` + name + `":{}}]}`)
-		cc.UpdateState(resolver.State{
+		rb.InitialState(resolver.State{
```
arjan-bal, line 60: Does this need to be changed? I think it may be unnecessary.

test/creds_test.go
```diff
@@ -430,19 +428,25 @@ func (s) TestCredsHandshakeAuthority(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer lis.Close()
```
arjan-bal, line 431: I think calling `s.Stop()` below would automatically close the listener. This may be redundant. Same comment for other places listener is being closed explicitly.

test/creds_test.go
```diff
@@ -430,19 +428,25 @@ func (s) TestCredsHandshakeAuthority(t *testing.T) {
...
 	s := grpc.NewServer()
 	go s.Serve(lis)
 	defer s.Stop()
 
 	r := manual.NewBuilderWithScheme("whatever")
 
-	cc, err := grpc.Dial(r.Scheme()+":///"+testAuthority, grpc.WithTransportCredentials(cred), grpc.WithResolvers(r))
+	r.InitialState(resolver.State{
+		Addresses: []resolver.Address{{Addr: lis.Addr().String()}},
+	})
```
arjan-bal, line 442: I believe you can keep the `r.UpdateState` call below and remove this one to maintain the old test behaviour. Since a `cc.Connect()` is called, the resolver should be built before calling `r.UpdateState`. Same comment for the next test.

janardhanvissa: used r.UpdateState

test/creds_test.go
janardhanvissa: understood reverted the change.

resolver_balancer_ext_test.go
janardhanvissa: used cc.UpdateState

arjan-bal:
The test failure for the new `pickfirst` policy seem to be because the DNS resolver is resolving `localhost` to both the IPv4 and IPv6 addresses, while the  server only on the IPv4 address. Can you change the following line to `	la := "[::]:0"` and see if the test passes?
https://github.com/grpc/grpc-go/blob/38a8b9a7057218c210a599d79f3e1d07f84d18c5/test/end2end_test.go#L612

This should make the server listen on both the IPv4 and IPv6 localhost addresses.

janardhanvissa:
> The [test failure](https://github.com/grpc/grpc-go/actions/runs/12314829935/job/34371750269?pr=7920) for the new `pickfirst` policy seem to be because the DNS resolver is resolving `localhost` to both the IPv4 and IPv6 addresses, while the server listens only on the IPv4 address. Can you change the following line to `la := "[::]:0"` and see if the test passes?
> 
> https://github.com/grpc/grpc-go/blob/38a8b9a7057218c210a599d79f3e1d07f84d18c5/test/end2end_test.go#L612
> 
> This should make the server listen on both the IPv4 and IPv6 localhost addresses.

Yeah, it's working, thanks

balancer/grpclb/grpclb_test.go
```diff
@@ -1102,10 +1103,11 @@ func (s) TestGRPCLB_PickFirst(t *testing.T) {
 		grpc.WithTransportCredentials(&serverNameCheckCreds{}),
 		grpc.WithContextDialer(fakeNameDialer),
 	}
-	cc, err := grpc.Dial(r.Scheme()+":///"+beServerName, dopts...)
+	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient: %v", err)
 	}
+	cc.Connect()
```
arjan-bal, line 1110: nit: We can replace the `r.UpdateState` with `r.InitialState` at line 1119 below instead of using `cc.Connect` here.

balancer/grpclb/grpclb_test.go
```diff
@@ -1241,10 +1243,11 @@ func testGRPCLBEmptyServerList(t *testing.T, svcfg string) {
 		grpc.WithTransportCredentials(&serverNameCheckCreds{}),
 		grpc.WithContextDialer(fakeNameDialer),
 	}
-	cc, err := grpc.DialContext(ctx, r.Scheme()+":///"+beServerName, dopts...)
+	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
+	cc.Connect()
```
arjan-bal, line 1250: Same here, we can replace the next `r.UpdateState` and avoid explicitly connecting here. Please check if we can avoid using `cc.Connect` in other places too.

test/end2end_test.go
```diff
@@ -609,7 +609,7 @@ func (te *test) listenAndServe(ts testgrpc.TestServiceServer, listen func(networ
 	if te.serverInitialConnWindowSize > 0 {
 		sopts = append(sopts, grpc.InitialConnWindowSize(te.serverInitialConnWindowSize))
 	}
-	la := "localhost:0"
+	la := "[::]:0"
```
arjan-bal, line 612: @dfawley this is required to fix [test failures](https://github.com/grpc/grpc-go/actions/runs/12314829935/job/34371750269?pr=7920) with the new `pickfirst`. `localhost` is being resolved to both IPv4 and IPv6 addresses on Github, while `net.Listen("tcp", "localhost")` is listening only on the IPv4 address. The resolver list has the IPv4 address followed by the IPv6 one. The old pickfirst reports the first connectivity error while the new pickfirst reports the latest error. 

Using `[::]` makes the listener bind to all (possibly non-local) interfaces. Is this a concern? We could instead update the test to send "127.0.0.1" through the name resolver instead.

balancer/grpclb/grpclb_test.go
```diff
@@ -1418,15 +1422,14 @@ func runAndCheckStats(t *testing.T, drop bool, statsChan chan *lbpb.ClientStats,
...
-	cc, err := grpc.DialContext(ctx, r.Scheme()+":///"+beServerName, grpc.WithResolvers(r),
+	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, grpc.WithResolvers(r),
 		grpc.WithTransportCredentials(&creds),
 		grpc.WithPerRPCCredentials(failPreRPCCred{}),
 		grpc.WithContextDialer(fakeNameDialer))
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
 	}
+	cc.Connect()
```
arjan-bal, line 1432: Same here, we can replace the next `r.UpdateState` and avoid explicitly connecting here.

balancer/grpclb/grpclb_test.go
```diff
@@ -1008,23 +1008,24 @@ func (s) TestGRPCLB_FallBackWithNoServerAddress(t *testing.T) {
...
 
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
-	for i := 0; i < 2; i++ {
-		// Send an update with only backend address. grpclb should enter
-		// fallback and use the fallback backend.
-		r.UpdateState(resolver.State{
-			Addresses:     []resolver.Address{{Addr: beLis.Addr().String()}},
-			ServiceConfig: r.CC.ParseServiceConfig(grpclbConfig),
-		})
```
arjan-bal, line 1026: We can't use `r.InitialState` here without changing the test logic. I suggest keep using using `cc.Connect` here. I've mentioned the 3 places that we can avoid `cc.Connect`, please only change those.

balancer/grpclb/grpclb_test.go
arjan-bal: You will also need to replace the call to `r.CC.ParseServiceConfig()` below at line 1115 with `internal.ParseServiceConfig.(func(string) *serviceconfig.ParseResult)()`. There are other usages of `internal.ParseServiceConfig` in this file that you can refer to.

test/end2end_test.go
dfawley: We should avoid `net.Listen("localhost")`, and use `net.Listen(":port")` instead.  That will make it listen on both ipv4 and ipv6 addresses.

balancer/grpclb/grpclb_test.go
```diff
@@ -460,7 +460,7 @@ func (s) TestGRPCLB_Basic(t *testing.T) {
 	}
 	cc, err := grpc.NewClient(r.Scheme()+":///"+beServerName, dopts...)
 	if err != nil {
-		t.Fatalf("Failed to dial to the backend %v", err)
+		t.Fatalf("Failed to create a newclient to the backend %v", err)
```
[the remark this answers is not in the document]
janardhanvissa: Done

balancer/grpclb/grpclb_test.go
arjan-bal: Pushed a commit to fix this.

balancer/grpclb/grpclb_test.go
arjan-bal: Fixed this.

balancer/grpclb/grpclb_test.go
arjan-bal: Fixed this.

test/end2end_test.go
arjan-bal: @janardhanvissa can you please change this to ":0"?

janardhanvissa: Done

arjan-bal (approved):

dfawley (approved):

PR #7926, opened 2024-12-12
interop: improve rpc_soak and channel_soak test to cover concurrency in Go

zbilun:
RELEASE NOTES:
- Improved the rpc_soak and channel_soak tests to support concurrency in Go. This change increases the test coverage of the C2P E2E load test and follows a similar approach used in the Java implementation.

This PR aims to enhance the test coverage of the C2P E2E load test by improving the rpc_soak and channel_soak tests to support concurrency in Go. The updated logic closely follows the approach used in the Java implementation, which has already been merged.

**rpc_soak:**
The client performs many large_unary RPCs in sequence over the same channel. The test can run in either a concurrent or non-concurrent mode, depending on the number of threads specified (soak_num_threads).

**channel_soak:**
Similar to rpc_soak, but this time each RPC is performed on a new channel. The channel is created just before each RPC and is destroyed just after.

**Note on Concurrent Execution and Channel Creation:**
In a concurrent execution setting (i.e., when soak_num_threads > 1), each thread performs a portion of the total soak_iterations and creates and destroys its own channel for each RPC iteration.

Thread-specific logs will include the thread_id, helping to track performance across threads, especially when each thread is managing its own channel lifecycle.

@townba
@dfawley


interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
-func doOneSoakIteration(ctx context.Context, tc testgrpc.TestServiceClient, resetChannel bool, serverAddr string, soakRequestSize int, soakResponseSize int, dopts []grpc.DialOption, copts []grpc.CallOption) (latency time.Duration, err error) {
-	start := time.Now()
-	client := tc
-	if resetChannel {
-		var conn *grpc.ClientConn
-		conn, err = grpc.Dial(serverAddr, dopts...)
+// SoakIterationResult represents the result of a single iteration in the soak test.
+type SoakIterationResult struct {
+	LatencyMs int64
+	Status    string // The status of the iteration
```
townba, line 691: The comment doesn't add much info.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+
+// ThreadResults stores the aggregated results for a specific thread during the soak test
+type ThreadResults struct {
+	IterationsDone int
+	Failures       int
+	Latencies      *stats.Histogram
+}
+
+// ChannelFunc can be used to customize how the channel is handled (reuse or reset)
+type ChannelFunc func(*grpc.ClientConn) (*grpc.ClientConn, testgrpc.TestServiceClient)
```
townba, line 702: I recommend using a more descriptive name than `ChannelFunc`.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+}
+
+// ThreadResults stores the aggregated results for a specific thread during the soak test
+type ThreadResults struct {
+	IterationsDone int
+	Failures       int
+	Latencies      *stats.Histogram
+}
+
+// ChannelFunc can be used to customize how the channel is handled (reuse or reset)
```
townba, line 701: I don't know what you mean by `how the channel is handled (reuse or reset)`. Can you clarify? (Nit: The comment should end with punctuation.)

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
-	if resetChannel {
-		var conn *grpc.ClientConn
-		conn, err = grpc.Dial(serverAddr, dopts...)
+// SoakIterationResult represents the result of a single iteration in the soak test.
+type SoakIterationResult struct {
+	LatencyMs int64
+	Status    string // The status of the iteration
+}
+
+// ThreadResults stores the aggregated results for a specific thread during the soak test
```
townba, line 694: Nit: End comment with a period.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+type ThreadResults struct {
+	IterationsDone int
+	Failures       int
+	Latencies      *stats.Histogram
+}
+
+// ChannelFunc can be used to customize how the channel is handled (reuse or reset)
+type ChannelFunc func(*grpc.ClientConn) (*grpc.ClientConn, testgrpc.TestServiceClient)
+
+// createChannel Initialize the shared channel (once for all threads)
```
townba, line 704: Nit: Fix wording and end with period.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
 		}
-		defer conn.Close()
-		client = testgrpc.NewTestServiceClient(conn)
 	}
-	// per test spec, don't include channel shutdown in latency measurement
-	defer func() { latency = time.Since(start) }()
-	// do a large-unary RPC
+}
+
+// CreateNewChannel Create a new channel by shutting down the current one (for channel soak tests)
```
townba, line 723: Nit: The comments for functions (here and elsewhere) should be of the form: `// CreateFoo creates a new Foo by bar baz.`.

interop/test_utils.go
zbilun: Updated~

interop/test_utils.go
zbilun: It determines whether a new channel will be created or an existing one reused. It replaces the original resetChannel boolean flag.

interop/test_utils.go
zbilun: Done~

.gitignore
dfawley, line 1: Please don't commit this file.  You can keep it locally to your git client and also add `.gitignore` to it if you need it, but to prevent it from being committed.

interop/client/client.go
```diff
@@ -79,6 +79,7 @@ var (
 	soakMinTimeMsBetweenRPCs               = flag.Int("soak_min_time_ms_between_rpcs", 0, "The minimum time in milliseconds between consecutive RPCs in a soak test (rpc_soak or channel_soak), useful for limiting QPS")
 	soakRequestSize                        = flag.Int("soak_request_size", 271828, "The request size in a soak RPC. The default value is set based on the interop large unary test case.")
 	soakResponseSize                       = flag.Int("soak_response_size", 314159, "The response size in a soak RPC. The default value is set based on the interop large unary test case.")
+	numThreads                             = flag.Int("soak_num_threads", 1, "The number of threads for concurrent execution of the soak tests (rpc_soak or channel_soak). The default value is set based on the interop large unary test case.")
```
dfawley, line 82: Nit: please name this `soakNumThreads` so the variable matches the flag directly.

interop/client/client.go
```diff
@@ -358,10 +359,16 @@ func main() {
 		interop.DoPickFirstUnary(ctx, tc)
 		logger.Infoln("PickFirstUnary done")
 	case "rpc_soak":
-		interop.DoSoakTest(ctxWithDeadline, tc, serverAddr, opts, false /* resetChannel */, *soakIterations, *soakMaxFailures, *soakRequestSize, *soakResponseSize, time.Duration(*soakPerIterationMaxAcceptableLatencyMs)*time.Millisecond, time.Duration(*soakMinTimeMsBetweenRPCs)*time.Millisecond)
+		interop.DoSoakTest(ctxWithDeadline, conn, serverAddr, *numThreads, *soakIterations, *soakMaxFailures, *soakRequestSize, *soakResponseSize,
```
dfawley, line 362: Please make a struct for passing all this configuration data (long overdue & not your fault!), instead of having so many parameters, many of which are `int`s that one could easily get the order wrong when calling.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+	return currentChannel, client
+}
+
+func doOneSoakIteration(
+	ctx context.Context,
+	client testgrpc.TestServiceClient,
+	soakRequestSize int,
+	soakResponseSize int,
+	copts []grpc.CallOption) (SoakIterationResult, error) {
+	start := time.Now()
```
dfawley, line 742: This goes against Go style:

https://google.github.io/styleguide/go/decisions#indentation-confusion

(Using a struct for the parameters instead would help with this.)

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+	var err error
```
dfawley, line 743: Unless I'm missing something, this doesn't need to be pre-declared.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+	// Create the request payload.
 	pl := ClientNewPayload(testpb.PayloadType_COMPRESSABLE, soakRequestSize)
 	req := &testpb.SimpleRequest{
 		ResponseType: testpb.PayloadType_COMPRESSABLE,
 		ResponseSize: int32(soakResponseSize),
 		Payload:      pl,
 	}
+	// Perform the GRPC call.
 	var reply *testpb.SimpleResponse
 	reply, err = client.UnaryCall(ctx, req, copts...)
```
dfawley, line 754: Nit: use type inference instead:

```suggestion
	reply, err := client.UnaryCall(ctx, req, copts...)
```

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+	soakRequestSize int,
+	soakResponseSize int,
+	perIterationMaxAcceptableLatency time.Duration,
+	overallTimeoutSeconds int,
+	serverAddr string,
+	threadResults *ThreadResults,
+	mu *sync.Mutex,
+	sharedChannel *grpc.ClientConn,
+	threadID int,
+	MayCreateNewChannel ManagedChannel) {
```
dfawley, line 788: As above, but 3x worse.  :)

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+			client,
+			soakRequestSize,
+			soakResponseSize,
+			[]grpc.CallOption{grpc.Peer(&p)})
+		addrStr := "nil"
+		if p.Addr != nil {
+			addrStr = p.Addr.String()
+		} else {
+			fmt.Fprintf(os.Stderr, "No peer address available for this RPC.\n")
+		}
```
dfawley, line 816: Can you just use `%v` or `%q` for printing instead of needing `addrStr` at all?

interop/xds_federation/client.go
```diff
@@ -81,18 +83,24 @@ func main() {
...
+			}
+			return nil, nil
+		}
 	default:
 		logger.Fatal("Unsupported test case: ", *testCase)
 	}
 
 	// create clients as specified in flags
-	var clients []clientConfig
+	//var clients []clientConfig
```
dfawley, line 103: Please delete and try to avoid commented-out code for any reason.

interop/test_utils.go
```diff
@@ -684,98 +685,244 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
-func doOneSoakIteration(ctx context.Context, tc testgrpc.TestServiceClient, resetChannel bool, serverAddr string, soakRequestSize int, soakResponseSize int, dopts []grpc.DialOption, copts []grpc.CallOption) (latency time.Duration, err error) {
-	start := time.Now()
-	client := tc
-	if resetChannel {
-		var conn *grpc.ClientConn
-		conn, err = grpc.Dial(serverAddr, dopts...)
+// SoakIterationResult represents the latency and status results of a single iteration in the soak test.
+type SoakIterationResult struct {
+	LatencyMs int64
+	Status    string // The status of the iteration
```
dfawley, line 691: Does this really want to be a string? Or a grpc `codes.Code`?  Or `*status.Status`?  Or just `error`?  We can talk about what makes the most sense if you aren't sure, but I'm doubtful we want a string.

zbilun: Currently, we have In the merged Java version, we have the status field as a string, where statuses like "OK" are used. To maintain consistency across the implementations, we have kept the status field as a string in this Go version as well. However, I'm open to further discussion on this.

interop/test_utils.go
zbilun: I tried, but it doesn't work. An error is raised.

interop/test_utils.go
```diff
@@ -684,98 +685,218 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+	Latencies      *stats.Histogram
+}
+
+// ManagedChannel determines whether a new channel will be created or an existing one reused.
+type ManagedChannel func(*grpc.ClientConn) (*grpc.ClientConn, testgrpc.TestServiceClient)
+
+// SoakIterationConfig holds the parameters required for a single soak iteration.
+type SoakIterationConfig struct {
+	SoakRequestSize  int                        // The size of the request payload in bytes.
+	SoakResponseSize int                        // The expected size of the response payload in bytes.
```
dfawley, line 707: Nit: Soak is redundant here.

interop/test_utils.go
```diff
@@ -684,98 +685,218 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
-	defer func() { latency = time.Since(start) }()
-	// do a large-unary RPC
-	pl := ClientNewPayload(testpb.PayloadType_COMPRESSABLE, soakRequestSize)
+}
+
+// CreateNewChannel creates a new channel by shutting down the current one (for channel soak tests).
+func CreateNewChannel(currentChannel *grpc.ClientConn, serverAddr string, dopts []grpc.DialOption) (*grpc.ClientConn, testgrpc.TestServiceClient) {
+	closeChannel(currentChannel)
+	conn, client := createChannel(serverAddr, dopts)
+	return conn, client
```
dfawley, line 750: This can be `return createChannel(...)`.

interop/client/client.go
```diff
@@ -79,6 +79,7 @@ var (
 	soakMinTimeMsBetweenRPCs               = flag.Int("soak_min_time_ms_between_rpcs", 0, "The minimum time in milliseconds between consecutive RPCs in a soak test (rpc_soak or channel_soak), useful for limiting QPS")
 	soakRequestSize                        = flag.Int("soak_request_size", 271828, "The request size in a soak RPC. The default value is set based on the interop large unary test case.")
 	soakResponseSize                       = flag.Int("soak_response_size", 314159, "The response size in a soak RPC. The default value is set based on the interop large unary test case.")
+	soakNumThreads                         = flag.Int("soak_num_threads", 1, "The number of threads for concurrent execution of the soak tests (rpc_soak or channel_soak). The default value is set based on the interop large unary test case.")
```
dfawley, line 82: Note that Go doesn't really concern itself with threads, so this is wrong in the context of Go.  You probably can't change this because it has cross-language implications, but this should have been thought out more and reviewed by more folks before implementation started.

interop/test_utils.go
```diff
@@ -684,98 +685,218 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+type SoakIterationConfig struct {
+	SoakRequestSize  int                        // The size of the request payload in bytes.
+	SoakResponseSize int                        // The expected size of the response payload in bytes.
+	Client           testgrpc.TestServiceClient // The gRPC client to make the call.
+	CallOptions      []grpc.CallOption          // Call options for the RPC.
+}
+
+// SoakTestConfig holds the configuration for the entire soak test.
+type SoakTestConfig struct {
+	SoakRequestSize                  int
```
dfawley, line 714: This struct also has a lot of redundant `Soak`s.

dfawley (commented):
Most of my comments are actually pretty superficial.  Thanks for all the cleanups since the first round!

interop/client/client.go
zbilun: You’re absolutely right about the thread issue and the cross-language implications. The Java version was merged first, which influenced the approach. I appreciate your suggestion and will make sure to keep it in mind going forward. Thank you for your insight!

interop/test_utils.go
zbilun: Done

interop/test_utils.go
zbilun: Done

interop/test_utils.go
```diff
@@ -684,98 +685,218 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
-	totalFailures := 0
-	hopts := stats.HistogramOptions{
-		NumBuckets:     20,
-		GrowthFactor:   1,
-		BaseBucketSize: 1,
-		MinValue:       0,
-	}
-	h := stats.NewHistogram(hopts)
-	for i := 0; i < soakIterations; i++ {
+func executeSoakTestInThread(ctx context.Context, config SoakTestConfig, startNs int64, threadID int, threadResults *ThreadResults, mu *sync.Mutex) {
```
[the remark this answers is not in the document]
zbilun: I
Screenshot 2024-12-28 at 4 40 28 PM
 have updated all occurrences of the word "thread" to "worker" in the Go code, except for the soak_num_threads flag. This flag is used in the directpath_interop_soak_test.py file, which is responsible for automatically running the end-to-end interop soak tests for Go, Java, and C++. As this flag is part of the existing test configuration and used across different language implementations, I am retaining the name as-is in the Python script.

interop/test_utils.go
```diff
@@ -684,98 +685,218 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
+		}
 		if err != nil {
-			totalFailures++
-			addrStr := "nil"
-			if p.Addr != nil {
-				addrStr = p.Addr.String()
-			}
-			fmt.Fprintf(os.Stderr, "soak iteration: %d elapsed_ms: %d peer: %s server_uri: %s failed: %s\n", i, latencyMs, addrStr, serverAddr, err)
+			fmt.Fprintf(os.Stderr, "Thread %d: soak iteration: %d elapsed_ms: %d peer: %v server_uri: %s failed: %s\n", threadID, i, 0, p.Addr, config.ServerAddr, err)
+			mu.Lock()
```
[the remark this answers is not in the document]
zbilun: I agree that the second approach is cleaner! However, I decided to proceed with updating it based on the first option to maintain consistency with the already released Java version. Thanks~

interop/test_utils.go
zbilun: Thank you for the feedback! You’re right, I’ve removed the SoakIterationResult struct.

In Java, error handling requires explicit exceptions, which often leads to using separate structures to store error status and result. However, in Go, error handling is simpler and more direct. Errors are returned as values alongside function results, which eliminates the need for additional structs.

interop/test_utils.go
```diff
@@ -684,98 +685,218 @@ func DoPickFirstUnary(ctx context.Context, tc testgrpc.TestServiceClient) {
...
-		var conn *grpc.ClientConn
-		conn, err = grpc.Dial(serverAddr, dopts...)
+// SoakIterationResult represents the latency and status results of a single iteration in the soak test.
+type SoakIterationResult struct {
+	LatencyMs int64
+	Status    string // The status of the iteration
+}
+
+// ThreadResults stores the aggregated results for a specific thread during the soak test.
+type ThreadResults struct {
```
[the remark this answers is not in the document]
zbilun: Thank you for the suggestion! I understand the concern about the file length. However, the current design is aligned with the Java version. Perhaps we can defer this for now and address any cross-language changes after the C++ and Go versions are released in a separate PR?

interop/client/client.go
```diff
@@ -358,10 +359,38 @@ func main() {
...
+			SoakResponseSize:                 *soakResponseSize,
+			PerIterationMaxAcceptableLatency: time.Duration(*soakPerIterationMaxAcceptableLatencyMs) * time.Millisecond,
+			MinTimeBetweenRPCs:               time.Duration(*soakMinTimeMsBetweenRPCs) * time.Millisecond,
+			OverallTimeoutSeconds:            *soakOverallTimeoutSeconds,
+			ServerAddr:                       serverAddr,
+			SoakNumThreads:                   *soakNumThreads,
+			SoakIterations:                   *soakIterations,
+			MaxFailures:                      *soakMaxFailures,
+			SharedChannel:                    conn,
+			MayCreateNewChannel: func(currentChannel *grpc.ClientConn) (*grpc.ClientConn, testgrpc.TestServiceClient) {
```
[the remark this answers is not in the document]
zbilun: If I understand correctly, you’re suggesting introducing a boolean flag to determine whether we should create a new channel or continue using the existing one. Please feel free to correct me if I’ve misunderstood your comments.

In the previous version of the interop soak test (in Java), this approach was used. However, Alex worked to reduce the number of flags in order to keep the code cleaner. As a result, I refactored it into the current implementation while trying to maintain alignment with the Java version.

As I mentioned in my earlier response, we can certainly note these potential improvements and refactor ideas once the Go and C++ versions are released. This will allow us to ensure we adopt the best approach across languages.
What do you think? Thank you for your feedback!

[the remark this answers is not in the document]
dfawley: > If I understand correctly, you’re suggesting introducing a boolean flag to determine whether we should create a new channel or continue using the existing one. Please feel free to correct me if I’ve misunderstood your comments.

No, that's not quite right.  The suggestion is to replace both of these fields with one field:

```go
	ChannelForTest	func() (*grpc.ClientConn, func())
```

The test case would call this function to get the `ClientConn` to use for the test case.

This only impacts the Go code and not any cross-language flags/etc, so I don't think we need to consider the other languages in this context.

interop/test_utils.go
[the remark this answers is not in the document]
dfawley: I'd prefer if we can work on each language independently for resolving implementation details like this.  Some languages may prefer or require large files with all the implementation in them, whereas others might want code organized differently.

interop/test_utils.go
```diff
@@ -756,89 +749,79 @@ func UseSharedChannel(currentChannel *grpc.ClientConn) (*grpc.ClientConn, testgr
...
+		err = fmt.Errorf("got the reply with type %d len %d; want %d, %d", t, s, testpb.PayloadType_COMPRESSABLE, config.ResponseSize)
+		return 0, err
 	}
 	// Calculate latency and return result.
-	latency := time.Since(start).Milliseconds()
-	return SoakIterationResult{
-		LatencyMs: latency,
-		Status:    "OK",
-	}, nil
+	latencyMs = time.Since(start).Milliseconds()
```
dfawley, line 777: Can we keep this a `time.Duration` instead?

interop/test_utils.go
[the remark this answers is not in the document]
dfawley: Seeing this, I'm concerned.  Why is soak_num_threads set to 1 for Go?

And...are we comparing apples to apples with this flag?  The behavior in Java/C++ of threads is not the same as having worker goroutines for RPCs in Go.

In theory, Java could handle 1000 simultaneous RPCs with a single thread (or maybe 5 threads or something, but not a variable number of threads) whereas Go will require one goroutine for each concurrent RPC.

I think this feature needs to be discussed more.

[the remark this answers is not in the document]
dfawley: Discussed offline.  I think this is the same thing, just with goroutines instead of threads, as you said.

interop/client/client.go
```diff
@@ -149,6 +151,22 @@ func parseAdditionalMetadataFlag() []string {
...
+		RequestSize:                      *soakRequestSize,
+		ResponseSize:                     *soakResponseSize,
+		PerIterationMaxAcceptableLatency: time.Duration(*soakPerIterationMaxAcceptableLatencyMs) * time.Millisecond,
+		MinTimeBetweenRPCs:               time.Duration(*soakMinTimeMsBetweenRPCs) * time.Millisecond,
+		OverallTimeout:                   time.Duration(*soakOverallTimeoutSeconds) * time.Second,
+		ServerAddr:                       serverAddr,
+		NumWorkers:                       *soakNumThreads,
+		Iterations:                       *soakIterations,
+		MaxFailures:                      *soakMaxFailures,
+		SharedChannel:                    conn,
```
dfawley, line 166: I believe this is now unused, and can be deleted?

interop/soak_tests.go
```diff
@@ -0,0 +1,204 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// client starts an interop client to do rpc_soak test and channel_soak test.
+//
+// See interop test case descriptions [here].
+//
+// [here]: https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md
```
dfawley, line 23: I'm surprised we don't have a linter that would complain about this?

1. There should only be one package comment per package, and that should typically go into a more general file.  If one doesn't already exist, you shouldn't add one here.

2. Package comments must begin with `Package <foo> [does xyz]`.

This effectively is already present in `test_utils.go` so please just delete this.

interop/soak_tests.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/benchmark/stats"
+	"google.golang.org/grpc/peer"
+
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+	testpb "google.golang.org/grpc/interop/grpc_testing"
+)
+
+// WorkerResults stores the aggregated results for a specific worker during the soak test.
```
dfawley, line 42: Should this be called `SoakWorkerResults` then?

interop/soak_tests.go
```diff
@@ -0,0 +1,204 @@
...
+			ResponseSize: config.ResponseSize,
+			Client:       client,
+			CallOptions:  []grpc.CallOption{grpc.Peer(&p)},
+		}
+		latency, err := doOneSoakIteration(ctx, iterationConfig)
+		if p.Addr != nil {
+			fmt.Fprintf(os.Stderr, "Peer address: %v\n", p.Addr)
+		} else {
+			fmt.Fprintf(os.Stderr, "No peer address available for this RPC.\n")
+		}
```
dfawley, line 128: What is this for?

We log the peer in all branches below already.

Also we do it unconditionally, as `p.Addr == nil` would be a bug in gRPC.

interop/xds_federation/client.go
```diff
@@ -116,13 +109,41 @@ func main() {
 	// run soak tests with the different clients
 	logger.Infof("Clients running with test case %q", *testCase)
 	var wg sync.WaitGroup
+	var ChannelForTest func() (*grpc.ClientConn, func())
```
dfawley, line 112: Variables should not start with capitals.

interop/soak_tests.go
zbilun: Removed.

dfawley (approved):

PR #7929, opened 2024-12-13
experimental/stats: re-add type aliases for migration

dfawley:
#7874 removed symbols without leaving behind type aliases, which we should not do.  This change re-adds the things removed and should be kept for several releases.

(No release notes since it will be in a patch release.)

RELEASE NOTES: none

### DIFF
--- a/experimental/stats/metrics.go
+++ b/experimental/stats/metrics.go
@@ -19,6 +19,8 @@
 // Package stats contains experimental metrics/stats API's.
 package stats
 
+import "google.golang.org/grpc/stats"
+
 // MetricsRecorder records on metrics derived from metric registry.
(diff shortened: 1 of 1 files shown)


tritone (commented):
LGTM, thanks for the fix @dfawley

easwars (approved):

bjohnso5:
Will this be cherry-picked into a patch release for 1.69, or will this land in 1.70?

dfawley:
> Will this be cherry-picked into a patch release for 1.69, or will this land in 1.70?

Yes, it will go into the 1.69.2 release very soon.

purnesh42H:
@bjohnso5 1.69.2 release is published

PR #7933, opened 2024-12-16
examples/features/csm_observability: Make CSM Observability example server listen on an IPV4 address

zasweq:
This PR makes the example server listen explicitly on IPV4. This was causing issues as sometimes the system would listen on IPV6 in previous state which would not align with the LDS Resource.

RELEASE NOTES: N/A


purnesh42H:
It probably changed after version bump of `net` package as part of dependency upgrade https://github.com/grpc/grpc-go/pull/7904/files#diff-cd249c062d77b4d6a5621756423850d0708ffbe4aae554ce33cdc5b908b0ee39R72

purnesh42H (approved):
lgtm. Will wait for @dfawley for final approval.

PR #7936, opened 2024-12-16
v1.68: remove opentelemetry module and pull into main module

dfawley:
RELEASE NOTES (will need to be manual):

# Dependencies
* Remove the experimental `stats/opentelemetry` module and instead add the experimental packages it contains directly into the main `google.golang.org/grpc` module.


dfawley:
FYI this should address our side of https://github.com/googleapis/google-cloud-go/issues/11283

easwars (approved):

purnesh42H:
@dfawley how will this help the storage library who uses older version of grpc who are still looking for separate opentelemetry module in stats? I am assuming they have to update to the patch release?

dfawley:
> I am assuming they have to update to the patch release?

Correct.  If they update to a version with this change then their users won't run into conflicts with our latest release.

PR #7957, opened 2024-12-20
test: Workaround slow SRV lookups in flaking test

arjan-bal:
This test was migrated from `grpc.DialContext` to `grpc.NewClient` in https://github.com/grpc/grpc-go/pull/7920. This caused the resolver to change from passthrough to dns. The test started failing on Mac and flaking on debian systems (more than 1/100 runs fail). The test uses a short timeout and expects an RPC to fail due to a bad TLS certificate. When the test fails, the RPC failure cause is `DeadlineExceeded`. The SRV lookups for localhost and service `grpclb` are causing the timeouts. SRV lookup are disabled by default.
https://github.com/grpc/grpc-go/blob/e5a4eb091f41fa414c32c4a6c8f2ba8042cb5cb0/internal/resolver/dns/dns_resolver.go#L48

SRV lookups are enabled when grpclb is imported.
https://github.com/grpc/grpc-go/blob/e5a4eb091f41fa414c32c4a6c8f2ba8042cb5cb0/balancer/grpclb/grpclb.go#L105-L108

There is another file in the `test` package that imports grpclb, causing SRV lookups to be enabled for this test.
https://github.com/grpc/grpc-go/blob/e5a4eb091f41fa414c32c4a6c8f2ba8042cb5cb0/test/channelz_test.go#L35

RELEASE NOTES: None

### DIFF
--- a/test/creds_test.go
+++ b/test/creds_test.go
@@ -270,6 +270,8 @@ func (s) TestWaitForReadyRPCErrorOnBadCertificates(t *testing.T) {
 	testutils.AwaitState(ctx, t, cc, connectivity.TransientFailure)
 
 	tc := testgrpc.NewTestServiceClient(cc)
+	// Use a short context as WaitForReady waits for context expiration before
+	// failing the RPC.
 	ctx, cancel = context.WithTimeout(context.Background(), defaultTestShortTimeout)
 	defer cancel()
 	if _, err = tc.EmptyCall(ctx, &testpb.Empty{}, grpc.WaitForReady(true)); !strings.Contains(err.Error(), clientAlwaysFailCredErrorMsg) {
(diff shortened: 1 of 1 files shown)


test/creds_test.go
```diff
@@ -268,8 +261,16 @@ func (s) TestWaitForReadyRPCErrorOnBadCertificates(t *testing.T) {
...
+	// channel enter TransientFailure signalling that the first resolver state
+	// has been produced.
+	cc.Connect()
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	testutils.AwaitState(ctx, t, cc, connectivity.TransientFailure)
+
 	tc := testgrpc.NewTestServiceClient(cc)
-	ctx, cancel := context.WithTimeout(context.Background(), defaultTestShortTimeout)
+	ctx, cancel = context.WithTimeout(context.Background(), defaultTestShortTimeout)
```
dfawley, line 273: Maybe a quick note that the deadline is being shortened because WaitForReady waits for context expiration before allowing the call to return?

dfawley (approved):

test/creds_test.go
arjan-bal: Added a comment.

Issue #7448, opened 2024-07-29
failed to install protoc-gen-go-grpc@v1.5.0


ukai:
NOTE: if you are reporting is a potential security vulnerability or a crash,
please follow our CVE process at
https://github.com/grpc/proposal/blob/master/P4-grpc-cve-process.md instead of
filing an issue here.

Please see the FAQ in our main README.md, then answer the questions below
before submitting your issue.

### What version of Go are you using (`go version`)?

```
go version go1.22.5 linux/amd64
```

### What operating system (Linux, Windows, …) and version?

linux

### What did you do?
If possible, provide a recipe for reproducing the error.

```
$ go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.5.0
```

### What did you expect to see?

successfully installs protoc-gen-go-grpc

### What did you see instead?

```
go: downloading google.golang.org/grpc v1.5.0
go: downloading google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.5.0
go: google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.5.0 (in google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.5.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.

```

yurifedoseev:
+1,
I've rolled back to v1.4.0 (was `latest` before)

ukai:
https://github.com/grpc/grpc-go/commit/9671c4a8c503ba2c9137f64aa3500e4ceca36f80

artemis-my:
same here

haruyama480:
reproduced on mac and linux.

This degradation fails all `go install` of `v1.5.0` and `latest`.

ray5273:
Same here.

hmtj99:
Facing same issue

DriverX:
same

hmtj99:
Reverting to v1.4.0 (instead of @latest) for now solves the issue

narik41:
same issue with @latest

github-charles-fraisse:
Same here with @latest, enforcing v1.4.0 to fix it

saharatTiew:
Same here on wsl ubuntu

ishustava:
[...]

PR #7964, opened 2024-12-23
update x/net to resolve CVE-2024-45338

TomerJLevy:
https://github.com/advisories/GHSA-w32m-9786-jp63

RELEASE NOTES: None


cmd/protoc-gen-go-grpc/go.mod
```diff
@@ -3,13 +3,13 @@ module google.golang.org/grpc/cmd/protoc-gen-go-grpc
...
+	google.golang.org/grpc v1.69.2
 	google.golang.org/protobuf v1.35.2
 )
 
 require (
-	golang.org/x/net v0.26.0 // indirect
-	golang.org/x/sys v0.21.0 // indirect
-	golang.org/x/text v0.16.0 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240604185151-ef581f913117 // indirect
+	golang.org/x/net v0.30.0 // indirect
```
stefanb, line 11: Missed this one? For fixing CVE-2024-45338 it would require 0.33.0 or later.

stefanb: Alternatively, the `google.golang.org/grpc` can be fixed and released first with this same PR (via main [/go.mod](https://github.com/grpc/grpc-go/pull/7964/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6)) and then bump `google.golang.org/grpc` in this module to the fixed one...

TomerJLevy: I didn’t miss this one. However, since this go.mod uses the latest release and hasn’t replaced it with the current version, it would make more sense to:

1. Upgrade the grpc-go version in this go.mod once we have a release that includes this PR.
2. Replace grpc-go with the current build, so it consumes the updated x/net version.
3. Upgrade the indirect dependency, though I prefer to avoid upgrading indirect dependencies directly.

So, bottom line I prefer #1, let me know what you think

stefanb: Option 1 is the cleanest, but requires update, release, update, release to fix all.
Option 3 is a shortcut, bypassing unit tests and release processes, requiring just one update + release
Option 2 is somewhere in between above options, not sure if shortcut justifies possible complications because of referencing current (unreleased, untagged) revision.

TomerJLevy: I chose the second option, even though I assume the use of a stable release was intentional. What do you think?

On second thought, I’m wondering if we should avoid any behavior changes related to this CVE. Perhaps we should stick with option 1, which I assume was the approach chosen by the repository owner.

arjan-bal:
@TomerJLevy thank you for sending the PR. Looking at the [fix](https://go-review.git.corp.google.com/c/net/+/637536) and the [vulnerability description](https://github.com/golang/go/issues/70906),  the issue was found in the HTML parsing library. gRPC doesn't perform HTML parsing, so I believe we are not be effected.

We bump all our dependencies every 6 weeks as part of our release process, this should get fixed as part of that. The next dependency bump will happen in January. 

I'm closing this PR, please feel free to re-open if you think this needs to be fixed urgently.

TomerJLevy:
Hi @arjan-bal, I understand that this shouldn’t be directly affected by the CVE, and of course, this upgrade can wait until January alongside other upgrades. However, could you respond to the conversation I had with @stefanb regarding the `grpc-go` version used in `cmd/protoc-gen-go-grpc/go.mod`? Is there a specific reason why this package is using the latest release instead of the current release?

arjan-bal:
> Hi @arjan-bal, I understand that this shouldn’t be directly affected by the CVE, and of course, this upgrade can wait until January alongside other upgrades. However, could you respond to the conversation I had with @stefanb regarding the `grpc-go` version used in `cmd/protoc-gen-go-grpc/go.mod`? Is there a specific reason why this package is using the latest release instead of the current release?

`protoc-gen-go-grpc` is a separate binary from the gRPC Go  library. It is used to generate client and server stubs. Prior to https://github.com/grpc/grpc-go/pull/7438, `protoc-gen-go-grpc` didn't depend on gRPC at all. This PR added a test dependency on gRPC Go using a replace directive to specify current version of gRPC Go. 

After releasing `protoc-gen-go-grpc`, we discovered that using a replace directive in a binary module is not allowed (https://github.com/grpc/grpc-go/issues/7448). To resolve this, the replace directive was removed in https://github.com/grpc/grpc-go/pull/7451. Since then, we haven't updated the grpc-go version since then. However, we should transition to using the the latest gRPC Go release instead remaining on `v1.65.0`.

FYI @dfawley

PR #7969, opened 2024-12-26
leastrequest: Delegate subchannel creation to pickfirst

arjan-bal:
This PR switches the least request balancer to defer to pick first instead of creating and handling SubConns itself. This is part of Dualstack, and makes this a petiole.

RELEASE NOTES:
* balancer/leastrequest: Least request uses pick first child policies to manage endpoints instead of creating SubConns.
* balancer/leastrequest: Least request uses the `Endpoints` field instead of the `Addresses` field in [resolver state](https://pkg.go.dev/google.golang.org/grpc@v1.69.2/resolver#State) updates.


balancer/leastrequest/leastrequest.go
```diff
@@ -80,104 +92,167 @@ func (bb) Name() string {
 }
 
 func (bb) Build(cc balancer.ClientConn, bOpts balancer.BuildOptions) balancer.Balancer {
-	b := &leastRequestBalancer{scRPCCounts: make(map[balancer.SubConn]*atomic.Int32)}
-	baseBuilder := base.NewBalancerBuilder(Name, b, base.Config{HealthCheck: true})
-	b.Balancer = baseBuilder.Build(cc, bOpts)
+	b := &leastRequestBalancer{
+		ClientConn:        cc,
+		endpointRPCCounts: resolver.NewEndpointMap(),
+		choiceCount:       2,
```
dfawley, line 98: Why is this initialized to 2?  This should be getting set from the config, and shouldn't ideally matter until the first time a ClientConn update happens.

balancer/leastrequest/leastrequest.go
```diff
@@ -80,104 +92,167 @@ func (bb) Name() string {
...
+	// Embeds balancer.ClientConn because needs to intercept UpdateState calls
+	// from the child balancer.
+	balancer.ClientConn
+	child  balancer.Balancer
+	logger *internalgrpclog.PrefixLogger
 
+	mu          sync.Mutex
 	choiceCount uint32
-	scRPCCounts map[balancer.SubConn]*atomic.Int32 // Hold onto RPC counts to keep track for subsequent picker updates.
+	// endpointRPCCounts holds  RPC counts to keep track for subsequent picker
```
dfawley, line 115: Nit: there's an extra space `holds  RPC`

balancer/leastrequest/leastrequest.go
```diff
@@ -80,104 +92,167 @@ func (bb) Name() string {
...
+	// updates.
+	endpointRPCCounts *resolver.EndpointMap // endpoint -> *atomic.Int32
+}
+
+// Close implements balancer.Balancer.
```
dfawley, line 120: These kinds of comments are unnecessary as `leastRequestBalancer` is not exported.

balancer/leastrequest/leastrequest.go
```diff
@@ -80,104 +92,167 @@ func (bb) Name() string {
...
 }
 
 type leastRequestBalancer struct {
-	// Embeds balancer.Balancer because needs to intercept UpdateClientConnState
-	// to learn about choiceCount.
-	balancer.Balancer
+	// Embeds balancer.ClientConn because needs to intercept UpdateState calls
+	// from the child balancer.
+	balancer.ClientConn
+	child  balancer.Balancer
```
dfawley, line 110: Why change `child` to not be embedded anymore, as it used to be?  At least two methods are still (effectively) directly propagated (`ResolverError` and `UpdateSubConnState`).

balancer/leastrequest/leastrequest.go
```diff
@@ -80,104 +92,167 @@ func (bb) Name() string {
...
+	})
 }
 
 type picker struct {
 	// choiceCount is the number of random SubConns to find the one with
 	// the least request.
 	choiceCount uint32
-	// Built out when receives list of ready RPCs.
-	subConns []scWithRPCCount
+	// Built out when receives list of ready child pickers.
```
dfawley, line 224: This is sort of a weird comment to have at all.  The field here is simply constructed when the picker is created.  The _picker itself_ is "built out when [the balancer] receives [an updated] list of ready child pickers".

Feel free to remove.

dfawley (commented):
It seems like some new test cases should be needed, in light of the changes here?

dfawley (approved):

PR #7974, opened 2024-12-30
deps: bump github.com/envoyproxy/go-control-plane to v0.13.4

mmorel-35:
#### Description

This updates `github.com/envoyproxy/go-control-plane` to v0.13.4 and import new `github.com/envoyproxy/go-control-plane/envoy` module

Related to https://github.com/envoyproxy/go-control-plane/issues/1074

RELEASE NOTES: None


cmd/protoc-gen-go-grpc/go.mod
```diff
@@ -1,6 +1,6 @@
 module google.golang.org/grpc/cmd/protoc-gen-go-grpc
 
-go 1.22
+go 1.22.8
```
arjan-bal, line 3: We got pushback from users when we specified a min Go minor version in our go.mod. It forced all users to bump their min versions to use the latest version of gRPC Go. So we decided to remove the minor version in https://github.com/grpc/grpc-go/pull/7831. Let me open an issue with go-control-plane to see if they are willing to drop the minor version.

arjan-bal: I've opened Opened https://github.com/envoyproxy/go-control-plane/issues/1077, let's see what they think.

scripts/vet.sh
```diff
@@ -163,7 +163,7 @@ XXXXX gRPC internal usage deprecation errors:
 BalancerAttributes is deprecated:
 CredsBundle is deprecated:
 internal.Logger is deprecated:
-Metadata is deprecated: use Attributes instead.
+Metadata is deprecated:
```
arjan-bal, line 166: Why was this needed? The deprecation message is still the same: https://github.com/grpc/grpc-go/blob/6f41085574082e83c928771cd494613568fd8a38/resolver/resolver.go#L119

arjan-bal: The new deprecation is coming from [envoy/config/rbac/v3/rbac.proto](
https://github.com/envoyproxy/envoy/blob/e294469598d0980a792221ce712eab14cc6511ab/api/envoy/config/rbac/v3/rbac.proto#L269-L272).

I think we should add a new entry instead of relaxing this pattern to avoid accidentally hiding other deprecation warnings in the future.
```
GetMetadata is deprecated: Marked as deprecated in envoy/config/rbac
```

arjan-bal: > By the way, would you be open to use golangci-lint to simplify linters maintenance ?

I think we've been able to achieve what we need using vet and staticcheck directly combined with regexes, so there wasn't a need to add a new dependency. If you think think using `golangci-lint` has significant benefits, please do file an issue and we can discuss this.

cmd/protoc-gen-go-grpc/go.mod
```diff
@@ -2,14 +2,16 @@ module google.golang.org/grpc/cmd/protoc-gen-go-grpc
 
 go 1.22
 
+replace google.golang.org/grpc => ../..
+
```
arjan-bal, line 6: `protoc-gen-go-grpc` is a separate binary from the gRPC Go  library. It is used to generate client and server stubs. Prior to https://github.com/grpc/grpc-go/pull/7438, `protoc-gen-go-grpc` didn't depend on gRPC at all. This PR added a test dependency on gRPC Go using a replace directive to specify current version of gRPC Go. 

After releasing `protoc-gen-go-grpc`, we discovered that using a replace directive in a binary module is not allowed (https://github.com/grpc/grpc-go/issues/7448). To resolve this, the replace directive was removed in https://github.com/grpc/grpc-go/pull/7451. Since then, we haven't updated the grpc-go version.

Updating the deps in this module may not be required.

arjan-bal (approved):
LGTM, adding another reviewers for a second approval.

examples/go.mod
```diff
@@ -43,7 +43,7 @@ require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/census-instrumentation/opencensus-proto v0.4.1 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
-	github.com/envoyproxy/go-control-plane v0.13.1 // indirect
+	github.com/envoyproxy/go-control-plane/envoy v1.32.3 // indirect
```
dfawley, line 46: 🎉

dfawley: (I'm extremely happy to see a v1 here -- thanks!!)

dfawley (approved):

Issue #7922, opened 2024-12-11
Updating from go-grpc v1.64.0 to v1.68.1 broke our AWS deployments, migration path is unclear


AnomalRoil:
### What version of gRPC are you using?
v1.68.1

### What version of Go are you using (`go version`)?
go1.22.10

### What operating system (Linux, Windows, …) and version?
Linux 4.14.348-265.565.amzn2.x86_64

### What did you do?
We routinely updated our dependencies and released a new patch version without thinking too much about it since it built locally, passed integration tests, built in Github actions and all...

We deployed it in our test AWS environment and it is now _somewhat_ broken:
- the nodes on AWS using the newest version were unable to talk to each other, failing with:
   ```
   transport: authentication handshake failed: credentials: cannot check peer: missing selected ALPN property
   ```
- the nodes on AWS using the newest version were still able to talk to nodes running the older version
- the nodes on AWS were still able to receive gRPC calls from the older nodes

So it was really just the newer nodes being unable to talk to each other, while being fine with the older ones, which led us to discover https://github.com/grpc/grpc-go/issues/7769 which in turn led us to discover that release [v1.67.0](https://github.com/grpc/grpc-go/releases/tag/v1.67.0) introduced the new behaviour:
> In accordance with [RFC 7540](https://tools.ietf.org/html/rfc7540#section-3.3), clients and servers will now reject TLS connections that don't support ALPN. This can be disabled by setting the environment variable GRPC_ENFORCE_ALPN_ENABLED to false (case insensitive). Please file a bug if you encounter any issues with this behavior.

Here is our current use case that broke:
- The NAT gateway has a public IP and this is effectively the IP that will be the source for gRPC connections from our node.
- The node is accessible through a different public IP via a AWS Classic Load Balancer. The load balancer acts as firewall, using an attached security group for allowlisting IPS, corresponding to the rest of the servers.
- The load balancer's IP has the hostname associated to it and performs TLS termination on the incoming gRPC connections.

It is unclear what the path forward is for us in order to support ALPN, and you said to file a bug if we encountered any issues with the new behaviour, so here am I.

But I don't think there's anything to do on your side, except maybe have a migration path / improved documentation around that error message since a little Google search didn't return anything recent about this issue. Hopefully this issue might point developers googling the issue in the right direction. 

### What did you expect to see?

Nodes continuing to connect to each others just fine.
[...]

PR #7980, opened 2025-01-06
credentials: Add experimental credentials that don't enforce ALPN

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/7922
Related Issue: https://github.com/grpc/grpc-go/issues/434

This change copies the existing TLS credentials, removing the code for ALPN enforcement. These credentials will be removed in a couple of releases and users who still need to disable ALPN can copy the credentials as is. Changes made:
* Copy `credentials/tls.go` to `experimental/credentials/tls.go`. Add suffix `WithALPNDisabled` to constructors.
* Copy required packages from `internal/credentials` to `experimental/credentials/internal`.
* Improve error message when connection fails due to ALPN.

RELEASE NOTES:
* experimental/credentials: Experimental transport credentials are added which don't enforce ALPN. These will be removed in upcoming grpc-go releases. Users must not rely on these credentials directly. Instead, they should either vendor a specific version of gRPC or copy the relevant credentials into their own codebase if absolutely necessary.

### DIFF
--- a/credentials/tls.go
+++ b/credentials/tls.go
@@ -32,6 +32,8 @@ import (
 	"google.golang.org/grpc/internal/envconfig"
 )
 
+const alpnFailureHelpMessage = "If you upgraded from a grpc-go version earlier than 1.67, your TLS connections may stopped working due to ALPN enforcement. For more details, see: https://github.com/grpc/grpc-go/issues/434. To disable ALPN enforcement, set the environment variable GRPC_ENFORCE_ALPN_ENABLED to false, or use the experimental credentials available under experimental/credentials. Note that these workarounds are intended for migration purposes and will be removed in future grpc-go versions."
+
 var logger = grpclog.Component("credentials")
 
 // TLSInfo contains the auth information for a TLS authenticated connection.
@@ -128,7 +130,7 @@ func (c *tlsCreds) ClientHandshake(ctx context.Context, authority string, rawCon
 	if np == "" {
 		if envconfig.EnforceALPNEnabled {
 			conn.Close()
-			return nil, nil, fmt.Errorf("credentials: cannot check peer: missing selected ALPN property")
+			return nil, nil, fmt.Errorf("credentials: cannot check peer: missing selected ALPN property. %s", alpnFailureHelpMessage)
 		}
 		logger.Warningf("Allowing TLS connection to server %q with ALPN disabled. TLS connections to servers with ALPN disabled will be disallowed in future grpc-go releases", cfg.ServerName)
 	}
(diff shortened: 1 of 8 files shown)


credentials/tls.go
```diff
@@ -32,6 +32,8 @@ import (
 	"google.golang.org/grpc/internal/envconfig"
 )
 
+const alpnFailureHelpMessage = "If you upgraded from a grpc-go version earlier than 1.67, your TLS connections may stopped working due to ALPN enforcement. For more details, see: https://github.com/grpc/grpc-go/issues/434. To disable ALPN enforcement, set the environment variable GRPC_ENFORCE_ALPN_ENABLED to false, or use the experimental credentials available under experimental/credentials. Note that these workarounds are intended for migration purposes and will be removed in future grpc-go versions."
```
dfawley, line 35: Nit: "may _have_ stopped working"

Also I think I'd remove everything after "see 434".  The rest of the explanation can be in there.

experimental/credentials/credentials_test.go
```diff
@@ -0,0 +1,252 @@
+/*
+ *
+ * Copyright 2016 gRPC authors.
```
dfawley, line 3: Update please

experimental/credentials/tls.go
```diff
@@ -0,0 +1,227 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package credentials contains experimental TLS credentials.
```
dfawley, line 19: This should have a big scary comment about how these credentials should only be used to maintain compatibility for people relying upon interoperability with clients violating the HTTP/2 specification, and that it will be deleted, so users should not use it directly.  Instead they either need to vendor gRPC or copy these credentials.

experimental/credentials/tls_ext_test.go
```diff
@@ -0,0 +1,393 @@
+/*
+ *
+ * Copyright 2023 gRPC authors.
```
dfawley, line 3: Date

experimental/credentials/tls.go
arjan-bal: Updated the package comment, please take a look.

experimental/credentials/tls_ext_test.go
arjan-bal: Fixed throughout.

dfawley (approved):

Issue #7923, opened 2024-12-11
Provide a way to adjust the initial window size that does not also disable BDP estimation


dfawley:
There are at least a couple options here:

1. Change the existing Dial/ServerOptions ([e.g.](https://pkg.go.dev/google.golang.org/grpc#WithInitialConnWindowSize)) so they do not disabled BDP estimation
  - Optionally, add another set of options that does disable BDP estimation.
2. Add new Dial/ServerOptions to set the size like the existing ones, but which do not also disable BDP esimation.

dfawley:
Proposal:

1. Add new settings:

```go
package grpc

func WithStaticStreamWindowSize(int32) DialOption
func WithStaticConnWindowSize(int32) DialOption

func StaticStreamWindowSize(int32) ServerOption
func StaticConnWindowSize(int32) ServerOption
```

These will behave the same as the current options.

2. After 2 releases, with a notice that the behavior is changing in the release notes, change the current options so they do not disable BDP estimation.  Also add:

```go
func WithInitialStreamWindowSize(int32) DialOption // just does "return WithInitialWindowSize(n)"

func InitialStreamWindowSize(int32) ServerOption // just does "return InitialWindowSize(n)"
```

And mark `WithInitialWindowSize` and `InitialWindowSize` as deprecated.

vinothkumarr227:
I'm looking into this. Please assign the issue to me.

arjan-bal:
@vinothkumarr227 is this issue completely addressed?

vinothkumarr227:
@arjan-bal Yeah, this addresses the issue and has also been merged

eshitachandwani:
Hey , @arjan-bal  I dont think the 2nd part of this has been addressed i.e. 
> After 2 releases, with a notice that the behavior is changing in the release notes, change the current options so they do not disable BDP estimation. Also add:
func WithInitialStreamWindowSize(int32) DialOption // just does "return WithInitialWindowSize(n)"
func InitialStreamWindowSize(int32) ServerOption // just does "return InitialWindowSize(n)"
And mark WithInitialWindowSize and InitialWindowSize as deprecated.

Currently the new and the old dialoptions have a same behavior of disabling the BDP estimation.
Correct me if I have misunderstood @vinothkumarr227

arjan-bal:
I see, thanks for catching this!

arjan-bal:
[...]

PR #7989, opened 2025-01-08
grpc: add new Dial/ServerOptions to set the size like the existing ones, but which do not also disable BDP esimation

janardhanvissa:
Partially addresses: https://github.com/grpc/grpc-go/issues/7923

RELEASE NOTES:
- grpc: add new Dial/ServerOptions to set the size like the existing ones, but which do not also disable BDP esimation

### DIFF
--- a/dialoptions.go
+++ b/dialoptions.go
@@ -94,6 +94,8 @@ type dialOptions struct {
 	idleTimeout                 time.Duration
 	defaultScheme               string
 	maxCallAttempts             int
+	staticConnWindowSize        int32
+	staticStreamWindowSize      int32
 }
 
 // DialOption configures how we set up the connection.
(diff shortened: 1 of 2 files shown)


purnesh42H:
@janardhanvissa could you add more details why is it partially addressing the issue?

janardhanvissa:
> @janardhanvissa could you add more details why is it partially addressing the issue?

 As per the proposal by Doug, the changes to the existing WithInitialWindowSize and InitialWindowSize options should be delayed for two releases. This delay ensures backward compatibility and gives users adequate time to adjust to the changes.

dfawley:
> What are the existing options? How do they disable bdp estimation currently?

They are mentioned briefly in the proposal in the second comment:

> And mark `WithInitialWindowSize` and `InitialWindowSize` as deprecated.

How do they disable it?  By passing configuration that the transport sees and disables BDP estimation.  E.g. for the client: https://github.com/grpc/grpc-go/blob/67bee55a47db557ad7074a73eb922c2f22fbc454/internal/transport/http2_client.go#L315


This PR, as-is, will not have any actual effect, so it needs work to pass these settings to the transport and make the transport use them correctly.

purnesh42H:
@janardhanvissa let's pause this and try to finish the otel tracing work.

Issue #7982, opened 2025-01-07
Flaky test: Test/AggregateCluster_BadEDSFromError_GoodToBadDNS


janardhanvissa:
Sample run: 

https://github.com/grpc/grpc-go/actions/runs/12644179299/job/35231337660?pr=7967

purnesh42H:
Putting as P2. We can bump the priority if its very frequent.

arjan-bal:
This seems to be a data race when clusterresolver changes the endpoints slice while pickfirstleaf LB Policy is holding a reference to it:
```
==================
WARNING: DATA RACE
Read at 0x00c00062ccc0 by goroutine 2594:
  runtime.slicecopy()
      /opt/hostedtoolcache/go/1.23.4/x64/src/runtime/slice.go:355 +0x0
  google.golang.org/grpc/balancer/pickfirst/pickfirstleaf.(*pickfirstBalancer).UpdateClientConnState()
      /home/runner/work/grpc-go/grpc-go/balancer/pickfirst/pickfirstleaf/pickfirstleaf.go:295 +0x1174
  google.golang.org/grpc/internal/balancer/gracefulswitch.(*Balancer).UpdateClientConnState()
      /home/runner/work/grpc-go/grpc-go/internal/balancer/gracefulswitch/gracefulswitch.go:193 +0x2b4
  google.golang.org/grpc/xds/internal/balancer/clusterimpl.(*clusterImplBalancer).UpdateClientConnState()
      /home/runner/work/grpc-go/grpc-go/xds/internal/balancer/clusterimpl/clusterimpl.go:290 +0x83a
  google.golang.org/grpc/internal/balancer/gracefulswitch.(*Balancer).UpdateClientConnState()
      /home/runner/work/grpc-go/grpc-go/internal/balancer/gracefulswitch/gracefulswitch.go:193 +0x2b4
  google.golang.org/grpc/xds/internal/balancer/outlierdetection.(*synchronizingBalancerWrapper).updateClientConnState()
      /home/runner/work/grpc-go/grpc-go/xds/internal/balancer/outlierdetection/balancer.go:884 +0xf7
  google.golang.org/grpc/xds/internal/balancer/outlierdetection.(*outlierDetectionBalancer).UpdateClientConnState()
      /home/runner/work/grpc-go/grpc-go/xds/internal/balancer/outlierdetection/balancer.go:321 +0xdc4
[...]

PR #7991, opened 2025-01-08
clustetresolver: Copy endpoints.Addresses slice from DNS updates to avoid data races

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/7982

## Background
`clusterresolver` uses the `lastUpate()` method in the resourceResolver interface that has implementations for both EDS and DNS.
https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/xds/internal/balancer/clusterresolver/resource_resolver.go#L60-L68

When the child balancer's state needs to be updated, `clusterresolver` calls `lastUpdate` on DNS and EDS. If there were no changes since the last time `lastUpdate` was called, the child resolver will return the same update object. In case of EDS, the update contains a list of xDS endpoints, while for DNS, since https://github.com/grpc/grpc-go/pull/7858 the update contains a list of `resolver.Endpoint` grouped by locality. 

In case of EDS, when the xDS endpoints are converted to resolver.Endpoints, new Endpoint objects are allocated that don't share any memory with the previous update sent to the child balancer.
https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/xds/internal/balancer/clusterresolver/configbuilder.go#L189

However, in case of DNS, the endpoint objects are copied by iterating on the original slice returned by DNS. As `Endpoint.Addresses` is a slice field, both the original and the copy share the same slice. This is the cause of the race.
https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/xds/internal/balancer/clusterresolver/configbuilder.go#L140-L153
If the child tries to access endpoints.Addresses while clusterresolver is creating the endpoints list for the next update, a race occurs.

## Tested
Reproduced the failure on forge, it failed within 1000 runs. Verified the test no longer races in 10000 runs with the fix.

RELEASE NOTES: N/A

### DIFF
--- a/xds/internal/balancer/clusterresolver/clusterresolver.go
+++ b/xds/internal/balancer/clusterresolver/clusterresolver.go
@@ -248,32 +248,37 @@ func (b *clusterResolverBalancer) updateChildConfig() {
 		b.logger.Infof("Built child policy config: %s", pretty.ToJSON(childCfg))
 	}
 
-	flattenedAddrs := make([]resolver.Address, len(endpoints))
-	for i := range endpoints {
-		for j := range endpoints[i].Addresses {
-			addr := endpoints[i].Addresses[j]
-			addr.BalancerAttributes = endpoints[i].Attributes
+	flattenedAddrs := make([]resolver.Address, 0, len(endpoints))
+	// Send a copy of the endpoints slice to the children as balancer may use
+	// the same endpoints slice in subsequent updates.
+	endpointsCopy := make([]resolver.Endpoint, 0, len(endpoints))
+	for _, endpoint := range endpoints {
+		addrs := make([]resolver.Address, 0, len(endpoint.Addresses))
+		for _, addr := range endpoint.Addresses {
+			addr.BalancerAttributes = endpoint.Attributes
+			addrs = append(addrs, addr)
 			// If the endpoint has multiple addresses, only the first is added
 			// to the flattened address list. This ensures that LB policies
 			// that don't support endpoints create only one subchannel to a
 			// backend.
-			if j == 0 {
-				flattenedAddrs[i] = addr
+			if len(addrs) == 1 {
+				flattenedAddrs = append(flattenedAddrs, addr)
 			}
-			// BalancerAttributes need to be present in endpoint addresses. This
-			// temporary workaround is required to make load reporting work
-			// with the old pickfirst policy which creates SubConns with multiple
-			// addresses. Since the addresses can be from different localities,
-			// an Address.BalancerAttribute is used to identify the locality of the
-			// address used by the transport. This workaround can be removed once
-			// the old pickfirst is removed.
-			// See https://github.com/grpc/grpc-go/issues/7339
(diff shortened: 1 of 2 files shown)


xds/internal/balancer/clusterresolver/clusterresolver.go
```diff
@@ -248,32 +248,37 @@ func (b *clusterResolverBalancer) updateChildConfig() {
...
 	}
 
-	flattenedAddrs := make([]resolver.Address, len(endpoints))
-	for i := range endpoints {
-		for j := range endpoints[i].Addresses {
-			addr := endpoints[i].Addresses[j]
-			addr.BalancerAttributes = endpoints[i].Attributes
+	flattenedAddrs := make([]resolver.Address, 0, len(endpoints))
+	// Send a copy of the endpoints slice to the children as balancer may use
+	// the same endpoints slice in subsequent updates.
```
dfawley, line 253: Can you explain more about this?  It seems unexpected to me that endpoints would ever be mutated.

dfawley: Sorry, I'm still not understanding this fully.  Why is the `clusterresolver` mutating the endpoints slice between updates instead of creating a new slice and replacing the old one?

That part seems wrong to me, but I'm probably missing something.

arjan-bal: `clusterresolver` uses the `lastUpate()` method in the resourceResolver interface that has implementations for both EDS and DNS.
https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/xds/internal/balancer/clusterresolver/resource_resolver.go#L60-L68

When the child balancer's state needs to be updated, `clusterresolver` calls `lastUpdate` on DNS and EDS. If there were no changes since the last time `lastUpdate` was called, the child resolver will return the same update object. In case of EDS, the update contains a list of xDS endpoints, while for DNS the update contains a list of `resolver.Endpoint` grouped by locality. 

In case of EDS, when the xDS endpoints are converted to resolver.Endpoints, new Endpoint objects are allocated that don't share any memory with the previous update sent to the child balancer.
https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/xds/internal/balancer/clusterresolver/configbuilder.go#L189

However, in case of DNS, the endpoint objects are copied by iterating on the original slice returned by DNS. As `Endpoint.Addresses` is a slice field, both the original and the copy share the same slice. This is the cause of the race.
https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/xds/internal/balancer/clusterresolver/configbuilder.go#L140-L153

I've change the PR to copy the `Endpoint.Addresses` field only for DNS which is the cause of the race.

dfawley (approved):
Thanks!  I think I like this fix better, but it's not clear to me which one to prefer.  Since it's all internal code, it doesn't really matter.

PR #7992, opened 2025-01-09
stats/opentelemetry: add trace event for name resolution delay

aranjans:
This PR adds a bonus feature to grfc [A72](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#a72-opentelemetry-tracing) to emit a trace event for when the initial RPCs block on the name resolver.

RELEASE NOTES:

- stats/opentelemetry: add trace event for name resolution delay.


clientconn.go
```diff
@@ -613,12 +613,18 @@ type ClientConn struct {
...
-	resolverWrapper *ccResolverWrapper         // Always recreated whenever entering idle to simplify Close.
-	balancerWrapper *ccBalancerWrapper         // Always recreated whenever entering idle to simplify Close.
-	sc              *ServiceConfig             // Latest service config received from the resolver.
-	conns           map[*addrConn]struct{}     // Set to nil on close.
-	mkp             keepalive.ClientParameters // May be updated upon receipt of a GoAway.
+	mu sync.RWMutex
+	// nameResolutionStartTime track the start time since name resolution started.
+	nameResolutionStartTime time.Time
+	// nameResolutionInProgress indicate if name resolution is in progress.
+	nameResolutionInProgress bool
```
dfawley, line 620: I don't think we want these to be fields on the `ClientConn`.  These are supposed to be part of the RPC's lifecycle since that's what we're tracing.

clientconn.go
```diff
@@ -674,13 +680,27 @@ func (cc *ClientConn) Connect() {
 // context expires.  Returns nil unless the context expires first; otherwise
 // returns a status error based on the context.
 func (cc *ClientConn) waitForResolvedAddrs(ctx context.Context) error {
+	// Set the start time for name resolution if it's not already set.
+	cc.mu.Lock()
```
dfawley, line 684: This is called on every RPC.

You're unconditionally setting this flag and timestamp at the start of every RPC.  That can't be right.

arjan-bal:
Closing this as author is no longer working on this. FYI @purnesh42H @dfawley.

PR #7994, opened 2025-01-09
chore: setup golangci-lint

mmorel-35:
#### Description 

Following https://github.com/grpc/grpc-go/pull/7974#discussion_r1908155660

Using golangci-lint will provide a common config file for all already used linters.
With this config comments can be provided.

Until now revive was only applied in the main module and not in submodules.
With one tool, this behavior will be avoided.

It will also provide a simpler way to provide new linters

RELEASE NOTES: None


purnesh42H:
@mmorel-35 thanks for the PR. Overall it looks good with added benefit of checking submodules. I am discussing with other maintainers for any concerns.

PR #8007, opened 2025-01-13
endpointsharding: Export parsed pickfirst config instead of json string

arjan-bal:
All the users of `PickFirstConfig` needed to handle parsing errors while using the exported config. This PR exports the parsed config making the usage simpler for users.

RELEASE NOTES:
* balancer/endpointsharding: The parsed pickfirst load balancer config is exported instead of the load balancer config json.

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -37,16 +37,24 @@ import (
 	"google.golang.org/grpc/balancer/base"
 	"google.golang.org/grpc/balancer/pickfirst/pickfirstleaf"
 	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/internal/balancer/gracefulswitch"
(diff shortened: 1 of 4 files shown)


balancer/endpointsharding/endpointsharding.go
```diff
@@ -37,16 +37,24 @@ import (
...
 	"google.golang.org/grpc/internal/balancer/gracefulswitch"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/serviceconfig"
 )
 
-// PickFirstConfig is a pick first config without shuffling enabled.
-var PickFirstConfig string
+var (
+	// PickFirstConfig is a pick first config without shuffling enabled.
+	PickFirstConfig serviceconfig.LoadBalancingConfig
```
arjan-bal, line 48: This is a pubic API change, but the package is experimental. @dfawley should I export a new var instead of replacing the existing one?

dfawley: I think we should be fine to just change it since it's experimental and very new.

PR #8010, opened 2025-01-15
github: Run deps workflow against base branch and add branch name suffixes for easier debuggability

arjan-bal:
Running the test against the base branch (instead of master) ensures the tests don't fail unnecessarily for PRs raised against release branches. 

Env vars available in Github actions: [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)

RELEASE NOTES: N/A

### DIFF
--- a/.github/workflows/deps.yml
+++ b/.github/workflows/deps.yml
@@ -34,7 +34,8 @@ jobs:
           AFTER="$(mktemp -d)"
 
           scripts/gen-deps.sh "${AFTER}"
-          git checkout origin/master
+          # GITHUB_BASE_REF is set when the job is triggered by a PR.
+          git checkout origin/"${GITHUB_BASE_REF:-master}"
           scripts/gen-deps.sh "${BEFORE}"
 
           echo "Comparing dependencies..."
--- a/.github/workflows/deps.yml
+++ b/.github/workflows/deps.yml
@@ -34,23 +34,16 @@ jobs:
           TEMP_DIR="$(mktemp -d)"
           # GITHUB_BASE_REF is set when the job is triggered by a PR.
           TARGET_REF="${GITHUB_BASE_REF:-master}"
-          # GITHUB_HEAD_REF is set when the job is triggered by a PR.
-          SOURCE_REF="${GITHUB_HEAD_REF:-source}"
 
-          # Replace special characters in the branch names.
-          BEFORE_DIR_NAME="$(echo "${SOURCE_REF}" | sed 's/[^a-zA-Z0-9._-]/_/g')"
-          AFTER_DIR_NAME="$(echo "${TARGET_REF}" | sed 's/[^a-zA-Z0-9._-]/_/g')"
-
-
-          mkdir "${TEMP_DIR}/${AFTER_DIR_NAME}"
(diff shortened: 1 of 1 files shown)


.github/workflows/deps.yml
```diff
@@ -30,11 +30,18 @@ jobs:
...
-          BEFORE="$(mktemp -d)"
-          AFTER="$(mktemp -d)"
+          set -eux
+          BASE_REF="${GITHUB_BASE_REF:-master}"
+          # Suffixes passed to mktemp must not contain "/".
+          BEFORE_SUFFIX="${BASE_REF//\//_}"
+          AFTER_SUFFIX="${GITHUB_HEAD_REF//\//_}"
+
+          BEFORE="$(mktemp -d --suffix="-${BEFORE_SUFFIX}")"
+          AFTER="$(mktemp -d --suffix="-${AFTER_SUFFIX}")"
```
dfawley, line 40: Idea:

How about have one `mktemp -d` and then make two subdirectories of it which are the same as the existing `BEFORE_SUFFIX` and `AFTER_SUFFIX` and then `cd` to the `mktemp` dir before  doing the `diff` so the relative path names are the names of the git refs?  WDYT?

arjan-bal: That's a good idea. Changed accordingly.

arjan-bal: I changed the subdir names to `before` and `after` to keep things simple. It avoids having to sanitize the branch names to ensure they are valid linux paths, handling same source and target ref names and other possible issues.

.github/workflows/deps.yml
```diff
@@ -30,14 +30,20 @@ jobs:
       # Run the commands to generate dependencies before and after and compare.
       - name: Compare dependencies
         run: |
-          BEFORE="$(mktemp -d)"
-          AFTER="$(mktemp -d)"
+          set -eux
```
dfawley, line 33: I would prefer not to have this (at least the `-x` (tracing) part) since the output _should_ generally be pretty human readable, and the commands should be pretty stable.

The reason I did `-x` in the vet script was mainly laziness around not wanting to output everything the script was doing before & after it did it.  (In hindsight that was probably a bad idea, as the output confuses most people.)

.github/workflows/deps.yml
```diff
@@ -30,14 +30,20 @@ jobs:
...
+
+          git checkout "origin/${TARGET_REF}"
+          mkdir "${TEMP_DIR}/before"
+          scripts/gen-deps.sh "${TEMP_DIR}/before"
 
           echo "Comparing dependencies..."
+          cd "${TEMP_DIR}"
           # Run grep in a sub-shell since bash does not support ! in the middle of a pipe
-          diff -u0 -r "${BEFORE}" "${AFTER}" | bash -c '! grep -v "@@"'
+          diff -u0 -r "./before" "./after" | bash -c '! grep -v "@@"'
```
dfawley, line 48: Nit: remove the `./` and I think the output will look a bit nicer

dfawley (approved):

.github/workflows/deps.yml
arjan-bal: Removed.

PR #8017, opened 2025-01-17
protoc-gen-go-grpc: skip TestUnimplementedEmbedding

purnesh42H:
We need to skip the `TestUnimplementedEmbedding` until google3 usages are fixed to embed `UnimplementedTestServiceServer`

RELEASE NOTES: None


dfawley:
Let's just not import this file, instead, since it's the only test in it.  We don't want to reduce OSS test coverage due to an internal issue.

PR #8025, opened 2025-01-22
interop: improve rpc_soak and channel_soak test to cover concurrency in Go

zbilun:
RELEASE NOTES:

Improved the rpc_soak and channel_soak tests to support concurrency in Go. This change increases the test coverage of the C2P E2E load test and follows a similar approach used in the Java implementation.
This PR aims to enhance the test coverage of the C2P E2E load test by improving the rpc_soak and channel_soak tests to support concurrency in Go. The updated logic closely follows the approach used in the Java implementation, which has already been merged. 
Noted. The interop_test.sh has been updated to include tests for rpc_soak and channel_soak.

rpc_soak:
The client performs many large_unary RPCs in sequence over the same channel. The test can run in either a concurrent or non-concurrent mode, depending on the number of threads specified (soak_num_threads).

channel_soak:
Similar to rpc_soak, but this time each RPC is performed on a new channel. The channel is created just before each RPC and is destroyed just after.

@dfawley

RELEASE NOTES: n/a

### DIFF
--- a/interop/interop_test.sh
+++ b/interop/interop_test.sh
@@ -92,6 +92,8 @@ CASES=(
   "unimplemented_service"
   "orca_per_rpc"
   "orca_oob"
+  "rpc_soak"
+  "channel_soak"
 )
 
 # Build server
--- a/interop/soak_tests.go
+++ b/interop/soak_tests.go
@@ -16,9 +16,7 @@
 *
  */
 
-
 package interop
-
 
 import (
 	"bytes"
@@ -28,16 +26,13 @@ import (
 	"sync"
 	"time"
 
-
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/benchmark/stats"
 	"google.golang.org/grpc/peer"
 
-
 	testgrpc "google.golang.org/grpc/interop/grpc_testing"
 	testpb "google.golang.org/grpc/interop/grpc_testing"
 )
-
 
 // SoakWorkerResults stores the aggregated results for a specific worker during the soak test.
 type SoakWorkerResults struct {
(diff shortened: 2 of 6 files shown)


dfawley:
@zbilun please run `gofmt` and make sure the `vet` check is passing.

Also can you show how this differs from the previous version?  Ideally it would be in two commits: one exactly as before and one with the new updates, but if you can point out the difference some other way, that's fine, too.

interop/soak_tests.go
```diff
@@ -0,0 +1,218 @@
...
+	// Calculate latency and return result.
+	latency = time.Since(start)
+	return latency, nil
+}
+
+
+func executeSoakTestInWorker(ctx context.Context, config SoakTestConfig, startTime time.Time, workerID int, soakWorkerResults *SoakWorkerResults) {
+	timeoutDuration := config.OverallTimeout
+	soakIterationsPerWorker := config.Iterations / config.NumWorkers
+	if soakWorkerResults.Latencies == nil {
```
zbilun, line 107: The root cause of the previous seg issue was that the Latencies field in SoakWorkerResults was not initialized, leading to a nil pointer dereference when attempting to call .Add(). Thus, to fix the issue,  in the executeSoakTestInWorker function, we initialize the Latencies histogram before using it.

interop/interop_test.sh
```diff
@@ -92,6 +92,8 @@ CASES=(
   "unimplemented_service"
   "orca_per_rpc"
   "orca_oob"
+  "rpc_soak"
```
zbilun, line 95: These two test cases have been added to improve  the interop_test.sh.

dfawley (approved):

PR #8026, opened 2025-01-22
balancer: Enforce embedding requirement for balancer.ClientConn

arjan-bal:
Similar to https://github.com/grpc/grpc-go/pull/7840, this PR ensures gRPC can add new methods to the ClientConn interface. This change prepares us for passing the metrics recorder to LB policies through the balancer.ClientConn interface instead of the `balancer.BuildOptions` as being done presently. Using `balancer.BuildOptions` doesn't allow gRPC to ensure users are propagating the metrics recorder to child LB Policies and can lead to nil pointer dereferences at runtime.

RELEASE NOTES:
* balancer: An internal method is added to the `balancer.ClientConn` interface to force implementors to embed a delegate implementation. This helps grpc-go add new methods to the interface without breaking builds for implementors of custom LB Policies.

PR #8027, opened 2025-01-22
balancer: Move metrics recorder from BuildOptions to ClientConn

arjan-bal:
This PR changes the way LB policies obtain a metrics recorder from the channel. Earlier LB policies obtained it through the `BuildOptions`. Now the LB policies need to call the `MetricsRecorder` method on the ClientConn.

## Why
When LB policies manage child policies, they need to correctly propagate the BuildOptions. If they fail to pass down the `MetricsRecorder` in the `BuildOptions`, it will read to nil pointer deferences when the child tries to record metrics. There is no way for gRPC to enforce LB polices propagate the MetricsRecorder.  https://github.com/grpc/grpc-go/pull/8026 is open to enforce LB policies embed an existing `ClientConn` object. Using a method on `ClientConn` ensures that the `MetricsRecorder` can be obtained from the channel, unless explicitly intercepted by an LB policy. The `MetricsRecorder` field in the `BuildOptions` was added in the past 6 months, so we don't expect too many breakages in external code.

RELEASE NOTES:
* balancer: Implementors of custom LB policies that record metrics must use the new `MetricsRecorder` method on `Balancer.ClientConn` instead of the removed `Balancer.BuildOptions.MetricsRecorder` field to obtain a metrics recorder.


balancer/balancer.go
```diff
@@ -167,6 +167,16 @@ type ClientConn interface {
...
+
+	// MetricsRecorder provides the metrics recorder that balancers can use to
+	// record metrics. Balancer implementations which do not register metrics on
+	// metrics registry and record on them can ignore this method.
+	MetricsRecorder() estats.MetricsRecorder
+
+	// EnforceClientConnEmbedding is included to force implementers to embed
+	// another implementation of this interface, allowing gRPC to add methods
+	// without breaking users.
+	internal.EnforceClientConnEmbedding
```
dfawley, line 179: Remove this for now since we know it will cause problems?

balancer/balancer.go
```diff
@@ -167,6 +167,16 @@ type ClientConn interface {
 	//
 	// Deprecated: Use the Target field in the BuildOptions instead.
 	Target() string
+
+	// MetricsRecorder provides the metrics recorder that balancers can use to
+	// record metrics. Balancer implementations which do not register metrics on
+	// metrics registry and record on them can ignore this method.
```
dfawley, line 173: How about also adding: `Guaranteed to never be nil` or `Must always be safe to call` or something?

balancer/balancer.go
```diff
@@ -327,6 +333,13 @@ type Picker interface {
 // UpdateClientConnState, ResolverError, UpdateSubConnState, and Close are
 // guaranteed to be called synchronously from the same goroutine.  There's no
 // guarantee on picker.Pick, it may be called anytime.
+//
+// NOTICE: This interface is intended to be implemented by gRPC, or intercepted
```
dfawley, line 337: This was supposed to be on `ClientConn`, not `Balancer`, right?

internal/testutils/test_metrics_recorder.go
```diff
@@ -16,8 +16,7 @@
  *
  */
 
-// Package stats implements a TestMetricsRecorder utility.
-package stats
+package testutils
```
dfawley, line 19: Why move this?  IMO we should not build a gigantic testing package with thousands of test helpers - it's better to have individual functionality in separate packages to avoid conflicts and maintain cleaner separation.

Moving also adds a bunch of churn which I'd rather avoid unless there's a good reason for it.

internal/internal.go
```diff
@@ -277,3 +277,9 @@ const RLSLoadBalancingPolicyName = "rls_experimental"
 type EnforceSubConnEmbedding interface {
 	enforceSubConnEmbedding()
 }
+
+// EnforceClientConnEmbedding is used to enforce proper ClientConn implementation
+// embedding.
+type EnforceClientConnEmbedding interface {
+	enforceSubConnEmbedding()
```
dfawley, line 284: s/SubConn/ClientConn/

balancer/balancer.go
arjan-bal: Added a line.

balancer/balancer.go
arjan-bal: Removed.

balancer/balancer.go
arjan-bal: Yes, I pasted it at the wrong place. Removed from this PR, fixing it in https://github.com/grpc/grpc-go/pull/8026.

internal/internal.go
arjan-bal: Fixing it in https://github.com/grpc/grpc-go/pull/8026.

internal/testutils/test_metrics_recorder.go
arjan-bal: I've mentioned this in the PR description.  It was done because TestMetricsRecorder uses `testutils.Channel`. When I make `testutils.TestClientConnn` return `NoOpMetricsRecorder` from `testutils/stats/`, it causes a circular dependency b/w `testutils` and `testutils/stats`. 

https://github.com/grpc/grpc-go/blob/2fd426d0919db7e9e7920e365dd30674c4c9fe90/internal/testutils/stats/test_metrics_recorder.go#L38-L42

Another option I can think of is to pull balancer into a separate package.

dfawley: Sorry.

Another option would be to just not use `NoopMetricsRecorder` (anywhere, really, but you don't need to delete & change other usages).  An empty [`MetricsRecorderList`](https://pkg.go.dev/google.golang.org/grpc/internal/stats#MetricsRecorderList) has the same behavior.

arjan-bal: `TestClientConnn` now returns an empty `MetricsRecorderList`. Moved `test_metrics_recorder.go` back to `testutils/stats/` and reverted associated changed.

arjan-bal:
Merged a PR to avoid breakages in grpc-gcp-go: https://github.com/GoogleCloudPlatform/grpc-gcp-go/pull/91
Waiting on release of https://github.com/GoogleCloudPlatform/grpc-gcp-go.

dfawley (approved):

PR #8031, opened 2025-01-23
endpointsharding: Allow children to remain idle if configured

arjan-bal:
This change allows the children of endpointsharding to remain in IDLE state. This allows lazy initialization of children and therefore lazy creation of subchannels for endpoints. This enables the usage of endpointsharding in the ringhash balancer for managing pickfirst children.

RELEASE NOTES:
* balancer/endpointsharding: Balancers created using the new `NewBalancerWithoutAutoReconnect` function will not attempt to call `ExitIdle` automatically on their children when the children report idle.

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -64,12 +64,13 @@ type ChildState struct {
 	State    balancer.State
 
 	// Balancer exposes only the ExitIdler interface of the child LB policy.
-	// Other methods on the child policy are called only by endpointsharding.
+	// Other methods of the child policy are called only by endpointsharding.
 	Balancer balancer.ExitIdler
 }
 
 // NewBalancer returns a load balancing policy that manages homogeneous child
-// policies each owning a single endpoint.
+// policies each owning a single endpoint. The balancer will automatically call
+// ExitIdle on its children if they report IDLE connectivity state.
 func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer {
 	return newBlanacer(cc, opts, true)
 }
@@ -175,7 +176,6 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 		bal := child.(*balancerWrapper)
 		if _, ok := newChildren.Get(e); !ok {
 			bal.Close()
-			bal.isClosed = true
 		}
 	}
 	es.children.Store(newChildren)
@@ -214,7 +214,6 @@ func (es *endpointSharding) Close() {
 	for _, child := range children.Values() {
 		bal := child.(*balancerWrapper)
 		bal.Close()
-		bal.isClosed = true
 	}
 }
 
@@ -313,13 +312,9 @@ func ChildStatesFromPicker(picker balancer.Picker) []ChildState {
 type balancerWrapper struct {
 	balancer.Balancer   // Simply forward balancer.Balancer operations.
 	balancer.ClientConn // embed to intercept UpdateState, doesn't deal with SubConns
-
-	es *endpointSharding
-
-	childState ChildState
-
-	// Users must hold `es.childMu`.
-	isClosed bool
+	es                  *endpointSharding
+	childState          ChildState
+	isClosed            bool
 }
 
 func (bw *balancerWrapper) UpdateState(state balancer.State) {
@@ -330,6 +325,11 @@ func (bw *balancerWrapper) UpdateState(state balancer.State) {
 		bw.ExitIdle()
 	}
 	bw.es.updateState()
+}
+
+func (bw *balancerWrapper) Close() {
+	bw.Balancer.Close()
+	bw.isClosed = true
 }
 
 // ExitIdle pings an IDLE child balancer to exit idle in a new goroutine to
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -62,14 +62,31 @@ func init() {
 type ChildState struct {
 	Endpoint resolver.Endpoint
 	State    balancer.State
+
+	// Balancer exposes only the ExitIdler interface of the child LB policy.
+	// Other methods on the child policy are called only by endpointsharding.
+	Balancer balancer.ExitIdler
 }
 
 // NewBalancer returns a load balancing policy that manages homogeneous child
 // policies each owning a single endpoint.
 func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer {
+	return newBlanacer(cc, opts, true)
+}
+
+// NewBalancerWithoutAutoReconnect returns a load balancing policy that manages
+// homogeneous child policies each owning a single endpoint. The balancer will
+// allow children to remain in IDLE state until triggered to exit idle state
+// using the ChildState obtained using the endpointsharding picker.
+func NewBalancerWithoutAutoReconnect(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer {
+	return newBlanacer(cc, opts, false)
+}
+
+func newBlanacer(cc balancer.ClientConn, opts balancer.BuildOptions, autoReconnect bool) balancer.Balancer {
 	es := &endpointSharding{
-		cc:    cc,
-		bOpts: opts,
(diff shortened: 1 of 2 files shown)


balancer/endpointsharding/endpointsharding.go
```diff
@@ -62,14 +62,31 @@ func init() {
...
 	Endpoint resolver.Endpoint
 	State    balancer.State
+
+	// Balancer exposes only the ExitIdler interface of the child LB policy.
+	// Other methods on the child policy are called only by endpointsharding.
+	Balancer balancer.ExitIdler
 }
 
 // NewBalancer returns a load balancing policy that manages homogeneous child
 // policies each owning a single endpoint.
```
dfawley, line 72: Probably a good idea to indicate that this version automatically requests reconnection for any IDLE children.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -150,9 +172,10 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 	// Delete old children that are no longer present.
 	for _, e := range children.Keys() {
 		child, _ := children.Get(e)
-		bal := child.(balancer.Balancer)
+		bal := child.(*balancerWrapper)
 		if _, ok := newChildren.Get(e); !ok {
 			bal.Close()
+			bal.isClosed = true
```
dfawley, line 178: Can we move this into `balancerWrapper`'s `Close` method?

Also since calls into a `balancer.Balancer` must not happen concurrently, I don't think it requires a lock.

arjan-bal: Added a Close method and moved this into the Close method.

balancer/endpointsharding/endpointsharding.go
arjan-bal: Added a line in the doc comment.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -124,12 +142,17 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 		var bal *balancerWrapper
 		if child, ok := children.Get(endpoint); ok {
 			bal = child.(*balancerWrapper)
+			// Endpoint attributes may have changes, update the stored endpoint.
+			es.mu.Lock()
+			bal.childState.Endpoint = endpoint
+			es.mu.Unlock()
```
dfawley, line 148: Note that these updates are racy with respect to calls back into `UpdateState` and what is observed from `ChildStatesFromPicker`.  I'm not sure this is significant, since we will get an eventually-consistent `UpdateState` call after this, and also asynchronous calls out of the child should be pretty rare, since notifications into the child are always synchronous.  The `inhibit` atomic would also cause us to skip these updates most of the time, but if the call to `UpdateState` checks the flag before `UpdateCCS` sets it, then it is possible.

I'm also a little more worried about lock ordering.  But `es.childMu`->`es.mu` is already happening by virtue of calls into the child holding `childMu` and those resulting in synchronous calls into `cc` that takes `es.mu`.  So maybe the mutexes could use some documenting?

- `childMu` must be held on any call into any child.  (to prevent simultaneous calls into any single child)
- `mu` must not be held on any call into a child.  (since synchronous calls back from the child may require taking `mu`)
- `childMu` may not be taken if `mu` is held.  (the ordering implied by the above)

balancer/endpointsharding/endpointsharding_test.go
```diff
@@ -153,3 +157,71 @@ func (s) TestEndpointShardingBasic(t *testing.T) {
...
+	if err = roundrobin.CheckRoundRobinRPCs(ctx, client, []resolver.Address{{Addr: backend1.Address}, {Addr: backend3.Address}}); err != nil {
+		t.Fatalf("error in expected round robin: %v", err)
+	}
+
+	// On closing the first server, the first child balancer should enter
+	// IDLE. Since endpointsharding is configured not to auto-reconnect, it will
+	// remain IDLE and will not try to connect to the second backend in the same
+	// endpoint.
+	backend1.Stop()
+	if err = roundrobin.CheckRoundRobinRPCs(ctx, client, []resolver.Address{{Addr: backend3.Address}}); err != nil {
```
dfawley, line 224: I'm not sure this leaves enough time for the potential reconnect to be attempted and succeed?  I think we may need to do something like `time.Sleep(defaultTestShortTimeout)`.

I suppose you could do an experiment: replace `NewBalancerWithoutAutoReconnect` here with just `NewBalancer` and see how reliably the test fails.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -79,19 +97,26 @@ func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Ba
...
 	children atomic.Pointer[resolver.EndpointMap]
-	closed   bool
 
 	// inhibitChildUpdates is set during UpdateClientConnState/ResolverError
 	// calls (calls to children will each produce an update, only want one
 	// update).
 	inhibitChildUpdates atomic.Bool
 
-	mu sync.Mutex // Sync updateState callouts and childState recent state updates
+	// mu synchronizes access to the stored children balancer states.
```
dfawley, line 115:
```suggestion
	// mu synchronizes access to the stored children balancer states in children.
```

?

Or "in the children field" or something to be more specific.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -124,16 +149,21 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 		var bal *balancerWrapper
```
dfawley, line 149: Could you rename this local to `childBalancerWrapper` or something more clear -- at least `child` or something?  `bal` to me makes me think this is the `endpointsharding` balancer itself when I see it later on.  I guess we use `es` for that, but the more common convention is that the `Balancer` implementation within the package is simply named `balancer`, and `b` is the receiver.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -150,9 +180,9 @@ func (es *endpointSharding) UpdateClientConnState(state balancer.ClientConnState
 	// Delete old children that are no longer present.
 	for _, e := range children.Keys() {
 		child, _ := children.Get(e)
-		bal := child.(balancer.Balancer)
+		bal := child.(*balancerWrapper)
```
dfawley, line 183: Similar nit about the name of things..  This local isn't even necessary:

```go
child, _ := children.Get(e)
if _, ok := newChildren.Get(e); !ok {
	child.(*balancerWrapper).closeLocked()
}
```

balancer/endpointsharding/endpointsharding.go
```diff
@@ -290,28 +312,38 @@ func ChildStatesFromPicker(picker balancer.Picker) []ChildState {
...
+	bw.Balancer.Close()
+	bw.isClosed = true
+}
+
+// ExitIdle pings an IDLE child balancer to exit idle in a new goroutine to
+// avoid deadlocks due to synchronous balancer state updates.
+func (bw *balancerWrapper) ExitIdle() {
+	if ei, ok := bw.Balancer.(balancer.ExitIdler); ok {
 		go func() {
 			bw.es.childMu.Lock()
```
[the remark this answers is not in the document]
dfawley: > Before this PR, calls to balancerWrapper.Close() and updates to endpointsharding.closed were holding es.ChildMu:

Oh, I see... I missed where it grabbed the lock in `Close` and thought it was just delegating directly to the wrappers.

This now seems good and makes it explicit what is required.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -310,11 +317,20 @@ func ChildStatesFromPicker(picker balancer.Picker) []ChildState {
 // balancerWrapper is a wrapper of a balancer. It ID's a child balancer by
 // endpoint, and persists recent child balancer state.
 type balancerWrapper struct {
-	balancer.Balancer   // Simply forward balancer.Balancer operations.
+	// The following fields are initialized at build time and read-only after
+	// that and therefore do not need to be guarded by a mutex.
+
+	// child contains the wrapped balancer. Access it's methods only through
```
dfawley, line 323: *its

dfawley (approved):

Issue #8023, opened 2025-01-20
panic: Cannot free freed buffer


clarkmcc:
NOTE: if you are reporting is a potential security vulnerability or a crash,
please follow our CVE process at
https://github.com/grpc/proposal/blob/master/P4-grpc-cve-process.md instead of
filing an issue here.

Please see the FAQ in our main README.md, then answer the questions below
before submitting your issue.

### What version of gRPC are you using?
```
google.golang.org/grpc v1.69.2
```

### What version of Go are you using (`go version`)?
```
go version go1.23.5 darwin/arm64
```

### What operating system (Linux, Windows, …) and version?
darwin 15.1.1

### What did you do?
If possible, provide a recipe for reproducing the error.

It seems like it might be related to large byte slices being sent over the wire. In at least two cases, I traced the issue back to endpoints that were dealing with these image byte slices. The error is random and not reproducible on every request, it appears that in some cases, the request works fine, and in other cases it crashes the server.

```proto
message InferenceRequest{
  bytes image = 1;
}

message OCRExtractionRequest {
  enum Engine {
    Default = 0;
    Tesseract = 1;
    AppleVision = 2;
  }
  Engine engine = 1;
  bytes image = 10;
}
```

### What did you expect to see?
Not a panic

### What did you see instead?
```
panic: Cannot free freed buffer

goroutine 59 [running]:
google.golang.org/grpc/mem.(*buffer).Free(0x14000aa7a18?)
        /Users/clarkmccauley/go/pkg/mod/google.golang.org/grpc@v1.69.2/mem/buffers.go:164 +0x12c
google.golang.org/grpc/mem.BufferSlice.Free(...)
        /Users/clarkmccauley/go/pkg/mod/google.golang.org/grpc@v1.69.2/mem/buffer_slice.go:70
google.golang.org/grpc.(*clientStream).commitAttemptLocked(0x14000a3e6c0)
        /Users/clarkmccauley/go/pkg/mod/google.golang.org/grpc@v1.69.2/stream.go:620 +0x74
google.golang.org/grpc.(*clientStream).withRetry(0x14000a3e6c0, 0x14000aa79f0, 0x14000aa79e0)
        /Users/clarkmccauley/go/pkg/mod/google.golang.org/grpc@v1.69.2/stream.go:800 +0x3b0
google.golang.org/grpc.(*clientStream).RecvMsg(0x14000a3e6c0, {0x1043286a0?, 0x14000053e50?})
[...]

PR #8033, opened 2025-01-23
grpc: Fix encoded message size reported in error message

arjan-bal:
This PR fixes two bugs in the encoded message size error message:
1. Using len(b) returns the number of buffers in a BufferSlice and not the number of bytes in the BufferSlice.
2. Finding the size after calling b.Free() would return 0 if there are no remaining references to the buffers within the slice.

RELEASE NOTES:
* grpc: Fix the number of bytes reported in the error message when encoded messages are larger than `math.MaxUint32`.

### DIFF
--- a/rpc_util.go
+++ b/rpc_util.go
@@ -692,9 +692,9 @@ func encode(c baseCodec, msg any) (mem.BufferSlice, error) {
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "grpc: error while marshaling: %v", err.Error())
 	}
-	if uint(b.Len()) > math.MaxUint32 {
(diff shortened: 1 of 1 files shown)


dfawley (approved):
Is there an open issue that this fixes or did you just happen to notice this?

arjan-bal:
> Is there an open issue that this fixes or did you just happen to notice this?

For debugging https://github.com/grpc/grpc-go/issues/8023, I changed the type of BufferSlice from a slice to a struct. I was storing stack traces along with the BufSlice objects. This broke the build and I spotted this issue when fixing all the places that relied on BufSlice being a slice.

Issue #7983, opened 2025-01-07
received context error while waiting for new LB policy update: context deadline exceeded


mayurkale22:
We're seeing an intermittent issue. This always happens randomly at app startup time, that prevents app from properly starting up

```
rpc error: code = DeadlineExceeded desc = received context error while waiting for new LB policy update: context deadline exceeded
```

how should we interpret this error? Does it signal an issue with connectivity, gRPC server/client configuration or something else entirely. Appreciate any feedback on this.

eshitachandwani:
Hey @mayurkale22 , only this error message doesn't provide much information about what could be causing it. The problem could be caused by a number of issues starting from slow network to some error in the application or issue with name resolution. To help identify the cause, you can enable debug logs using 
```
$ export GRPC_GO_LOG_VERBOSITY_LEVEL=99
$ export GRPC_GO_LOG_SEVERITY_LEVEL=info
```
and that will give us more idea about the root cause.

itzmanish:
Hi @mayurkale22 are you using DNS resolver? 

I am also getting the same error on the latest GRPC client. 
grpc-go v1.67.3

I can see the resolver resolving 3 endpoints that point to my NLB and are correct. The only difference I can see is that it's using LB policy to "pick_first" instead of round-robin.

itzmanish:
I have more information on this issue. I got it fixed for my usecase ***SOMEHOW***, not sure if this really fixed it or just mitigated it for the time being. 

In my setup I had a DNS endpoint resolving to multiple A address of NLB and my client was using `pick_first` loadbalancer for some reason (maybe it's default loadbalancer). I changed the loadbalancer to `round_robin` and I don't find the context deadline issue anymore. My guess is for some reason one of the NLB endpoint was taking longer or GRPC lib was not able update the resolver state for that endpoint. 

I am still poking around the library for the flow and how it creates a client connection. I will update here if I find anything else.

purnesh42H:
@mayurkale22 could you provide more information in following format https://github.com/grpc/grpc-go/issues/new?template=bug.md with debugging enabled? Also, specify if you are using a different load balancing policy or name resolver than the default one.

purnesh42H:
As per your other question of how to interpret `rpc error: code = DeadlineExceeded desc = received context error while waiting for new LB policy update: context deadline exceeded`

[...]

PR #8035, opened 2025-01-24
picker_wrapper: simplify picker error when timing out waiting for con…

purnesh42H:
Fixes: #7983

RELEASE NOTES: None

### DIFF
--- a/picker_wrapper.go
+++ b/picker_wrapper.go
@@ -123,7 +123,7 @@ func (pw *pickerWrapper) pick(ctx context.Context, failfast bool, info balancer.
 				if lastPickErr != nil {
 					errStr = "latest balancer error: " + lastPickErr.Error()
 				} else {
-					errStr = fmt.Sprintf("received context error due to timing out waiting for connection: %s", ctx.Err().Error())
+					errStr = fmt.Sprintf("%v while waiting for connections to become ready", ctx.Err())
 				}
 				switch ctx.Err() {
 				case context.DeadlineExceeded:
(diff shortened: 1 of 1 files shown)


picker_wrapper.go
```diff
@@ -123,7 +123,7 @@ func (pw *pickerWrapper) pick(ctx context.Context, failfast bool, info balancer.
 				if lastPickErr != nil {
 					errStr = "latest balancer error: " + lastPickErr.Error()
```
purnesh42H, line 124: @dfawley should we change this as well? or is it fine to mention balancer here?

dfawley: I think this is probably fine.  The problem with below is not that it's mentioning LB policies, but that it's unclear to most users what might trigger an LB policy update to occur.

picker_wrapper.go
```diff
@@ -123,7 +123,7 @@ func (pw *pickerWrapper) pick(ctx context.Context, failfast bool, info balancer.
...
 				} else {
-					errStr = fmt.Sprintf("received context error while waiting for new LB policy update: %s", ctx.Err().Error())
+					errStr = fmt.Sprintf("received context error due to timing out waiting for connection: %s", ctx.Err().Error())
```
dfawley, line 126: How about:

`fmt.Sprinft("%v while waiting for connections to become ready", ctx.Err())`

This would result in either:

`rpc error: code = DeadlineExceeded desc = context deadline exceeded while waiting for connections to become ready`

OR

`rpc error: code = Canceled desc = context canceled while waiting for connections to become ready`


For reference we used to use `ctx.Err().Error()` directly, which resulted in, e.g.:

`rpc error: code = DeadlineExceeded desc = context deadline exceeded`

purnesh42H: Got it. Updated

dfawley (approved):

PR #8051, opened 2025-01-29
interop: Add empty_unary_error, rpc_soak_error test cases

nimf:
This PR adds two new test cases (`empty_unary_error` and `rpc_soak_error`) to the interop client.
The test cases verify that RPCs end up with status code equal to the `expected_error` provided via flag and the error message contains what is provided via `expecter_error_message_contains` flag.

We need these these cases to verify that the client reacts as expected to the configured failure injection in our test environment. I.e., we configure a failure injection so that RPCs are expected to fail with a specific code and message.

RELEASE NOTES: None

### DIFF
--- a/interop/client/client.go
+++ b/interop/client/client.go
@@ -37,6 +37,7 @@ import (
 
 	"golang.org/x/oauth2"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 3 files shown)


nimf:
@dfawley does the test failure look like a flake? PTAL at the PR.

nimf:
sorry, no need to look at it right now. We may use a different approach that doesn't require that.

dfawley:
No problem - please reopen if it's something that we need to review.

PR #8052, opened 2025-01-30
endpointsharding, lazy: Remove intermediary gracefulswitch balancers

arjan-bal:
Presently both `lazy` and `endpointsharding` create `gracefulswtich` balancers to manage their children. The `gracefulswitch` balancer is unnecessary as the type of the child balancer in not expected to change in existing use cases.  Having the `gracefulswitch` child requires the child balancer's builder to be registered in the global LB registry. The `lazy` LB is not a complete LB policy and should not be registered globally.

This PR makes the constructors of `endpointpointsharding` and `lazy` accept the child balancer's builder to avoid using the global registry. This PR also removes the `ParseConfig` functions/methods in `lazy` and `endpointsharding` as they will simply forward the `LoadBalancingConfig`s to their children. The children are responsible to parse the `LoadBalancingConfig` JSON. Since `pickfirst` can use a `nil` `LoadBalancingConfig`, the existing users of `endpointsharding` don't need to provide a `LoadBalancingConfig`.

RELEASE NOTES:
* balancer/endpointsharding: The constructor accepts the child balancer's builder and a struct with optional configuration.
* balancer/endpointsharding: Balancers created with the new `DisableAutoReconnect` option will not attempt to call `ExitIdle` automatically on their children when the children report idle.

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -59,14 +59,14 @@ type Options struct {
 	DisableAutoReconnect bool
 }
 
-// childBuilderFunc creates a new balancer with the ClientConn. It has the same
+// ChildBuilderFunc creates a new balancer with the ClientConn. It has the same
 // type as the balancer.Builder.Build method.
-type childBuilderFunc = func(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer
+type ChildBuilderFunc func(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer
 
 // NewBalancer returns a load balancing policy that manages homogeneous child
 // policies each owning a single endpoint. The endpointsharding balancer
 // forwards the LoadBalancingConfig in ClientConn state updates to its children.
-func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions, childBuilder childBuilderFunc, esOpts Options) balancer.Balancer {
+func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions, childBuilder ChildBuilderFunc, esOpts Options) balancer.Balancer {
 	es := &endpointSharding{
 		cc:           cc,
 		bOpts:        opts,
@@ -84,7 +84,7 @@ type endpointSharding struct {
 	cc           balancer.ClientConn
 	bOpts        balancer.BuildOptions
 	esOpts       Options
-	childBuilder childBuilderFunc
+	childBuilder ChildBuilderFunc
 
 	// childMu synchronizes calls to any single child. It must be held for all
 	// calls into a child. To avoid deadlocks, do not acquire childMu while
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -59,10 +59,14 @@ type Options struct {
 	DisableAutoReconnect bool
 }
 
+// childBuilderFunc creates a new balancer with the ClientConn. It has the same
+// type as the balancer.Builder.Build method.
+type childBuilderFunc = func(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer
+
 // NewBalancer returns a load balancing policy that manages homogeneous child
 // policies each owning a single endpoint. The endpointsharding balancer
 // forwards the LoadBalancingConfig in ClientConn state updates to its children.
-func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions, childBuilder balancer.Builder, esOpts Options) balancer.Balancer {
+func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions, childBuilder childBuilderFunc, esOpts Options) balancer.Balancer {
 	es := &endpointSharding{
 		cc:           cc,
 		bOpts:        opts,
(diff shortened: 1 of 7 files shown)


examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go
```diff
@@ -100,7 +99,7 @@ func (crr *customRoundRobin) UpdateClientConnState(state balancer.ClientConnStat
 	// is guaranteed to happen since the aggregator will always call
 	// UpdateChildState in its UpdateClientConnState.
 	return crr.Balancer.UpdateClientConnState(balancer.ClientConnState{
-		BalancerConfig: gracefulSwitchPickFirst,
+		BalancerConfig: nil, // pickfirst can handle nil configs.
```
dfawley, line 102: Delete the field entirely to remove complications from example?

balancer/weightedroundrobin/balancer.go
```diff
@@ -236,7 +233,7 @@ func (b *wrrBalancer) UpdateClientConnState(ccs balancer.ClientConnState) error
 	// This causes child to update picker inline and will thus cause inline
 	// picker update.
 	return b.child.UpdateClientConnState(balancer.ClientConnState{
-		BalancerConfig: endpointShardingLBConfig,
+		BalancerConfig: nil, // pickfirst can handle nil configs.
```
dfawley, line 236: And then I'd delete it from all our code to avoid clutter.  Most LB policies should handle not having configs, except some special xds things.

balancer/endpointsharding/endpointsharding.go
```diff
@@ -68,26 +48,26 @@ type ChildState struct {
...
+	// the endpointsharding balancer will automatically call ExitIdle on child
+	// connections that report IDLE.
+	DisableAutoReconnect bool
 }
 
-func newBlanacer(cc balancer.ClientConn, opts balancer.BuildOptions, autoReconnect bool) balancer.Balancer {
+// NewBalancer returns a load balancing policy that manages homogeneous child
+// policies each owning a single endpoint. The endpointsharding balancer
+// forwards the LoadBalancingConfig in ClientConn state updates to its children.
+func NewBalancer(cc balancer.ClientConn, opts balancer.BuildOptions, childBuilder balancer.Builder, esOpts Options) balancer.Balancer {
```
dfawley, line 65: I'm wondering about the signature here.

There's a lot of options:

1. Keep this & make anyone wanting to use es+lazy wrap lazy in a Builder implementation.
2. Keep this & implement a Builder in `lazy`, still.
3. Create `type ChildBuilder interface { Build(..) .. }` without `Name()` and make `lazy` implement that.
4. Accept a function matching the signature of `Build` here.

None of these are great.  I think I might like (4) the best for usability?  But I dislike it for understandability.  I suppose that could be helped with:

```go
type ChildBuilder = func(...) ... // matching Build
```

?

balancer/lazy/lazy.go
```diff
@@ -26,44 +26,31 @@
...
-	// Name is the name of the lazy balancer.
-	Name      = "lazy"
 	logPrefix = "[lazy-lb %p] "
 )
 
-type builder struct{}
-
-func (builder) Build(cc balancer.ClientConn, bOpts balancer.BuildOptions) balancer.Balancer {
+// NewBalancer is the constructor for the lazy balancer.
+func NewBalancer(cc balancer.ClientConn, bOpts balancer.BuildOptions, childBuilder balancer.Builder) balancer.Balancer {
```
dfawley, line 49: The same concerns as above apply here, too.

dfawley (commented):
Mostly looks fine, but the minor API thing should be decided.

balancer/lazy/lazy.go
```diff
@@ -45,8 +45,12 @@ const (
 	logPrefix = "[lazy-lb %p] "
 )
 
+// childBuilderFunc creates a new balancer with the ClientConn. It has the same
+// type as the balancer.Builder.Build method.
+type childBuilderFunc = func(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer
```
dfawley, line 50: This really should be exported, otherwise you expose unexported symbols.  The user would have to look at the source code to learn how to use it.

Also I'm completely unclear what the difference is in making this a type alias vs. just a type definition (w/ vs. w/o the `=`).

dfawley (approved):

balancer/lazy/lazy.go
arjan-bal: A type alias provides a new name for an existing type without introducing a new type that has a different identity. So the unexported `childBuilderFunc` isn't it's own type and doesn't need to be exposed to package users. The language server expands the `childBuilderFunc` type to  `func(cc balancer.ClientConn, opts balancer.BuildOptions) balancer.Balancer`.

image

Using a type declaration introduces a new type. For a type declaration, it makes sense to export the type because external users need to create variables of  the type to call this function. The language server also mentions the type now.
image

arjan-bal: Changing to a type declaration and exporting the type. The benefit is the shorter function signature and the godoc of the type being visible.

balancer/endpointsharding/endpointsharding.go
arjan-bal: Updated to use a type declaration instead of an alias.

PR #8060, opened 2025-01-31
interop: Fix logging and totalIterations issues in soak_tests.go

zbilun:
Two issues handled here:
- Updated the logging print approach to fail the test rather than only print the information when soak test total failures exceed max failures threshold.
- Renamed the successfully run soak iterations and updated the total soak iteration calculations.

@dfawley 

RELEASE NOTES: n/a

### DIFF
--- a/interop/soak_tests.go
+++ b/interop/soak_tests.go
@@ -190,7 +190,7 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
 		soakConfig.ServerAddr, totalSuccesses, soakConfig.Iterations, totalFailures, b.String())
 
 	if totalIterations != soakConfig.Iterations {
-		fmt.Fprintf(os.Stderr, "Soak test consumed all %v of time and quit early, ran %d out of %d iterations.\n", soakConfig.OverallTimeout, totalSuccesses, soakConfig.Iterations)
+		logger.Fatalf("Soak test consumed all %v of time and quit early, ran %d out of %d iterations.\n", soakConfig.OverallTimeout, totalSuccesses, soakConfig.Iterations)
 	}
 
 	if totalFailures > soakConfig.MaxFailures {
--- a/interop/soak_tests.go
+++ b/interop/soak_tests.go
@@ -183,12 +183,13 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
 		}
 	}
 	var b bytes.Buffer
+	totalIterations := totalSuccesses + totalFailures
 	latencies.Print(&b)
 	fmt.Fprintf(os.Stderr,
 		"(server_uri: %s) soak test ran: %d / %d iterations. Total failures: %d. Latencies in milliseconds: %s\n",
 		soakConfig.ServerAddr, totalSuccesses, soakConfig.Iterations, totalFailures, b.String())
 
-	if totalSuccesses+totalFailures != soakConfig.Iterations {
+	if totalIterations != soakConfig.Iterations {
 		fmt.Fprintf(os.Stderr, "Soak test consumed all %v of time and quit early, ran %d out of %d iterations.\n", soakConfig.OverallTimeout, totalSuccesses, soakConfig.Iterations)
 	}
 
--- a/interop/soak_tests.go
+++ b/interop/soak_tests.go
@@ -166,7 +166,7 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
 	wg.Wait()
 
 	//Handle results.
-	totalIterations := 0
+	totalSuccesses := 0
 	totalFailures := 0
 	latencies := stats.NewHistogram(stats.HistogramOptions{
 		NumBuckets:     20,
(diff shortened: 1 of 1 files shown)


interop/soak_tests.go
```diff
@@ -175,25 +175,26 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
 		MinValue:       0,
 	})
 	for _, worker := range soakWorkerResults {
-		totalIterations += worker.IterationsDone
+		totalSuccesses += worker.IterationsDone
```
dfawley, line 178: Should we rename the `IterationsDone` field, if it's only counting successful iterations, to make this kind of bug more obvious?

interop/soak_tests.go
```diff
@@ -175,25 +175,26 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
...
 		totalFailures += worker.Failures
 		if worker.Latencies != nil {
 			// Add latencies from the worker's Histogram to the main latencies.
 			latencies.Merge(worker.Latencies)
 		}
 	}
 	var b bytes.Buffer
+	totalIterations := totalSuccesses + totalFailures
```
dfawley, line 186: What's the difference between this and `soakConfig.Iterations`?  Because you print the latter just below instead of `totalIterations`.  I am guessing `totalIterations == soakConfig.Iterations * soakConfig.NumWorkers`, unless it ended early?

interop/soak_tests.go
```diff
@@ -175,25 +175,26 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
...
 	latencies.Print(&b)
 	fmt.Fprintf(os.Stderr,
 		"(server_uri: %s) soak test ran: %d / %d iterations. Total failures: %d. Latencies in milliseconds: %s\n",
```
dfawley, line 189: Is this string standardized?

I think it's a bit hard to understand what the first two numbers represent without looking at the code.  IMO it should say: `"soak test successes" instead of "soak test ran".  But if this is standardized then OK.

interop/soak_tests.go
```diff
@@ -175,25 +175,26 @@ func DoSoakTest(ctx context.Context, soakConfig SoakTestConfig) {
...
-		soakConfig.ServerAddr, totalIterations, soakConfig.Iterations, totalFailures, b.String())
+		soakConfig.ServerAddr, totalSuccesses, soakConfig.Iterations, totalFailures, b.String())
 
 	if totalIterations != soakConfig.Iterations {
-		fmt.Fprintf(os.Stderr, "Soak test consumed all %v of time and quit early, ran %d out of %d iterations.\n", soakConfig.OverallTimeout, totalIterations, soakConfig.Iterations)
+		logger.Fatalf("Soak test consumed all %v of time and quit early, ran %d out of %d iterations.\n", soakConfig.OverallTimeout, totalSuccesses, soakConfig.Iterations)
```
dfawley, line 193: This looks like it was meaning to print how many iterations actually occurred vs. how many were expected.  So now it's different.

interop/soak_tests.go
zbilun: `totalIterations` represents the total number of completed iterations, which includes both successful and failed runs.
`soakConfig.Iterations` is the parameter defined at the start, indicating the expected number of iterations to be completed within the required time range.

Under ideal conditions, we expect `totalIterations` to equal `soakConfig.Iterations`. However, if this isn't the case, we check for a mismatch. If totalIterations is not equal to soakConfig.Iterations, we log an error:

```
if totalIterations != soakConfig.Iterations {
    logger.Fatalf("Soak test ran out of time (overall timeout: %v) and quit early. Completed %d out of %d iterations.\n", soakConfig.OverallTimeout, totalSuccesses, soakConfig.Iterations)
}
```

interop/soak_tests.go
zbilun: It is not standardized. I have updated it based on your suggestion.

interop/soak_tests.go
zbilun: Yes, we want the test to fail if totalIterations != soakConfig.Iterations.

dfawley: But you changed it to print the number of _successes_, not the number of iterations.  So I think this is wrong now.  Or the text should be changed.  But if you don't print `totalIterations` we can't see why the `if` just above failed.

zbilun: I will keep testing locally to make sure all the soak tests are stable before merging this PR. I will provide updates afterwards. Thanks!

dfawley (approved):

interop/soak_tests.go
dfawley: OK, let me know when you want us to merge this PR, or if there are more changes that need reviewing.

zbilun: I've made two minor changes (see commits from Feb 7, 2025), and this PR is now ready to be merged. The longer latency in each soak iteration compared to Java will be addressed in a separate PR, if necessary. Thanks!

PR #8066, opened 2025-02-05
xds: introduce simple grpc transport for generic xds clients

purnesh42H:
This PR is adding a simple grpc-based transport implementation of generic xds clients transport builder which can be provided as transport builder in xDS and LRS client. 

The `build()` method currently creates a new grpc channel every time. However, in future we will incorporate reference count map for existing transports and deduplicate transports based on provided ServerConfig so that transport channel to same server can be shared among xDS and LRS client.

[POC](https://github.com/purnesh42H/grpc-go/pull/8)
[Internal Design](https://docs.google.com/document/d/11RtR8CDgBzhSHImgJda1AJxw6uUUW-2LA8HX0UPQmyE)

RELEASE NOTES: None

### DIFF
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -48,17 +48,17 @@ type Builder struct{}
 // The Extension field of the ServerConfig must be a ServerConfigExtension.
 func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	if sc.ServerURI == "" {
-		return nil, fmt.Errorf("ServerConfig's ServerURI field cannot be empty")
+		return nil, fmt.Errorf("grpctransport: ServerURI is not set in ServerConfig")
 	}
 	if sc.Extensions == nil {
-		return nil, fmt.Errorf("ServerConfig's Extensions field cannot be nil for gRPC transport")
+		return nil, fmt.Errorf("grpctransport: Extensions is not set in ServerConfig")
 	}
 	sce, ok := sc.Extensions.(ServerConfigExtension)
 	if !ok {
-		return nil, fmt.Errorf("ServerConfig Extensions field is %T, but must be %T", sc.Extensions, ServerConfigExtension{})
+		return nil, fmt.Errorf("grpctransport: Extensions field is %T, but must be %T in ServerConfig", sc.Extensions, ServerConfigExtension{})
 	}
 	if sce.Credentials == nil {
-		return nil, fmt.Errorf("ServerConfigExtensions's Credentials field cannot be nil for gRPC transport")
+		return nil, fmt.Errorf("grptransport: Credentials field is not set in ServerConfigExtension")
 	}
 
 	// TODO: Incorporate reference count map for existing transports and
@@ -76,7 +76,7 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	})
 	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(sce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
 	if err != nil {
-		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
+		return nil, fmt.Errorf("grpctransport: failed to create transport to server %q: %v", sc.ServerURI, err)
 	}
 
 	return &grpcTransport{cc: cc}, nil
@@ -125,7 +125,7 @@ func (c *byteCodec) Marshal(v any) ([]byte, error) {
 	if b, ok := v.([]byte); ok {
 		return b, nil
 	}
-	return nil, fmt.Errorf("message must be a byte slice")
+	return nil, fmt.Errorf("message is %T, but must be a byte slice ", v)
 }
 
 func (c *byteCodec) Unmarshal(data []byte, v any) error {
@@ -133,9 +133,9 @@ func (c *byteCodec) Unmarshal(data []byte, v any) error {
 		*b = data
 		return nil
 	}
-	return fmt.Errorf("target must be a pointer to a byte slice")
+	return fmt.Errorf("target is %T, but must be a pointer to a byte slice", v)
 }
 
 func (c *byteCodec) Name() string {
-	return "grpc.xds.internal.clients.grpctransport.byte_codec"
+	return "grpctransport.byteCodec"
 }
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -77,7 +77,6 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
 	}
-	cc.Connect()
 
 	return &grpcTransport{cc: cc}, nil
 }
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -40,7 +40,7 @@ type ServerConfigExtension struct {
 }
 
 // Builder creates gRPC-based Transports. It must be paired with ServerConfigs
-// that contain Extension field of type ServerConfigExtension.
+// that contain an Extension field of type ServerConfigExtension.
 type Builder struct{}
 
 // Build returns a gRPC-based clients.Transport.
@@ -88,11 +88,16 @@ type grpcTransport struct {
 
 // NewStream creates a new gRPC stream to the server for the specified method.
 func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.Stream, error) {
-	s, err := g.cc.NewStream(ctx, &grpc.StreamDesc{StreamName: method, ClientStreams: true, ServerStreams: true}, method)
+	s, err := g.cc.NewStream(ctx, &grpc.StreamDesc{ClientStreams: true, ServerStreams: true}, method)
 	if err != nil {
 		return nil, err
 	}
 	return &stream{stream: s}, nil
+}
+
+// Close closes the gRPC channel to the server.
+func (g *grpcTransport) Close() error {
+	return g.cc.Close()
 }
 
 type stream struct {
@@ -109,14 +114,9 @@ func (s *stream) Recv() ([]byte, error) {
 	var typedRes []byte
 
 	if err := s.stream.RecvMsg(&typedRes); err != nil {
-		return typedRes, err
+		return nil, err
 	}
 	return typedRes, nil
-}
-
-// Close closes the gRPC channel to the server.
-func (g *grpcTransport) Close() error {
-	return g.cc.Close()
 }
 
 type byteCodec struct{}
@@ -125,7 +125,7 @@ func (c *byteCodec) Marshal(v any) ([]byte, error) {
 	if b, ok := v.([]byte); ok {
 		return b, nil
 	}
-	return nil, fmt.Errorf("message is %T, but must be a byte slice ", v)
+	return nil, fmt.Errorf("message is %T, but must be a []byte", v)
 }
 
 func (c *byteCodec) Unmarshal(data []byte, v any) error {
@@ -133,7 +133,7 @@ func (c *byteCodec) Unmarshal(data []byte, v any) error {
 		*b = data
 		return nil
 	}
-	return fmt.Errorf("target is %T, but must be a pointer to a byte slice", v)
+	return fmt.Errorf("target is %T, but must be *[]byte", v)
 }
 
 func (c *byteCodec) Name() string {
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -40,7 +40,7 @@ type ServerConfigExtension struct {
 }
 
 // Builder creates gRPC-based Transports. It must be paired with ServerConfigs
-// that contain its ServerConfigExtension.
+// that contain Extension field of type ServerConfigExtension.
 type Builder struct{}
 
 // Build returns a gRPC-based clients.Transport.
@@ -53,11 +53,11 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	if sc.Extensions == nil {
 		return nil, fmt.Errorf("ServerConfig's Extensions field cannot be nil for gRPC transport")
 	}
-	gtsce, ok := sc.Extensions.(ServerConfigExtension)
+	sce, ok := sc.Extensions.(ServerConfigExtension)
 	if !ok {
 		return nil, fmt.Errorf("ServerConfig Extensions field is %T, but must be %T", sc.Extensions, ServerConfigExtension{})
 	}
-	if gtsce.Credentials == nil {
+	if sce.Credentials == nil {
 		return nil, fmt.Errorf("ServerConfigExtensions's Credentials field cannot be nil for gRPC transport")
 	}
 
@@ -66,15 +66,15 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	// transport channel to same server can be shared between xDS and LRS
 	// client.
 
-	// Create a new gRPC client/channel for the xDS management server with the
-	// provided credentials, server URI, and a byte codec to send and receive
-	// messages. Aso set a static keepalive configuration that is common across
-	// gRPC language implementations.
+	// Create a new gRPC client/channel for the server with the provided
+	// credentials, server URI, and a byte codec to send and receive messages.
+	// Also set a static keepalive configuration that is common across gRPC
+	// language implementations.
 	kpCfg := grpc.WithKeepaliveParams(keepalive.ClientParameters{
 		Time:    5 * time.Minute,
 		Timeout: 20 * time.Second,
 	})
-	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
+	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(sce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
 	if err != nil {
 		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
 	}
@@ -107,14 +107,14 @@ func (s *stream) Send(msg []byte) error {
 // Recv receives a message from the server.
 func (s *stream) Recv() ([]byte, error) {
 	var typedRes []byte
-	err := s.stream.RecvMsg(&typedRes)
-	if err != nil {
+
+	if err := s.stream.RecvMsg(&typedRes); err != nil {
 		return typedRes, err
 	}
 	return typedRes, nil
 }
 
-// Close closes all the gRPC streams to the server.
+// Close closes the gRPC channel to the server.
 func (g *grpcTransport) Close() error {
 	return g.cc.Close()
 }
@@ -137,5 +137,5 @@ func (c *byteCodec) Unmarshal(data []byte, v any) error {
 }
 
 func (c *byteCodec) Name() string {
-	return "byteCodec"
+	return "grpc.xds.internal.clients.grpctransport.byte_codec"
 }
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -1,0 +1,142 @@
+/*
+ *
+ * Copyright 2025 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package grpctransport provides an implementation of the
+// clients.TransportBuilder interface using gRPC.
+package grpctransport
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/xds/internal/clients"
+)
+
+// ServerConfigExtension allows extending the clients.ServerConfig for the
+// gRPC-based transport builder. Any implementation of this must implement
+// the ServerConfig() method.
+//
+// This interface must be implemented by the grpc transport builder provided
+// in the Extensions field of the clients.ServerConfig.
+type ServerConfigExtension interface {
+	ServerConfig() *ServerConfig
+}
+
+// ServerConfig holds the settings for connecting to an xDS management server
+// using gRPC.
+type ServerConfig struct {
+	// Credentials is the credential bundle containing the gRPC credentials for
+	// connecting to the xDS management server.
+	Credentials credentials.Bundle
+}
+
+// ServerConfig returns the ServerConfig itself. This method is designed
+// to satisfy [ServerConfigExtension] interface requirement.
+func (s *ServerConfig) ServerConfig() *ServerConfig {
+	return s
+}
+
+// Builder provides a way to build a gRPC-based transport to an xDS management
+// server.
+type Builder struct{}
+
+// Build creates a new gRPC-based transport to an xDS management server using
+// the provided clients.ServerConfig. This involves creating a
+// grpc.ClientConn to the server using the provided credentials and server URI.
+//
+// If any of ServerURI or Extensions of `sc` are not present, Build() will return
+// an error.
+func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
+	if sc.ServerURI == "" {
+		return nil, fmt.Errorf("xds: ServerConfig's ServerURI field cannot be empty")
+	}
+	if sc.Extensions == nil {
+		return nil, fmt.Errorf("xds: ServerConfig's Extensions field cannot be nil for gRPC transport")
+	}
+	gtsce, ok := sc.Extensions.(ServerConfigExtension)
+	if !ok {
+		return nil, fmt.Errorf("xds: ServerConfig's Extensions field cannot be anything other than grpctransport.ServerConfigExtension for gRPC transport")
+	}
+	gtsc := gtsce.ServerConfig()
+	if gtsc.Credentials == nil {
+		return nil, fmt.Errorf("xsd: ServerConfigExtensions's Credentials field cannot be nil for gRPC transport")
+	}
+
+	// TODO: Incorporate reference count map for existing transports and
+	// deduplicate transports based on server URI and credentials so that
+	// transport channel to same server can be shared between xDS and LRS
+	// client.
+
+	// Dial the xDS management server with the provided credentials, server URI,
+	// and a static keepalive configuration that is common across gRPC language
+	// implementations.
+	kpCfg := grpc.WithKeepaliveParams(keepalive.ClientParameters{
+		Time:    5 * time.Minute,
+		Timeout: 20 * time.Second,
+	})
+	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsc.Credentials))
+	if err != nil {
+		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
+	}
+	cc.Connect()
+
+	return &grpcTransport{cc: cc}, nil
+}
+
+type grpcTransport struct {
+	cc *grpc.ClientConn
+}
+
+// NewStream creates a new gRPC stream to the xDS management server for the
+// specified method.  The returned Stream interface can be used to send and
+// receive messages on the stream.
+func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.Stream, error) {
+	s, err := g.cc.NewStream(ctx, &grpc.StreamDesc{StreamName: method, ClientStreams: true, ServerStreams: true}, method)
+	if err != nil {
+		return nil, err
+	}
+	return &stream{stream: s}, nil
+}
+
+type stream struct {
+	stream grpc.ClientStream
+}
+
+// Send sends a message to the xDS management server.
+func (s *stream) Send(msg []byte) error {
+	return s.stream.SendMsg(msg)
+}
+
+// Recv receives a message from the xDS management server.
+func (s *stream) Recv() ([]byte, error) {
+	var typedRes []byte
+	err := s.stream.RecvMsg(&typedRes)
+	if err != nil {
+		return typedRes, err
+	}
+	return typedRes, nil
+}
+
+// Close closes the gRPC stream to the xDS management server.
+func (g *grpcTransport) Close() error {
+	return g.cc.Close()
+}
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -32,10 +32,10 @@ import (
 )
 
 // ServerConfigExtension holds settings for connecting to a gRPC server,
-// such as an xDS or LRS server.
+// such as an xDS management or an LRS server.
 type ServerConfigExtension struct {
-	// Credentials will be used for all gRPC transports. If is unset, transport
-	// creation will fail.
+	// Credentials will be used for all gRPC transports. If it is unset,
+	// transport creation will fail.
 	Credentials credentials.Bundle
 }
 
@@ -62,13 +62,14 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	}
 
 	// TODO: Incorporate reference count map for existing transports and
-	// deduplicate transports based on server URI and credentials so that
+	// deduplicate transports based on the provided ServerConfig so that
 	// transport channel to same server can be shared between xDS and LRS
 	// client.
 
-	// Dial the xDS management server with the provided credentials, server URI,
-	// and a static keepalive configuration that is common across gRPC language
-	// implementations.
+	// Create a new gRPC client/channel for the xDS management server with the
+	// provided credentials, server URI, and a byte codec to send and receive
+	// messages. Aso set a static keepalive configuration that is common across
+	// gRPC language implementations.
 	kpCfg := grpc.WithKeepaliveParams(keepalive.ClientParameters{
 		Time:    5 * time.Minute,
 		Timeout: 20 * time.Second,
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -27,6 +27,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/encoding"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/xds/internal/clients"
 )
@@ -43,9 +44,15 @@ type ServerConfigExtension struct {
 // server.
 type Builder struct{}
 
+func init() {
+	encoding.RegisterCodec(&byteCodec{})
+}
+
 // Build creates a new gRPC-based transport to an xDS management server using
 // the provided clients.ServerConfig. This involves creating a
-// grpc.ClientConn to the server using the provided credentials and server URI.
+// grpc.ClientConn to the server using the provided credentials and server URI,
+// and byteCodec which is byte-based implementation of encoding.Codec to send
+// and receive messages as bytes on the stream.
 //
 // If any of ServerURI or Extensions of `sc` are not present, Build() will return
 // an error.
@@ -76,7 +83,7 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 		Time:    5 * time.Minute,
 		Timeout: 20 * time.Second,
 	})
-	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsce.Credentials))
+	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
 	if err != nil {
 		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
 	}
@@ -122,4 +129,25 @@ func (s *stream) Recv() ([]byte, error) {
 // Close closes the gRPC stream to the xDS management server.
 func (g *grpcTransport) Close() error {
 	return g.cc.Close()
+}
+
+type byteCodec struct{}
+
+func (c *byteCodec) Marshal(v any) ([]byte, error) {
+	if b, ok := v.([]byte); ok {
+		return b, nil
+	}
+	return nil, fmt.Errorf("message must be a byte slice")
+}
+
+func (c *byteCodec) Unmarshal(data []byte, v any) error {
+	if b, ok := v.(*[]byte); ok {
+		*b = data
+		return nil
+	}
+	return fmt.Errorf("target must be a pointer to a byte slice")
+}
+
+func (c *byteCodec) Name() string {
+	return "byteCodec"
 }
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -31,28 +31,12 @@ import (
 	"google.golang.org/grpc/xds/internal/clients"
 )
 
-// ServerConfigExtension allows extending the clients.ServerConfig for the
-// gRPC-based transport builder. Any implementation of this must implement
-// the ServerConfig() method.
-//
-// This interface must be implemented by the grpc transport builder provided
-// in the Extensions field of the clients.ServerConfig.
-type ServerConfigExtension interface {
-	ServerConfig() *ServerConfig
-}
-
-// ServerConfig holds the settings for connecting to an xDS management server
+// ServerConfigExtension holds the settings for connecting to an xDS management server
 // using gRPC.
-type ServerConfig struct {
+type ServerConfigExtension struct {
 	// Credentials is the credential bundle containing the gRPC credentials for
 	// connecting to the xDS management server.
 	Credentials credentials.Bundle
-}
-
-// ServerConfig returns the ServerConfig itself. This method is designed
-// to satisfy [ServerConfigExtension] interface requirement.
-func (s *ServerConfig) ServerConfig() *ServerConfig {
-	return s
 }
 
 // Builder provides a way to build a gRPC-based transport to an xDS management
@@ -76,8 +60,7 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	if !ok {
 		return nil, fmt.Errorf("xds: ServerConfig's Extensions field cannot be anything other than grpctransport.ServerConfigExtension for gRPC transport")
 	}
-	gtsc := gtsce.ServerConfig()
-	if gtsc.Credentials == nil {
+	if gtsce.Credentials == nil {
 		return nil, fmt.Errorf("xsd: ServerConfigExtensions's Credentials field cannot be nil for gRPC transport")
 	}
 
@@ -93,7 +76,7 @@ func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 		Time:    5 * time.Minute,
 		Timeout: 20 * time.Second,
 	})
-	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsc.Credentials))
+	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsce.Credentials))
 	if err != nil {
 		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
 	}
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -125,7 +125,7 @@ func (c *byteCodec) Marshal(v any) ([]byte, error) {
 	if b, ok := v.([]byte); ok {
 		return b, nil
 	}
-	return nil, fmt.Errorf("message is %T, but must be a []byte", v)
+	return nil, fmt.Errorf("grpctransport: message is %T, but must be a []byte", v)
 }
 
 func (c *byteCodec) Unmarshal(data []byte, v any) error {
@@ -133,7 +133,7 @@ func (c *byteCodec) Unmarshal(data []byte, v any) error {
 		*b = data
 		return nil
 	}
-	return fmt.Errorf("target is %T, but must be *[]byte", v)
+	return fmt.Errorf("grpctransport: target is %T, but must be *[]byte", v)
 }
 
 func (c *byteCodec) Name() string {
--- a/xds/internal/clients/grpctransport/grpc_transport.go
+++ b/xds/internal/clients/grpctransport/grpc_transport.go
@@ -27,48 +27,38 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
-	"google.golang.org/grpc/encoding"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/xds/internal/clients"
 )
 
-// ServerConfigExtension holds the settings for connecting to an xDS management server
-// using gRPC.
+// ServerConfigExtension holds settings for connecting to a gRPC server,
+// such as an xDS or LRS server.
 type ServerConfigExtension struct {
-	// Credentials is the credential bundle containing the gRPC credentials for
-	// connecting to the xDS management server.
+	// Credentials will be used for all gRPC transports. If is unset, transport
+	// creation will fail.
 	Credentials credentials.Bundle
 }
 
-// Builder provides a way to build a gRPC-based transport to an xDS management
-// server.
+// Builder creates gRPC-based Transports. It must be paired with ServerConfigs
+// that contain its ServerConfigExtension.
 type Builder struct{}
 
-func init() {
-	encoding.RegisterCodec(&byteCodec{})
-}
-
-// Build creates a new gRPC-based transport to an xDS management server using
-// the provided clients.ServerConfig. This involves creating a
-// grpc.ClientConn to the server using the provided credentials and server URI,
-// and byteCodec which is byte-based implementation of encoding.Codec to send
-// and receive messages as bytes on the stream.
+// Build returns a gRPC-based clients.Transport.
 //
-// If any of ServerURI or Extensions of `sc` are not present, Build() will return
-// an error.
+// The Extension field of the ServerConfig must be a ServerConfigExtension.
 func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
 	if sc.ServerURI == "" {
-		return nil, fmt.Errorf("xds: ServerConfig's ServerURI field cannot be empty")
+		return nil, fmt.Errorf("ServerConfig's ServerURI field cannot be empty")
 	}
 	if sc.Extensions == nil {
-		return nil, fmt.Errorf("xds: ServerConfig's Extensions field cannot be nil for gRPC transport")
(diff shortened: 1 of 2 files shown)


xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,142 @@
...
+	return &stream{stream: s}, nil
+}
+
+type stream struct {
+	stream grpc.ClientStream
+}
+
+// Send sends a message to the xDS management server.
+func (s *stream) Send(msg []byte) error {
+	return s.stream.SendMsg(msg)
```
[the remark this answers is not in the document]
dfawley: grpc.ClientStream works on `any`s.  So we will need a byte-based codec to make this work.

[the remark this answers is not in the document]
purnesh42H: yeah by binary codec i meant byte-based codec. So, we need similar implementation as proto for Codec (v1) interface https://github.com/grpc/grpc-go/blob/master/encoding/proto/proto.go#L43?

Where should that live? in xds/clients/grpctransport or under encoding?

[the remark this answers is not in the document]
dfawley: > Where should that live? in xds/clients/grpctransport

^^ Yes, this is what I was thinking.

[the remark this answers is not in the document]
purnesh42H: I have added the byte-based transport and wrote a simple test for Send. Let me know how it looks overall. Thanks.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+	// connecting to the xDS management server.
+	Credentials credentials.Bundle
+}
+
+// Builder provides a way to build a gRPC-based transport to an xDS management
+// server.
+type Builder struct{}
+
+func init() {
+	encoding.RegisterCodec(&byteCodec{})
```
dfawley, line 48: As a library doing this for internal reasons, I don't think it's appropriate to register something globally.

We can use `ForceCodec` on the calls instead.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Build creates a new gRPC-based transport to an xDS management server using
+// the provided clients.ServerConfig. This involves creating a
```
dfawley, line 52: ServerConfig is for any kind of server.  LRS is not necessarily the same as the xds management server.  We should be more careful about using this phrase in places where it doesn't apply.  Also in `ServerConfig` itself.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+// grpc.ClientConn to the server using the provided credentials and server URI,
+// and byteCodec which is byte-based implementation of encoding.Codec to send
+// and receive messages as bytes on the stream.
+//
+// If any of ServerURI or Extensions of `sc` are not present, Build() will return
+// an error.
```
dfawley, line 58:
```suggestion
// Build returns a gRPC-based clients.Transport.
//
// The Extension field of the ServerConfig must be a ServerConfigExtension.
```

No need to go into implementation details about codecs in docstrings.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/xds/internal/clients"
+)
+
+// ServerConfigExtension holds the settings for connecting to an xDS management server
+// using gRPC.
+type ServerConfigExtension struct {
+	// Credentials is the credential bundle containing the gRPC credentials for
+	// connecting to the xDS management server.
+	Credentials credentials.Bundle
```
dfawley, line 40:
```suggestion
	// Credentials will be used for all gRPC transports. If is unset, transport creation will fail.
	Credentials credentials.Bundle
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Builder provides a way to build a gRPC-based transport to an xDS management
+// server.
+type Builder struct{}
```
dfawley, line 45:
```suggestion
// Builder creates gRPC-based Transports.  It must be paired with ServerConfigs that contain its
// ServerConfigExtension.
type Builder struct{}
```

Or something like that.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+// the provided clients.ServerConfig. This involves creating a
+// grpc.ClientConn to the server using the provided credentials and server URI,
+// and byteCodec which is byte-based implementation of encoding.Codec to send
+// and receive messages as bytes on the stream.
+//
+// If any of ServerURI or Extensions of `sc` are not present, Build() will return
+// an error.
+func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
+	if sc.ServerURI == "" {
+		return nil, fmt.Errorf("xds: ServerConfig's ServerURI field cannot be empty")
```
dfawley, line 61: Please remove the "xds" prefix here.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	if sc.Extensions == nil {
+		return nil, fmt.Errorf("xds: ServerConfig's Extensions field cannot be nil for gRPC transport")
+	}
+	gtsce, ok := sc.Extensions.(ServerConfigExtension)
+	if !ok {
+		return nil, fmt.Errorf("xds: ServerConfig's Extensions field cannot be anything other than grpctransport.ServerConfigExtension for gRPC transport")
+	}
```
dfawley, line 69:
```suggestion
	if !ok {
		return nil, fmt.Errorf("ServerConfig Extensions field is %T, but must be %T", sc.Extensions, ServerConfigExtension{})
	}
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+	// client.
+
+	// Dial the xDS management server with the provided credentials, server URI,
+	// and a static keepalive configuration that is common across gRPC language
+	// implementations.
+	kpCfg := grpc.WithKeepaliveParams(keepalive.ClientParameters{
+		Time:    5 * time.Minute,
+		Timeout: 20 * time.Second,
+	})
+	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(gtsce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
```
dfawley, line 86: Oh you already are using `ForceCodec`...So then we just don't need to register it.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
+	}
+	cc.Connect()
```
dfawley, line 90: This shouldn't be necessary.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+	return &grpcTransport{cc: cc}, nil
+}
+
+type grpcTransport struct {
+	cc *grpc.ClientConn
+}
+
+// NewStream creates a new gRPC stream to the xDS management server for the
+// specified method.  The returned Stream interface can be used to send and
+// receive messages on the stream.
```
dfawley, line 101: Optional: this can be omitted since it's not in the docstrings and it's pretty clear what it does.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, err
+	}
+	return &stream{stream: s}, nil
+}
+
+type stream struct {
+	stream grpc.ClientStream
+}
+
+// Send sends a message to the xDS management server.
```
dfawley, line 114: Similar to above.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,198 @@
...
+// - Credentials are nil.
+// - Success cases.
+func (s) TestBuild(t *testing.T) {
+	tests := []struct {
+		name      string
+		serverCfg clients.ServerConfig
+		wantErr   bool
+	}{
+		{
+			name: "ServerURI_empty",
```
dfawley, line 64: Nit: `s/_/ /` throughout?  `t.Run` will convert them to underscores to make it a legal test name -- just make sure they're human-readable here.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	// Send a discovery request message on the stream.
+	req, err := proto.Marshal(&v3discoverypb.DiscoveryRequest{})
+	if err != nil {
+		t.Fatalf("failed to marshal DiscoveryRequest: %v", err)
+	}
+	if err := stream.Send(req); err != nil {
+		t.Fatalf("failed to send message: %v", err)
+	}
```
dfawley, line 197: Can you also make this do a `Recv`?  Codecov shows that's not covered.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+package grpctransport
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/internal/grpctest"
```
dfawley, line 27: We should avoid using anything from grpc/internal.  That includes grpctest and e2e.  We won't be able to move this out like this.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: yeah i forgot to remove that. Its not needed if we are using ForceCodec

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: I have removed the mention of xDS from all docstrings.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Removed. I though we want to prefix xds to all error messages for xds clients.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: yeah right

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Simplified and removed mention of xDS

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Simplified.

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: I have made it more like sentence. Was trying to follow go/go-style/decisions#subtest-names

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: yeah. I wanted to check on this that for Recv, we need a grpc enabled xDS server with byte-based handlers. Should we add an xDS server with byte-based handler or just have a test grpc server to test this? Eventually for e2e tests I think we do need a test xDS server because go-control-plane one proto based.

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: yeah intend it remove. Regarding grpctest, should we just remove subtests for now or later? Regarding e2e, as i mentioned I think we need to add grpc enabeld xds server.

xds/internal/clients/grpctransport/grpc_transport_test.go
dfawley: For this test, we could do literally any kind of grpc server.  It could be like the stub server, or whatever.  Byte-based is not relevant.  The server should ideally do proto as is the default, and we can serialize/deserialize proto messages into byte buffers on the client manually for the testing.

True, for e2e tests of xdsclient, we will need to use go-control-plane.  If we have testing helpers that make it easy to use, that's fine but, I'd prefer to copy them into the `clients` subtree and try to make sure we don't import things from the rest of the grpc tree.  Otherwise it will be really hard to separate when it's time to move it out.

purnesh42H: For this test. I have created a test server using the same byteCodec to receive and send bytes and written a single test for send and receive. Let me know if that's good enough.

For e2e tests I think we will need to have management server using go-control-plane but that can be separate PR

dfawley: > > The server should ideally do proto as is the default

> I have created a test server using the same byteCodec

Is there a reason you didn't follow the way I suggested?  A normal proto server would be less testing code (which is good) and better proof the byte codec is working as intended.

purnesh42H: Kept the test logic same and change the test server to implement AggregatedDiscoveryServiceServer and handle StreamAggregatedResources

xds/internal/clients/grpctransport/grpc_transport_test.go
easwars: This is something that I learnt recently as well, and though it feels a little weird in the beginning to have subtest names that have underscrores, when it comes time to run them individually, it makes a lot of sense as direct copy paste works.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+// ServerConfigExtension holds settings for connecting to a gRPC server,
+// such as an xDS management or an LRS server.
+type ServerConfigExtension struct {
+	// Credentials will be used for all gRPC transports. If it is unset,
+	// transport creation will fail.
+	Credentials credentials.Bundle
+}
+
+// Builder creates gRPC-based Transports. It must be paired with ServerConfigs
+// that contain its ServerConfigExtension.
```
easwars, line 43: Nit: that contain an extension of type `ServerConfigExtension`? or something similar to that?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+//
+// The Extension field of the ServerConfig must be a ServerConfigExtension.
+func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
+	if sc.ServerURI == "" {
+		return nil, fmt.Errorf("ServerConfig's ServerURI field cannot be empty")
+	}
+	if sc.Extensions == nil {
+		return nil, fmt.Errorf("ServerConfig's Extensions field cannot be nil for gRPC transport")
+	}
+	gtsce, ok := sc.Extensions.(ServerConfigExtension)
```
easwars, line 56: `gtsce` is not very readable. Maybe just `sce` might do?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	if gtsce.Credentials == nil {
+		return nil, fmt.Errorf("ServerConfigExtensions's Credentials field cannot be nil for gRPC transport")
+	}
+
+	// TODO: Incorporate reference count map for existing transports and
+	// deduplicate transports based on the provided ServerConfig so that
+	// transport channel to same server can be shared between xDS and LRS
+	// client.
+
+	// Create a new gRPC client/channel for the xDS management server with the
```
easwars, line 69: Here also it can be an xDS server or an LRS server. So, I'd probably skip the xDS part.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+// Send sends a message to the server.
+func (s *stream) Send(msg []byte) error {
+	return s.stream.SendMsg(msg)
+}
+
+// Recv receives a message from the server.
+func (s *stream) Recv() ([]byte, error) {
+	var typedRes []byte
+	err := s.stream.RecvMsg(&typedRes)
+	if err != nil {
```
easwars, line 111: Nit: Assignment and conditional in the same line.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+		return typedRes, err
+	}
+	return typedRes, nil
+}
+
+// Close closes all the gRPC streams to the server.
```
easwars, line 117: Maybe: Close closes the gRPC channel to the server?

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+func Test(t *testing.T) {
+	grpctest.RunSubTests(t, s{})
+}
+
+// testServer implements the AggregatedDiscoveryServiceServer interface to test
+// the gRPC transport implementation.
+type testServer struct {
+	v3discoverygrpc.UnimplementedAggregatedDiscoveryServiceServer
+
+	lis         net.Listener                         // listener used by the test server
```
easwars, line 62: If the only reason to store the `net.Listener` here is to get the address out of it, then just store the address string here instead.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+)
+
+const (
+	defaultTestTimeout = 10 * time.Second
+)
+
+var (
+	testDiscoverRequest  = &v3discoverypb.DiscoveryRequest{VersionInfo: "1"}
+	testDiscoverResponse = &v3discoverypb.DiscoveryResponse{VersionInfo: "1"}
+)
```
easwars, line 47: Please avoid global state that is shared across tests.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			b := &Builder{}
+			tr, err := b.Build(test.serverCfg)
+			if (err != nil) != test.wantErr {
+				t.Fatalf("Build() error = %v, wantErr %v", err, test.wantErr)
+			}
+			if tr != nil {
+				defer tr.Close()
+			}
+			if !test.wantErr && tr == nil {
```
easwars, line 179: I think this conditional only captures one case: the test does not expect an error, but transport is `nil`. We also want to catch the other case: the test expects an error, but transport is non-nil.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+				t.Fatalf("got non-nil transport from Build(), want nil")
+			}
+		})
+	}
+}
+
+// TestNewStream verifies that grpcTransport.NewStream() successfully creates a
+// new client stream for the server.
+func (s) TestNewStream(t *testing.T) {
+	ts := setupTestServer(t)
```
easwars, line 189: Please move this to be inside `t.Run`. As much as possible, sub-tests should not share state among each other.

easwars: And once you do this, you wont need the `serverURI` field in the table.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			serverCfg := clients.ServerConfig{
+				ServerURI:  test.serverURI,
+				Extensions: ServerConfigExtension{Credentials: insecure.NewBundle()},
+			}
+			builder := Builder{}
+			transport, err := builder.Build(serverCfg)
+			if err != nil {
+				t.Fatalf("failed to build transport: %v", err)
```
easwars, line 216: Nit: s/failed/Failed/

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+		wantErr   bool
+	}{
+		{
+			name:      "success",
+			serverURI: ts.lis.Addr().String(),
+			wantErr:   false,
+		},
+		{
+			name:      "error",
+			serverURI: "invalid-server-uri",
```
easwars, line 203: Hmm.. ok, i see what you are doing. Maybe you can just split it into two separate tests, each with no table. So, in the second test, you wont even start a server and instead you will pass an invalid server uri to `Build`.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// Build a grpc-based transport to the above server.
+	serverCfg := clients.ServerConfig{
+		ServerURI:  ts.lis.Addr().String(),
+		Extensions: ServerConfigExtension{Credentials: insecure.NewBundle()},
+	}
+	builder := Builder{}
+	transport, err := builder.Build(serverCfg)
+	if err != nil {
+		t.Fatalf("failed to build transport: %v", err)
```
easwars, line 253: Here and everywhere else, the error strings that are part of log messages and test failure messages need not obey the rules governing general error strings in Go. See: go/go-style/decisions#error-strings

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+		t.Fatalf("failed to send message: %v", err)
+	}
+
+	// Verify that the DiscoveryRequest received on the server was same as
+	// sent.
+	select {
+	case gotReq, ok := <-ts.requestChan:
+		if !ok {
+			t.Fatalf("ts.requestChan is closed")
+		}
```
easwars, line 278: `requestChan` is never closed. I would get rid of this to simplify the test logic.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+		if !cmp.Equal(testDiscoverRequest, gotReq, protocmp.Transform()) {
```
easwars, line 279: Please use the same order, "got before want" in your comparison (as well as the error string).

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,301 @@
...
+		t.Fatalf("failed to receive message: %v", err)
+	}
+
+	// Verify that the DiscoveryResponse received was same as sent from the
+	// server.
+	var gotRes v3discoverypb.DiscoveryResponse
+	if err := proto.Unmarshal(res, &gotRes); err != nil {
+		t.Fatalf("failed to unmarshal response from ts.requestChan to DiscoveryRequest: %v", err)
+	}
+	if !cmp.Equal(testDiscoverResponse, &gotRes, protocmp.Transform()) {
```
easwars, line 298: There is just way too much coupling between the test logic and the logic in the test server. This results in very brittle tests.

What is the eventual plan? Is it to use the go-control-plane as the management server here?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+func (c *byteCodec) Unmarshal(data []byte, v any) error {
+	if b, ok := v.(*[]byte); ok {
+		*b = data
+		return nil
+	}
+	return fmt.Errorf("target must be a pointer to a byte slice")
+}
+
+func (c *byteCodec) Name() string {
+	return "byteCodec"
```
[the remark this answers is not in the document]
purnesh42H: Make sense. Changed.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Modified

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Modified

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Changed to address only

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: yes, moved the response to be part of test server struct. request can be declared within test itself since we already have channel for exchange from test server.

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Added

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Separated into 2 different tests

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Separated into 2 different tests

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: yeah, once we have xds client implemented, we will be able to use go-control-plane in e2e tests. For now, having the code coverage with simple grpc server is enough.

After last commit, response to be returned by StreamAggregatedResources handler can be controlled by test as it is a param in testServer. Regarding having test response as part of testServer struct, we can have implementation of StreamAggregatedResources within test inline so then we don't need response field in testServer.

xds/internal/clients/grpctransport/grpc_transport.go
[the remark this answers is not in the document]
easwars: Upon thinking more about this, I'm not sure how I feel about the name that I suggested. This string will be part of every HEADERS frame, sent for every new stream being created. I'm wondering if we should at all be concerned about sending such a long name in the header, for performance reasons.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+// Builder creates gRPC-based Transports. It must be paired with ServerConfigs
+// that contain Extension field of type ServerConfigExtension.
+type Builder struct{}
+
+// Build returns a gRPC-based clients.Transport.
+//
+// The Extension field of the ServerConfig must be a ServerConfigExtension.
+func (b *Builder) Build(sc clients.ServerConfig) (clients.Transport, error) {
+	if sc.ServerURI == "" {
+		return nil, fmt.Errorf("ServerConfig's ServerURI field cannot be empty")
```
easwars, line 51: go/go-style/decisions#error-strings

The style guide does say "Error strings should not be capitalized unless beginning with an exported name". But this error is happening deep inside the stack, and will definitely be wrapped in some other error. So, this capitalization will not look great when the final error is passed to the user.

Can we do better here?

It is customary for Go packages to prefix their errors with their package name. So, one option is to do that for all errors generated by this package. So, errors here can be `fmt.Errorf("grpctransport: missing field ServerConfig.ServerURI")`. Or if we take a leaf out of the `tls` package, they could be `fmt.Errorf("grpctransport: ServerURI not set in ServerConfig")`

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	return g.cc.Close()
+}
+
+type byteCodec struct{}
+
+func (c *byteCodec) Marshal(v any) ([]byte, error) {
+	if b, ok := v.([]byte); ok {
+		return b, nil
+	}
+	return nil, fmt.Errorf("message must be a byte slice")
```
easwars, line 128: Can we identify the type of `v` in the error message?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (c *byteCodec) Unmarshal(data []byte, v any) error {
+	if b, ok := v.(*[]byte); ok {
+		*b = data
+		return nil
+	}
+	return fmt.Errorf("target must be a pointer to a byte slice")
```
easwars, line 136: Same here. Identify the offending type.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	// credentials, server URI, and a byte codec to send and receive messages.
+	// Also set a static keepalive configuration that is common across gRPC
+	// language implementations.
+	kpCfg := grpc.WithKeepaliveParams(keepalive.ClientParameters{
+		Time:    5 * time.Minute,
+		Timeout: 20 * time.Second,
+	})
+	cc, err := grpc.NewClient(sc.ServerURI, kpCfg, grpc.WithCredentialsBundle(sce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
+	if err != nil {
+		return nil, fmt.Errorf("error creating grpc client for server uri %s, %v", sc.ServerURI, err)
```
easwars, line 79: This error string could be `grpctransport: failed to create transport to server %q: %v", sc.ServerURI, err`

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	address     string                               // address of the server
+	requestChan chan *v3discoverypb.DiscoveryRequest // channel to send the received requests on for verification
+	response    *v3discoverypb.DiscoveryResponse     // response to send back to the client from handler
+}
+
+// setupTestServer set up the gRPC server for AggregatedDiscoveryService. It
+// creates an instance of testServer that returns the provided response from
+// the StreamAggregatedResources() handler and registers it with a gRPC server.
+func setupTestServer(t *testing.T, response *v3discoverypb.DiscoveryResponse) *testServer {
```
easwars, line 65: Please mark as a test helper. See: go/go-style/decisions#mark-test-helpers

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			b := &Builder{}
+			tr, err := b.Build(test.serverCfg)
+			if (err != nil) != test.wantErr {
+				t.Fatalf("Build() error = %v, wantErr %v", err, test.wantErr)
+			}
+			if tr != nil {
```
easwars, line 174: Do you think we can simplify all this complex logic by splitting the success and failure cases into separate tests? See: https://g3doc.corp.google.com/company/teams/go-community/gotip/episodes/50.md?cl=head

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+		// Send the response back to the client
+		if err := stream.Send(s.response); err != nil {
+			return err
+		}
+	}
+}
+
+// TestBuild verifies that the grpctransport.Builder creates a new
+// grpc.ClientConn every time Build() is called.
```
easwars, line 115: `every time Build() is called`
I don't see this being verified or even exercised here.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	_, err = transport.NewStream(ctx, "/envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources")
+	if err == nil {
+		t.Fatal("transport.NewStream() succeeded, want failure")
+	}
+}
+
+// TestStream_SendAndRecv verifies that grpcTransport.Stream.Send()
```
easwars, line 233: Minor nits. I don't think we have to identify function/method names with their package and type prefixes like `grpcTransport.Stream.Send`. Just `Send` and `Recv` should be good enough I feel.

Applies to many such comments across this file.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+// and grpcTransport.Stream.Recv() successfully send and receive messages
+// on the stream to and from the gRPC server.
+//
+// It starts a gRPC test server using setupTestServer(). The test then sends a
+// testDiscoverRequest on the stream and verifies that the received discovery
```
easwars, line 238: Nit: s/received discovery/received discovery request/

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Verify that the DiscoveryRequest received on the server was same as
+	// sent.
+	select {
+	case gotReq := <-ts.requestChan:
+		if !cmp.Equal(gotReq, testDiscoverRequest, protocmp.Transform()) {
+			t.Fatalf("<-ts.requestChan = %v, want %v", &gotReq, &testDiscoverRequest)
+		}
+	case <-ctx.Done():
+		t.Fatalf("timeout waiting for request to reach server")
```
easwars, line 284: s/timeout/Timeout

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := stream.Send(msg); err != nil {
+		t.Fatalf("Failed to send message: %v", err)
+	}
+
+	// Verify that the DiscoveryRequest received on the server was same as
+	// sent.
+	select {
+	case gotReq := <-ts.requestChan:
+		if !cmp.Equal(gotReq, testDiscoverRequest, protocmp.Transform()) {
+			t.Fatalf("<-ts.requestChan = %v, want %v", &gotReq, &testDiscoverRequest)
```
easwars, line 281: Test error strings should be readable. `<-ts.requestChan` is not.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Thanks. Modified all the error messages in format "grpctransport: ....not set in ....."

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Simplified. For success, also added check to verify if grpc.ClientConn is not nil.

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Not applicable now.

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Thanks for catching. Modified to display diff.

xds/internal/clients/grpctransport/grpc_transport.go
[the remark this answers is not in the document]
dfawley: > This string will be part of every HEADERS frame, sent for every new stream being created

Why?  I don't think we will set the content sub-type unless `CallContentSubtype` is used, will we?

The name really shouldn't matter -- the codec isn't registered anywhere, so there should be no chance it conflicts with anything the user is doing.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	return &grpcTransport{cc: cc}, nil
+}
+
+type grpcTransport struct {
+	cc *grpc.ClientConn
+}
+
+// NewStream creates a new gRPC stream to the server for the specified method.
+func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.Stream, error) {
+	s, err := g.cc.NewStream(ctx, &grpc.StreamDesc{StreamName: method, ClientStreams: true, ServerStreams: true}, method)
```
dfawley, line 91: Per our godoc:
```
	// StreamName and Handler are only used when registering handlers on a
	// server.
```

```suggestion
	s, err := g.cc.NewStream(ctx, &grpc.StreamDesc{ClientStreams: true, ServerStreams: true}, method)
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Recv receives a message from the server.
+func (s *stream) Recv() ([]byte, error) {
+	var typedRes []byte
+
+	if err := s.stream.RecvMsg(&typedRes); err != nil {
+		return typedRes, err
+	}
+	return typedRes, nil
```
dfawley, line 114:
```suggestion
	if err := s.stream.RecvMsg(&typedRes); err != nil {
		return nil, err
	}
	return typedRes, nil
```

OR

```suggestion
	err := s.stream.RecvMsg(&typedRes)
	return typedRes, err
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Close closes the gRPC channel to the server.
+func (g *grpcTransport) Close() error {
+	return g.cc.Close()
+}
```
dfawley, line 120: Nit: move this to be next to the other grpcTransport methods.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+
+type byteCodec struct{}
+
+func (c *byteCodec) Marshal(v any) ([]byte, error) {
+	if b, ok := v.([]byte); ok {
+		return b, nil
+	}
+	return nil, fmt.Errorf("message is %T, but must be a byte slice ", v)
```
dfawley, line 128:
```suggestion
	return nil, fmt.Errorf("message is %T, but must be a byte slice", v)
```

or even

```suggestion
	return nil, fmt.Errorf("message is %T, but must be []byte", v)
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (c *byteCodec) Unmarshal(data []byte, v any) error {
+	if b, ok := v.(*[]byte); ok {
+		*b = data
+		return nil
+	}
+	return fmt.Errorf("target is %T, but must be a pointer to a byte slice", v)
```
dfawley, line 136: Maybe

```suggestion
	return fmt.Errorf("target is %T, but must be *[]byte", v)
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	stream grpc.ClientStream
+}
+
+// Send sends a message to the server.
+func (s *stream) Send(msg []byte) error {
+	return s.stream.SendMsg(msg)
+}
+
+// Recv receives a message from the server.
+func (s *stream) Recv() ([]byte, error) {
```
dfawley, line 108: We never need to stop a stream, I guess?  So this only has 2 methods?

So then when we add the connection sharing, we'll need to keep track of which streams are in flight for the instance of the transport, and close them all when the transport is closed.  Just FYI for now.

xds/internal/clients/grpctransport/grpc_transport_test.go
dfawley: I'm fine with just removing the grpctest magic for now.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,306 @@
...
+import (
+	"context"
+	"io"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
```
dfawley, line 30: Is there a reason to use `insecure` instead of `local`?

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,306 @@
...
+		address:     lis.Addr().String(),
+		response:    response,
+	}
+
+	s := grpc.NewServer()
+
+	v3discoverygrpc.RegisterAggregatedDiscoveryServiceServer(s, ts)
+	go func() {
+		if err := s.Serve(lis); err != nil {
+			t.Logf("Server exited with error: %v", err)
```
dfawley, line 83: `t.Errorf`?

Note that this line races with the test case exiting entirely and `t` becoming invalid.  So if this is not just `go s.Serve(lis)` then we probably should have a waitgroup that the `Cleanup` function blocks on after calling stop.

xds/internal/clients/grpctransport/grpc_transport.go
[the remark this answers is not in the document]
purnesh42H: For now i have kept it "grpctransport.byteCodec"

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+// ServerConfigExtension holds settings for connecting to a gRPC server,
+// such as an xDS management or an LRS server.
+type ServerConfigExtension struct {
+	// Credentials will be used for all gRPC transports. If it is unset,
+	// transport creation will fail.
+	Credentials credentials.Bundle
+}
+
+// Builder creates gRPC-based Transports. It must be paired with ServerConfigs
+// that contain Extension field of type ServerConfigExtension.
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Thanks for info. Changed.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Changed to `return nil, err`

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Moved

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Changed to `...[]byte`

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: I hadn't come across `local`. I checked it and changed to use local.TransportCredentials() in the test credentials implementing credentials.Bundle

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Changed to `go s.Serve(lis)`

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: i think we can keep it because in the next PR I am going to copy it over to clients anyways.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: so, the right now xds client implementation is handling the stream closure https://github.com/purnesh42H/grpc-go/pull/10/files#diff-ca92835519a714b27f2accab838042e37299d2ed453c4ddb522a711c5bb802d0 and its triggered when the xds client wants to close the transport channel.

But yeah once we have refcounting on grpctransport and share the channel with xds and lrs client LoadStore, we will have to check before closing for the instance of grpctransport.

dfawley: As discussed offline, a stream close isn't necessary since cancelling the stream's context does that already.

xds/internal/clients/grpctransport/grpc_transport_test.go
dfawley: OK that's fine then.

dfawley (approved):

dfawley:
You might need to push an empty commit to make the new test case run.  Let me know if it's not working and I can force the merge.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	return &grpcTransport{cc: cc}, nil
+}
+
+type grpcTransport struct {
+	cc *grpc.ClientConn
+}
+
+// NewStream creates a new gRPC stream to the server for the specified method.
+func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.Stream, error) {
+	s, err := g.cc.NewStream(ctx, &grpc.StreamDesc{ClientStreams: true, ServerStreams: true}, method)
```
easwars, line 91: This means that this transport only support bidirectional streams. It doesn't matter because both ADS and LRS are bidirectional streams. But I'm wondering if this should be documented?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -0,0 +1,141 @@
...
+	return typedRes, nil
+}
+
+type byteCodec struct{}
+
+func (c *byteCodec) Marshal(v any) ([]byte, error) {
+	if b, ok := v.([]byte); ok {
+		return b, nil
+	}
+	return nil, fmt.Errorf("message is %T, but must be a []byte", v)
```
easwars, line 128: Nit: Adding a package prefix to this error message would also be useful I think. So, maybe something like:
fmt.Errorf("grpctransport: attempting to send message of type %T, but must be a []byte", v)
and similarly for the error returned by the  `Unmarshal` method as well.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,313 @@
...
+	for {
+		// Receive a DiscoveryRequest from the client
+		req, err := stream.Recv()
+		if err == io.EOF {
+			return nil // Stream closed by client
+		}
+		if err != nil {
+			return err // Handle other errors
+		}
+		s.requestChan <- req
```
easwars, line 105: Might be minor. But, if this method is blocked writing to the channel, canceling the stream context will not unblock the server handler. So, we might want to use a `select` here with one `case` for the stream context getting done.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,313 @@
...
+	transport, err := builder.Build(serverCfg)
+	if err != nil {
+		t.Fatalf("Failed to build transport: %v", err)
+	}
+	defer transport.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	_, err = transport.NewStream(ctx, "/envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources")
+	if err != nil {
```
easwars, line 218: Nit: same line.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,313 @@
...
+	transport, err := builder.Build(serverCfg)
+	if err != nil {
+		t.Fatalf("Failed to build transport: %v", err)
+	}
+	defer transport.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	_, err = transport.NewStream(ctx, "/envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources")
+	if err == nil {
```
easwars, line 240: Nit: same here.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if err := stream.Send(msg); err != nil {
+		t.Fatalf("Failed to send message: %v", err)
+	}
+
+	// Verify that the DiscoveryRequest received on the server was same as
+	// sent.
+	select {
+	case gotReq := <-ts.requestChan:
+		if diff := cmp.Diff(gotReq, testDiscoverRequest, protocmp.Transform()); diff != "" {
```
easwars, line 291: I think you got the order wrong in the call to `cmp.Diff` based on what is used in the error message. See: go/go-style/decisions#types-of-equality

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,313 @@
...
+	res, err := stream.Recv()
+	if err != nil {
+		t.Fatalf("Failed to receive message: %v", err)
+	}
+
+	// Verify that the DiscoveryResponse received was same as sent from the
+	// server.
+	var gotRes v3discoverypb.DiscoveryResponse
+	if err := proto.Unmarshal(res, &gotRes); err != nil {
+		t.Fatalf("Failed to unmarshal response from ts.requestChan to DiscoveryResponse: %v", err)
```
easwars, line 308: Nit: s/ts.requestChan/server/ ?

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if diff := cmp.Diff(&gotRes, ts.response, protocmp.Transform()); diff != "" {
+		t.Fatalf("proto.Unmarshal(res, &gotRes) returned unexpected diff (-want +got):\n%s", diff)
```
easwars, line 311: Same here, about the order or arguments to `cmp.Diff` and the order of want and got in the error message.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Added

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: thanks. Yeah changed to select with ctx.Done()

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: i had it want, got before then i think it got changed to want, got. I think want, got is correct way. Changed.

xds/internal/clients/grpctransport/grpc_transport_test.go
purnesh42H: i had it want, got before then i think it got changed to want, got. I think want, got is correct way. Changed.

PR #8067, opened 2025-02-05
deps: bump envoyproxy/go-control-plane/envoy and synchronize go.mods

mmorel-35:
#### Description

synchronise go.mods with 
* google.golang.org/grpc@v1.70.0
* google.golang.org/protobuf@v1.36.4
* github.com/envoyproxy/go-control-plane/envoy@v1.32.4

RELEASE NOTES: None


arjan-bal:
Hi @mmorel-35, I believe the aim is to bump go-control-plane/envoy to v1.32.4, and the other modules were bumped to achieve this. Is there a bug fix in this version of Envoy that needs to be included in the next grpc-go release?

We bump all the dependencies after cutting release branches every six weeks. If a dependency bump isn't urgent, we wait for the next release.

arjan-bal:
The latest release brach doesn't have the `github.com/envoyproxy/go-control-plane/envoy` dependency, this change is needed only for `master`. 
https://github.com/grpc/grpc-go/blob/v1.70.x/go.mod

arjan-bal (approved):

arjan-bal:
LGTM, thank you for sending the contribution!

PR #8068, opened 2025-02-06
credentials, transport, grpc : add a call option to override the :authority header on a per-RPC basis

eshitachandwani:
Fixes: #5361 

RELEASE NOTES: 
- Added a `CallAuthority` callOption that can be used to overwrite the http `:authority` header on per-RPC basis.


credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,365 @@
...
+				grpc.WithTransportCredentials(&FakeCredsWithAuthValidator{}))
+			if err != nil {
+				t.Fatalf("Failed to create gRPC client connection: %v", err)
+			}
+			defer clientConn.Close()
+
+			// Perform a test RPC with a specified call authority.
+			ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+			defer cancel()
+
```
purnesh42H, line 356: nix new line

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,365 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	if _, err = testgrpc.NewTestServiceClient(cc).EmptyCall(ctx, &testpb.Empty{}, grpc.CallAuthority(expectedAuthority)); err != nil {
+		t.Fatalf("EmptyCall() rpc failed: %v", err)
+	}
+}
+
+// FakeCredsNoAuthValidator is a test credential that does not implement AuthorityValidator.
+type FakeCredsNoAuthValidator struct {
```
purnesh42H, line 198: the FakeCreds can be at the top

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,365 @@
...
+// ServerHandshake performs the server-side handshake.
+// Returns a test AuthInfo object to satisfy the interface requirements.
+func (c *FakeCredsNoAuthValidator) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	return rawConn, TestAuthInfo{}, nil
+}
+
+// TestCallOptionWithNoAuthorityValidator tests the CallAuthority call option
+// with custom credentials that do not implement AuthorityValidator and verifies
+// that it fails with `UNAVAILABLE` status code.
+func (s) TestCallOptionWithNoAuthorityValidator(t *testing.T) {
```
purnesh42H, line 236: Same. This can be another test case in test with CustomCreds

purnesh42H (commented):
Overall lgtm. Just few minor comments. I think we can combine some individual into t-tests.

purnesh42H:
Also, don't feel strongly but may be we can utilize the setup in tls_ext_test or move the tls tests to tls_ext_test.go

credentials/credentials.go
```diff
@@ -120,6 +120,13 @@ type AuthInfo interface {
 	AuthType() string
 }
 
+// AuthorityValidator validates the authority used to override the `:authority`
+// header. A struct implementing AuthInfo should also implement
+// AuthorityValidator if the credentials need to support per-RPC authority overrides.
+type AuthorityValidator interface {
```
easwars, line 126: Also, I think we should focus on documenting a couple of things here:
- that this is an option interface
- will be invoked only when the appropriate call option is specified by the application

Something to this effect could work (might need some cleanup)
```
// AuthorityValidator validates  `:authority` header overrides. This is an
// optional interface for transport credentials to implement.
type AuthorityValidator interface {
	// ValidateAuthority validates the authority used to override the
	// `:authority` header. If implemented by the AuthInfo type returned by the
	// transport credentials, this method will be invoked if the client
	// application attempts to override the `:authority` header by using the
	// CallAuthority call option.
	ValidateAuthority(authority string) error
}
```

credentials/tls.go
```diff
@@ -50,6 +50,17 @@ func (t TLSInfo) AuthType() string {
...
 
+// ValidateAuthority validates the authority by checking it against the peer certificates.
+func (t TLSInfo) ValidateAuthority(authority string) error {
+	var err error
+	for _, cert := range t.State.PeerCertificates {
+		if err = cert.VerifyHostname(authority); err == nil {
+			return nil
+		}
+	}
+	return fmt.Errorf("credentials: failed to verify authority %s", err)
```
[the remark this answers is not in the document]
easwars: Nit: Also, the error message could be something like
`fmt.Errorf(credentials: invalid authority %q: %s", authority, err)`

rpc_util.go
```diff
@@ -365,6 +366,36 @@ func (o MaxRecvMsgSizeCallOption) before(c *callInfo) error {
...
 
+// CallAuthority returns a CallOption which sets the authority to override the
+// `:authority` pseudoheader in a RPC. If this is set, the RPC will use only
+// this authority.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a
+// later release.
+func CallAuthority(auth string) CallOption {
```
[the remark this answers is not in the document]
easwars: We went through this in the design doc. We had consensus on `CallAuthority`.

easwars, line 377: Nit: s/auth/authority/

Please note that this is part of our public API. Parameter names also serve as documentation. So using good parameter names is critical.

rpc_util.go
```diff
@@ -365,6 +366,36 @@ func (o MaxRecvMsgSizeCallOption) before(c *callInfo) error {
 }
 func (o MaxRecvMsgSizeCallOption) after(*callInfo, *csAttempt) {}
 
+// CallAuthority returns a CallOption which sets the authority to override the
+// `:authority` pseudoheader in a RPC. If this is set, the RPC will use only
```
easwars, line 370: This is the only place where we use the term `pseudoheader`. Either use it everywhere (like in the dosctrings of the `AuthorityValidator` interface), or don't use it here. Thanks.

easwars: How about something more simpler and direct

"CallAuthority creates a CallOption that sets the HTTP/2 :authority header of an RPC to the specified value."

rpc_util.go
```diff
@@ -365,6 +366,36 @@ func (o MaxRecvMsgSizeCallOption) before(c *callInfo) error {
...
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a
+// later release.
+func CallAuthority(auth string) CallOption {
+	return AuthorityOverrideCallOption{Authority: auth}
+}
+
+// AuthorityOverrideCallOption is a CallOption which sets the authority that the
```
easwars, line 381: Same here

AuthorityOverrideCallOption is a CallOption that indicates the HTTP/2 :authority header value to use for the call.

internal/transport/http2_client.go
```diff
@@ -749,6 +749,24 @@ func (t *http2Client) NewStream(ctx context.Context, callHdr *CallHdr) (*ClientS
...
 
+	// Authority specified in the `CallAuthority` CallOption takes the highest
+	// precedence to determine the :authority header. Any value in Host field of
+	// CallHdr is overwritten. But before overriding, we validate the authority
+	// string against the peer certificates and fail the RPC with `UNAVAILABLE`
+	// status code if either of the condition fails.
+	if callHdr.Authority != "" {
+		auth, ok := t.authInfo.(credentials.AuthorityValidator)
+		if !ok {
+			return nil, &NewStreamError{Err: status.Error(codes.Unavailable, fmt.Sprintf("credentials type %s does not implement the AuthorityValidator interface", t.authInfo.AuthType())), AllowTransparentRetry: false}
```
easwars, line 760: Extend the error message like

credentials type %s does not implement the AuthorityValidator interface, but authority override specified with CallAuthority call option

internal/transport/transport.go
```diff
@@ -540,6 +540,9 @@ type CallHdr struct {
 	PreviousAttempts int // value of grpc-previous-rpc-attempts header to set
 
 	DoneFunc func() // called when the stream is finished
+
+	// Authority is used to override the `:authority` header.
```
easwars, line 544: I understand this is an internal package, but I feel it makes sense to mention here that if this value is populated, it will be preferred over the `Host` field to be used as the value for the `:authority` header.

internal/transport/http2_client.go
```diff
@@ -749,6 +749,24 @@ func (t *http2Client) NewStream(ctx context.Context, callHdr *CallHdr) (*ClientS
...
+		}
+		if err := auth.ValidateAuthority(callHdr.Authority); err != nil {
+			return nil, &NewStreamError{Err: status.Error(codes.Unavailable, err.Error()), AllowTransparentRetry: false}
```
easwars, line 763: Here and above, do we need to set `AllowTransparentRetry: false` give that the default for boolean is `false`?

internal/transport/http2_client.go
```diff
@@ -749,6 +749,24 @@ func (t *http2Client) NewStream(ctx context.Context, callHdr *CallHdr) (*ClientS
 		callHdr = &newCallHdr
 	}
 
+	// Authority specified in the `CallAuthority` CallOption takes the highest
+	// precedence to determine the :authority header. Any value in Host field of
+	// CallHdr is overwritten. But before overriding, we validate the authority
+	// string against the peer certificates and fail the RPC with `UNAVAILABLE`
```
easwars, line 755: `against peer certificates` is only true for TLS creds. So, we probably can omit that here.

internal/transport/http2_client.go
easwars: Also, might be useful to include the the authority value that failed validation here, since there is no guarantee that the error string returned by the credentials contains that.

test/end2end_test.go
```diff
@@ -5013,7 +5016,8 @@ func (s) TestInterceptorCanAccessCallOptions(t *testing.T) {
 		grpc.PerRPCCredentials(testPerRPCCredentials{}),
 		grpc.Header(&headers),
 		grpc.Trailer(&trailers),
-		grpc.Peer(&pr))
+		grpc.Peer(&pr),
+		grpc.CallAuthority("test-authority"))
```
easwars, line 5020: Please consider writing a simple e2e test that only uses this feature. This file requires a lot of cleanup because it has a ton of options that tests end up setting, and eventually these tests become so complicated. Writing a separate test will make the test more readable and will also ensure that when the test fails, it is easy to debug. Also, when writing the new test, do not use this `test` type provided in this file. Instead bring up your own test server using the stubserver package, and create a grpc channel and make an RPC and validate the scenarios.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+	"net"
+	"testing"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/internal/stubserver"
+	testgrpc "google.golang.org/grpc/interop/grpc_testing"
+	testpb "google.golang.org/grpc/interop/grpc_testing"
```
easwars, line 35: Nit: Please group proto imports separately.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+func (s) TestCorrectAuthorityWithCustomCreds(t *testing.T) {
+	tests := []struct {
+		name         string
+		creds        credentials.TransportCredentials
+		expectedAuth string
+		wantRPCError bool
+	}{
+		{
+			name:         "CorrectAuthorityWithFakeCreds",
+			expectedAuth: "auth.test.example.com",
```
easwars, line 289: Prefer want over expected (in variable/field names, as well in error strings). Throughout this file.

go/go-style/decisions#got-before-want

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+func (s) TestAuthorityCallOptionsWithTLSCreds(t *testing.T) {
+	tests := []struct {
+		name         string
+		expectedAuth string
+		wantRPCError bool
+	}{
+		{
+			name:         "CorrectAuthorityWithTLS",
+			expectedAuth: "auth.test.example.com",
+			wantRPCError: false,
```
easwars, line 102: What if you simplified this by making this field `wantStatus`. So, in this subtest the field will be set to `codes.Ok` and the next one will have `codes.Unavailable`. And you wont need the `checkUnavailableRPCError` helper function at all.

The last line can be rewritten as
if _, err :=client.EmptyCall(ctx, &testpb.Empty{}, grpc.CallAuthority(tt.expectedAuth)); status.Code(err) != tt.wantStatus {
  t.Fatalf("EmptyCall returned status: %v, want %v", status.Code(err), tt.wantStatus)
}

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+			if tt.wantRPCError {
+				checkUnavailableRPCError(t, err)
+			} else if err != nil {
+				t.Fatalf("EmptyCall() rpc failed: %v", err)
+			}
+		})
+	}
+}
+
+func (s) TestTLSCredsWithNoAuthorityOverride(t *testing.T) {
```
easwars, line 141: Please add a test comment for this.

What scenario is this testing? If this test fails, what does it indicate about the authority override feature?

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+			name:         "IncorrectAuthorityWithTLS",
+			expectedAuth: "auth.example.com",
+			wantRPCError: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ss := &stubserver.StubServer{
+				EmptyCallF: func(ctx context.Context, _ *testpb.Empty) (*testpb.Empty, error) {
+					return authorityChecker(ctx, tt.expectedAuth)
```
easwars, line 114: Hmmm.. I'm not convinced about checking the authority header on the server handler, because when the validation fails, the RPC will not even make it to the server. Maybe, checking from the server handler is OK when you actually expect validation to succeed on the client and expect the RPC to reach the server.

You know the certs you are using for the server. So, you can specify an authority override on the client that you expect to work and one that you dont expect to work, because it will fail validation with the peer certificate.

credentials/credentials_ext_test.go
easwars: Please rename this `testCreds`. We use the term fake when we use a test double that actually provides a fake implementation of something that the code under test depends on.

Also, the AuthInfo types need to be named such that it is clear from their names that they either implement the `AuthorityValidator` interface or not.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+// ClientHandshake performs the client-side handshake.
+func (c *FakeCredsNoAuthValidator) ClientHandshake(ctx context.Context, authority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	return rawConn, TestAuthInfo{}, nil
+}
+
+// TestAuthInfo implements the AuthInfo interface.
+type TestAuthInfo struct{}
+
+// AuthType returns the authentication type.
+func (TestAuthInfo) AuthType() string { return "test" }
```
easwars, line 209: Nit: The definition of this type is sandwiched between the implementation of the test creds type. Please move it to be separate.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+type FakeCredsNoAuthValidator struct {
+}
+
+// ClientHandshake performs the client-side handshake.
+func (c *FakeCredsNoAuthValidator) ClientHandshake(ctx context.Context, authority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	return rawConn, TestAuthInfo{}, nil
+}
+
+// TestAuthInfo implements the AuthInfo interface.
+type TestAuthInfo struct{}
```
easwars, line 206: Does this have to be exported?

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+// OverrideServerName overrides the server name used for verification.
+func (c *FakeCredsWithAuthValidator) OverrideServerName(serverName string) error {
+	return nil
+}
+
+// ServerHandshake performs the server-side handshake.
+// Returns a test AuthInfo object to satisfy the interface requirements.
+func (c *FakeCredsWithAuthValidator) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
```
easwars, line 273: Basically, we can have a single `testCreds` type that contains a boolean field that indicates whether or not it should return an `AuthInfo` that implements `AuthorityValidator` or not. We don't need two test credentials types here.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+// TestCorrectAuthorityWithCustomCreds tests the CallAuthority call option
+// with custom credentials that implement AuthorityValidator and verifies
+// it with both correct and incorrect authority override.
+func (s) TestCorrectAuthorityWithCustomCreds(t *testing.T) {
+	tests := []struct {
+		name         string
+		creds        credentials.TransportCredentials
+		expectedAuth string
+		wantRPCError bool
```
easwars, line 285: Same here. You can directly use the status that you want the RPC to return.

test/end2end_test.go
eshitachandwani: Right! I already have the e2e tests , just thought this is could serve as a unit test for callOption , but removing it.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,365 @@
...
+			if tt.expectRPCError {
+				checkUnavailableRPCError(t, err)
+			} else if err != nil {
+				t.Fatalf("EmptyCall() rpc failed: %v", err)
+			}
+		})
+	}
+}
+
+func (s) TestTLSCredsWithNoAuthorityOverride(t *testing.T) {
```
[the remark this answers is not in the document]
eshitachandwani: I had added that after our discussion to have a base test , to check if the authority is not being override without calling the callOption , but I think the authority tests might already be testing that . So I am removing it.

purnesh42H:
> Added a CallAuthority callOption that can be used to overwrite the http :authority header on per-RPC basis.
Introduces an option AuthorityValidator interface which needs to be implemented by credentials that want to allow authority overwrite.
Adds implementations of AuthorityValidator interface for Insecure and TLS credentials.

these aren't the release notes. Are they? Release notes needs to make sense to the user which is the new call option so it can be something close to what you have in the PR title. The above bullets can just be moved to PR description.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "failed to parse metadata")
+	}
+	auths, ok := md[":authority"]
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "no authority header")
+	}
+	if len(auths) != 1 {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("no authority header, auths = %v", auths))
```
purnesh42H, line 67: nit: can't we just use `status.Errorf` instead of having to use another `fmt.Sprintf`? same for below

credentials/tls.go
```diff
@@ -50,6 +51,22 @@ func (t TLSInfo) AuthType() string {
 	return "tls"
 }
 
+// ValidateAuthority validates that the provided authority being used to
+// override the :authority header is valid by verifying it against the peer
```
purnesh42H, line 55:
```suggestion
// override the :authority header by verifying it against the peer
```

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+	return credentials.ProtocolInfo{}
+}
+
+// OverrideServerName overrides the server name used for verification.
+func (c *testCreds) OverrideServerName(serverName string) error {
+	return nil
+}
+
+// TestCorrectAuthorityWithCustomCreds tests the `grpc.CallAuthority` call
+// option using custom credentials. It verifies behavior both, when the
```
purnesh42H, line 218: nit: the paragraph is hard to read since there are multiple cases. Can use bullets. Something like

```
// TestCorrectAuthorityWithCustomCreds tests the `grpc.CallAuthority` call
// option using custom credentials. It verifies following behaviors that are expected
// fail with `UNAVAILABLE` status code:
// - when the credentials implement AuthorityValidator with correct and incorrect authority
// - when credentials do not implement AuthorityValidator but dial option `grpc.CallAuthority` is used
// - when the override authority is invalid
```

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	if auths[0] != wantAuthority {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid authority header %v, want %v", auths[0], wantAuthority))
+	}
+	return &testpb.Empty{}, nil
+}
+
+// Tests the grpc.CallAuthority option with TLS credentials. This test verifies
+// that the provided authority is correctly propagated to the server when using
+// TLS. It covers both positive and negative cases: correct authority and
```
purnesh42H, line 77:
```suggestion
// TLS when a correct authority is used anf fail with `UNAVAILABLE` status code when incorrect authority is used`
```

rpc_util.go
```diff
@@ -365,6 +366,35 @@ func (o MaxRecvMsgSizeCallOption) before(c *callInfo) error {
 }
 func (o MaxRecvMsgSizeCallOption) after(*callInfo, *csAttempt) {}
 
+// CallAuthority creates a CallOption that sets the HTTP/2 :authority header of
```
purnesh42H, line 369: I think we should mention here that if this dial option is used, the credentials in use must implement `AuthorityValidator`

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+// are expected to fail with `UNAVAILABLE` status code.
+func (s) TestCorrectAuthorityWithCustomCreds(t *testing.T) {
+	tests := []struct {
+		name       string
+		creds      credentials.TransportCredentials
+		wantAuth   string
+		wantStatus codes.Code
+	}{
+		{
+			name:       "CorrectAuthorityWithFakeCreds",
```
purnesh42H, line 232: nit: can use _ or just space between words for better readability. Applies to all tabular test cases

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	return &testpb.Empty{}, nil
+}
+
+// Tests the grpc.CallAuthority option with TLS credentials. This test verifies
+// that the provided authority is correctly propagated to the server when using
+// TLS. It covers both positive and negative cases: correct authority and
+// incorrect authority, expecting the RPC to fail with `UNAVAILABLE` status code
+// error in the later case.
+func TestAuthorityCallOptionsWithTLSCreds(t *testing.T) {
```
easwars, line 80: What do you think about splitting this test into two:
- Where the authority override specified by the caller in the call option is accepted by the credentials, and therefore you can verify that the `:authority` header on the server RPC handler is set appropriately
- Where the authority override specified by the caller in the call option is not accepted by the credentials, and therefore the RPC fails on the client side, and you can therefore verify that the server RPC handler is not even invoked.

I feel that calling `authorityChecker` from the server handler (for both cases) makes it look like the error happens on the server, which is not the case.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+	tests := []struct {
+		name       string
+		wantAuth   string
```
easwars, line 83: Also, it is a little confusing to call this `wantAuth` since this is the same field that is used to populate the call option on the client.

credentials/credentials_ext_test.go
easwars: I think what we have currently is good. Please see: go/go-style/decisions#subtest-names and https://g3doc.corp.google.com/company/teams/go-community/gotip/episodes/117.md?cl=head. The tl;dr is to think of subtest names like other function or variable names (i.e., identifiers in your code). If you need a more descriptive description for every entry in the table, you could add a different field (maybe call it `desc`, where you have a more free form text description), and you can add that to error messages.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+type testAuthInfoWithValidator struct{}
+
+// AuthType returns the authentication type.
+func (testAuthInfoWithValidator) AuthType() string {
+	return "test"
+}
+
+// ValidateAuthority implements credentials.AuthorityValidator.
+func (testAuthInfoWithValidator) ValidateAuthority(authority string) error {
+	if authority == "auth.test.example.com" {
```
easwars, line 174: Please have a field in `testAuthInfoWithValidator` that stores the supported authority, instead of using a magic string here.

credentials/credentials_ext_test.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil
+	}
+	return fmt.Errorf("invalid authority")
+}
+
+// testCreds is a test TransportCredentials that can optionally support
+// authority validation.
+type testCreds struct {
+	WithValidator bool
```
easwars, line 183: Since this is the type that creates the `testAuthInfoWithValidator`, you would need to have a field for the supported authority here too.

Doing so makes it clear when reading the test as to what authority is supported by the creds, and what is specified by the test, and therefore whether the case is expected to pass or not.

credentials/credentials_ext_test.go
easwars: All cases don't expect to fail with `UNAVAILABLE` status code.

And `grpc.CallAuthority` is a call option and not a dial option.

credentials/insecure/insecure.go
```diff
@@ -71,6 +71,10 @@ func (info) AuthType() string {
 	return "insecure"
 }
 
+func (info) ValidateAuthority(string) error {
```
easwars, line 74: A comment for this method please, mentioning that insecure credentials allows override the HTTP/2 `:authority` header to any value.

credentials/tls.go
```diff
@@ -50,6 +51,22 @@ func (t TLSInfo) AuthType() string {
...
+// certificates. It returns nil on successful validation, or a non-nil error if
```
easwars, line 56: Same nit here. You can just say "It returns a non-nil error if validation fails".

credentials/tls.go
```diff
@@ -50,6 +51,22 @@ func (t TLSInfo) AuthType() string {
...
+func (t TLSInfo) ValidateAuthority(authority string) error {
+	var errs []error
+	for _, cert := range t.State.PeerCertificates {
+		var err error
+		if err = cert.VerifyHostname(authority); err == nil {
+			return nil
+		}
+		errs = append(errs, err)
+	}
+	return fmt.Errorf("credentials: invalid authority %q: %w", authority, errors.Join(errs...))
```
easwars, line 67: Embedding the error from the tls package using the `%w` format string means that we are making that error a part of our API. I don't think we want that.

Please read this section: go/go-style/best-practices#error-extra-info

internal/transport/http2_client.go
```diff
@@ -749,6 +749,25 @@ func (t *http2Client) NewStream(ctx context.Context, callHdr *CallHdr) (*ClientS
...
+	// The authority specified via the `CallAuthority` CallOption takes the
+	// highest precedence when determining the `:authority` header. It overrides
+	// any value present in the Host field of CallHdr. Before applying this
+	// override, the authority string is validated. If the credentials do not
+	// implement the AuthorityValidator interface, or if validation fails, the
+	// RPC is failed with a status code of `UNAVAILABLE`.
+	if callHdr.Authority != "" {
+		auth, ok := t.authInfo.(credentials.AuthorityValidator)
+		if !ok {
+			return nil, &NewStreamError{Err: status.Error(codes.Unavailable, fmt.Sprintf("credentials type %s does not implement the AuthorityValidator interface, but authority override specified with CallAuthority call option", t.authInfo.AuthType()))}
```
easwars, line 761: Same here too. Use `status.Errorf`.

easwars: And use `%q` formatting directive for the credentials type.

internal/transport/http2_client.go
```diff
@@ -749,6 +749,25 @@ func (t *http2Client) NewStream(ctx context.Context, callHdr *CallHdr) (*ClientS
...
+		}
+		if err := auth.ValidateAuthority(callHdr.Authority); err != nil {
+			return nil, &NewStreamError{Err: status.Error(codes.Unavailable, fmt.Sprintf("failed to validate authority %s : %s", callHdr.Authority, err))}
```
easwars, line 764: Both the above comments here too. And use `%v` for `err`.

easwars (approved):
LGTM modulo minor nits in this last pass

PR #8071, opened 2025-02-10
transport: Send RST stream from the server when deadline expires

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/2886

This PR makes the server cancel an RPC when it sees the deadline expiring instead of waiting forn the client to cancel the stream. 

RELEASE NOTES:
* transport: Servers send an HTTP/2 RST_STREAM frame to cancel a stream when the deadline expires.


internal/grpctest/grpctest.go
```diff
@@ -58,11 +58,13 @@ func (Tester) Setup(t *testing.T) {
 	//  completely addressed, and this can be turned back on as soon as this issue is
 	//  fixed.
 	leakcheck.SetTrackingBufferPool(logger{t: t})
+	leakcheck.SetTimerTracker(logger{t: t})
 }
 
 // Teardown performs a leak check.
 func (Tester) Teardown(t *testing.T) {
 	leakcheck.CheckTrackingBufferPool()
+	leakcheck.CheckTimers(5 * time.Second)
```
dfawley, line 67: This is additive with the 10s for checking goroutines.  It might be nice to pass a context (or timer) to both of these instead, so they can share the 10s timeout?

internal/leakcheck/leakcheck.go
```diff
@@ -298,3 +285,123 @@ func NewLeakChecker(logger Logger) *LeakChecker {
 	SetTrackingBufferPool(logger)
 	return &LeakChecker{logger: logger}
 }
+
+type timerFactory struct {
+	logger Logger
+
+	originalTimeAfterFunc func(time.Duration, func()) internal.Timer
+	mu                    sync.Mutex
+	bufferCount           int
```
dfawley, line 294: Delete

internal/leakcheck/leakcheck.go
```diff
@@ -298,3 +285,123 @@ func NewLeakChecker(logger Logger) *LeakChecker {
 	SetTrackingBufferPool(logger)
 	return &LeakChecker{logger: logger}
 }
+
+type timerFactory struct {
+	logger Logger
+
+	originalTimeAfterFunc func(time.Duration, func()) internal.Timer
```
dfawley, line 292: Is there any need for this?  As opposed to just directly calling `time.AfterFunc`

internal/leakcheck/leakcheck.go
```diff
@@ -298,3 +285,123 @@ func NewLeakChecker(logger Logger) *LeakChecker {
...
+		originalTimeAfterFunc: internal.TimeAfterFunc,
+	}
+	internal.TimeAfterFunc = globalTimerTracker.timeAfterFunc
+}
+
+// CheckTimers undoes the effects of SetTimerTracker, and fails unit tests if
+// not all timers were cancelled or executed. It is invalid to invoke this
+// function without previously having invoked SetTimerTracker.
+func CheckTimers(timeout time.Duration) {
+	// Reset the internal function.
```
dfawley, line 356: This happens at the end?

internal/transport/http2_client.go
```diff
@@ -1238,7 +1238,7 @@ func (t *http2Client) handleRSTStream(f *http2.RSTStreamFrame) {
 	if statusCode == codes.Canceled {
 		if d, ok := s.ctx.Deadline(); ok && !d.After(time.Now()) {
 			// Our deadline was already exceeded, and that was likely the cause
-			// of this cancellation.  Alter the status code accordingly.
+			// of this cancellation. Alter the status code accordingly.
```
dfawley, line 1241: Please revert.

internal/leakcheck/leakcheck.go
```diff
@@ -298,3 +285,123 @@ func NewLeakChecker(logger Logger) *LeakChecker {
...
+
+func (t *trackingTimer) Stop() bool {
+	t.parent.remove(t.Timer)
+	return t.Timer.Stop()
+}
+
+// SetTimerTracker replaces internal.TimerAfterFunc  with a one that tracks
+// timer creations. CheckTimers should then be invoked at the end of the test to
+// validate that all timers created have either executed or are cancelled.
+func SetTimerTracker(logger Logger) {
```
dfawley, line 343: Can you add a test for this functionality?

internal/transport/http2_server.go
```diff
@@ -598,6 +599,15 @@ func (t *http2Server) operateHeaders(ctx context.Context, frame *http2.MetaHeade
...
 		t.idle = time.Time{}
 	}
+	// Start a timer to close the stream on reaching the deadline.
+	if timeoutSet {
+		timer := internal.TimeAfterFunc(timeout, func() {
+			t.closeStream(s, true, http2.ErrCodeCancel, false)
+		})
+		s.deadlineTimerCancel = func() {
+			timer.Stop()
+		}
```
dfawley, line 609:
```suggestion
		s.deadlineTimerCancel = timer.Stop
```

internal/transport/http2_server.go
```diff
@@ -1268,6 +1278,9 @@ func (t *http2Server) Close(err error) {
 	channelz.RemoveEntry(t.channelz.ID)
 	// Cancel all active streams.
 	for _, s := range streams {
+		if s.deadlineTimerCancel != nil {
+			s.deadlineTimerCancel()
+		}
```
dfawley, line 1283: Should we make sure this is always a valid function so we can just call it without the nil check?

internal/transport/http2_server.go
```diff
@@ -1324,7 +1341,13 @@ func (t *http2Server) closeStream(s *ServerStream, rst bool, rstCode http2.ErrCo
 	// called to interrupt the potential blocking on other goroutines.
 	s.cancel()
 
-	s.swapState(streamDone)
+	oldState := s.swapState(streamDone)
+	if oldState == streamDone {
```
dfawley, line 1345: Was this a bug before?  Or did we have something that prevented it?  Or was the code tolerant of multiple calls?

internal/transport/server_stream.go
```diff
@@ -34,9 +34,10 @@ import (
 type ServerStream struct {
 	*Stream // Embed for common stream functionality.
 
-	st      internalServerTransport
-	ctxDone <-chan struct{}    // closed at the end of stream.  Cache of ctx.Done() (for performance)
-	cancel  context.CancelFunc // invoked at the end of stream to cancel ctx.
+	st                  internalServerTransport
+	ctxDone             <-chan struct{}    // closed at the end of stream.  Cache of ctx.Done() (for performance)
+	cancel              context.CancelFunc // invoked at the end of stream to cancel ctx.
+	deadlineTimerCancel func()             // Invoked at the end of stream.
```
dfawley, line 40: Do we always call `cancel`?  Can we piggyback the timer cancellation with it if so?

internal/transport/http2_client.go
dfawley: It's not a typo.  :)

"Two spaces after periods" was recommended when typewriters were used, because they were fixed-width, and it helped with readability:

https://graphicdesign.stackexchange.com/questions/2160/typesetting-two-spaces-after-the-period

I prefer continuing the tradition with source code since it isn't rendered in a variable-width font (also it's a habit), but I am fine with people wanting to use one space.  We shouldn't go around changing it from one way to the other, though, as it creates churn and is just a preference, not a typo.

internal/leakcheck/leakcheck.go
```diff
@@ -298,3 +278,119 @@ func NewLeakChecker(logger Logger) *LeakChecker {
...
+	internal.Timer
+	parent *timerFactory
+}
+
+func (t *trackingTimer) Stop() bool {
+	t.parent.remove(t.Timer)
+	return t.Timer.Stop()
+}
+
+// TrackTimers replaces internal.TimerAfterFunc  with one that tracks timer
```
dfawley, line 329:
```suggestion
// TrackTimers replaces internal.TimerAfterFunc with one that tracks timer
```

(This one is a typo :stuck_out_tongue: )

internal/transport/http2_server.go
```diff
@@ -598,6 +599,22 @@ func (t *http2Server) operateHeaders(ctx context.Context, frame *http2.MetaHeade
...
+		timer := internal.TimeAfterFunc(timeout, func() {
+			<-cancelUpdated
+			t.closeStream(s, true, http2.ErrCodeCancel, false)
+		})
+		oldCancel := s.cancel
+		s.cancel = func() {
+			oldCancel()
+			timer.Stop()
+		}
+		close(cancelUpdated)
```
dfawley, line 616: Does this work?  `oldCancel` could be called multiple times, but I think that's OK if it's just doing a context cancelation.

```suggestion
		oldCancel := s.cancel
		timer := internal.TimeAfterFunc(timeout, func() {
			oldCancel()
			t.closeStream(s, true, http2.ErrCodeCancel, false)
		})
		s.cancel = func() {
			timer.Stop()
			oldCancel()
		}
```

arjan-bal: The data race happens when reading and writing `s.cancel`. The timer fires, calling `t.closeStream` which in turn calls `s.cancel`.  The race is still possible with this change because the timer is started in line 605 and s.cancel is updated afterwards in line 609. 

<details>
<summary>Stack trace</summary>

```
WARNING: DATA RACE
Read at 0x00c000cc40e0 by goroutine 30741:
  google.golang.org/grpc/internal/transport.(*http2Server).closeStream()
      grpc-go/internal/transport/http2_server.g
o:1337 +0x44
  google.golang.org/grpc/internal/transport.(*http2Server).operateHeaders.func4()
      grpc-go/internal/transport/http2_server.g
o:607 +0x57
  google.golang.org/grpc/internal/leakcheck.(*timerFactory).timeAfterFunc.func1()
      grpc-go/internal/leakcheck/leakcheck.go:2
92 +0x3c

Previous write at 0x00c000cc40e0 by goroutine 30738:
  google.golang.org/grpc/internal/transport.(*http2Server).operateHeaders()
      grpc-go/internal/transport/http2_server.g
o:609 +0x2e32
  google.golang.org/grpc/internal/transport.(*http2Server).HandleStreams()
      grpc-go/internal/transport/http2_server.g
o:680 +0x3fb
  google.golang.org/grpc.(*Server).serveStreams()
      grpc-go/server.go:1029 +0x67a
  google.golang.org/grpc.(*Server).handleRawConn.func1()
      grpc-go/server.go:964 +0x86

Goroutine 30741 (running) created at:
  time.goFunc()
      /usr/lib/google-golang/src/time/sleep.go:215 +0x44

Goroutine 30738 (running) created at:
  google.golang.org/grpc.(*Server).handleRawConn()
      grpc-go/server.go:963 +0x2e9
  google.golang.org/grpc.(*Server).Serve.func3()
      grpc-go/server.go:932 +0x7b
```
</details>

Running all the tests a couple of times should surface the failure. One way to avoid synchronization is to pass the cancellation function to `closeStream` as a parameter. The closure executed by the timee can pass `oldCancel`, existing callers can pass `s.cancel`.

internal/transport/http2_client.go
arjan-bal: Thanks for sharing, wasn't aware of this.

dfawley (approved):

dfawley (approved):

PR #8073, opened 2025-02-10
advancedtls: update CRL provider certificates

eshitachandwani:
RELEASE NOTES: N/A


security/advancedtls/testdata/crl/provider_client_trust_cert.srl
arjan-bal, line 1: You need to remove all the untracked files from your PR. There are temporary files generated by the scripts. The failing test is counting the number of files in the directory. This is causing the failure. I was able to get all the tests to pass locally after removing the temp files.

PR #8077, opened 2025-02-11
stats/openetelemetry: refactor and make e2e test stats verification deterministic

purnesh42H:
Fixes: https://github.com/grpc/grpc-go/issues/8053 https://github.com/grpc/grpc-go/issues/7423 #8076 

The flakiness in stats/opentelemetry e2e tests flakiness is related to asynchronous metrics and trace collection. gRPC server metrics, specifically those generated by `stats.End` calls, are processed asynchronously. This means that immediately after a client-side RPC completes, the corresponding server-side metrics may not be available yet. To handle this asynchronicity, the tests employ a polling mechanism that repeatedly checks for the expected metrics until they appear or a timeout occurs. This polling strategy ensures that the tests wait for the server to finish processing the metrics, preventing premature validation against incomplete data.

However, the following limitations are still there which are causing the flakiness
1. the validation logic for histogram metrics, particularly `sent_total_compressed_message_size`, did not adequately account for asynchronous data point population. The tests would sometimes check the histogram before all expected data points were present, resulting in false negatives. 
2. the validation logic for `duration` metrics did not use the above polling mechanism, specifically for histogram metrics.
3. the trace exporter was only queried once for collected spans, which could lead to missing spans due to the asynchronous nature of trace exporting. This resulted in incomplete trace verification and occasional test failures.

This PR fixes these flakiness issues by adding the appropriate polling before the metric and trace validation logic. 
1. the `waitForServerCompletedRPCs` now not only wait for required metric to be present but also wait for the number of histogram data points to match the expected values before proceeding with validation. 
2. The `duration` metrics validation check now also `waitForServerCompletedRPCs`.
3. the trace exporter is polled repeatedly until all expected spans are available, ensuring complete trace collection.

RELEASE NOTES: None


zasweq:
Sorry, lots of work on new team so no bandwidth to review this. Glad you worked on this though!

stats/opentelemetry/e2e_test.go
```diff
@@ -678,248 +666,8 @@ func (s) TestMetricsAndTracesOptionEnabled(t *testing.T) {
...
-					t.Errorf("Got attribute key for span name %q with event name %v, as %v, want %v", span.Name, event.Name, got, want)
-				}
-				if got, want := att.Value, wantSI[index].events[eventIdx].Attributes[idx].Value; got != want {
-					t.Errorf("Got attribute value for span name %v with event name %v, as %v, want %v", span.Name, event.Name, got, want)
-				}
-			}
-		}
-	}
+	wantSI := testutils.TraceDataWithCompressor()
+	testutils.VerifyAndCompareTraces(ctx, t, exporter, 6, wantSI)
```
arjan-bal, line 670: I don't like the idea of moving the expected output and the validation logic into a utils package. The expected output should be visible in the test function even if it results in duplication. This makes it easy for readers to catch bugs in the test logic. 

The style guide discourages the use of helpers for validation, however in grpc-go we still use them. I would suggest moving the validation helper into the same file.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -59,7 +74,16 @@ func waitForServerCompletedRPCs(ctx context.Context, t *testing.T, reader metric
...
 			continue
 		}
-		metricdatatest.AssertEqual(t, wantMetric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars())
+		if _, ok := val.Data.(metricdata.Histogram[int64]); ok {
+			if len(wantMetric.Data.(metricdata.Histogram[int64]).DataPoints) != len(val.Data.(metricdata.Histogram[int64]).DataPoints) {
+				continue
+			}
+		}
+		if _, ok := val.Data.(metricdata.Histogram[float64]); ok {
+			if len(wantMetric.Data.(metricdata.Histogram[float64]).DataPoints) != len(val.Data.(metricdata.Histogram[float64]).DataPoints) {
```
arjan-bal, line 83: I think a safer check would be to compare for `len(wantMetric.Data.(metricdata.Histogram[float64]).DataPoints) < len(val.Data.(metricdata.Histogram[float64]).DataPoints)`.

In failure cases when there are more metrics then expected, this would result in a validation error later instead of deadline exceeded error.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -59,7 +74,16 @@ func waitForServerCompletedRPCs(ctx context.Context, t *testing.T, reader metric
 		if !ok {
 			continue
 		}
-		metricdatatest.AssertEqual(t, wantMetric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars())
+		if _, ok := val.Data.(metricdata.Histogram[int64]); ok {
+			if len(wantMetric.Data.(metricdata.Histogram[int64]).DataPoints) != len(val.Data.(metricdata.Histogram[int64]).DataPoints) {
```
arjan-bal, line 78: Same comment as below, IMO it's better to check for `<` instead of `!=`.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -59,7 +74,16 @@ func waitForServerCompletedRPCs(ctx context.Context, t *testing.T, reader metric
 		if !ok {
 			continue
 		}
-		metricdatatest.AssertEqual(t, wantMetric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars())
```
arjan-bal, line 62: Should this be added just before the return statement below?

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -766,20 +790,25 @@ func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader,
...
 			// stats.End show up.
 			var err error
-			if gotMetrics, err = waitForServerCompletedRPCs(ctx, t, mr, metric); err != nil { // move to shared helper
+			if gotMetrics, err = waitForServerCompletedRPCs(ctx, mr, metric); err != nil { // move to shared helper
 				t.Fatal(err)
 			}
+			val := gotMetrics[metric.Name]
+			if !metricdatatest.AssertEqual(t, metric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars()) {
+				t.Fatalf("Metrics data type not equal for metric: %v", metric.Name)
+			}
```
arjan-bal, line 799: Is there any benefit of moving this out of `waitForServerCompletedRPCs`? If not, I think we should keep it the way it was earlier.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -766,20 +790,25 @@ func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader,
...
-		val, ok := gotMetrics[metric.Name]
-		if !ok {
-			t.Fatalf("Metric %v not present in recorded metrics", metric.Name)
-		}
 		if metric.Name == "grpc.client.attempt.duration" || metric.Name == "grpc.client.call.duration" || metric.Name == "grpc.server.call.duration" {
+			var err error
+			if gotMetrics, err = waitForServerCompletedRPCs(ctx, mr, metric); err != nil { // move to shared helper
+				t.Fatal(err)
+			}
+			val := gotMetrics[metric.Name]
```
arjan-bal, line 811: IIUC the intended use of `waitForServerCompletedRPCs` is to wait for metrics that mark the end of an RPC. If we need to wait for metrics again, does it mean there are metrics that come after metrics that we consider terminal? If so, is this a bug?

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -766,20 +790,25 @@ func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader,
...
 		// If one of the duration metrics, ignore the bucket counts, and make
 		// sure it count falls within a bucket <= 5 seconds (maximum duration of
 		// test due to context).
-		val, ok := gotMetrics[metric.Name]
-		if !ok {
-			t.Fatalf("Metric %v not present in recorded metrics", metric.Name)
-		}
 		if metric.Name == "grpc.client.attempt.duration" || metric.Name == "grpc.client.call.duration" || metric.Name == "grpc.server.call.duration" {
+			var err error
+			if gotMetrics, err = waitForServerCompletedRPCs(ctx, mr, metric); err != nil { // move to shared helper
```
arjan-bal, line 808: We are changing the metrics that are passed as params to the function. So we are not actually comparing the values passed. This seems like unexpected behaviour.

stats/opentelemetry/e2e_test.go
purnesh42H: I had just followed the same approach as done for metrics comparison. I have moved the tracing validation and expected data, which was added in this PR, to e2e_test but not the metrics.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: I don't think the way waitForServerCompletedRPCs is written right now is suitable for failure cases or any situation where the number of expected data points might be less than the number of data points actually collected because then there are higher chances of false positives leading to flakiness again.

I actually tried < instead of != but that again bring back the same flakiness for histogram metrics. I haven't fully gone through the assertion and equality logic but looks like waiting for the exact number of data points is less prone to race from async metrics collection.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: It had to be moved because now we are using for both `duration` and `sent_total_compressed_message_size` metric and assertion option for `duration` is different

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: Replied above

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: We are not waiting for same metrics again. `TestMethodAttributeFilter` doesn't have `sent_total_compressed_message_size` in wantMetrics but have `duration`. Hence, we need to have same for `duration` metrics as well. If both are present, the `duration` one will probably return first time itself.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: Replied above. If `sent_total_compressed_message_size` is present in wantMetric then second call for `duration` metrics will return first time itself without changing anything as we would have already waited for metric collection after `stats.End`

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -789,8 +818,554 @@ func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader,
...
+// Returns the collected spans or an error if the context deadline is exceeded
+// before the expected number of spans are exported.
+func waitForCompleteTraceSpans(ctx context.Context, exporter *tracetest.InMemoryExporter, wantSpans int) (tracetest.SpanStubs, error) {
+	var spans []tracetest.SpanStub
+	for ; ctx.Err() == nil; <-time.After(time.Millisecond) {
+		spans = exporter.GetSpans()
+		if len(spans) == wantSpans {
+			return spans, nil
+		}
+	}
```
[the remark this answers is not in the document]
purnesh42H: Actually the current wait logic was not complete. We can't just rely on number of expected spans to be equal. We actually should poll until we get all expected spans by name and span kind. In this case the number of expected spans worked because that's the total we actually emit. However, if we were emitting more, relying on number of expected spans to be equal wouldn't have worked.

I have modified `waitForTraceSpans` to wait until the in-memory span exporter has received the
// expected trace spans based on span name and kind

stats/opentelemetry/e2e_test.go
```diff
@@ -1569,9 +839,16 @@ func (s) TestRPCSpanErrorStatus(t *testing.T) {
...
+	var spans []tracetest.SpanStub
+	for ; ctx.Err() == nil; <-time.After(time.Millisecond) {
+		spans = exporter.GetSpans()
+		if len(spans) == wantSpans {
+			break
+		}
+	}
+	if len(spans) != wantSpans {
+		t.Fatalf("got %d spans, want %d", len(spans), wantSpans)
 	}
```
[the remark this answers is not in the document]
purnesh42H: Not applicable now. For this case we just need to wait until the first span has been collected because we are throwing the error in the first step of tracing itself.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -789,8 +818,554 @@ func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader,
...
 }
+
+// waitForCompleteTraceSpans waits until the in-memory span exporter has
+// received all the expected number of spans.  It polls the exporter at a short
+// interval until the desired number of spans are available or the context is
+// cancelled.
+//
+// Returns the collected spans or an error if the context deadline is exceeded
+// before the expected number of spans are exported.
+func waitForCompleteTraceSpans(ctx context.Context, exporter *tracetest.InMemoryExporter, wantSpans int) (tracetest.SpanStubs, error) {
```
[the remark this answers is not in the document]
purnesh42H: Done

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -789,8 +818,554 @@ func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader,
...
+			if streamServer != nil && streamServer.Parent.SpanID() != streamAttempt.SpanContext.SpanID() {
+				t.Error("Streaming server span parent does not match attempt span ID")
+			}
+		}
+	}
+
+	compareTraces(t, spans, wantSpanInfos)
+}
+
+func compareTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []TraceSpanInfo) {
```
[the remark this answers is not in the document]
purnesh42H: Not as is i think. We would need to add Equal method to the traceSpanInfo which would be the same code as here.

stats/opentelemetry/e2e_test.go
```diff
@@ -143,102 +143,662 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+		return nil, fmt.Errorf("error waiting for complete trace spans %v: %v", wantSpans, ctx.Err())
+	}
+	return exporter.GetSpans(), nil
+}
 
-	wantMetrics := []metricdata.Metrics{
+// traceDataWithCompressor returns a traceSpanInfo for a unary RPC and
+// streaming RPC with certain compression and message flow sent, when
+// compressor is used
+func traceDataWithCompressor() []traceSpanInfo {
```
arjan-bal, line 185: Can this be inlined, like how it was before? The input and output are important parts of the test that should not be abstracted.

stats/opentelemetry/e2e_test.go
```diff
@@ -82,6 +78,10 @@ type traceSpanInfo struct {
 	attributes []attribute.KeyValue
 }
 
+func Test(t *testing.T) {
+	grpctest.RunSubTests(t, s{})
+}
+
```
arjan-bal, line 84: Can this be reverted? It adds an extra diff to the PR which isn't required.

stats/opentelemetry/e2e_test.go
```diff
@@ -143,102 +143,662 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
-// OpenTelemetry SDK's Meter Provider. It then makes an RPC that is unregistered
-// on the Client (no StaticMethodCallOption set) and Server. The method
-// attribute on subsequent metrics should be bucketed in "other".
-func (s) TestAllMetricsOneFunction(t *testing.T) {
-	mo, reader := defaultMetricsOptions(t, nil)
-	ss := setupStubServer(t, mo, nil)
+// traceDataWithoutCompressor returns a traceSpanInfo for a unary RPC and
+// streaming RPC with certain compression and message flow sent, when
+// compressor is not used.
+func traceDataWithoutCompressor() []traceSpanInfo {
```
arjan-bal, line 367: Can this be inlined, like how it was before? The input and output are important parts of the test that should not be abstracted.

stats/opentelemetry/internal/testutils/testutils.go
arjan-bal: Are the `sent_total_compressed_message_size` always emitted after the duration metrics? If so, that is probably the reason this function was only waiting for the compression metric.

stats/opentelemetry/internal/testutils/testutils.go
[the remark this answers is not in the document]
arjan-bal: When using [cmp.Equal](https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal), you can pass a [Comparer](https://pkg.go.dev/github.com/google/go-cmp/cmp#Comparer) as an option which specifies how to compare objects of a particular type. This means that you only need to specify how to compare structs that have special comparison logic and cmp will take care of the rest. You can also pass options to ignore unexported fields or specify fields to ignore.

stats/opentelemetry/internal/testutils/testutils.go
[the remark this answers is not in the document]
arjan-bal: Are we expected to emit more spans? Should the test assert we get exactly the specified spans as sending extra spans could be a regression?

stats/opentelemetry/internal/testutils/testutils.go
arjan-bal: > I don't think the way waitForServerCompletedRPCs is written right now is suitable for failure cases or any situation where the number of expected data points might be less than the number of data points actually collected.

This is expected because the job of the function is to wait and not validate the metrics. Validation should be done at the end of the test.

purnesh42H: How will you wait for the metric that is not expected in a failure test case? I don't think it was intended for failure test cases the way the it is written right now.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: It could be but TestMethodAttributeFilter doesn't have `sent_total_compressed_message_size` in wantMetric, so we have to wait for duration metrics as well.

stats/opentelemetry/internal/testutils/testutils.go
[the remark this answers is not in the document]
purnesh42H: As of now we are verifying all the spans that are being emitted. In future, we could add more trace spans to be emitted for some cases. However, not all test cases will expect them. So, for a test to wait for only expected spans to collect is fine IMO even if more spans are being emitted by grpc.

> Should the test assert we get exactly the specified spans as sending extra spans could be a regression?
That's good thought but it might become hard to maintain long term when we add more spans because everytime we add a new span we have to change all test cases. We don't do that for metrics as well.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: So, i moved the waiting loop to another helper and call it explicitly before comparing at the start of `CompareMetrics`. I think that's a good middle ground to avoid every test to call both wait and compare everytime because the docstring of `CompareMetrics` already state that it waits for eventual metrics.

In the comparison, I changed the `t.FatalF` to `t.ErrorF` so that test doesn't stop just after comparing failed metric and we get to see all comparison failures because we have already waited for all server metrics.

stats/opentelemetry/internal/testutils/testutils.go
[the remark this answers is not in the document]
purnesh42H: I tried writing it but its not saving that much code lines. We have to write comparer for event and attributes. Then we need to make sure that both got and want are same type and the downside is compromising the readability of error output from `cmp.Diff`. With current code, we explicitly say which part of metric failed comparison which is more readable. Metrics are also being compared in similar fashion. I am guessing the main motivation is to have better readability.

stats/opentelemetry/e2e_test.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +143,202 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+	for _, span := range spans {
+		// Check that the attempt span has the correct status
+		if got, want := span.Status.Code, otelcodes.Ok; got != want {
+			t.Errorf("Got status code %v, want %v", got, want)
+		}
+
+		var want traceSpanInfo
+		switch {
+		case span.Name == "grpc.testing.TestService.UnaryCall" && span.SpanKind == oteltrace.SpanKindServer:
+			want = wantSpanInfos[0] // Reference expected unary server span
```
arjan-bal, line 289: Using integers to encode `span.Name ` and `span.SpanKind` is manual and error prone. Can we instead change `wantSpanInfos` to map[keyType] traceSpanInfo` where `keyType` is a struct that holds name and spanKind?
```go
type keyType struct {
  name string
  spanKind int
}
```

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +143,202 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+			}
+		}
+	}
+
+	compareTraces(t, spans, wantSpanInfos)
+}
+
+func compareTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []traceSpanInfo) {
+	// Validate attributes/events by span type instead of index
+	for _, span := range spans {
```
arjan-bal, line 280: Can we convert `spans` into a `map[keyType] traceSpanInfo`? Then we can do a `cmp.Diff` to compare two objects of the same type.

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +143,202 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+}
+
+// verifyAndCompareTraces first waits for the exporter to receive the expected
+// number of spans. It then groups the received spans by their TraceID. For
+// each trace group, it identifies the client, server, and attempt spans for
+// both unary and streaming RPCs. It checks that the expected spans are
+// present and that the server spans have the correct parent (attempt span).
+// Finally, it compares the content of each span (name, kind, attributes,
+// events) against the provided expected spans information.
+func verifyAndCompareTraces(ctx context.Context, t *testing.T, exporter *tracetest.InMemoryExporter, wantSpanInfos []traceSpanInfo) {
```
arjan-bal, line 189: `verify` and `compare` seem to be redundant terms. Can we call it `validateTraces`?

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +143,202 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+// expected trace spans based on span name and kind. It polls the exporter at a
+// short interval until the desired spans are available or the context is
+// cancelled.
+//
+// Returns the collected spans or an error if the context deadline is exceeded
+// before the expected spans are exported.
+func waitForTraceSpans(ctx context.Context, exporter *tracetest.InMemoryExporter, wantSpans []traceSpanInfo) (tracetest.SpanStubs, error) {
+	for ; ctx.Err() == nil; <-time.After(time.Millisecond) {
+		spans := exporter.GetSpans()
+		allFound := true
```
arjan-bal, line 156: This can be simplified by removing the `allFound` variable. In the inner loop we can track if any span is missing.

```go
	for ; ctx.Err() == nil; <-time.After(time.Millisecond) {
		spans := exporter.GetSpans()

		for _, wantSpan := range wantSpans {
			missingAnySpan := false
			for _, span := range spans {
				if span.Name == wantSpan.name && span.SpanKind.String() == wantSpan.spanKind {
					continue
				}
				missingAnySpan = true
				break
			}
			if !missingAnySpan {
				return spans, nil
			}
		}
	}
	return nil, fmt.Errorf("error waiting for complete trace spans %v: %v", wantSpans, ctx.Err())
```

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +143,202 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+// each trace group, it identifies the client, server, and attempt spans for
+// both unary and streaming RPCs. It checks that the expected spans are
+// present and that the server spans have the correct parent (attempt span).
+// Finally, it compares the content of each span (name, kind, attributes,
+// events) against the provided expected spans information.
+func verifyAndCompareTraces(ctx context.Context, t *testing.T, exporter *tracetest.InMemoryExporter, wantSpanInfos []traceSpanInfo) {
+	spans, err := waitForTraceSpans(ctx, exporter, wantSpanInfos)
+	if err != nil {
+		t.Fatal(err)
+	}
```
arjan-bal, line 193: nit: I think it's better if the call to `waitForTraceSpans` is moved out of this function and the resultant `spans` are passed as an argument. This makes it clearer that `verifyAndCompareTraces` validates static data and `waitForTraceSpans` waits for data to be avaiable.

stats/opentelemetry/e2e_test.go
```diff
@@ -937,12 +1068,12 @@ func (s) TestMetricsAndTracesOptionEnabled(t *testing.T) {
...
 	mo, _ := defaultMetricsOptions(t, nil)
 	// Using defaultTraceOptions to set up OpenTelemetry with an in-memory exporter.
-	to, spanExporter := defaultTraceOptions(t)
+	to, exporter := defaultTraceOptions(t)
 	// Start the server with trace options.
 	ss := setupStubServer(t, mo, to)
 	defer ss.Stop()
 
-	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), 100*defaultTestTimeout)
```
arjan-bal, line 1076: 500 seconds is too much time to wait. Do we really need this much time?

stats/opentelemetry/e2e_test.go
```diff
@@ -678,11 +874,6 @@ func (s) TestMetricsAndTracesOptionEnabled(t *testing.T) {
 	testutils.CompareMetrics(ctx, t, reader, gotMetrics, wantMetrics)
 
 	// Verify traces
```
arjan-bal, line 876: nit: This comment can be removed, the function name is self explanatory.

stats/opentelemetry/e2e_test.go
```diff
@@ -962,12 +1093,7 @@ func (s) TestSpan(t *testing.T) {
 		t.Fatalf("stream.Recv received an unexpected error: %v, expected an EOF error", err)
 	}
 
-	// Get the spans from the exporter
-	spans := spanExporter.GetSpans()
-	if got, want := len(spans), 6; got != want {
-		t.Fatalf("got %d spans, want %d", got, want)
-	}
-
+	// Verify traces
```
arjan-bal, line 1096: nit: This comment can be removed, the function name is self explanatory.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -59,7 +59,16 @@ func waitForServerCompletedRPCs(ctx context.Context, t *testing.T, reader metric
...
-		metricdatatest.AssertEqual(t, wantMetric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars())
+		if _, ok := val.Data.(metricdata.Histogram[int64]); ok {
+			if len(wantMetric.Data.(metricdata.Histogram[int64]).DataPoints) > len(val.Data.(metricdata.Histogram[int64]).DataPoints) {
+				continue
+			}
+		}
+		if _, ok := val.Data.(metricdata.Histogram[float64]); ok {
+			if len(wantMetric.Data.(metricdata.Histogram[float64]).DataPoints) > len(val.Data.(metricdata.Histogram[float64]).DataPoints) {
+				continue
+			}
```
arjan-bal, line 70: nit: Can use a type switch instead of `if` statements:
```go
switch data := val.Data.(type) {
case metricdata.Histogram[int64]:
	if len(wantMetric.Data.(metricdata.Histogram[int64]).DataPoints) > len(data.DataPoints) {
		continue
	}
case metricdata.Histogram[float64]:
	if len(wantMetric.Data.(metricdata.Histogram[float64]).DataPoints) > len(data.DataPoints) {
		continue
	}
}
```

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -760,37 +769,66 @@ func MetricData(options MetricDataOptions) []metricdata.Metrics {
...
+		if metric.Name == "grpc.server.call.sent_total_compressed_message_size" || metric.Name == "grpc.server.call.rcvd_total_compressed_message_size" {
+			// Sync the metric reader to see the event because stats.End is
+			// handled async server side. Thus, poll until metrics created from
+			// stats.End show up.
+			var err error
+			if gotMetrics, err = waitForServerCompletedRPCs(ctx, mr, metric); err != nil { // move to shared helper
+				t.Fatal(err)
+			}
+			continue
+		}
```
arjan-bal, line 820: Both these if blocks look similar. We can create a list of metric names and check if metric.Name exists in the list:
```go
	terminalMetrics := []string{
		"grpc.server.call.sent_total_compressed_message_size",
		"grpc.server.call.rcvd_total_compressed_message_size",
		"grpc.client.attempt.duration",
		"grpc.client.call.duration",
		"grpc.server.call.duration",
	}
	for _, metric := range wantMetrics {
		if slices.Contains(terminalMetrics, metric.Name) {
			// Sync the metric reader to see the event because stats.End is
			// handled async server side. Thus, poll until metrics created from
			// stats.End show up.
			var err error
			if gotMetrics, err = waitForServerCompletedRPCs(ctx, mr, metric); err != nil { // move to shared helper
				t.Fatal(err)
			}
			continue
		}
	}
```

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -760,37 +769,66 @@ func MetricData(options MetricDataOptions) []metricdata.Metrics {
 // and for duration metrics makes sure the data point is within possible testing
 // time (five seconds from context timeout).
 func CompareMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader, gotMetrics map[string]metricdata.Metrics, wantMetrics []metricdata.Metrics) {
+	gotMetrics = waitForEventualServerMetrics(ctx, t, mr, gotMetrics, wantMetrics)
```
arjan-bal, line 772: nit: Can we moved this call to `waitForEventualServerMetrics` to the callers of `CompareMetrics`. This makes the distinction about between `action` and `assert`  parts of the test clearer.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -760,37 +769,66 @@ func MetricData(options MetricDataOptions) []metricdata.Metrics {
...
 		if !metricdatatest.AssertEqual(t, metric, val, metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreExemplars()) {
-			t.Fatalf("Metrics data type not equal for metric: %v", metric.Name)
+			t.Errorf("Metrics data type not equal for metric: %v", metric.Name)
 		}
 	}
 }
+
+// waitForEventualServerMetrics waits for eventual server metrics (not emitted
+// synchronously with client side rpc returning).
+func waitForEventualServerMetrics(ctx context.Context, t *testing.T, mr *metric.ManualReader, gotMetrics map[string]metricdata.Metrics, wantMetrics []metricdata.Metrics) map[string]metricdata.Metrics {
```
arjan-bal, line 809: nit: `Eventual` in the function name seems redundant. IMO, it can be omitted.

stats/opentelemetry/internal/testutils/testutils.go
arjan-bal: This is better, thanks.

stats/opentelemetry/e2e_test.go
purnesh42H: Added the struct and code to create map of traceSpanInfos in validateTrace. Then retrieve expected span based on name and span kind

stats/opentelemetry/e2e_test.go
purnesh42H: Done as mentioned above. Regarding `cmp.Diff` I mentioned before that comparer is getting too complicated because of Value field in attribute. I am keeping the check simple for now similar as metrics.

However, I did find that we were not comparing the attributes values before because the bools were set as int64s in wants. I changed the wants to have correct value types and added check for comparing attribute values as well.

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +143,202 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+			if streamServer != nil && streamServer.Parent.SpanID() != streamAttempt.SpanContext.SpanID() {
+				t.Error("Streaming server span parent does not match attempt span ID")
+			}
+		}
+	}
+
+	compareTraces(t, spans, wantSpanInfos)
+}
+
+func compareTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []traceSpanInfo) {
```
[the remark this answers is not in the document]
purnesh42H: Moved within validateTraces

stats/opentelemetry/e2e_test.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
purnesh42H: This seems to spin forever? Was getting timeouts. I think its order agnostic. We need to check for wantSpan in entire span list and only then call it missing.

I think the current one is also simple enough. Ofcourse there can be simpler ways to write it but not a big deal for testing code.

stats/opentelemetry/e2e_test.go
purnesh42H: Moved. Although i don't feel strongly about this because the docstring was mentioning that.

stats/opentelemetry/e2e_test.go
purnesh42H: had changed for some debugging. Forgot to remove. Removed now.

stats/opentelemetry/e2e_test.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
```diff
@@ -1568,13 +1571,11 @@ func (s) TestRPCSpanErrorStatus(t *testing.T) {
...
 	}})
 
-	// Verify traces
-	spans := exporter.GetSpans()
-	if got, want := len(spans), 3; got != want {
-		t.Fatalf("got %d spans, want %d", got, want)
-	}
-
 	// Verify spans has error status with rpcErrorMsg as error message.
+	for ; len(exporter.GetSpans()) == 0 && ctx.Err() == nil; <-time.After(time.Millisecond) {
```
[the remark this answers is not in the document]
purnesh42H: Was not needed as we are only checking the first span anyway. The error is enforced on first event itself.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: Done. Thanks for suggestion.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: Moved. Although i don't feel strongly about this because the docstring was mentioning that.

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
purnesh42H: Actually nvm about `cmp.Diff`. After some more modifications, I was able to get it working. Changed both attributes and events to cmp.Diff

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -755,42 +765,65 @@ func MetricData(options MetricDataOptions) []metricdata.Metrics {
...
 			if err := checkDataPointWithinFiveSeconds(val); err != nil {
-				t.Fatalf("Data point not within five seconds for metric %v: %v", metric.Name, err)
+				t.Errorf("Data point not within five seconds for metric %v: %v", metric.Name, err)
 			}
 			continue
 		}
 
+		val, ok := gotMetrics[metric.Name]
+		if !ok {
+			t.Errorf("Metric %v not present in recorded metrics", metric.Name)
```
arjan-bal, line 797: I think we `continue` after this check fails.

stats/opentelemetry/e2e_test.go
arjan-bal: > This seems to spin forever? 

I didn't actually run the code earlier. The declaration for `missingAnySpans` should have been in the outer scope. I update the code and tested it verified that it works.

> I think the current one is also simple enough. Ofcourse there can be simpler ways to write it but not a big deal for testing code.

The number of nested if blocks makes the code difficult to follow. This complexity increases the risk of bugs in the tests. Since we don't have tests for our tests, it's crucial that the code remains simple enough for readers to easily verify its correctness.

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +152,174 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+		})
+		eventsTimeIgnore := cmpopts.IgnoreFields(trace.Event{}, "Time")
+
+		// attributes
+		if diff := cmp.Diff(want.attributes, span.Attributes, attributesSort, attributesValueIgnoreUnexported); diff != "" {
+			t.Errorf("Attributes mismatch for span %s (-want +got):\n%s", span.Name, diff)
+		}
+		// events
+		if diff := cmp.Diff(want.events, span.Events, eventsSort, attributesSort, attributesValueIgnoreUnexported, eventsTimeIgnore); diff != "" {
+			t.Errorf("Events m mismatch for span %s (-want +got):\n%s", span.Name, diff)
```
arjan-bal, line 318: nit: Extra "m" in "Events m mismatch".

stats/opentelemetry/e2e_test.go
purnesh42H: Yeah the updated code works. Updated.

stats/opentelemetry/internal/testutils/testutils.go
```diff
@@ -755,42 +765,65 @@ func MetricData(options MetricDataOptions) []metricdata.Metrics {
...
+		}
```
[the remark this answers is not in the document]
purnesh42H: Moved to the top

stats/opentelemetry/internal/testutils/testutils.go
purnesh42H: Moved to the top with continue

stats/opentelemetry/e2e_test.go
purnesh42H: Done

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +152,174 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+		if !ok {
+			t.Errorf("Unexpected span: %v", span)
+			continue
+		}
+
+		// comparers
+		attributesSort := cmpopts.SortSlices(func(a, b attribute.KeyValue) bool {
+			return a.Key < b.Key
+		})
+		attributesValueIgnoreUnexported := cmpopts.IgnoreUnexported(attribute.KeyValue{}.Value)
```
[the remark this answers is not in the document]
arjan-bal: All the fields in `attribute.Value` are unexported, so ignoring unexported fields was the same as ignoring `attribute.Value`.

stats/opentelemetry/e2e_test.go
```diff
@@ -143,6 +153,163 @@ func setupStubServer(t *testing.T, metricsOptions *opentelemetry.MetricsOptions,
...
+		}
+
+		// comparers
+		attributesSort := cmpopts.SortSlices(func(a, b attribute.KeyValue) bool {
+			return a.Key < b.Key
+		})
+		attributesValueComparable := cmpopts.EquateComparable(attribute.KeyValue{}.Value)
+		eventsSort := cmpopts.SortSlices(func(a, b trace.Event) bool {
+			return a.Name < b.Name
+		})
```
arjan-bal, line 299: Is it correct to ignore the order of events during comparison? From the test case, I see that the expected order for unary clients is:
1. Outbound compressed message
2. Inbound compressed message
For unary servers the event order is:
1. Inbound compressed message
2. Outbound compressed message

This seems to follow the order of messages being exchanged b/w the client and servers. So it may be worth verifying the event order.

stats/opentelemetry/e2e_test.go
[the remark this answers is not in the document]
arjan-bal: It would change the test behaviour though. Earlier the test was asserting that exactly 3 spans are received and the first span's status message is as expected. Now it's just asserting the second part. Is it okay to skip checking the span count?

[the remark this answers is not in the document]
purnesh42H: I really don't see any reason unless we are verifying what those spans are. 3 seem to be an arbitrary number. That's why I modified the other tests to go through all emitted spans instead of 6 because in case we add more spans in future, we want the current tests to take care of that. So, other success test cases are already verifying the expected spans. For this test case we are only interested in the error being present in first rpc attempt span from client.

stats/opentelemetry/e2e_test.go
purnesh42H: I don't think it matters that much because the event names are pretty explicit and they are determined by the order of the top level stats event which are InPayload and OutPayload. However, i tested without eventSort and tests don't flake so i went ahead and removed it.

arjan-bal (approved):
LGTM

stats/opentelemetry/trace.go
```diff
@@ -46,7 +46,7 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
 		// correctness.
 		span.SetAttributes(
 			attribute.Bool("Client", rs.Client),
-			attribute.Bool("FailFast", rs.Client),
+			attribute.Bool("FailFast", rs.FailFast),
```
arjan-bal, line 49: Is this a bugfix? If yes, should we have a release note and test, possibly in a separate PR?

purnesh42H: I think we never released the opentelemetry tracing api anyways. This will be the first time. However, it looks like the release notes were not picked up for it https://github.com/grpc/grpc-go/releases

PR #8084, opened 2025-02-12
Add xds host_rewrite_literal support

rob05c:
Adds support to `grpc-go` for xDS [host_rewrite_literal](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-routeaction).

This PR is intended to inform a forthcoming RFC.

This doesn't affect the public interface. All symbols added are internal, and to avoid changing any existing behaviour the stream host is not overridden in any other scenario, via an internal context type only set when host_rewrite_literal is received from RDS.

We've tested this internally. The xDS control plane sends the `host_rewrite_literal` in RDS (which we incidentally use for Envoy xDS in a very similar fashion as our grpc-go clients), and with this changeset the clients make a request to the upstream cluster with the correct host, and everything works as-expected.

See https://github.com/grpc/grpc/issues/38062 for the motivation. TLDR we have an Envoy instance acting as a Reverse Proxy, for which clients need to receive and set the host from xDS in their request (via host_rewrite_literal). This host differs from the reverse proxy's own host in the Cluster, and there are numerous upstream hosts it forwards to, without this the Reverse Proxy doesn't know where to send the request (relatively normal/common Reverse Proxy behaviour).

### DIFF
--- a/internal/serviceconfig/serviceconfig.go
+++ b/internal/serviceconfig/serviceconfig.go
@@ -149,6 +149,8 @@ type MethodConfig struct {
 	MaxRespSize *int
 	// RetryPolicy configures retry options for the method.
 	RetryPolicy *RetryPolicy
+
+	Host string
 }
 
 // RetryPolicy defines the go-native version of the retry policy defined by the
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -733,6 +733,10 @@ func (e NewStreamError) Error() string {
 	return e.Err.Error()
 }
 
+// CtxValAllowHostOverride can be added to the context given to NewStream with
+// a bool value of true to allow the CallHdr host to override the server name.
+type CtxValAllowHostOverride struct{}
+
 // NewStream creates a stream and registers it into the transport as "active"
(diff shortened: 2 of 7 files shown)


eshitachandwani:
Hey , @rob05c , thank you for the PR , can you open an issue for the same and link this to it to have a discussion there. We generally do not encourage draft PRs. Thank you!

internal/serviceconfig/serviceconfig.go
```diff
@@ -149,6 +149,8 @@ type MethodConfig struct {
 	MaxRespSize *int
 	// RetryPolicy configures retry options for the method.
 	RetryPolicy *RetryPolicy
+
+	Host string
```
atollena, line 153: I don't think we'll want to expose this in the service config. Have you considered using the `iresolver.RPCConfig` context to pass this information along instead?

atollena:
> Hey , @rob05c , thank you for the PR , can you open an issue for the same and link this to it to have a discussion there. We generally do not encourage draft PRs. Thank you!

We plan to open a gRFC for this, see https://github.com/grpc/grpc/issues/38062.

PR #8100, opened 2025-02-18
chore: enable early-return and unnecessary-stmt and useless-break from revive

mmorel-35:
#### Description

This add early-return and unnecessary-stmt and useless-break rules and fix them as they were mostly applied.
It also provide "preserveScope" argument for indent-error-flow and superfluous-else rules to avoid suggestions of refactorings that would increase variable scope.

RELEASE NOTES: None

### DIFF
--- a/scripts/revive.toml
+++ b/scripts/revive.toml
@@ -1,33 +1,60 @@
 # Enabled rules
 [rule.blank-imports]
+
 [rule.context-as-argument]
+
 [rule.context-keys-type]
+
 [rule.dot-imports]
+
 [rule.errorf]
+
 [rule.error-return]
+
 [rule.error-strings]
+
 [rule.error-naming]
+
 [rule.exported]
+
 [rule.increment-decrement]
+
 [rule.indent-error-flow]
+    arguments = ["preserveScope"]
+
 [rule.package-comments]
+
 [rule.range]
+
 [rule.receiver-naming]
+
 [rule.superfluous-else]
+    arguments = ["preserveScope"]
+
 [rule.time-naming]
-[rule.var-naming]
+
 [rule.unexported-return]
-[rule.unused-parameter]
+
+[rule.unnecessary-stmt]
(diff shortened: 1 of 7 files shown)


examples/features/metadata/client/main.go
```diff
@@ -71,15 +71,14 @@ func unaryCallWithMetadata(c pb.EchoClient, message string) {
...
-		}
-	} else {
+	t, ok := trailer["timestamp"]
+	if !ok {
 		log.Fatal("timestamp expected but doesn't exist in trailer")
 	}
+	fmt.Printf("timestamp from trailer:\n")
+	for i, e := range t {
+		fmt.Printf(" %d. %s\n", i, e)
+	}
```
arjan-bal, line 81: This refactoring doesn't increase the scope of `t, ok` immediately. However, when more more code is added at the end of the function, `t, ok` would be accessible. With the `if t, ok := trailer["timestamp"]; ok` this wouldn't happen. 

I don't think the benefit of this rule is worth the risk of causing a potential bug because of accidentally accessing a variable define higher in the function.

scripts/revive.toml
```diff
@@ -1,33 +1,61 @@
...
+
 [rule.use-any]
 
-# Disabled rules
-[rule.empty-block] # Disabled to allow intentional no-op blocks (e.g., channel draining).
-    Disabled = true
-[rule.import-shadowing] # Disabled to allow intentional reuse of variable names that are the same as package imports.
-    Disabled = true
-[rule.redefines-builtin-id] # Disabled to allow intentional reuse of variable names that are the same as built-in functions.
-    Disabled = true
```
arjan-bal, line 31: Can we keep the disabled rules together?

arjan-bal (commented):
Thanks for sending the PR!

arjan-bal (approved):

easwars (approved):

Issue #8119, opened 2025-02-25
Cardinaltiy violation in stream interface implementation


eshitachandwani:
There are 2 issues : 

1. If we have a client streaming RPC , and the server doesn't send a message before returning status `OK` we expect an cardinality violation error and and `INTERNAL` status code , but currently we receive `EOF` if this happens.

Issue #7570, opened 2024-08-29
Create tests for two potential cardinality violation bugs


dfawley:
When looking into the generic stream interface documentation, I noticed a couple _potential_ bugs in our code through inspection:

1. Client should ensure a cardinality violation error is returned for client-streaming RPCs if the server doesn't send a message before returning status OK.

Test case: Using stubserver, create a client-streaming RPC handler (will need to be added similar to the unary/bidi streaming handlers) that just does `return nil` without first calling `SendAndClose`.  This should result in the client receiving an error when making this call and calling `CloseAndRecv`.  Also in this case, we should receive `Error` log messages indicating what happened and that it is illegal behavior.  Client streaming RPCs must terminate with either no message and a non-OK status, or a message and an OK status.

We can also test the reverse case where it calls `SendAndClose` and returns an error from the handler that results in an RPC status.  This is not necessarily "wrong" since `SendAndClose` itself could return an error, but the behavior of the RPC should be that the RPC is terminated when `SendAndClose` is called, so the return value of the handler should be ignored.

2. Also for client-streaming RPCs, when `SendAndClose` is called, the server should perform a RST_STREAM() after sending the response message successfully.  We should confirm this is happening.

Test case: Using stubserver, create a client-streaming RPC handler (like above) that calls `SendAndClose` and then calls `Recv`.  Those operations should fail with an error indicating the stream was already closed.  On the client side, just send messages continuously.  An error should eventually propagate to the client's `Send` calls.

Implementation plan:

Check if similar tests exist first.  Then create these tests if not.  If they fail, let's `t.Skip` them and check them in, and then file issues and fix the bugs and re-enable the tests.

dfawley:
(2) is actually tricky to fix, because there is a behavior change that is required:

```go
func ClientStreamingMethodHandler(stream ___) error {
	// ...
	stream.SendAndClose(responseMessage)
	stream.SetTrailer(...)
	return nil
}
```

This used to work (trailers in `SetTrailer` are sent upon return), but wouldn't work if we make `SendAndClose` send trailers and `RST_STREAM`.

Maybe there's another way to prevent `Recv` from continuing to work after `SendAndClose` is called.

dfawley:
Somewhat related to #7286

eshitachandwani:
Created a issue for both : https://github.com/grpc/grpc-go/issues/8119

PR #8120, opened 2025-02-25
grpc_test: add tests for cardinality violation

eshitachandwani:
Fixes: #7570

Added the following tests for client streaming RPC : 
1. Client should ensure a status `internal` error is returned for client-streaming RPCs if the server doesn't send a message before returning status OK. This is a cardinality violation. Skipping this test for now till the issues are fixed: #8119 
2.   Server can continue to receive messages after calling SendAndClose.
3. Recv() on client streaming client blocks till the server handler returns even after calling SendAndClose from the server handler
4. Client receives nil message and non-nil error when server handler returns an error after calling SendAndClose.

RELEASE NOTES: N/A

### DIFF
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -3586,10 +3586,9 @@ func testClientStreamingError(t *testing.T, e env) {
 	}
 }
 
-// Tests the scenario where a client-streaming server returns a status OK
-// before calling `SendAndClose()`, while the client invokes
-// `CloseAndReceive()`. An error is expected if no message is sent along with
-// the status OK.
+// Tests  to verify that client receive a cardinality violation error for
+// client-streaming RPCs if the server doesn't send a message before returning
+// status OK.
 func (s) TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClose(t *testing.T) {
 	// TODO : https://github.com/grpc/grpc-go/issues/8119 - remove `t.Skip()`
 	// after this is fixed.
@@ -3614,16 +3613,16 @@ func (s) TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClos
 
 	_, err = stream.CloseAndRecv()
 	if err == nil {
-		t.Fatalf("%v.CloseAndRecv() = %v, want an error", stream, err)
+		t.Fatalf("stream.CloseAndRecv() = %v, want an error", err)
 	}
 	if status.Code(err) != codes.Internal {
-		t.Fatalf("%v.CloseAndRecv() = %v, want error %s", stream, err, codes.Internal)
+		t.Fatalf("stream.CloseAndRecv() = %v, want error %s", err, codes.Internal)
 	}
 }
 
-// Tests the scenario where a client-streaming server calls `Recv()` after calling
-// `SendAndClose()`. And error is expected if `Recv()` is called after calling
-// SendAndClose()` i.e the RPC should be terminated.
+// Test to verify for client-streaming RPCs, when SendAndClose is called, the server
+// should reset stream after sending the response message successfully.
+// If server calls Recv after, those operations should fail.
 func (s) TestClientStreamingCardinalityViolation_ServerHandlerRecvAfterSendAndClose(t *testing.T) {
 	// TODO : https://github.com/grpc/grpc-go/issues/8119 - remove `t.Skip()`
 	// after this is fixed.
@@ -3658,27 +3657,24 @@ func (s) TestClientStreamingCardinalityViolation_ServerHandlerRecvAfterSendAndCl
 		t.Fatal(err)
 	}
 
-	req := &testpb.StreamingInputCallRequest{
-		Payload: payload,
-	}
-
+	req := &testpb.StreamingInputCallRequest{Payload: payload}
+	//eshita break on eof , any other error t.Fatal
 	for {
-		if err = stream.Send(req); err == nil {
-			continue
+		if err = stream.Send(req); err == io.EOF {
+			break
 		}
 		select {
 		case <-ctx.Done():
 			t.Fatal("timed out waiting for error from server")
 		default:
 		}
-		break
 	}
 	_, err = stream.CloseAndRecv()
 	if err == nil {
-		t.Fatalf("%v.CloseAndRecv() = %v, want error", stream, err)
+		t.Fatalf("stream.CloseAndRecv() = %v, want error", err)
 	}
 	if status.Code(err) != codes.Internal {
-		t.Fatalf("%v.CloseAndRecv() = %v, want error %s", stream, err, codes.Internal)
+		t.Fatalf("stream.CloseAndRecv() = %v, want error %s", err, codes.Internal)
 	}
 }
 
@@ -3723,22 +3719,17 @@ func (s) TestClientStreamingCardinalityViolation_IgnoreServerHandlerErrorAfterSe
 	}
 
 	for {
-		if err = stream.Send(req); err == nil {
-			continue
+		if err = stream.Send(req); err == io.EOF {
+			break
 		}
 		select {
 		case <-ctx.Done():
 			t.Fatal("timed out waiting for error from server")
 		default:
 		}
-		break
 	}
-	_, err = stream.CloseAndRecv()
-	if err == nil {
-		t.Fatalf("%v.CloseAndRecv() = %v, want error", stream, err)
-	}
-	if status.Code(err) != codes.Internal {
-		t.Fatalf("%v.CloseAndRecv() = %v, want error %s", stream, err, codes.Internal)
+	if _, err = stream.CloseAndRecv(); err != nil {
+		t.Fatalf("stream.CloseAndRecv() = %v, want nil", err)
 	}
 }
 
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -3594,6 +3594,8 @@ func (s) TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClos
 	t.Skip()
 	ss := &stubserver.StubServer{
 		StreamingInputCallF: func(_ testgrpc.TestService_StreamingInputCallServer) error {
+			// Returning status OK without sending a response message.This is a
+			// cardinality violation.
 			return nil
 		},
 	}
@@ -3619,16 +3621,16 @@ func (s) TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClos
 	}
 }
 
-// Tests the behavior where Recv() can be called after SendAndClose(). Although
+// Tests that the server can continue to receive messages after calling SendAndClose. Although
 // this is unexpected, we retain it for backward compatibility.
 func (s) TestClientStreaming_ServerHandlerRecvAfterSendAndClose(t *testing.T) {
 	ss := stubserver.StubServer{
 		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
 			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
-				t.Fatalf("stream.SendAndClose(_) = %v, want <nil>", err)
+				t.Errorf("stream.SendAndClose(_) = %v, want <nil>", err)
 			}
-			if resp, err := stream.Recv(); err != nil || resp == nil {
-				t.Fatalf("stream.Recv() = %v, %v, want non-nil empty response, <nil>", resp, err)
+			if resp, err := stream.Recv(); err != nil || !proto.Equal(resp, &testpb.StreamingInputCallRequest{}) {
+				t.Errorf("stream.Recv() = %s, %v, want non-nil empty response, <nil>", resp, err)
 			}
 			return nil
 		},
@@ -3647,20 +3649,19 @@ func (s) TestClientStreaming_ServerHandlerRecvAfterSendAndClose(t *testing.T) {
 	if err := stream.Send(&testpb.StreamingInputCallRequest{}); err != nil {
 		t.Fatalf("stream.Send(_) = %v, want <nil>", err)
 	}
-	if resp, err := stream.CloseAndRecv(); err != nil || resp == nil {
+	if resp, err := stream.CloseAndRecv(); err != nil || !proto.Equal(resp, &testpb.StreamingInputCallResponse{}) {
 		t.Fatalf("stream.CloseSend() = %v , %v, want non-nil empty response, <nil>", resp, err)
 	}
 }
 
-// Tests the behavior where Recv() on client streaming client blocks till the
-// server handler returns even after calling SendAndClose from the server
-// handler.
+// Tests that Recv() on client streaming client blocks till the server handler
+// returns even after calling SendAndClose from the server handler.
 func (s) TestClientStreaming_RecvWaitsForServerHandlerRetrun(t *testing.T) {
 	waitForReturn := make(chan struct{})
 	ss := stubserver.StubServer{
 		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
 			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
-				t.Fatalf("stream.SendAndClose(_) = %v, want <nil>", err)
+				t.Errorf("stream.SendAndClose(_) = %v, want <nil>", err)
 			}
 			<-waitForReturn
 			return nil
@@ -3706,14 +3707,14 @@ func (s) TestClientStreaming_RecvWaitsForServerHandlerRetrun(t *testing.T) {
 }
 
 // Tests the behavior where server handler returns an error after calling
-// SendAndClose. It verifies the behavior that client receives nil message and
+// SendAndClose. It verifies the that client receives nil message and
 // non-nil error.
 func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	wantError := status.Error(codes.Internal, "error for testing")
 	ss := stubserver.StubServer{
 		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
 			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
-				t.Fatalf("stream.SendAndClose(_) = %v, want <nil>", err)
+				t.Errorf("stream.SendAndClose(_) = %v, want <nil>", err)
 			}
 			return wantError
 		},
@@ -3729,7 +3730,12 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	if err != nil {
 		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
 	}
-	if resp, err := stream.CloseAndRecv(); status.Code(err) != codes.Internal || resp != nil {
+	resp, err := stream.CloseAndRecv()
+
+	wantStatus, _ := status.FromError(wantError)
+	gotStatus, _ := status.FromError(err)
+
+	if gotStatus.Code() != wantStatus.Code() || gotStatus.Message() != wantStatus.Message() || resp != nil {
 		t.Fatalf("stream.CloseSend() = %v , %v, want <nil>, %s", resp, err, wantError)
 	}
 }
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -3622,24 +3622,18 @@ func (s) TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClos
 // Test to verify for client-streaming RPCs, when SendAndClose is called, the server
 // should reset stream after sending the response message successfully.
 // If server calls Recv after, those operations should fail.
-func (s) TestClientStreamingCardinalityViolation_ServerHandlerRecvAfterSendAndClose(t *testing.T) {
+func TestClientStreamingCardinalityViolation_ServerHandlerRecvAfterSendAndClose(t *testing.T) {
 	// TODO : https://github.com/grpc/grpc-go/issues/8119 - remove `t.Skip()`
 	// after this is fixed.
-	t.Skip()
+	// t.Skip()
 	ss := &stubserver.StubServer{
 		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
-			sum := 0
-			in, _ := stream.Recv()
-			p := in.GetPayload().GetBody()
-			sum += len(p)
-			stream.SendAndClose(&testpb.StreamingInputCallResponse{
-				AggregatedPayloadSize: int32(sum),
-			})
+			stream.SendAndClose(&testpb.StreamingInputCallResponse{})
 			_, err := stream.Recv()
 			if err == nil {
 				t.Errorf("stream.Recv() = nil, want an error")
 			}
-			return nil
+			return err
 		},
 	}
 	if err := ss.Start(nil); err != nil {
@@ -3677,6 +3671,14 @@ func (s) TestClientStreamingCardinalityViolation_ServerHandlerRecvAfterSendAndCl
 	if err != nil {
 		t.Fatalf("stream.CloseAndRecv() = %v, want error", err)
 	}
+
+	te := newTest(t, tcpClearEnv)
+	te.startServer(ss)
+	defer te.tearDown()
+	te.withServerTester(func(st *serverTester) {
+		st.wantRSTStream(http2.ErrCodeNo) // Expect RST_STREAM with code NO_ERROR (success)
+	})
+
 }
 
 // Tests the scenario where a client-streaming server sends an error after calling
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -289,11 +289,10 @@ func (s *testServer) StreamingInputCall(stream testgrpc.TestService_StreamingInp
 		return nil
 	}
 	if s.recvAfterClose {
-		fmt.Printf("reAfetclose")
-		err := stream.SendAndClose(&testpb.StreamingInputCallResponse{
+		stream.SendAndClose(&testpb.StreamingInputCallResponse{
 			AggregatedPayloadSize: int32(sum),
 		})
-		_, err = stream.Recv()
+		_, err := stream.Recv()
 		return err
 	}
 	for {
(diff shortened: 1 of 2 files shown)


test/end2end_test.go
```diff
@@ -3585,6 +3597,75 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	}
+
+	req := &testpb.StreamingInputCallRequest{
+		Payload: payload,
+	}
+
+	for {
+		if err = stream.Send(req); err == nil {
+			continue
+		}
```
arjan-bal, line 3662: If send returns `io.EOF`, you need to call `Recv()` to get the RPC status. See https://pkg.go.dev/google.golang.org/grpc#ClientStream

eshitachandwani: Right. Thank you for pointing it out. Changed.

test/end2end_test.go
```diff
@@ -145,6 +145,8 @@ type testServer struct {
 	setHeaderOnly      bool   // whether to only call setHeader, not sendHeader.
 	multipleSetTrailer bool   // whether to call setTrailer multiple times.
 	unaryCallSleepTime time.Duration
+	earlyNil           bool // whether to return nil without calling SendAndClose().
+	recvAfterClose     bool // whether to call Recv() after calling SendAndClose().
```
dfawley, line 149: We really really don't want to extend `testServer`.  Could you use the `stubserver` package instead?  `testServer` has a lot of bad properties (especially: it splits the testing code into two different places: the test and the `testServer`), and this only complicates it further by adding more ways to conditionalize its behavior.

You can follow this kind of example to use `stubserver` instead:

https://github.com/grpc/grpc-go/blob/d48317fafe7361f172fc243bbde7e5594f12ebd5/test/end2end_test.go#L4485

Also please add a comment on `testServer` that it should not be used for any new tests so that this doesn't happen again.

Sorry!

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	stream, err := ss.Client.StreamingInputCall(ctx)
+	if err != nil {
+		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+
+	if _, err := stream.CloseAndRecv(); status.Code(err) != codes.Internal {
```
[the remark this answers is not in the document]
purnesh42H: Also, it might be good to have separate if condition for both error and status violation. Although they are always supposed to happen in conjuction but checking them separately make it easier to understand.

```
if _, err := stream.CloseAndRecv(); err == nil {
      t.Fatalf("CloseAndRecv should have returned an error")
}

if  status.Code(err) == status.ok {
    t.Fatalf("CloseAndRecv should have returned a non-ok")
}
```

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
 	}
 }
 
+func (s) TestClientStreamingMissingSendAndCloseError(t *testing.T) {
```
purnesh42H, line 3589: suggestion for name to be more explicit `TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClose`

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	if err != nil {
+		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+
+	if _, err := stream.CloseAndRecv(); status.Code(err) != codes.Internal {
+		t.Fatalf("%v.CloseAndRecv() = %v, want error %s", stream, err, codes.Internal)
+	}
+}
+
+func (s) TestClientStreamingRecvAfterCloseError(t *testing.T) {
```
purnesh42H, line 3620: suggestion for name to be more explicit TestClientStreamingCardinalityViolation_IgnoreServerHandlerErrorAfterSendAndClose

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	ss := &stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			sum := 0
+			in, _ := stream.Recv()
+			p := in.GetPayload().GetBody()
+			sum += len(p)
+			stream.SendAndClose(&testpb.StreamingInputCallResponse{
+				AggregatedPayloadSize: int32(sum),
+			})
+			_, err := stream.Recv()
```
purnesh42H, line 3637: nit: do we need to do this? What if just return a static error with status code like `status.Error(codes.Internal, "something went wrong")).
`? Because calling stream.Recv() adding one more possibility to the test case

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
...
+
+	req := &testpb.StreamingInputCallRequest{
+		Payload: payload,
+	}
+
+	for {
+		if err = stream.Send(req); err == nil {
+			continue
+		}
+		if _, err := stream.CloseAndRecv(); status.Code(err) != codes.Internal {
```
purnesh42H, line 3666: nit: We should probably just break out of loop when server closes the stream and call the CloseAndRecv() outside

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
...
+			t.Fatalf("%v.CloseAndRecv() = %v, want error %s", stream, err, codes.Internal)
```
purnesh42H, line 3667: Similar comment here as above, probably good to write 2 separate ifs but the opposite way in this case. 1) raise t.Fatalf if error is not nil. 2) raise t.Fatalf if status is not ok

purnesh42H (commented):
Looks like we need one more test case when "Server Doesn't RST_STREAM" 

Server-Side:
- Create a client-streaming RPC handler.
- The handler calls SendAndClose with a valid response message.
- The handler then attempts to call Recv.

Client-Side:
- Make a client-streaming RPC call.
- Continuously send messages.

Expected Outcome:
- The server's Recv call should return an error indicating that the stream is closed.
- The client's Send calls should eventually return an error indicating that the stream has been reset.

eshitachandwani:
> Looks like we need one more test case when "Server Doesn't RST_STREAM"
> 
> Server-Side:
> 
> * Create a client-streaming RPC handler.
> * The handler calls SendAndClose with a valid response message.
> * The handler then attempts to call Recv.
> 
> Client-Side:
> 
> * Make a client-streaming RPC call.
> * Continuously send messages.
> 
> Expected Outcome:
> 
> * The server's Recv call should return an error indicating that the stream is closed.
> * The client's Send calls should eventually return an error indicating that the stream has been reset.

This looks like the test I have written already , the `TestClientStreamingRecvAfterCloseError` test , can you tell me what is the difference between the two?

test/end2end_test.go
```diff
@@ -3585,6 +3586,90 @@ func testClientStreamingError(t *testing.T, e env) {
 	}
 }
 
+func (s) TestClientStreamingMissingSendAndCloseError(t *testing.T) {
+	// TODO : https://github.com/grpc/grpc-go/issues/8119 - remove `t.Skip()`
+	// after this is fixed.
+	t.Skip()
+	testClientStreamingMissingSendAndCloseError(t)
```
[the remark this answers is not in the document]
eshitachandwani: Done.

test/end2end_test.go
eshitachandwani: Done.

test/end2end_test.go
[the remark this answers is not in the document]
eshitachandwani: This looks repetitive. Changed it to check error nil and status not internal.

test/end2end_test.go
```diff
@@ -3585,6 +3586,162 @@ func testClientStreamingError(t *testing.T, e env) {
...
+		}
+		select {
+		case <-ctx.Done():
+			t.Fatal("timed out waiting for error from server")
+		default:
+		}
+		break
+	}
+	_, err = stream.CloseAndRecv()
+	if err == nil {
```
purnesh42H, line 3677: same here. Isn't client expected to ignore the error?

test/end2end_test.go
```diff
@@ -3585,6 +3586,162 @@ func testClientStreamingError(t *testing.T, e env) {
...
+		t.Fatalf("%v.CloseAndRecv() = %v, want error", stream, err)
+	}
+	if status.Code(err) != codes.Internal {
```
purnesh42H, line 3680: same question about expected status being OK

test/end2end_test.go
```diff
@@ -3585,6 +3586,162 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	defer cancel()
+
+	stream, err := ss.Client.StreamingInputCall(ctx)
+	if err != nil {
+		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+
+	_, err = stream.CloseAndRecv()
+	if err == nil {
+		t.Fatalf("%v.CloseAndRecv() = %v, want an error", stream, err)
```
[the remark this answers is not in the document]
eshitachandwani: right ! Changed.

test/end2end_test.go
```diff
@@ -3585,6 +3586,152 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	ss := &stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			sum := 0
+			in, _ := stream.Recv()
+			p := in.GetPayload().GetBody()
+			sum += len(p)
+			stream.SendAndClose(&testpb.StreamingInputCallResponse{
+				AggregatedPayloadSize: int32(sum),
+			})
+			_, err := stream.Recv()
```
purnesh42H, line 3639: so, we should verify if `stream.Recv()` actually failed. Can do `logger.Fatalf()` if err is nil. Then just return as the stream is no longer available. See here https://github.com/grpc/grpc-go/blob/eb744dec5da784e7e2fc140ddaa6d2aa645b5371/stream_interfaces.go#L94

I don't think for this test case we need to return the error from handler

test/end2end_test.go
purnesh42H: Not sure if it was missed or not clarified. If you do as mentioned above, the client is not expected to receive any error. So, you should check for err != nil and terminate with failure if that's the case.

test/end2end_test.go
```diff
@@ -3585,6 +3586,152 @@ func testClientStreamingError(t *testing.T, e env) {
...
+		t.Fatalf("StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+	payload, err := newPayload(testpb.PayloadType_COMPRESSABLE, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	req := &testpb.StreamingInputCallRequest{Payload: payload}
+	for {
+		if err = stream.Send(req); err == io.EOF {
```
purnesh42H, line 3662: we should also `t.fatalf` with failure if err is anything else apart from `io.EOF`?

test/end2end_test.go
purnesh42H: same. should either check for `OK` status not being there or check for error code being there. Should do `t.fatalf` if not ok status is there.

test/end2end_test.go
```diff
@@ -3585,6 +3586,152 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	req := &testpb.StreamingInputCallRequest{
+		Payload: payload,
+	}
+
+	for {
+		if err = stream.Send(req); err == io.EOF {
```
purnesh42H, line 3721: same. t.fatalf if any other error apart from `io.EOF`

test/end2end_test.go
eshitachandwani: done.

test/end2end_test.go
eshitachandwani: done.

test/end2end_test.go
eshitachandwani: done.

test/end2end_test.go
eshitachandwani: done.

eshitachandwani (commented):
Looks like I had not published the earlier comments.

eshitachandwani (commented):
Looks like I had not published the earlier comments.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+		if err != nil {
+			t.Fatalf("Stream.Send(req) = %v, want <nil>", err)
+		}
+		select {
+		case <-ctx.Done():
+			t.Fatal("timed out waiting for error from server")
+		default:
+		}
+	}
+	if _, err = stream.CloseAndRecv(); err != nil {
```
purnesh42H, line 3739: nit: should check the status ok after this well.

purnesh42H (approved):
LGTM

purnesh42H:
@dfawley for second review

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
 	}
 }
 
+// Tests  to verify that client receive a cardinality violation error for
+// client-streaming RPCs if the server doesn't send a message before returning
+// status OK.
```
dfawley, line 3591:
```suggestion
// Tests that a client receives a cardinality violation error for
// client-streaming RPCs if the server doesn't send a message before returning
// status OK.
```

(Please rewrap)

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+			sum := 0
+			in, _ := stream.Recv()
+			p := in.GetPayload().GetBody()
+			sum += len(p)
+			stream.SendAndClose(&testpb.StreamingInputCallResponse{
+				AggregatedPayloadSize: int32(sum),
+			})
+			_, err := stream.Recv()
+			if err == nil {
+				logger.Fatalf("stream.Recv() = %v, want an error", err)
```
dfawley, line 3641: You want `t.Errorf` not `logger.Fatalf` here.  `logger.Fatalf` kills the whole process.  `t.Fatalf` kills the current goroutine.  This won't run in the main test goroutine, so `t.Fatalf` would just cause problems, not terminate the test.

dfawley, line 3641: Nit: `"= %v", err` -- `err` is `nil`.  Just put `nil` in the string.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	if err == nil {
+		t.Fatalf("stream.CloseAndRecv() = %v, want an error", err)
+	}
+	if status.Code(err) != codes.Internal {
+		t.Fatalf("stream.CloseAndRecv() = %v, want error %s", err, codes.Internal)
+	}
+}
+
+// Test to verify for client-streaming RPCs, when SendAndClose is called, the server
+// should reset stream after sending the response message successfully.
```
dfawley, line 3624: You don't appear to be validating RST_STREAM is written to the connection.  You'll need to use the `serverTester` for that.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+		default:
+		}
+	}
+	_, err = stream.CloseAndRecv()
+	if err != nil {
+		t.Fatalf("stream.CloseAndRecv() = %v, want error", err)
+	}
+	if status.Code(err) != codes.OK {
+		t.Fatalf("stream.CloseAndRecv() = %v, want error %s", err, codes.OK)
+	}
```
dfawley, line 3683: `status.Code(nil)` is always `codes.OK` so this doesn't do anything.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+}
+
+// Tests the scenario where a client-streaming server sends an error after calling
+// `SendAndClose()`. The error should be ignored by client because RPC should be
```
dfawley, line 3687:
```suggestion
// `SendAndClose()`. The error should be ignored by the client because the RPC should be
```

Please re-wrap.

test/end2end_test.go
dfawley: A `nil` error is an `OK` status, so I don't think that's a good idea.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+// closed when `SendAndClose()` is called.
+func (s) TestClientStreamingCardinalityViolation_IgnoreServerHandlerErrorAfterSendAndClose(t *testing.T) {
+	// TODO : https://github.com/grpc/grpc-go/issues/8119 - remove `t.Skip()`
+	// after this is fixed.
+	t.Skip()
```
dfawley, line 3692: Based on the test description above...why does this fail today?  I don't think this should be a problem -- I was expecting that it's impossible for the server to do anything else.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	ss := &stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			sum := 0
+			in, _ := stream.Recv()
+			p := in.GetPayload().GetBody()
+			sum += len(p)
```
dfawley, line 3698: `sum := len(p)` ?  But why is it a `sum`?  Probably this was left over from an old iteration.  You can just delete it and return `len(p)` directly in the response message, too.  Or, better, just return 0 in the response.  The value is unimportant.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	defer cancel()
+
+	stream, err := ss.Client.StreamingInputCall(ctx)
+	if err != nil {
+		t.Fatalf("StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+	payload, err := newPayload(testpb.PayloadType_COMPRESSABLE, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
```
dfawley, line 3720: Seems easier to just delete this.  Send empty requests instead...

Less code is usually better code.

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+			break
+		}
+		if err != nil {
+			t.Fatalf("Stream.Send(req) = %v, want <nil>", err)
+		}
+		select {
+		case <-ctx.Done():
+			t.Fatal("timed out waiting for error from server")
+		default:
+		}
```
dfawley, line 3737: This can be replaced by doing:

```go
for ctx.Err() == nil {
	if err := stream.Send()....
}
```

test/end2end_test.go
```diff
@@ -3585,6 +3586,161 @@ func testClientStreamingError(t *testing.T, e env) {
...
+
+	req := &testpb.StreamingInputCallRequest{
+		Payload: payload,
+	}
+
+	for {
+		if err = stream.Send(req); err == io.EOF {
+			break
+		}
+		if err != nil {
```
dfawley, line 3730: If you use an `else if` here, then you can use `err :=` in the `if`.  That's more natural unless it's important to keep `err` beyond the loop -- but it isn't.

test/end2end_test.go
eshitachandwani: Done.

test/end2end_test.go
eshitachandwani: Done.

test/end2end_test.go
eshitachandwani: Done.

test/end2end_test.go
eshitachandwani: I am not sure what you mean , because if I send an error code , example `Unimplemented` , according to the test , it is being reflected in the client when `CloseAndRecv()` is called by client. 
Also in the the code, `SendAndCLose()` doesnt look like its doing anything to close , just calling [`SendMsg()`](https://github.com/grpc/grpc-go/blob/732f3f32f5365f4f5b89a3f1be19534cd28db794/stream_interfaces.go#L225C1-L227C2) . It also says to return nil after calling `SendAndCLose()` or return an error ,look [here](https://github.com/grpc/grpc-go/blob/732f3f32f5365f4f5b89a3f1be19534cd28db794/stream_interfaces.go#L99) so I would think that returning an error after `SendAndClose()` would surely propogate to client and this is the intended behavior and this test is not needed.

dfawley: I see.  As mentioned offline, I'm worried that we won't be able to ever change the behavior of `SendAndClose` to end the RPC, even though that is what it claims to do.  So we should instead confirm that `SendAndClose` _doesn't_ send the TRAILERS on success.  And that returning an error from the handler after `SendAndClose` results in a TRAILERS frame being sent that contains the error.

How the client handles that may depend on the type of stream from the client's view.  E.g. if the client is using it as a bidi streaming RPC, then it should see the error trailers.  But if it's a client streaming call, then it might get some other kind of error, since a message _and_ an error is not expected for client streaming RPCs.

test/end2end_test.go
```diff
@@ -3585,6 +3586,154 @@ func testClientStreamingError(t *testing.T, e env) {
...
 
+// Tests that a client receives a cardinality violation error for client-streaming
+// RPCs if the server doesn't send a message before returning status OK.
+func (s) TestClientStreamingCardinalityViolation_ServerHandlerMissingSendAndClose(t *testing.T) {
+	// TODO : https://github.com/grpc/grpc-go/issues/8119 - remove `t.Skip()`
+	// after this is fixed.
+	t.Skip()
+	ss := &stubserver.StubServer{
+		StreamingInputCallF: func(_ testgrpc.TestService_StreamingInputCallServer) error {
+			return nil
```
dfawley, line 3597: There should be comments here and there explaining why the test is doing what it is.  Like  "// Immediately return a success without first sending a response message, which is illegal."

test/end2end_test.go
```diff
@@ -3585,6 +3586,154 @@ func testClientStreamingError(t *testing.T, e env) {
...
+// this is unexpected, we retain it for backward compatibility.
+func (s) TestClientStreaming_ServerHandlerRecvAfterSendAndClose(t *testing.T) {
+	ss := stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
+				t.Fatalf("stream.SendAndClose(_) = %v, want <nil>", err)
+			}
+			if resp, err := stream.Recv(); err != nil || resp == nil {
+				t.Fatalf("stream.Recv() = %v, %v, want non-nil empty response, <nil>", resp, err)
+			}
```
dfawley, line 3632: Can you assert that the response matches some expectation, not just "not nil"?  I.e. `proto.Equal(&testpb.StreamingInputCallRequest{})`?

test/end2end_test.go
```diff
@@ -3585,6 +3586,154 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	stream, err := ss.Client.StreamingInputCall(ctx)
+	if err != nil {
+		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+	if err := stream.Send(&testpb.StreamingInputCallRequest{}); err != nil {
+		t.Fatalf("stream.Send(_) = %v, want <nil>", err)
+	}
+	if resp, err := stream.CloseAndRecv(); err != nil || resp == nil {
```
dfawley, line 3650: Similar here.

test/end2end_test.go
```diff
@@ -3585,6 +3586,154 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	if err == nil {
+		t.Fatalf("stream.CloseAndRecv() = %v, want an error", err)
+	}
+	if status.Code(err) != codes.Internal {
+		t.Fatalf("stream.CloseAndRecv() = %v, want error %s", err, codes.Internal)
+	}
+}
+
+// Tests the behavior where Recv() can be called after SendAndClose(). Although
+// this is unexpected, we retain it for backward compatibility.
```
dfawley, line 3623: The test description shouldn't simply say "tests this scenario", it should say "tests that in this scenario this thing happens".  Here it should say something like  "Tests that the server can continue to receive messages after calling SendAndClose".

test/end2end_test.go
```diff
@@ -3585,6 +3586,154 @@ func testClientStreamingError(t *testing.T, e env) {
...
+
+// Tests the behavior where server handler returns an error after calling
+// SendAndClose. It verifies the behavior that client receives nil message and
+// non-nil error.
+func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
+	wantError := status.Error(codes.Internal, "error for testing")
+	ss := stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
+				t.Fatalf("stream.SendAndClose(_) = %v, want <nil>", err)
```
dfawley, line 3716: This needs to be `Errorf` since it's not in the main test goroutine.  Otherwise it just kills the current goroutine, and keeps the test running.  (I see the same thing above.)

test/end2end_test.go
```diff
@@ -3585,6 +3586,154 @@ func testClientStreamingError(t *testing.T, e env) {
...
+	}
+	defer ss.Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	stream, err := ss.Client.StreamingInputCall(ctx)
+	if err != nil {
+		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+	if resp, err := stream.CloseAndRecv(); status.Code(err) != codes.Internal || resp != nil {
```
dfawley, line 3732: You should also check `status.Message(err) == status.Message(wantError)`.  And it would be better to use `status.Code(wantError)` instead of repeating the code in it here.

dfawley (approved):
Looks great, thanks!  Just a bunch of little things.

dfawley (approved):
Looks great, thanks!  Just a bunch of little things.

test/end2end_test.go
eshitachandwani: Done.

test/end2end_test.go
eshitachandwani: Done.

PR #8124, opened 2025-02-25
[Security] Add support for parsing SPIFFE Bundle Maps

gtcooke94:
This adds a dependency on go-spiffe in order to parse SPIFFE bundles. More specifically, that library does not yet support SPIFFE bundle maps, but it does support SPIFFE bundles. This adds parsing of these maps to grpc-go

RELEASE NOTES: N/A


credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,45 @@
...
+	json.Unmarshal([]byte(byteValue), &result)
+	if result.Bundles == nil {
+		return nil, errors.New("no content in spiffe bundle map file")
+	}
+	bundleMap := map[string]*spiffebundle.Bundle{}
+	for trustDomain, jsonBundle := range result.Bundles {
+		bundle, err := spiffebundle.Parse(spiffeid.RequireTrustDomainFromString(trustDomain), jsonBundle)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse bundle in map: %v", err)
+
```
matthewstevenson88, line 40: nit: Unnecessary blank line.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+		bundleMap[trustDomain] = bundle
```
matthewstevenson88, line 42: Do we need to worry about duplicate keys?

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,45 @@
...
+// This API is experimental.
+func LoadSpiffeBundleMap(filePath string) (map[string]*spiffebundle.Bundle, error) {
+	bundleMapFile, err := os.Open(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open spiffe bundle map file: %v", err)
+	}
+	defer bundleMapFile.Close()
+	byteValue, _ := io.ReadAll(bundleMapFile)
+	var result partialParsedSpiffeBundleMap
+	json.Unmarshal([]byte(byteValue), &result)
```
matthewstevenson88, line 31: Should we check the return value?

gtcooke94: Yes, good catch

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: This is a good question - I was looking into it when writing the code - the spec says
"Producers of SPIFFE Bundle Maps MUST ensure uniqueness amongst trust domain names in a given “trust_domains” element. Consumers MUST reject maps in which duplicate trust domain name keys are detected, provided that the underlying libraries are capable of detecting such duplications."

So yes, we should. However, because this happens via json unmarshaling, we are semi-trapped by Go's handling of duplicate keys for json->map (which is to just replace with the last occurring key). However, there are plenty of options that involve more complex parsing/parsing multiple times. I think given the language of the spec we are technically valid either way? My take is that the Go json unmarshaler is not capable of detecting these duplications.

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Removed

credentials/tls/certprovider/spiffe/spiffe.go
matthewstevenson88: SGTM. Could you document how we're handling this in a comment?

matthewstevenson88 (approved):

matthewstevenson88 (approved):

easwars:
@dfawley : I can pick this up to reduce your load a little bit.

testdata/spiffe/spiffe-openssl.cnf
easwars, line 1: For how long are these newly added certs valid? Is that controllable by some parameter in this config file?

easwars, line 1: And could we please have a README file in this directory that talks about how to generate these files (if and when they become invalid) and what each of these represent? Thanks.

scripts/vet-proto.sh
```diff
@@ -32,7 +32,7 @@ elif [[ "$#" -ne 0 ]]; then
 fi
 
 # - Check that generated proto files are up to date.
-go generate google.golang.org/grpc/... && git status --porcelain 2>&1 | fail_on_output || \
+go generate ./... && git status --porcelain 2>&1 | fail_on_output || \
```
easwars, line 35: Again why this change? If the reason is to regenerate protos in submodules, then we should probably be explicit about it.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package spiffe implements utilities for grpc-go for work with SPIFFE bundles
```
easwars, line 19: Nit: skip the `for grpc-go` part?

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+// https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#4-spiffe-bundle-format
+// If duplicate keys are encountered in the JSON parsing, Go's default unmarshal
+// behavior occurs which causes the last processed entry to be the entry in the
+// parsed map.
+//
+// This API is experimental.
+func LoadSpiffeBundleMap(filePath string) (map[string]*spiffebundle.Bundle, error) {
+	bundleMapFile, err := os.Open(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open spiffe bundle map file: %v", err)
```
easwars, line 49: Nit: `os.Open` already spits out an error message which would probably say something like "failed to open file". Could you please ensure that there is no duplicate in the eventual error returned from this function? Thanks.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	defer bundleMapFile.Close()
+	byteValue, _ := io.ReadAll(bundleMapFile)
```
easwars, line 52: How about using a single API to both open and read the file? See: https://pkg.go.dev/os#ReadFile

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+	var result partialParsedSpiffeBundleMap
+	err = json.Unmarshal([]byte(byteValue), &result)
+	if err != nil {
```
easwars, line 55: The assignment and the conditional can be on the same line.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+// This API is experimental.
+func LoadSpiffeBundleMap(filePath string) (map[string]*spiffebundle.Bundle, error) {
+	bundleMapFile, err := os.Open(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open spiffe bundle map file: %v", err)
+	}
+	defer bundleMapFile.Close()
+	byteValue, _ := io.ReadAll(bundleMapFile)
+	var result partialParsedSpiffeBundleMap
+	err = json.Unmarshal([]byte(byteValue), &result)
```
easwars, line 54: Why do we need the type cast to `[]byte`? Doesn't the API to read the contents of the file already return a `[]byte`?

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if result.Bundles == nil {
+		return nil, errors.New("no content in spiffe bundle map file")
```
easwars, line 59: Nit: Please include a package name prefix in all error strings returned from this package.

easwars, line 59: Does is make sense to include the file name in this error message (to be more informative)?

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	bundleMap := map[string]*spiffebundle.Bundle{}
+	for trustDomainString, jsonBundle := range result.Bundles {
+		trustDomain, err := spiffeid.TrustDomainFromString(trustDomainString)
+		if err != nil {
+			return nil, fmt.Errorf("invalud trust domain found when parsing map: %v", err)
```
easwars, line 65: Looks like this error condition is not covered by any test. At least that is what codecov is saying. Could you add a test for this? Thanks.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+	var result partialParsedSpiffeBundleMap
+	err = json.Unmarshal([]byte(byteValue), &result)
+	if err != nil {
+		return nil, err
+	}
+	if result.Bundles == nil {
+		return nil, errors.New("no content in spiffe bundle map file")
+	}
+	bundleMap := map[string]*spiffebundle.Bundle{}
+	for trustDomainString, jsonBundle := range result.Bundles {
```
easwars, line 62: It is not common to include the type in the variable name. And variable names that are used as loop indices are generally short. So, maybe something like `domain` or `td`. See: https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#variable-names

credentials/tls/certprovider/spiffe/spiffe.go
easwars, line 65: Nit: s/invalud/invalid.

Also, please consider either adding the file name or the trust domain name in the error message to make it more meaningful and actionable.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+		trustDomain, err := spiffeid.TrustDomainFromString(trustDomainString)
+		if err != nil {
+			return nil, fmt.Errorf("invalud trust domain found when parsing map: %v", err)
+		}
+		bundle, err := spiffebundle.Parse(trustDomain, jsonBundle)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse bundle in map: %v", err)
```
easwars, line 69: Same here. Please make error message more meaningful and actionable.

credentials/tls/certprovider/spiffe/spiffe.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// LoadSpiffeBundleMap loads a SPIFFE Bundle Map from a file. See the SPIFFE
+// Bundle Map spec for more detail -
+// https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#4-spiffe-bundle-format
+// If duplicate keys are encountered in the JSON parsing, Go's default unmarshal
+// behavior occurs which causes the last processed entry to be the entry in the
+// parsed map.
+//
+// This API is experimental.
```
easwars, line 45: Is this package going to be used only by grpc-go and not for consumption by external users? If so, please move this into an internal directory. Maybe something like `credentials/tls/certprovider/internal/spiffe`.

Also, would it be possible to move the test certs into a subdirectory that lives along with the code, instead of in the top-level testdata directory. Thanks.

.github/workflows/testing.yml
```diff
@@ -107,7 +107,7 @@ jobs:
         if: matrix.type == 'tests'
         run: |
           go version
-          go test ${{ matrix.testflags }} -cpu 1,4 -timeout 7m google.golang.org/grpc/...
+          go test ${{ matrix.testflags }} -cpu 1,4 -timeout 7m ./...
```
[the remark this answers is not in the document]
gtcooke94: The CI was failing due to weird issues with go.sum entries, discussed with Arjan and Doug - Arjan found this https://github.com/golang/go/issues/59458#issuecomment-1497938718 and found that changing to ./... worked.
The issue seems to stem from the go-spiffe repo taking a direct dependency on grpc-go/examples: https://github.com/spiffe/go-spiffe/blob/dd15542826b6e12943e74670a8807a74b4da09d9/v2/go.mod#L11

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Done

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: I can see a case where a user might want to implement their own specific spiffe provider behavior, but that might not actually be reasonable. As for now, this is meant to be used internally.
I found there are actually already a few spiffe utils in internal/credentials/spiffe.go, would you like to see these APIs added there instead?

testdata/spiffe/spiffe-openssl.cnf
gtcooke94: It depends on the commands used to generate them - I've largely copied these from the equivalent work in the java repo which (unfortunately) doesn't have detailed instructions like that. I'll make a best effort at it, I think I roughly know how it was done.

gtcooke94: Will do, see more detail in the other thread

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Swapped to just returning the error

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Done

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Fixed, I think at some point it wasn't happy with a string there but I changed things around

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Done

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Added spiffe: to error string, is that what you meant?

gtcooke94: Done

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Swapped to `td` - was saving `trustDomain` for the same value but actually converted to a `TrustDomain` type

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Will do

gtcooke94: Done

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Done

scripts/vet-proto.sh
gtcooke94: Same for the other `./...` change that Arjan discovered

.github/workflows/testing.yml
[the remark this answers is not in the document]
dfawley: I really would like to understand this better before proceeding.  Yes this fixes our CI but could it mean something is broken that will affect external usages upon the next release?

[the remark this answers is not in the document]
gtcooke94: @arjan-bal have you by chance been able to investigate this further?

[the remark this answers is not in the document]
arjan-bal: I found an issue on Github with a similar error: https://github.com/golang/go/issues/59458. From my understanding, the reason for the problem  mentioned was that the to resolve query with absolute path `go list` was looking for a checksum of the parent module in the submodule. This case seems to be the a little different. Running go list with the absolute path in the main module fails.
```sh
$ go list google.golang.org/grpc/...
pattern google.golang.org/grpc/...: google.golang.org/grpc/examples@v0.0.0-20230224211313-3775f633ce20: missing go.sum entry
```
Running `go mod tidy` in all the submodules does nothing. Running `go mod list ./...` in all submodules works. 

I verified that the main module doesn't depend on the examples submodule by running the following command.
```sh
$ go list -f '{{with .Module}}{{.Path}}{{end}}' -deps -test all | sort -u | grep example

```

It seems to me like the `examples` submodule isn't needed to build the main module, so `go mod tidy` doesn't add it. `go mod list` needs a checksum for the examples submodule to resolve the absolute path query. so it errors out. The exact root cause probably depends on the logic of resolving the absolute path query. Maybe we should open an issue on the golang/go repo or ask a question on go-nuts, YAQs.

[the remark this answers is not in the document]
arjan-bal: Asked on yaqs/4031798088380186624 (internal).

scripts/vet-proto.sh
easwars: Let's use the other thread to get to the bottom of this.

credentials/tls/certprovider/spiffe/spiffe.go
easwars: Sure, that makes sense.

And after looking at that file, I think we need to capitalize SPIFFE in the name of this function. See: go/go-style/decisions#initialisms

credentials/tls/certprovider/spiffe/spiffe.go
easwars: Yes, but now if we are moving it into the `internal/credentials` package, maybe these error strings should have a `credentials: ` prefix. But I'm ok with `spiffe: ` as well.

credentials/tls/certprovider/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,136 @@
...
+func loadX509Cert(filePath string) *x509.Certificate {
+	certFile, _ := os.Open(filePath)
+	certRaw, _ := io.ReadAll(certFile)
+	block, _ := pem.Decode([]byte(certRaw))
+	if block == nil {
+		panic("failed to parse certificate PEM")
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		panic("failed to parse certificate: " + err.Error())
```
easwars, line 69: Instead of calling `panic`, please pass `t.Testing` to this function and call `t.Fatalf` here. Also mark the function as a test helper. See: go/go-style/decisions#mark-test-helpers

credentials/tls/certprovider/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return cert
+}
+
+func TestLoadSpiffeBundleMapFailures(t *testing.T) {
+	testCases := []struct {
+		filePath     string
+		expectError  bool
```
easwars, line 77: s/expect/want. See: go/go-style/decisions#got-before-want

credentials/tls/certprovider/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,136 @@
...
+		{
+			filePath:    testdata.Path("spiffe/spiffebundle_invalid_trustdomain.json"),
+			expectError: true,
+		},
+		{
+			// SPIFFE Bundles only support a use of x509-svid and jwt-svid. If a
+			// use other than this is specified, the parser does not fail, it
+			// just doesn't add an x509 authority or jwt authority to the bundle
+			filePath:     testdata.Path("spiffe/spiffebundle_wrong_use.json"),
+			expectNoX509: true,
```
easwars, line 121: Do we need this field in the table since all subtests are expected to fail, and therefore this field is always set to `true`.

credentials/tls/certprovider/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.filePath, func(t *testing.T) {
+			bundle, err := LoadSpiffeBundleMap(tc.filePath)
+			if tc.expectError && err == nil {
+				t.Fatalf("Did not fail but should have.")
```
easwars, line 129: Please make the error strings more actionable by identifying the function and its input. See: go/go-style/decisions#identify-the-function and go/go-style/decisions#identify-the-input

credentials/tls/certprovider/spiffe/spiffe.go
gtcooke94: Done

internal/credentials/spiffe.go
```diff
@@ -73,3 +78,42 @@ func SPIFFEIDFromCert(cert *x509.Certificate) *url.URL {
...
+}
+
+// LoadSPIFFEBundleMap loads a SPIFFE Bundle Map from a file. See the SPIFFE
+// Bundle Map spec for more detail -
+// https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#4-spiffe-bundle-format
+// If duplicate keys are encountered in the JSON parsing, Go's default unmarshal
+// behavior occurs which causes the last processed entry to be the entry in the
+// parsed map.
+//
+// This API is experimental.
```
easwars, line 93: Nit: This can be removed now that this code lives in an `internal` directory.

internal/credentials/spiffe_test.go
```diff
@@ -231,3 +233,110 @@ func (s) TestSPIFFEIDFromCert(t *testing.T) {
 		})
 	}
 }
+
+var SPIFFETrustBundle map[string]*spiffebundle.Bundle
```
easwars, line 237: Where is this used?

Why is this exported?

internal/credentials/spiffe_test.go
```diff
@@ -231,3 +233,110 @@ func (s) TestSPIFFEIDFromCert(t *testing.T) {
...
+
+func TestKnownSPIFFEBundle(t *testing.T) {
+	bundles, err := LoadSPIFFEBundleMap(testdata.Path("spiffe/spiffebundle.json"))
+	if err != nil {
+		t.Fatalf("Error during parsing: %v", err)
```
easwars, line 242: Please make the error strings more actionable by identifying the function and its input. See: go/go-style/decisions#identify-the-function and go/go-style/decisions#identify-the-input

internal/credentials/spiffe_test.go
```diff
@@ -231,3 +233,110 @@ func (s) TestSPIFFEIDFromCert(t *testing.T) {
...
+	}
+	if len(bundles) != 2 {
+		t.Fatal("did not parse correct bundle length")
```
easwars, line 245: Strings that are part of log messages or test errors don't have to comply to the error string capitalization style guide recommendation. Please see here: go/go-style/decisions#error-strings. This comment applies to other test error strings in this file as well.

easwars, line 245: Also, please include the actual length as seen by the test. See: go/go-style/decisions#got-before-want

internal/credentials/spiffe_test.go
```diff
@@ -231,3 +233,110 @@ func (s) TestSPIFFEIDFromCert(t *testing.T) {
...
+		t.Fatal("expected bundle for example.com")
+	}
+	if bundles["test.example.com"] == nil {
+		t.Fatal("expected bundle for test.example.com")
+	}
+
+	expectedExampleComCert := loadX509Cert(t, testdata.Path("spiffe/spiffe_cert.pem"))
+	expectedTestExampleComCert := loadX509Cert(t, testdata.Path("spiffe/server1_spiffe.pem"))
+	if !bundles["example.com"].X509Authorities()[0].Equal(expectedExampleComCert) {
+		t.Fatalf("expected cert for example.com bundle not correct.")
```
easwars, line 257: What is not correct about the bundle? Please make test errors actionable.

internal/credentials/spiffe_test.go
```diff
@@ -231,3 +233,110 @@ func (s) TestSPIFFEIDFromCert(t *testing.T) {
...
+
+}
+
+func loadX509Cert(t *testing.T, filePath string) *x509.Certificate {
+	t.Helper()
+	certFile, _ := os.Open(filePath)
+	certRaw, _ := io.ReadAll(certFile)
+	block, _ := pem.Decode([]byte(certRaw))
+	if block == nil {
+		panic("failed to parse certificate PEM")
```
easwars, line 271: Please use t.Fatalf() instead of `panic`. Here and below.

credentials/tls/certprovider/spiffe/spiffe_test.go
easwars: This needs to be taken care of still.

testdata/spiffe/README.md
```diff
@@ -0,0 +1,55 @@
...
+  -extfile spiffe-openssl.cnf -days 3650 -sha256
+$ openssl req -new -key server1.key -out spiffe-cert.csr \
+ -subj /C=US/ST=CA/L=SVL/O=gRPC/CN=*.test.google.com/ \
+ -config spiffe-openssl.cnf -reqexts spiffe_server_e2e
+$ openssl x509 -req -CA ca.pem -CAkey ca.key -CAcreateserial \
+ -in spiffe-cert.csr -out server1_spiffe.pem -extensions spiffe_server_e2e \
+  -extfile spiffe-openssl.cnf -days 3650 -sha256
+```
+
+Additionally, the SPIFFE trust bundle map files (spiffebundle*.json) are manually created for end to end testing. The
```
easwars, line 39: Please wrap README file contents to 80-cols. That's typically what we have done in our repo.

gtcooke94: Done

gtcooke94 (commented):
Sorry I think I missed submitting this review and my comments

internal/credentials/spiffe.go
```diff
@@ -73,3 +78,40 @@ func SPIFFEIDFromCert(cert *x509.Certificate) *url.URL {
 	}
 	return spiffeID
 }
+
+type partialParsedSPIFFEBundleMap struct {
+	Bundles map[string]json.RawMessage `json:"trust_domains"`
```
easwars, line 83: Nit: Sorry, didn't notice this earlier. But any specific reason for not naming it as `TrustDomains` to match the JSON field name?

internal/credentials/spiffe.go
```diff
@@ -73,3 +78,40 @@ func SPIFFEIDFromCert(cert *x509.Certificate) *url.URL {
...
+	}
+	bundleMap := map[string]*spiffebundle.Bundle{}
+	for td, jsonBundle := range result.Bundles {
+		trustDomain, err := spiffeid.TrustDomainFromString(td)
+		if err != nil {
+			return nil, fmt.Errorf("spiffe: invalid trust domain (%v) found when parsing map file (%v): %v", td, filePath, err)
+		}
+		bundle, err := spiffebundle.Parse(trustDomain, jsonBundle)
+		if err != nil {
+			return nil, fmt.Errorf("spiffe: failed to parse bundle in map file (%v): %v", filePath, err)
```
easwars, line 112: Nit: also please include the trust domain name here for a more specific error message. Thanks.

credentials/tls/certprovider/spiffe/spiffe_test.go
easwars: I see. Had a squint a bit to figure that out :)

Can we please move that out to a separate test to simplify the test logic in both of them. Please see: https://goto.google.com/gotip/episodes/50

easwars (approved):
LGTM modulo minor changes mentioned in this round of review.

.github/workflows/testing.yml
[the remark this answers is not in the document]
matloob: Hi, I'm the maintainer of the `go` command.

`go test google.golang.org/grpc/...` not working doesn't mean that anything is broken.

The `go test google.golang.org/grpc/...` and `go test ./...` commands have surprisingly different semantics. The first command says: "Test every package in a module in the [build list](https://go.dev/ref/mod#glos-build-list) used for the that has the path prefix `google.golang.org/grpc.`" The second command, when run from the root of the module, says: "test every package in this module".

If your goal is to test all the packages in all the modules in your repo, you may want to run `go test ./...` from the root of each of the modules contained in your repo. 

If you only want to test the packages in the top level `google.golang.org/grpc`, but not any of its sub-modules, `go test ./...` from the root of your repo is sufficient.

`go test google.golang.org/grpc/...` does not test all the packages in all the modules in your repo: only those reachable through the dependency graph of the module it's run from (which we call the work module), so you probably don't want to run it. But to explain what's happening when you do that: `go mod tidy` only keeps the dependencies needed to test the the packages in the work module and their dependencies. That means, for example, that we don't keep around the dependencies needed to test a dependency of a test of a dependency. See the [module graph pruning section](https://go.dev/ref/mod#graph-pruning) in the modules reference for more details.

You can use `go get` to get the dependencies needed to run `go test google.golang.org/grpc/...`: If you run `go get google.golang.org/grpc/...` before the test command, it will edit `go.mod` to add the requirements on the necessary modules to test all the matching packages. But running `go mod tidy` after that will then clean up the module requirements and sums to the smaller set needed to build the packages in the work module and their dependencies.

Let me know if you have any additional questions.

scripts/vet-proto.sh
arjan-bal, line 35: I think the complete fix would be to run `go generate` on all directories that contain a `go.mod` file.

.github/workflows/testing.yml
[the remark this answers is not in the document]
dfawley: Thanks for the explanation @matloob.  So essentially it worked before because we had no dependencies that were pruned from our main module but that were required for testing a submodule before?

Did this change (previously) create such a situation?  I thought it just added a new dependency which made its way, not just into our main module, but our root level package as well.

Either way it sounds like we really want the behavior of `./...`, so we can just change it either in this PR (if needed to keep it working) or a separate PR.  The behavior of the other version definitely seems surprising to me.

scripts/vet-proto.sh
gtcooke94: Placed this in a loop

scripts/vet-proto.sh
```diff
@@ -31,9 +31,14 @@ elif [[ "$#" -ne 0 ]]; then
...
 
-# - Check that generated proto files are up to date.
-go generate ./... && git status --porcelain 2>&1 | fail_on_output || \
-(git status; git --no-pager diff; exit 1)
+for MOD_FILE in $(find . -name 'go.mod'); do
+  MOD_DIR=$(dirname ${MOD_FILE})
+  pushd ${MOD_DIR}
+  # - Check that generated proto files are up to date.
+  go generate ./... && git status --porcelain 2>&1 | fail_on_output || \
+  (git status; git --no-pager diff; exit 1)
```
arjan-bal, line 39: nit: We can avoid redundant git operations by carrying out the git check only once, after the loop.
```sh
for MOD_FILE in $(find . -name 'go.mod'); do
  MOD_DIR=$(dirname ${MOD_FILE})
  pushd ${MOD_DIR}
  go generate ./...
  popd
done

# - Check that generated proto files are up to date.
git status --porcelain 2>&1 | fail_on_output || \
  (git status; git --no-pager diff; exit 1)
```

.github/workflows/testing.yml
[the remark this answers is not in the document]
matloob: > Thanks for the explanation @matloob. So essentially it worked before because we had no dependencies that were pruned from our main module but that were required for testing a submodule before?
> 
> Did this change (previously) create such a situation? I thought it just added a new dependency which made its way, not just into our main module, but our root level package as well.

I think the easiest way to think about it is that it before this change `go test google.golang.org/grpc/...` had the same meaning as `go test ./...`. And that was the case because the only module with the `google.golang.org/grpc` prefix in the build list was `google.golang.org/grpc` itself. (You can run `go list -m all` to verify this`.

By adding the dependency on `github.com/spiffe/go-spiffe/v2`, this CL expands the build list to include the `google.golang.org/grpc/examples` module, among others. Now the meaning of `go test google.golang.org/grpc/...` is different from the meaning of `go test ./...`.

Given that the module `google.golang.org/grpc/examples` is now included in the build, what caused it to stop working is that `google.golang.org/grpc/examples` is not needed to test `google.golang.org/gprc` or any of its dependencies in the package dependency tree. While non-test code in `github.com/spiffe/go-spiffe/v2` does import `google.golang.org/gprc/examples`, none of that code is imported by `google.golang.org/grpc`, so we can prune out dependencies needed to build it.

internal/credentials/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,129 @@
...
+		},
+		{
+			filePath: testdata.Path("spiffe/spiffebundle_invalid_trustdomain.json"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.filePath, func(t *testing.T) {
+			_, err := LoadSPIFFEBundleMap(tc.filePath)
+			if err == nil {
```
easwars, line 110: Nit: The assignment and the conditional can be on a single line.

internal/credentials/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,129 @@
...
+		{
+			filePath: testdata.Path("spiffe/spiffebundle_wrong_seq_type.json"),
+		},
+		{
+			filePath: testdata.Path("NOT_A_REAL_FILE"),
+		},
+		{
+			filePath: testdata.Path("spiffe/spiffebundle_invalid_trustdomain.json"),
+		},
+	}
```
easwars, line 105: You could get rid of the `testCases` struct and replace it with a slice of strings.

```golang
filePaths := []string{
	filePath: testdata.Path("spiffe/spiffebundle_corrupted_cert.json"),
	filePath: testdata.Path("spiffe/spiffebundle_malformed.json"),
	filePath: testdata.Path("spiffe/spiffebundle_wrong_kid.json"),
	filePath: testdata.Path("spiffe/spiffebundle_wrong_kty.json"),	
	filePath: testdata.Path("spiffe/spiffebundle_wrong_multi_certs.json"),	
	filePath: testdata.Path("spiffe/spiffebundle_wrong_root.json"),		
	filePath: testdata.Path("spiffe/spiffebundle_wrong_seq_type.json"),		
	filePath: testdata.Path("NOT_A_REAL_FILE"),	
	filePath: testdata.Path("spiffe/spiffebundle_invalid_trustdomain.json"),
}
for _, path := range filePaths {
  t.Run(path, ...)
}
```

internal/credentials/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,129 @@
...
+		t.Run(tc.filePath, func(t *testing.T) {
+			_, err := LoadSPIFFEBundleMap(tc.filePath)
+			if err == nil {
+				t.Fatalf("LoadSPIFFEBundleMap(%v) did not fail but should have.", tc.filePath)
+			}
+		})
+	}
+}
+
+func TestLoadSPIFFEBundleMapFailuresTODO(t *testing.T) {
```
easwars, line 117: What is this TODO about?

security/advancedtls/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	golang.org/x/crypto v0.33.0
 	google.golang.org/grpc v1.70.0
-	google.golang.org/grpc/examples v0.0.0-00010101000000-000000000000
+	google.golang.org/grpc/examples v0.0.0-20230224211313-3775f633ce20
```
easwars, line 9: Why does the advancedtls module depend on examples?

security/advancedtls/examples/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.23
 
 require (
 	google.golang.org/grpc v1.70.0
-	google.golang.org/grpc/examples v0.0.0-00010101000000-000000000000
+	google.golang.org/grpc/examples v0.0.0-20230224211313-3775f633ce20
```
easwars, line 7: Same here. Why is there a dependency on the examples submodule?

internal/credentials/spiffe/spiffe_test.go
gtcooke94: Done

internal/credentials/spiffe/spiffe_test.go
gtcooke94: Done

internal/credentials/spiffe/spiffe_test.go
gtcooke94: Naming it better, fixed

security/advancedtls/go.mod
gtcooke94: The integration test uses protos from the example

https://github.com/grpc/grpc-go/blob/5ac904279519840b8bd2f620be54d3d86d43e695/security/advancedtls/advancedtls_integration_test.go#L36

PR #7970, opened 2024-12-26
cleanup: replace dial with newclient

janardhanvissa:
Partially address: https://github.com/grpc/grpc-go/issues/7049

RELEASE NOTES: None

PR #8141, opened 2025-03-05
interop: Wait for server to become ready in alts tests

arjan-bal:
In #7970, the [alts client was changed](https://github.com/grpc/grpc-go/pull/7970/files#diff-7f71e4c7ed0cdd9defd79e0b7424066d90939c53063b5d413ed5aecb3e65e9b1) from using `grpc.Dial` with `grpc.WithBlock` to `grpc.NewClient` without `WithBlock`.  The tests that run in g3 don't wait for the server to start before starting the client. If WithBlock is not used, the client enters TRANSIENT_FAILURE after getting a `connection refused` error in the first connection attempt. This PR adds `grpc.WaitForReady` while making the RPC to restore the previous behaviour.

RELEASE NOTES: N/A

### DIFF
--- a/interop/alts/client/client.go
+++ b/interop/alts/client/client.go
@@ -47,7 +47,6 @@ func main() {
 		opts.HandshakerServiceAddress = *hsAddr
 	}
 	altsTC := alts.NewClientCreds(opts)
-	// Block until the server is ready.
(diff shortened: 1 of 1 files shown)


purnesh42H (approved):
lgtm

Issue #8146, opened 2025-03-06
"no children to pick from" errors in v1.71.0 when using client-side load-balancing


KnutZuidema:
### What version of gRPC are you using?

v1.71.0

### What version of Go are you using (`go version`)?

1.23.5

### What operating system (Linux, Windows, …) and version?

Linux

### What did you do?

updated from v1.70.0 to v1.71.0

### What did you expect to see?

no functional change

### What did you see instead?

gRPC calls started returning these errors:

```
rpc error: code = Unavailable desc = no children to pick from"
```

### Further information

I am assuming that the error is connected to recent changes in client-side load-balancing, particularly [this change](https://github.com/grpc/grpc-go/pull/7966).

I am using the round-robin load-balancing policy, using a service config like this one:
```
{
  "loadBalancingPolicy": "round_robin"
}
```

arjan-bal:
Hi @KnutZuidema, the error message "no children to pick from" is due to #7966. It means that the resolver produced no addresses for the specified target hostname. Before the change, the same situation would return the error message "produced zero addresses". Are you seeing new failures after updating or only a change in the error message?

arjan-bal:
If you're seeing new failures after upgrading, can you please [turn on gRPC debug logs](https://github.com/grpc/grpc-go?tab=readme-ov-file#how-to-turn-on-logging) and share the logs?

KnutZuidema:
@arjan-bal the errors are new, there were none before the upgrade (i.e. no "produced zero addresses")

I'll try to get the debug logs for this soon.

Another supporting piece of information would potentially be that I am using [kuberesolver](https://github.com/sercand/kuberesolver) for name resolution, but I am not sure whether that would be at all connected. However, I've been using kuberesolver before the update to the new version, and there was no update to kuberesolver itself either.

arjan-bal:
[...]

PR #8149, opened 2025-03-06
grpc: Add endpoints in resolverWrapper.NewAddresses

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/8146

The resolver wrapper has code in the UpdateState method to convert Addresses to Endpoints, but this handling was missing in the NewAddresses API. This PR adds it.

RELEASE NOTES: 
* grpc: Fix bug causing failures with message "no children to pick from" with custom resolvers that use the deprecated [NewAddresses API](https://pkg.go.dev/google.golang.org/grpc@v1.71.0/resolver#ClientConn).

### DIFF
--- a/scripts/vet.sh
+++ b/scripts/vet.sh
@@ -197,7 +197,8 @@ GetSafeRegexMatch
 GetSuffixMatch
 GetTlsCertificateCertificateProviderInstance
 GetValidationContextCertificateProviderInstance
-XXXXX PleaseIgnoreUnused'
+XXXXX PleaseIgnoreUnused
+NewAddress is deprecated:'
   popd
 done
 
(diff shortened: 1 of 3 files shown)


scripts/vet.sh
```diff
@@ -197,7 +197,8 @@ GetSafeRegexMatch
 GetSuffixMatch
 GetTlsCertificateCertificateProviderInstance
 GetValidationContextCertificateProviderInstance
-XXXXX PleaseIgnoreUnused'
+XXXXX PleaseIgnoreUnused
+NewAddress is deprecated:'
```
dfawley, line 201:
```suggestion
NewAddress is deprecated:
XXXXX PleaseIgnoreUnused'
```

That last line is there to avoid need to move the `'` around.  Also, let's try to keep this list sorted, so actually, please move this to right before `NewSubConn is deprecated:`.

dfawley (approved):

PR #8183, opened 2025-03-19
xds: generic xds client resource watching e2e

purnesh42H:
This is the change to make generic xDS client do resource watching from xDS management server. The xDS client uses the  ADS (Aggregated Data Service) stream transport channel  that was created in https://github.com/grpc/grpc-go/pull/8144. The e2e tests are written using the listener resource type from #8144 and grpctransport from #8066.

The PR copies the existing `clientimpl`, `authority`, `clientimpl_watcher`  from internal xdsclient code along with the helpers that are part of gRPC and then modify them to use the generic client types and interfaces.

The PR also adds `String()` and `Equal()` methods for `ServerIdentifierExtension` and implements them for `grpctransport.ServerIdentifierExtension`. It also adds a custom map implementation similar to `resolver.AddrMap` to re-use existing channels for similar server configurations.

Recommend to review each commit of the PR separately in the order.

Commits with copy prefix are copy paste of internal xdsclient code and are followed by modifications. Review only the modification commit.

RELEASE NOTES: None


purnesh42H:
@dfawley could you do a pass on the implementation? For now, I have removed the metrics from client as it requires some more thought. It can be added in separate commit or PR. Otherwise the main implementation can be reviewed to make sure we are following the right approach and change before we go further. Also, I will be adding more e2e tests in subsequent commits

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -39,6 +40,64 @@ type ServerIdentifierExtension struct {
...
+		return true
+	case (sie.Credentials != nil) != (sie2.Credentials != nil):
+		return false
+	case (sie.Credentials.TransportCredentials() != nil) != (sie2.Credentials.TransportCredentials() != nil):
+		return false
+	case (sie.Credentials.PerRPCCredentials() != nil) != (sie2.Credentials.PerRPCCredentials() != nil):
+		return false
+	}
+
+	if sie.Credentials.TransportCredentials() != nil {
```
dfawley, line 81: This is not a guaranteed correct equality check.

We should see if the credentials implements `Equal` and call it.  If not, if they are pointers we could compare those.  But if they are not, we would just have to consider them not equal and maybe log a warning or something.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -39,6 +40,64 @@ type ServerIdentifierExtension struct {
...
+		case sie.Credentials.TransportCredentials().Info().ServerName != sie2.Credentials.TransportCredentials().Info().ServerName:
+			return false
+		}
+	}
+
+	if sie.Credentials.PerRPCCredentials() != nil {
+		if sie.Credentials.PerRPCCredentials().RequireTransportSecurity() != sie2.Credentials.PerRPCCredentials().RequireTransportSecurity() {
+			return false
+		}
+	}
```
dfawley, line 96: Same comment as above, but even more strongly.  `RequireTransportSecurity` is a boolean.  If two credentials both require transport security, there is no reason to assume they are the same credential.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -39,6 +40,64 @@ type ServerIdentifierExtension struct {
 	Credentials credentials.Bundle
 }
 
+// String returns a string representation of the ServerIdentifierExtension.
+func (sie *ServerIdentifierExtension) String() string {
+	if sie.Credentials == nil {
+		return ""
+	}
+	var tcParts []string
+	if sie.Credentials.TransportCredentials() != nil {
```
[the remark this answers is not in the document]
purnesh42H: So, user will have to provide String() implementation for both TransportCredentials and PerRPCCredentials?

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Same, so user is expected provide Equal() implementation for Credentials?

dfawley: No.  Only if they want the debugging messages we would log including them to look pretty.

xds/internal/clients/grpctransport/grpc_transport.go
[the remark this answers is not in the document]
dfawley: No.  Only if they want the debugging messages we would log including them to look pretty.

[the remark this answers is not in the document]
purnesh42H: Modifed String() for both helper and grpctransport. PTAL

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Added comment on docstring for requirement of Equal in `ServerIdentifierConfig.Extensions`. Modified Equal() for both helper and grpctransport. PTAL.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done as suggested above

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,114 @@ package grpctransport
...
 // such as an xDS management or an LRS server.
 type ServerIdentifierExtension struct {
 	// Credentials will be used for all gRPC transports. If it is unset,
 	// transport creation will fail.
+	//
+	// It is recommended to implement `Equal(other any) bool` for
+	// TransportCredentials and PerRPCCredentials, if present. Two
+	// ServerIdentifierExtension will be considered unequal if either does not
+	// implement this method.
 	Credentials credentials.Bundle
```
dfawley, line 49: With the above, this wording should be changed.

Also consider this pattern instead to make it easier to use correctly with the above restrictions:


```go
type SIExt struct {
	Credentials string // The name of the credentials to use for this connection; must be present in Builder.Credentials
}

type Builder struct {
	Credentials map[string]credentials.Bundle
}
```

xds/internal/clients/internal/internal.go
```diff
@@ -33,11 +32,38 @@ import (
...
+	if stringer, ok := si.Extensions.(fmt.Stringer); ok {
+		return fmt.Sprintf("%s-%s", si.ServerURI, stringer)
+	}
+	return fmt.Sprintf("%s-%p", si.ServerURI, si.Extensions)
+}
+
+// ServerIdentifierEqual returns true if si1 and si2 are considered equal. If
+// Extensions are non-nil in both si1 and si2, and any of them don't implement
+// the Equal method, then false is returned.
+func ServerIdentifierEqual(si1, si2 clients.ServerIdentifier) bool {
```
dfawley, line 51: Now we can delete this!

xds/internal/clients/internal/map.go
dfawley, line 1: And this whole file goes away.

xds/internal/clients/xdsclient/xdsclient.go
```diff
@@ -32,15 +30,58 @@
...
+
+	v3statuspb "github.com/envoyproxy/go-control-plane/envoy/service/status/v3"
+	"google.golang.org/protobuf/proto"
+)
+
 // XDSClient is a client which queries a set of discovery APIs (collectively
 // termed as xDS) on a remote management server, to discover
 // various dynamic resources.
 type XDSClient struct {
+	client *clientImpl
```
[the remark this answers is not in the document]
purnesh42H: yeah may be not needed? Wanted to keep the diff from copy paste less. Will check if can be removed without much changes.

xds/internal/clients/config.go
```diff
@@ -59,6 +59,10 @@ type ServerIdentifier struct {
 	//
 	// For example, a custom TransportBuilder might use this field to
 	// configure a specific security credentials.
+	//
+	// If present, Extensions must implement the `Equal(other any) bool`
+	// method. Two ServerIdentifiers with Extensions will be considered unequal
+	// if either Extensions does not implement this method.
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done. This makes the map key simple since Extensions just have string.

xds/internal/clients/internal/internal.go
purnesh42H: Deleted.

xds/internal/clients/internal/map.go
purnesh42H: Deleted.

purnesh42H:
@dfawley i have made the change to use ServerConfig and ServerIdentifier as map key for xdsclient and transport respectively. Now we don't need equal. Only thing is for grpcTransport, we return runtime error if extensions are not `*ServerIdentifierExtension` which is probably okay since that's our transport?

Could you do the pass on last commit and confirm if we can finalize this approach?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,59 @@ package grpctransport
...
+}
+
+// String returns a string representation of the ServerIdentifierExtension.
+//
+// This method is primarily intended for logging and testing purposes. The
+// output returned by this method is not guaranteed to be stable and may change
+// at any time. Do not rely on it for production use.
+func (sie *ServerIdentifierExtension) String() string {
+	return sie.Credentials
 }
```
dfawley, line 55: Why not delete this and just rely on auto-formatting?

xds/internal/clients/xdsclient/xdsclient.go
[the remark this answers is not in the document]
purnesh42H: Removed clientImpl. Moved all methods under XDSClient

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: If we remove this then we will have to switch to auto-formatting here as well `ServerIdentifierEqual` http://github.com/grpc/grpc-go/pull/8183/files#diff-86488560c4bc0bfecb05e056137dbaa8034a4efc79a02c1b07e063d3c4e901f9R38

Should we always just do `fmt.Sprintf("%s-%v", si.ServerURI, si.Extensions`?

dfawley: We shouldn't need `ServerIdentifierEqual` anymore.

dfawley: Why not just `logger.Info("blah blah %v blah blah", si)`?

dfawley: (OR `%+v` to get field names if desired)

purnesh42H: Done

purnesh42H:
@dfawley I have made the change to restrict ServerConfigExtension to be added as value and have added csds dump tests. PTAL. Thanks

xds/internal/clients/config_test.go
```diff
@@ -0,0 +1,310 @@
...
+	grpctest.Tester
+}
+
+func Test(t *testing.T) {
+	grpctest.RunSubTests(t, s{})
+}
+
+type testServerIdentifierExtension struct{ x int }
+
+func (s) TestServerIdentifierMap_Exist(t *testing.T) {
```
dfawley, line 37: What are these tests for?  It feels like it's testing `map` itself.  I think we could delete these, but maybe I'm missing something.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,55 @@ package grpctransport
...
 
+var logger = grpclog.Component("grpctransport")
+
 // ServerIdentifierExtension holds settings for connecting to a gRPC server,
 // such as an xDS management or an LRS server.
+//
+// It must be added as value to the clients.ServerIdentifier.Extensions field.
+//
+// Example:
+// clients.ServerIdentifier{ServerURI: "localhost:5678", Extensions: grpctransport.ServerIdentifierExtension{Credentials: "local"}}
```
dfawley, line 44: Examples should be in an `_test.go` file, in a separate package (in this directory, `package grpc_transport_test`).

https://go.dev/blog/examples

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,55 @@ package grpctransport
...
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/xds/internal/clients"
 )
 
+var logger = grpclog.Component("grpctransport")
+
 // ServerIdentifierExtension holds settings for connecting to a gRPC server,
 // such as an xDS management or an LRS server.
+//
+// It must be added as value to the clients.ServerIdentifier.Extensions field.
```
dfawley, line 41:
```suggestion
// It must be set by value (not pointer) in the clients.ServerIdentifier.Extensions
// field. (See examples.)
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,55 @@ package grpctransport
...
-type Builder struct{}
+type Builder struct {
+	// Credentials is a map of credentials names to credentials.Bundle which
+	// can be used to connect to the server.
+	Credentials map[string]credentials.Bundle
+
+	mu sync.Mutex
+	// serverIdentifierMap is a map of clients.ServerIdentifiers in use by the
+	// Builder to connect to different servers.
+	serverIdentifierMap map[clients.ServerIdentifier]any
```
dfawley, line 63: Why `any`?  Surely we know the type?  (`*grpcTransport`)?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -74,12 +114,15 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 		Timeout: 20 * time.Second,
 	})
-	cc, err := grpc.NewClient(si.ServerURI, kpCfg, grpc.WithCredentialsBundle(sce.Credentials), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
+	cc, err := grpc.NewClient(si.ServerURI, kpCfg, grpc.WithCredentialsBundle(credentialsToUse), grpc.WithDefaultCallOptions(grpc.ForceCodec(&byteCodec{})))
 	if err != nil {
 		return nil, fmt.Errorf("grpctransport: failed to create transport to server %q: %v", si.ServerURI, err)
 	}
+	tc := &grpcTransport{cc: cc}
+	// Add the newly created transport to the map to re-use the connection.
+	b.serverIdentifierMap[si] = tc
```
dfawley, line 123: You're never deleting this.  You need to ref count the transports and delete when there are zero refs.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -57,15 +87,25 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
 	if !ok {
 		return nil, fmt.Errorf("grpctransport: Extensions field is %T, but must be %T in ServerIdentifier", si.Extensions, ServerIdentifierExtension{})
 	}
-	if sce.Credentials == nil {
-		return nil, fmt.Errorf("grptransport: Credentials field is not set in ServerIdentifierExtension")
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	var credentialsToUse credentials.Bundle
```
dfawley, line 94: Nit:

```suggestion
	var creds credentials.Bundle
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -57,15 +87,25 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
+	if credentialsToUse, ok = b.Credentials[sce.Credentials]; !ok {
+		return nil, fmt.Errorf("grptransport: Credentials %s in ServerIdentifierExtension is not present in the Builder", sce.Credentials)
 	}
 
-	// TODO: Incorporate reference count map for existing transports and
-	// deduplicate transports based on the provided ServerIdentifier so that
-	// transport channel to same server can be shared between xDS and LRS
-	// client.
+	if value, ok := b.serverIdentifierMap[si]; ok {
```
dfawley, line 99: Nit: do this before looking up the credentials.  You don't need the creds if there's a match.

xds/internal/clients/xdsclient/authority.go
```diff
@@ -0,0 +1,862 @@
...
+func (a *authority) xdsChannelToUse() (*xdsChannelWithConfig, error) {
+	if a.activeXDSChannel != nil {
+		return a.activeXDSChannel, nil
+	}
+
+	sc := a.xdsChannelConfigs[0].serverConfig
+	xc, cleanup, err := a.getChannelForADS(sc, a)
+	if err != nil {
+		return nil, err
+	}
```
dfawley, line 753: We should have testing for failed channel construction.

Or we can say that channel construction must not fail, and return a channel that returns streams that only return errors.

I'm fine with either option.  Java and C++ don't actually fail channel construction like Go does.

xds/internal/clients/xdsclient/xdsclient.go
```diff
@@ -52,16 +102,26 @@ func New(config Config) (*XDSClient, error) {
...
 func (c *XDSClient) WatchResource(typeURL, name string, watcher ResourceWatcher) (cancel func()) {
-	panic("unimplemented")
+	return c.watchResource(typeURL, name, watcher)
 }
 
 // Close closes the xDS client.
 func (c *XDSClient) Close() error {
-	panic("unimplemented")
+	c.close()
+	return nil
```
dfawley, line 111: Delete and rename the other one?

xds/internal/clients/xdsclient/xdsclient.go
```diff
@@ -52,16 +102,26 @@ func New(config Config) (*XDSClient, error) {
...
 
 // DumpResources returns the status and contents of all xDS resources being
 // watched by the xDS client.
-func (c *XDSClient) DumpResources() []byte {
-	panic("unimplemented")
+func (c *XDSClient) DumpResources() ([]byte, error) {
+	resp := &v3statuspb.ClientStatusResponse{}
+	cfg := c.dumpResources()
+	resp.Config = append(resp.Config, cfg)
+	return proto.Marshal(resp)
```
dfawley, line 120: Move to `dumpResources`?

xds/internal/clients/xdsclient/authority.go
```diff
@@ -0,0 +1,862 @@
...
+
+// serverIndexForConfig returns the index of the xdsChannelConfig matching the
+// provided server config, panicking if no match is found (which indicates a
+// programming error).
+func (a *authority) serverIndexForConfig(sc *ServerConfig) int {
+	for i, cfg := range a.xdsChannelConfigs {
+		if isServerConfigEqual(sc, cfg.serverConfig) {
+			return i
+		}
+	}
```
dfawley, line 252: This can be a map now?

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Added. Thanks for info!

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: yeah it can be `*grpcTransport`

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Moved

xds/internal/clients/xdsclient/xdsclient.go
purnesh42H: Deleted this and exported the other one

xds/internal/clients/xdsclient/authority.go
purnesh42H: I don't think we can use the map here because order of servers matter here. For each authority, first index server is primary and then then rest are fallbacks with decreasing priority at higher indexes. See https://github.com/grpc/grpc-go/pull/8183/files#diff-210975ef13af9d19ee41dbc773691f6275759787edaa2ea8e680b2673fd48392R108

dfawley: Oh, this is _that_ part...

I think this can be restructured to simplify it and match the semantics better.  Within an authority, we should only be using a single server at a time.  If it fails, we fall back to the next one.  So we should always know which one is active, and they should know their own indices, so we should never need to look them up by their config.

But I'm fine with saving that as a cleanup for later if the changes aren't obvious.  It's probably not simple.

xds/internal/clients/config.go
```diff
@@ -59,6 +59,11 @@ type ServerIdentifier struct {
 	//
 	// For example, a custom TransportBuilder might use this field to
 	// configure a specific security credentials.
+	//
+	// Extensions must be able to be used as a map key, or the client may
```
easwars, line 63: How about:

// Extensions may be any type that is comparable, as they are used as map keys internally.
// See: https://go.dev/ref/spec#Comparison_operators.

xds/internal/clients/config.go
```diff
@@ -59,6 +59,11 @@ type ServerIdentifier struct {
...
+	// panic. Any equivalent extensions in all ServerIdentifiers present in a
```
easwars, line 64: Can you please help me understand this.

	// Any equivalent extensions in all ServerIdentifiers present in a
	// single client's configuration should have the same value.  Not following
	// this restriction may result in excess resource usage.

xds/internal/clients/grpctransport/example_server_identifier_extension_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+
+	"google.golang.org/grpc/xds/internal/clients"
+	"google.golang.org/grpc/xds/internal/clients/grpctransport"
+)
+
+func ExampleServerIdentifierExtension() {
+	si := clients.ServerIdentifier{ServerURI: "localhost:5678", Extensions: grpctransport.ServerIdentifierExtension{Credentials: "local"}}
+	fmt.Printf("%v", si)
```
easwars, line 30: There needs to a comment with the expected output. Otherwise, this does nothing.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,53 @@ package grpctransport
...
 	"google.golang.org/grpc/xds/internal/clients"
 )
 
+var logger = grpclog.Component("grpctransport")
+
 // ServerIdentifierExtension holds settings for connecting to a gRPC server,
 // such as an xDS management or an LRS server.
+//
+// It must be set by value (not pointer) in the
+// clients.ServerIdentifier.Extensions field (See Example).
```
easwars, line 42: Why? Pointer types are comparable too? Maybe I'm missing something.

https://go.dev/ref/spec#Comparison_operators

easwars:
> The PR copies the existing clientimpl, authority, clientimpl_watcher from internal xdsclient code along with the helpers that are part of gRPC and then modify them to use the generic client types and interfaces.

Didn't we at some point say that we will move the code as-in in one PR and then make changes in a follow-up, so that we can clearly see what is changing. If that is not feasible, could you please describe the changes in detail in each of those non-test files that are being copied over. Thanks.

xds/internal/clients/xdsclient/authority.go
purnesh42H: I have added the tests to cover the channel creation failure by copying the `misc_watcher_test.go` and modifying it. There was one test in that file `TestNodeProtoSentOnlyInFirstRequest` which had slightly different expectations because of watchers of two different types. In our case though since we only have listener type, there is a slight modification.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Added ref count support in separate commit

xds/internal/clients/config_test.go
purnesh42H: yeah they are not needed. Removed.

xds/internal/clients/config.go
purnesh42H: this basically means if all the fields in ServerIdentifier are not equal between 2 server identifiers, then they will create a separate transport to the server. So, even if they are pointing to same server but some fields are different they will create two separate transports to that server.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: With pointer there are 2 issues
1) two different pointer with same values are unequal
2) same pointers with different values are equal

So, we decided to go by value

xds/internal/clients/config.go
purnesh42H: Added. Thanks

xds/internal/clients/grpctransport/example_server_identifier_extension_test.go
purnesh42H: Oh actually, you need to mention the output as a comment below. I added. See https://go.dev/blog/examples

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,53 @@ package grpctransport
...
+	Credentials string
 }
 
 // Builder creates gRPC-based Transports. It must be paired with ServerIdentifiers
 // that contain an Extension field of type ServerIdentifierExtension.
-type Builder struct{}
+type Builder struct {
+	// Credentials is a map of credentials names to credentials.Bundle which
+	// can be used to connect to the server.
+	Credentials map[string]credentials.Bundle
```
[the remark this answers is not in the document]
purnesh42H: No. Unexported

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,53 @@ package grpctransport
...
+
+	mu sync.Mutex
+	// serverIdentifierMap is a map of clients.ServerIdentifiers in use by the
+	// Builder to connect to different servers.
+	serverIdentifierMap map[clients.ServerIdentifier]*grpcTransport
```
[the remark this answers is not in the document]
purnesh42H: Renamed.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -57,15 +85,25 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 	}
-	if sce.Credentials == nil {
-		return nil, fmt.Errorf("grptransport: Credentials field is not set in ServerIdentifierExtension")
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if value, ok := b.serverIdentifierMap[si]; ok {
+		if logger.V(2) {
+			logger.Info("Reusing existing connection to the server for ServerIdentifier: %v", si)
```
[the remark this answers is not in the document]
purnesh42H: Renamed

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -57,15 +85,25 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
-	// deduplicate transports based on the provided ServerIdentifier so that
-	// transport channel to same server can be shared between xDS and LRS
-	// client.
+	var creds credentials.Bundle
+	if creds, ok = b.Credentials[sce.Credentials]; !ok {
+		return nil, fmt.Errorf("grptransport: Credentials %s in ServerIdentifierExtension is not present in the Builder", sce.Credentials)
+	}
 
+	if logger.V(2) {
+		logger.Info("Creating a new connection to the server for ServerIdentifier: %v", si)
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -123,32 +123,93 @@ func (s *testServer) StreamAggregatedResources(stream v3discoverygrpc.Aggregated
...
+		t.Fatalf("Got nil `tr1` transport from Build(serverID1), want non-nil")
+	}
+	if tr1.(*grpcTransport).cc == nil {
+		t.Fatalf("Got nil grpc.ClientConn in transport `tr1`, want non-nil")
+	}
+	if len(b.serverIdentifierMap) != 1 {
+		t.Fatalf("Builder.serverIdentifierMap has unexpected length %d, want 1", len(b.serverIdentifierMap))
+	}
+	if b.serverIdentifierMap[serverID1] != tr1 {
+		t.Fatalf("Builder.serverIdentifierMap[serverID1] = %v, want %v", b.serverIdentifierMap[serverID1], tr1)
```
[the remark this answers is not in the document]
purnesh42H: yeah i simplified this test to just create transports. The dedupe part is being tested in refcounted_transport_test.go which is using hooks.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -91,16 +89,16 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 	if value, ok := b.serverIdentifierMap[si]; ok {
 		if logger.V(2) {
 			logger.Info("Reusing existing connection to the server for ServerIdentifier: %v", si)
 		}
-		return value.(*grpcTransport), nil
+		return value, nil
+	}
+
+	var creds credentials.Bundle
+	if creds, ok = b.Credentials[sce.Credentials]; !ok {
```
dfawley: Nit: this would more commonly be written to use type inference:
```go
creds, ok := b.Credentials[sce.Credentials]
if !ok {
```

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -116,15 +125,24 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
-	return tc, nil
+	b.serverIdentifierMap[si] = tr
+	grpcTransportCreateHook()
+	return &transportRef{grpcTransport: tr}, nil
 }
 
 type grpcTransport struct {
-	cc *grpc.ClientConn
+	cc                           *grpc.ClientConn
+	refCount                     int32  // accessed atomically
```
dfawley, line 144: If atomic, then it should be `atomic.Int32`.  But you need a lock otherwise there will be races between transport re-use and dereferencing.  In which case you don't need an atomic here.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -138,9 +156,38 @@ func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.S
 
 // Close closes the gRPC channel to the server.
 func (g *grpcTransport) Close() error {
+	if g.decrRef() != 0 {
+		return nil
+	}
+
+	g.deleteFromServerIdentiferMap()
```
dfawley, line 163: Yeah this is a race.  If you dec the counter then take the lock, then re-use happens, then you're going to delete a used transport.  Best to take the lock and remove the atomic and keep things simple.  Performance does not matter on these code paths.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -138,9 +156,38 @@ func (g *grpcTransport) NewStream(ctx context.Context, method string) (clients.S
...
+	*grpcTransport
+
+	closed sync.Once
+}
+
+func (tr *transportRef) Close() error {
+	var err error
+	tr.closed.Do(func() {
+		err = tr.grpcTransport.Close()
+	})
```
dfawley, line 187: I think we can guarantee this is only called once, so the Once shouldn't be needed.  No?

xds/internal/clients/grpctransport/transport_refcounted_test.go
```diff
@@ -0,0 +1,268 @@
...
+		"local": &testCredentials{transportCredentials: local.NewCredentials()},
+	}
+
+	// Override the gRPC transport creation hook to get notified.
+	origGRPCTransportCreateHook := grpcTransportCreateHook
+	grpcTransportCreateCh := testutils.NewChannelWithSize(1)
+	grpcTransportCreateHook = func() {
+		grpcTransportCreateCh.Replace(struct{}{})
+	}
+	defer func() { grpcTransportCreateHook = origGRPCTransportCreateHook }()
```
dfawley, line 56: Maybe you don't need the hooks if you just check the pointer of the returned transport to confirm you are getting the same one when expected, or a new when when re-use is not expected?

xds/internal/clients/grpctransport/example_server_identifier_extension_test.go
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/example_server_identifier_extension_test.go
purnesh42H: Added the docstring. Still printing though but combined pink and initialization in single line.

purnesh42H: By examples, do you mean add more examples?

xds/internal/clients/grpctransport/example_server_identifier_extension_test.go
```diff
@@ -0,0 +1,31 @@
...
+package grpctransport_test
+
+import (
+	"fmt"
+
+	"google.golang.org/grpc/xds/internal/clients"
+	"google.golang.org/grpc/xds/internal/clients/grpctransport"
+)
+
+func ExampleServerIdentifierExtension() {
```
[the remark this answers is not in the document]
purnesh42H: Looks like it requires to follow format "Example<struct name>" else it gives warning that the struct is not defined.

xds/internal/clients/grpctransport/example_server_identifier_extension_test.go
```diff
@@ -0,0 +1,31 @@
...
+	si := clients.ServerIdentifier{ServerURI: "localhost:5678", Extensions: grpctransport.ServerIdentifierExtension{Credentials: "local"}}
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: incrRef and decRef were doing atomic add and subtract so the var itself need not be atomic? But yeah you correctly pointed out the race condition of deleting the transport and at the someone else getting it. I have switched to lock so now both decrement and deletion are serialized.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: yeah fixed using lock. Thanks for pointing it out.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: How will you guarantee? I think we can guarantee in our implementation may be but this transportRef is returned to the user of grpcTransport which can be used by anyone who is using our xdsclient so we should make sure its idempotent.

xds/internal/clients/grpctransport/transport_refcounted_test.go
purnesh42H: I am not sure if that's a better alternative. As @easwars pointed out, then we are being tied to implementation for assertion. So, in this case we will have a get method in transportReference which will return underlying transport but then to compare we will have to type cast etc. to compare. Also, for close we still need a hook? So, better to be consistent?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -74,16 +112,33 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 	if err != nil {
 		return nil, fmt.Errorf("grpctransport: failed to create transport to server %q: %v", si.ServerURI, err)
 	}
-
-	return &grpcTransport{cc: cc}, nil
+	tr := &grpcTransport{
+		cc:       cc,
+		refCount: 1,
+		deleteTransport: func() {
+			b.mu.Lock()
```
dfawley, line 123: This nests mutexes, which makes me nervous.

Can we eliminate `grpcTransport.mu` and use `b.mu` for everything instead?

Otherwise, we may need to document that we can never take a `grpcTransport.mu` while `Builder.mu` is held.  ...Which, we actually _are_ doing.  So I think there's already a deadlock race:

thread 1: `Builder.Build` called: `b.mu.Lock()` (line 91)
thread 2: `grpcTransport.Close` called: `g.mu.Lock()` (line 155)
thread 1: `tr.incrRef` called (line 98), `g.mu.Lock()` (line 168) blocks because thread 2 holds it
thread 2: `g.deleteTransport` called (line 163), `b.mu.Lock()` (here) blocks because thread 1 holds it

purnesh42H: I have moved both incr and decr ref count under b.mu along with deleting the transport from map.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -119,11 +119,23 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 		cc:       cc,
 		refCount: 1,
-		deleteTransport: func() {
-			b.mu.Lock()
-			defer b.mu.Unlock()
-			delete(b.transports, si)
-		}}
+	}
+	// Register a cleanup function that decrements the refCount to the gRPC
+	// transport each time Close() is called or close it and remove from
```
dfawley, line 124:
```suggestion
	// transport each time Close() is called to close it and remove from
```

?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -136,9 +148,12 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
 type grpcTransport struct {
 	cc *grpc.ClientConn
 
-	mu              sync.Mutex
-	refCount        int32  // accessed atomically
-	deleteTransport func() // called when refCount reaches 0
+	// refCount is accessed atomically to update the count of references to
+	// the gRPC transport.
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Changed

xds/internal/clients/xdsclient/test/authority_test.go
```diff
@@ -0,0 +1,379 @@
...
+	// Stop the secondary.
+	secondaryLis.Close()
+
+	// Ensure that the connectivity error callback is called.
+	if _, err := watcher.errCh.Receive(ctx); err != nil {
+		t.Fatal("Timeout when waiting for error callback on the watcher")
+	}
+}
+
+type listenerWatcherV2 struct {
```
easwars, line 340: We have a listener watcher defined in the lds_watchers_test.go. Can we have a single watcher (given that everything now is in the same package).

xds/internal/clients/xdsclient/clientimpl.go
```diff
@@ -0,0 +1,357 @@
...
+		channelToClose := state.channel
+		c.channelsMu.Unlock()
+
+		channelToClose.close()
+	})
+}
+
+// DumpResources returns the status and contents of all xDS resources being
+// watched by the xDS client.
+func (c *XDSClient) DumpResources() ([]byte, error) {
```
easwars, line 278: Why are we returning a byte slice here instead of the actual proto?

xds/internal/clients/xdsclient/test/dump_test.go
```diff
@@ -0,0 +1,491 @@
...
+	mgmtServer2 := e2e.StartManagementServer(t, e2e.ManagementServerOptions{})
+
+	// The first of the above management servers will be the top-level xDS
+	// server in the bootstrap configuration, and the second will be the xDS
+	// server corresponding to the test authority.
+	nodeID := uuid.New().String()
+
+	resourceTypes := map[string]xdsclient.ResourceType{}
+	listenerType := listenerType
+	resourceTypes[xdsresource.V3ListenerURL] = listenerType
```
easwars, line 321: For tests in this file too.

xds/internal/clients/xdsclient/test/dump_test.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+
+	var lastErr error
+	for ; ctx.Err() == nil; <-time.After(defaultTestShortTimeout) {
+		b, err := client.DumpResources()
+		if err != nil {
+			lastErr = err
+			continue
+		}
+		var got v3statuspb.ClientStatusResponse
```
easwars, line 79: If this line is changed to
```golang
var got := &v3statuspb.ClientStatusResponse{}
```
then you can avoid the `&` operator in all places where `got` is used below.

xds/internal/clients/xdsclient/clientimpl_watchers.go
```diff
@@ -0,0 +1,119 @@
...
+	// Return early if the client is already closed.
+	//
+	// The client returned from the top-level API is a ref-counted client which
+	// contains a pointer to `XDSClient`. When all references are released, the
+	// ref-counted client sets its pointer to `nil`. And if any watch APIs are
+	// made on such a closed client, we will get here with a `nil` receiver.
+
+	watcher = &wrappingWatcher{
+		ResourceWatcher: watcher,
+		nodeID:          c.config.Node.ID,
```
easwars, line 60: This will panic if `c` is `nil`. We probably don't have the case of `c` being `nil` in the new world where the xDS client is no longer reference counted, but the check for `c.done.HasFired` definitely needs to be done up front.

xds/internal/clients/xdsclient/test/misc_watchers_test.go
```diff
@@ -0,0 +1,538 @@
...
+	// Start an xDS management server and set the option to allow it to respond
+	// to requests which only specify a subset of the configured resources.
+	mgmtServer := e2e.StartManagementServer(t, e2e.ManagementServerOptions{AllowResourceSubset: true})
+
+	nodeID := uuid.New().String()
+	authority := makeAuthorityName(t.Name())
+
+	resourceTypes := map[string]xdsclient.ResourceType{}
+	listenerType := listenerType
+	resourceTypes[xdsresource.V3ListenerURL] = listenerType
```
easwars, line 111: And this file too.

xds/internal/clients/xdsclient/test/misc_watchers_test.go
```diff
@@ -0,0 +1,538 @@
...
+			if !strings.Contains(gotErr.Error(), nodeID) {
+				t.Fatalf("Unexpected error: %v, want error with node ID: %q", err, nodeID)
+			}
+		}
+	})
+}
+
+// testTransportBuilder is a transport builder which always returns a nil
+// transport along with an error.
+type testTransportBuilder struct{}
```
easwars, line 475: Probably name it something like `erroringTransportBuilder` so that it is very clear from its name that it is supposed to return errors.

easwars:
Apologies for the delay.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -23,25 +23,55 @@ package grpctransport
...
+// It must be set by value (not pointer) in the
+// clients.ServerIdentifier.Extensions field (See Example).
 type ServerIdentifierExtension struct {
-	// Credentials will be used for all gRPC transports. If it is unset,
-	// transport creation will fail.
-	Credentials credentials.Bundle
+	// Credentials is name of the credentials to use for this transport to the
+	// server.
+	//
+	// It must be present in Builder.Credentials.
```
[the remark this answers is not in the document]
purnesh42H: Changed

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -57,14 +87,22 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 	}
 
-	// TODO: Incorporate reference count map for existing transports and
-	// deduplicate transports based on the provided ServerIdentifier so that
-	// transport channel to same server can be shared between xDS and LRS
-	// client.
+	creds, ok := b.credentials[sce.Credentials]
+	if !ok {
+		return nil, fmt.Errorf("grpctransport: unknown credentials type %q specified in extensions", sce.Credentials)
+	}
```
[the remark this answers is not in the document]
purnesh42H: yeah i think had it before but moved it below because of another comment calling out that we don't need to check for credentials if we are reusing anyway. But yeah with mutex it make sense to keep it separate. Moved above.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -74,16 +112,47 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
 		return nil, fmt.Errorf("grpctransport: failed to create transport to server %q: %v", si.ServerURI, err)
 	}
+	tr := &grpcTransport{
+		cc:       cc,
+		refCount: 1,
+	}
+	// Register a cleanup function that decrements the refCount to the gRPC
+	// transport each time Close() is called to close it and remove from
+	// transports map if last reference is being released.
+	tr.cleanup = func() {
```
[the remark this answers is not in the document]
purnesh42H: I have removed transportRef. I did what you suggested in other comment to maintain count of refs in a map. The sync.Once is now in grpcTransport.

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -74,16 +112,47 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
+		logger.Info("Created a new transport to the server for ServerIdentifier: %v", si)
+	}
+	return &transportRef{grpcTransport: tr}, nil
 }
 
 type grpcTransport struct {
 	cc *grpc.ClientConn
+
+	// refCount is guarded by the parent Builder.mu.
+	refCount int32
```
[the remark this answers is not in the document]
purnesh42H: yeah that's a better approach. The only difference I did here is to maintain serverIdentifier -> clientConn map so that each Build() call returns a new grpctransport instance but keep the underlying clientconn same.

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -123,33 +124,55 @@ func (s *testServer) StreamAggregatedResources(stream v3discoverygrpc.Aggregated
...
 }
+func (tc *testCredentials) PerRPCCredentials() credentials.PerRPCCredentials {
+	return tc.perRPCCredentials
+}
 
 // TestBuild_Success verifies that the Builder successfully creates a new
-// Transport with a non-nil grpc.ClientConn.
+// Transport in both cases when provided clients.ServerIdentifer is same
+// one of the existing transport or a new one.
 func (s) TestBuild_Success(t *testing.T) {
```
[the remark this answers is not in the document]
purnesh42H: NA now? i have modified the refcount tests to use listener wrapper.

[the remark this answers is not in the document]
purnesh42H: yeah having dialOption was there in the discussion but probably good to do when we really need it.

xds/internal/clients/grpctransport/transport_refcounted_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"google.golang.org/grpc/xds/internal/clients"
+
+	v3discoverypb "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+)
+
+// TestBuild_Single tests that multiple calls to Build() with the same
+// clients.ServerIdentifier returns the same transport. Also verifies that
+// only when all references to the newly created transport are released,
+// the underlying transport is closed.
+func (s) TestBuild_Single(t *testing.T) {
```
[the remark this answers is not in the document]
purnesh42H: i have the refcount test in separate package so its probably fine to keep it. Its just verifying the creation of transport without actually connecting. To connect, we need to start the stream.

xds/internal/clients/grpctransport/transport_refcounted_test.go
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package grpctransport
```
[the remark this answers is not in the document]
purnesh42H: yes, done. I think that's a better approach.

xds/internal/clients/grpctransport/transport_refcounted_test.go
```diff
@@ -0,0 +1,209 @@
...
+		func() {
+			transports[i], err = builder.Build(serverID)
+			if err != nil {
+				t.Fatalf("Failed to build transport: %v", err)
+			}
+			if (tr.(*transportRef)).grpcTransport != (transports[i].(*transportRef)).grpcTransport {
+				t.Fatalf("Wanted underlying gRPC transport to be reused, but got different transport: %v", (transports[i].(*transportRef)).grpcTransport)
+			}
+		}()
+	}
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/grpctransport/grpc_transport_test.go
```diff
@@ -123,33 +124,55 @@ func (s *testServer) StreamAggregatedResources(stream v3discoverygrpc.Aggregated
 type testCredentials struct {
 	credentials.Bundle
 	transportCredentials credentials.TransportCredentials
+	perRPCCredentials    credentials.PerRPCCredentials
```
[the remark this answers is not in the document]
purnesh42H: removed perRPCCredentials. But we need testCredentials because local.NewCredentials() returns a transportCredentials only but we pass credentials.Bundle to Builder

xds/internal/clients/xdsclient/clientimpl_watchers.go
purnesh42H: Added. Thanks for catching this.

xds/internal/clients/xdsclient/clientimpl_watchers.go
```diff
@@ -0,0 +1,119 @@
...
+//
+// The returned function cancels the watch and prevents future calls to the
+// watcher.
+func (c *XDSClient) WatchResource(typeURL, resourceName string, watcher ResourceWatcher) (cancel func()) {
+	// Return early if the client is already closed.
+	//
+	// The client returned from the top-level API is a ref-counted client which
+	// contains a pointer to `XDSClient`. When all references are released, the
+	// ref-counted client sets its pointer to `nil`. And if any watch APIs are
+	// made on such a closed client, we will get here with a `nil` receiver.
```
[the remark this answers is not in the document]
purnesh42H: yeah added c.done.HasFired check.

xds/internal/clients/xdsclient/clientimpl_watchers.go
```diff
@@ -0,0 +1,119 @@
...
+// A registry of ResourceType implementations indexed by their corresponding
+// type URLs. Registration of an ResourceType happens the first time a watch
+// for a resource of that type is invoked.
+type resourceTypeRegistry struct {
+	types map[string]ResourceType
+}
+
+func newResourceTypeRegistry(resourceTypes map[string]ResourceType) *resourceTypeRegistry {
+	return &resourceTypeRegistry{types: resourceTypes}
+}
```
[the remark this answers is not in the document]
purnesh42H: Removed and directly used config.ResourceTypes

xds/internal/clients/xdsclient/clientimpl.go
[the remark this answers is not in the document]
purnesh42H: Moved.  Most of the the things like this was mainly done to keep the diff lesser for easier review.

xds/internal/clients/xdsclient/test/helpers_test.go
```diff
@@ -0,0 +1,277 @@
...
+	listenerResourceTypeName = "ListenerResource"
+
+	ldsName         = "xdsclient-test-lds-resource"
+	rdsName         = "xdsclient-test-rds-resource"
+	cdsName         = "xdsclient-test-cds-resource"
+	edsName         = "xdsclient-test-eds-resource"
+	ldsNameNewStyle = "xdstp:///envoy.config.listener.v3.Listener/xdsclient-test-lds-resource"
+	rdsNameNewStyle = "xdstp:///envoy.config.route.v3.RouteConfiguration/xdsclient-test-rds-resource"
+	cdsNameNewStyle = "xdstp:///envoy.config.cluster.v3.Cluster/xdsclient-test-cds-resource"
+	edsNameNewStyle = "xdstp:///envoy.config.endpoint.v3.ClusterLoadAssignment/xdsclient-test-eds-resource"
```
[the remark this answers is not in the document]
purnesh42H: Removed.

xds/internal/clients/xdsclient/test/lds_watchers_test.go
```diff
@@ -0,0 +1,1432 @@
...
+	onDone()
+}
+
+type listenerUpdateErrTuple struct {
+	update      listenerUpdate
+	resourceErr error
+	ambientErr  error
+}
+
+type listenerWatcher struct {
```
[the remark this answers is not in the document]
purnesh42H: Done and removed the extra listenerWatcher

xds/internal/clients/xdsclient/test/helpers_test.go
```diff
@@ -0,0 +1,277 @@
...
+	// Singleton instantiation of the resource type implementation.
+	listenerType = xdsclient.ResourceType{
+		TypeURL:                    xdsresource.V3ListenerURL,
+		TypeName:                   listenerResourceTypeName,
+		AllResourcesRequiredInSotW: true,
+		Decoder:                    listenerDecoder{},
+	}
+)
+
+func unmarshalListenerResource(r []byte) (string, listenerUpdate, error) {
```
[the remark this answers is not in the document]
purnesh42H: You are right. I removed some of the unused things. The only remaining part to simplify is `processClientSideListener`. This is mainly being used to differentiate between bad and good listener based on routeSpecifier. I was thinking of simplifying this after fully migrating and had it already in my fast follow up checklist.

xds/internal/clients/xdsclient/test/lds_watchers_test.go
```diff
@@ -0,0 +1,1432 @@
...
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			mgmtServer := e2e.StartManagementServer(t, e2e.ManagementServerOptions{})
+
+			nodeID := uuid.New().String()
+
+			resourceTypes := map[string]xdsclient.ResourceType{}
+			listenerType := listenerType
+			resourceTypes[xdsresource.V3ListenerURL] = listenerType
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/xdsclient/test/authority_test.go
```diff
@@ -0,0 +1,379 @@
...
+	nonDefaultAuthorityServer := e2e.StartManagementServer(t, e2e.ManagementServerOptions{Listener: lisNonDefault})
+
+	// Create a bootstrap configuration with two non-default authorities which
+	// have empty server configs, and therefore end up using the default server
+	// config, which points to the above management server.
+	nodeID := uuid.New().String()
+
+	resourceTypes := map[string]xdsclient.ResourceType{}
+	listenerType := listenerType
+	resourceTypes[xdsresource.V3ListenerURL] = listenerType
```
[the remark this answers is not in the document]
purnesh42H: Done

xds/internal/clients/xdsclient/test/authority_test.go
```diff
@@ -0,0 +1,379 @@
...
+	authorityTestResourceName3 = buildResourceName(listenerResourceTypeName, testAuthority3, ldsName+"3", nil)
+)
+
+// setupForAuthorityTests spins up two management servers, one to act as the
+// default and the other to act as the non-default. It also creates a
+// xDS client configuration with three authorities (the first two pointing to
+// the default and the third one pointing to the non-default).
+//
+// Returns two listeners used by the default and non-default management servers
+// respectively, and the xDS client and its close function.
```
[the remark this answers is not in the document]
purnesh42H: Removed

xds/internal/clients/xdsclient/test/authority_test.go
purnesh42H: Yes, done as part of making the watcher with 3 channels.

xds/internal/clients/xdsclient/clientimpl.go
purnesh42H: That's what we decided for the user facing API similar to ResourceType `Bytes` function

xds/internal/clients/xdsclient/test/dump_test.go
purnesh42H: Done

xds/internal/clients/xdsclient/test/dump_test.go
purnesh42H: Done

xds/internal/clients/xdsclient/test/misc_watchers_test.go
purnesh42H: Done

xds/internal/clients/xdsclient/test/misc_watchers_test.go
purnesh42H: Done

purnesh42H:
> Apologies for the delay.

thanks @easwars. I have addressed all the comments. Could you please do another pass?

xds/internal/clients/grpctransport/grpc_transport.go
```diff
@@ -114,45 +116,48 @@ func (b *Builder) Build(si clients.ServerIdentifier) (clients.Transport, error)
...
+	b.refs[si] = 1
+
+	if logger.V(2) {
+		logger.Info("Created a new transport to the server for ServerIdentifier: %v", si)
+	}
+	return tr, nil
+}
+
+func (b *Builder) cleanupFunc(si clients.ServerIdentifier, tr *grpcTransport) func() {
+	return func() {
```
easwars, line 139: If you use `sync.OnceFunc` here, you would be guaranteed that the returned function will execute only once, how many ever times you call it. So, you will not need a `closed` field in the `grpcTransport` type. It can simply call the `cleanup` function every time `Close` is called, and the actual function will be called only once.

xds/internal/clients/grpctransport/grpc_transport_test.go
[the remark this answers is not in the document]
easwars: I still don't see the point of the `TestBuild_Success`. 

What is it verifying?

If the test fails, on what conditions will it fail?

xds/internal/clients/xdsclient/clientimpl_watchers.go
```diff
@@ -67,7 +66,7 @@ func (c *XDSClient) WatchResource(typeURL, resourceName string, watcher Resource
 		return func() {}
 	}
 
-	if c == nil || c.done.HasFired() {
+	if c.done.HasFired() {
```
easwars, line 69: Why do we need this same check twice?

xds/internal/clients/xdsclient/clientimpl.go
easwars: I thought we were going to require byte slices for the send and recv functionalities on the transport. But why do we need to return a byte slice here? We are marshaling the proto to bytes here and the caller would do exactly the opposite. So, I'm not seeing the reason. Maybe I'm missing something?

easwars (commented):
Mostly LGTM.

Just a couple of open comments that needs addressing.

xds/internal/clients/xdsclient/clientimpl_watchers.go
purnesh42H: Removed. Was leftover from copy paste i think.

xds/internal/clients/grpctransport/grpc_transport.go
purnesh42H: Yeah that's better. Done.

xds/internal/clients/xdsclient/clientimpl.go
purnesh42H: I think the reason was similar to keep the API generic and prevent protobuf from being in the API https://github.com/grpc/grpc-go/pull/8042/files#diff-07006a5717489af4cddbe3acd6230bc5ecc4c4ce5ba469b09ad6c3b83af1be8eR65. The underlying `XdsConfig` of `v3statuspb.ClientConfig_GenericXdsConfig` is `pb.any` which contains the bytes of actual resource. The other values are just for debugging purpose. So, yeah to get the debugging information, the bytes has to be eventually unmarshalled into `v3statuspb.ClientConfig_GenericXdsConfig`

easwars (approved):

PR #8187, opened 2025-03-20
resolver: create AddressMapV2 with generics to replace AddressMap

dfawley:
RELEASE NOTES:
* resolver: Add experimental `AddressMapV2` with generics to ultimately replace `AddressMap`.  Deprecate `AddressMap` for deletion.

### DIFF
--- a/balancer/base/balancer.go
+++ b/balancer/base/balancer.go
@@ -122,8 +122,7 @@ func (b *baseBalancer) UpdateClientConnState(s balancer.ClientConnState) error {
 		}
 	}
 	for _, a := range b.subConns.Keys() {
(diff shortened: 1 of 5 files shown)


arjan-bal (approved):

arjan-bal:
> I don't have any reason to believe this is used widely (if at all) outside our project, and it's experimental, and it's very easy to migrate to, so I believe it's fine to just swap in place without a migration period. Let me know if you feel strongly otherwise.
> 
> RELEASE NOTES:
> 
> * resolver: convert experimental AddressMap to use generics

Discussed offline and we found usages in SourceGraph. The PR now keeps the old AddressMap and deprecates it.

PR #8189, opened 2025-03-21
resolver: convert EndpointMap to use generics

dfawley:
This is #8187 but for EndpointMap.  In this case, I could find zero external usages of this type, since it is so new, so I thought it would be OK to just convert it in place.

RELEASE NOTES:
- resolver: convert EndpointMap to use generics

### DIFF
--- a/resolver/map_test.go
+++ b/resolver/map_test.go
@@ -72,11 +72,11 @@ func (s) TestAddressMap_Length(t *testing.T) {
 }
 
 func (s) TestAddressMap_Get(t *testing.T) {
-	addrMap := NewAddressMapV2[any]()
+	addrMap := NewAddressMapV2[int]()
 	addrMap.Set(addr1, 1)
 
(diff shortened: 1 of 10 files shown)


resolver/map_test.go
```diff
@@ -174,7 +174,7 @@ func (s) TestAddressMap_Values(t *testing.T) {
 }
 
 func (s) TestEndpointMap_Length(t *testing.T) {
-	em := NewEndpointMap()
+	em := NewEndpointMap[any]()
```
arjan-bal, line 177: We can use `struct{}` as a more precise type parameter here, and `int` in the test cases below. Any reason to use `any`?

arjan-bal (approved):
LGTM!

resolver/map_test.go
dfawley: Good idea.  I just did the simplest change possible for the tests, but using the proper type is better.

Issue #8168, opened 2025-03-13
Panic seen in google.golang.org/grpc/internal/resolver/delegatingresolver.(*delegatingResolver).updateProxyResolverState


nkawli:
NOTE: if you are reporting is a potential security vulnerability or a crash,
please follow our CVE process at
https://github.com/grpc/proposal/blob/master/P4-grpc-cve-process.md instead of
filing an issue here.

Please see the FAQ in our main README.md, then answer the questions below
before submitting your issue.

### What version of gRPC are you using?
v1.71.0

### What version of Go are you using (`go version`)?
1.23.7

### What operating system (Linux, Windows, …) and version?
Linux

### What did you do?
If possible, provide a recipe for reproducing the error.

We are trying to connect to non existing server like test:8882
There is no host/domain like test in the setup. Its just a dummy value to execute some piece of code in unit test.

### What did you expect to see?
Test works fine as earlier and there is no panic. v1.69.4 and previous versions did not face this panic issue. So seems to be seeing this post v1.69.4 .

### What did you see instead?
Below panic is seen in one of our unit-test:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x55e73a5a3456]

goroutine 1007 [running]:
google.golang.org/grpc/internal/resolver/delegatingresolver.(*delegatingResolver).updateProxyResolverState(0xc000fc2690, {{0xc00095a080, 0x1, 0x1}, {0x0, 0x0, 0x0}, 0x0, 0x0})
	/opt/workspace/test-app/pkg/mod/google.golang.org/grpc@v1.71.0/internal/resolver/delegatingresolver/delegatingresolver.go:274 +0x2b6
google.golang.org/grpc/internal/resolver/delegatingresolver.(*wrappingClientConn).UpdateState(0xc000fd0300?, {{0xc00095a080, 0x1, 0x1}, {0x0, 0x0, 0x0}, 0x0, 0x0})
	/opt/workspace/test-app/pkg/mod/google.golang.org/grpc@v1.71.0/internal/resolver/delegatingresolver/delegatingresolver.go:311 +0x68
google.golang.org/grpc/internal/resolver/dns.(*dnsResolver).watcher(0xc000fd0300)
	/opt/workspace/test-app/pkg/mod/google.golang.org/grpc@v1.71.0/internal/resolver/dns/dns_resolver.go:212 +0x138
[...]

PR #8193, opened 2025-03-24
resolver/delegatingresolver: Add lock in close to avoid race 

eshitachandwani:
Fixes : #8168 

RELEASE NOTES: N/A

### DIFF
--- a/internal/resolver/delegatingresolver/delegatingresolver.go
+++ b/internal/resolver/delegatingresolver/delegatingresolver.go
@@ -170,6 +170,9 @@ func (r *delegatingResolver) ResolveNow(o resolver.ResolveNowOptions) {
 }
 
 func (r *delegatingResolver) Close() {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	
 	r.targetResolver.Close()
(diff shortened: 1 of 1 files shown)


internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -170,6 +170,9 @@ func (r *delegatingResolver) ResolveNow(o resolver.ResolveNowOptions) {
 }
 
 func (r *delegatingResolver) Close() {
+	r.mu.Lock()
+	defer r.mu.Unlock()
```
arjan-bal, line 174: This is not sufficient to avoid the panic in the issue.  It will serialize the calls, but `Close` still lock the mutex before `updateTargetResolverState` and `updateTargetResolverState` will panic.

I have a possible fix [here](https://github.com/grpc/grpc-go/compare/master...arjan-bal:fix-delegating-resolver-race?expand=1). I'm working on adding a test that verifies this panic doesn't happen.

arjan-bal:
Raised a PR with the fix and test. Closing this one.

PR #8195, opened 2025-03-24
delegatingresolver: Stop calls into delegates once the parent resolver is closed.

arjan-bal:
Fixes: #8168

While handling updates from child resolvers, the delegating resolver locks it's mutex. If the channel returns an error on updating the state, the delegating resolver calls `ResolveNow` on the other child. This can result in a deadlock if the other child attempts to call `cc.UpdateState` synchronously while handling the `ResolveNow` call. This change makes delegatingResovler call `ResolveNow` in a new goroutine to avoid such deadlocks.

This change introduced a second mutex in delegating resolver to serialize calls into child resolvers. This ensures no calls to `ResolveNow` are made after the child resolvers are closed.

RELEASE NOTES:
* resolver/delegatingresolver: Avoid deadlocks when an http proxy is configured and a resolver calls `UpdateState` synchronously while handling a `ResolveNow` call.
* resolver/delegatingresolver: Fix race that may cause panics when an http proxy is configured and the resolver is closing.

### DIFF
--- a/internal/resolver/delegatingresolver/delegatingresolver.go
+++ b/internal/resolver/delegatingresolver/delegatingresolver.go
@@ -53,6 +53,11 @@ type delegatingResolver struct {
 	mu                  sync.Mutex         // protects all the fields below
 	targetResolverState *resolver.State    // state of the target resolver
 	proxyAddrs          []resolver.Address // resolved proxy addresses; empty if no proxy is configured
+	// isClosed is set to "true" when the channel closes the resolver. Once the
+	// target and proxy resolvers are closed, this boolean is used to avoid
+	// calling "ResolveNow" on the target resolver when the proxy resolver
+	// reports an error or vice-versa.
+	isClosed bool
 }
 
 // nopResolver is a resolver that does nothing.
@@ -170,6 +175,10 @@ func (r *delegatingResolver) ResolveNow(o resolver.ResolveNowOptions) {
 }
 
 func (r *delegatingResolver) Close() {
+	r.mu.Lock()
+	r.isClosed = true
+	r.mu.Unlock()
+
 	r.targetResolver.Close()
 	r.targetResolver = nil
 
@@ -249,6 +258,9 @@ func (r *delegatingResolver) updateClientConnStateLocked() error {
 func (r *delegatingResolver) updateProxyResolverState(state resolver.State) error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
+	if r.isClosed {
+		return nil
+	}
 	if logger.V(2) {
 		logger.Infof("Addresses received from proxy resolver: %s", state.Addresses)
 	}
@@ -267,7 +279,7 @@ func (r *delegatingResolver) updateProxyResolverState(state resolver.State) erro
 	err := r.updateClientConnStateLocked()
 	// Another possible approach was to block until updates are received from
 	// both resolvers. But this is not used because calling `New()` triggers
-	// `Build()`  for the first resolver, which calls `UpdateState()`. And the
+	// `Build()` for the first resolver, which calls `UpdateState()`. And the
 	// second resolver hasn't sent an update yet, so it would cause `New()` to
 	// block indefinitely.
 	if err != nil {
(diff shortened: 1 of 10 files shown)


internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -93,6 +97,11 @@ func New(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOpti
 	r := &delegatingResolver{
 		target: target,
 		cc:     cc,
+		// Child resolvers may send a state update as soon as they're built.
+		// Initialize children with no-op resolvers. When both children are nil,
+		// the delegatingResolver is considered closed.
```
dfawley, line 102: Related: do we not need to hold `childMu` when calling `Build`?

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -44,15 +44,19 @@ var (
...
+	cc       resolver.ClientConn // gRPC ClientConn
+	proxyURL *url.URL            // proxy URL, derived from proxy environment and target
 
 	mu                  sync.Mutex         // protects all the fields below
 	targetResolverState *resolver.State    // state of the target resolver
 	proxyAddrs          []resolver.Address // resolved proxy addresses; empty if no proxy is configured
+
+	// childMu serializes calls into child resolvers. It also protects access to
+	// the following fields.
+	childMu        sync.Mutex
```
dfawley, line 57: Can we not share `mu` for this?  Why not?

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -267,15 +274,27 @@ func (r *delegatingResolver) updateProxyResolverState(state resolver.State) erro
...
-	// `Build()`  for the first resolver, which calls `UpdateState()`. And the
+	// `Build()` for the first resolver, which calls `UpdateState()`. And the
 	// second resolver hasn't sent an update yet, so it would cause `New()` to
 	// block indefinitely.
 	if err != nil {
-		r.targetResolver.ResolveNow(resolver.ResolveNowOptions{})
+		go func() {
+			r.childMu.Lock()
+			defer r.childMu.Unlock()
+			if !r.isClosedLocked() {
```
dfawley, line 284: Probably a nit, but IMO it is simpler to delete the helper and here just check if `r.targetResolver != nil` since that's what really matters.

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -291,7 +310,13 @@ func (r *delegatingResolver) updateTargetResolverState(state resolver.State) err
 	r.targetResolverState = &state
 	err := r.updateClientConnStateLocked()
 	if err != nil {
-		r.proxyResolver.ResolveNow(resolver.ResolveNowOptions{})
+		go func() {
+			r.childMu.Lock()
+			defer r.childMu.Unlock()
+			if !r.isClosedLocked() {
```
dfawley, line 316: As above.

arjan-bal: Removed the helper function.

internal/resolver/delegatingresolver/delegatingresolver.go
arjan-bal: Deleted the helper and added a check.

internal/resolver/delegatingresolver/delegatingresolver.go
arjan-bal: Yes, we need to. Updates from one child can trigger ResolveNow calls in the other child. This also avoids the need to initialize children here. Added code to acquire the mutex and release it.

internal/resolver/delegatingresolver/delegatingresolver.go
arjan-bal: `childMu` is being used to serialize calls into the child resolvers, similar to the [serailzer queue in resolver_wrapper.go](https://github.com/grpc/grpc-go/blob/a51009d1d7074ee1efcd323578064cbe44ef87e5/resolver_wrapper.go#L98-L124).

If we re-use the existing mutex, we would need to hold `mu` while calling `childResolver.ResolveNow()`. If this calls `cc.UpdateState()` synchronously, it will try to re-acquire `mu` and deadlock.

* `mu` is used to serialize `resolver -> channel` calls.
* `childMu` is used to serialize `channel -> resolver` calls.

dfawley (approved):

PR #8198, opened 2025-03-24
credentials: expose NewRequestInfoContext publicly

### LINKED ISSUE #8080: Expose credentials.NewRequestInfoContext as a public (experimental) API
We have a Go library that reuses gRPC code generator and many of APIs, but uses an alternative transport layer (simple HTTP request/responses): http://go.chromium.org/luci/grpc/prpc The can historically be traced to lack of HTTP2 support on Google Appengine and it solves the same problem as solved by grpc-web proxy, except without a proxy.

We are morphing this library to be closer and closer to the real gRPC client, to do a full switch at some point.

As part of that, it would be nice to be able to use PerRPCCredentials. Unfortunately, many implementations of this interface call `credentials.RequestInfoFromContext(ctx)` to check the security level etc. And this is impossible to setup without access to internal https://pkg.go.dev/google.golang.org/grpc/internal/credentials#NewRequestInfoContext

### Proposed Solution
Please consider making this API public. This will also likely help with unit-testing PerRPCCredentials implementations.

vadimsht:
The primary use case is to allow testing PerRPCCredentials implementations that live outside of google.golang.org/grpc.

The secondary (somewhat weirder) use case is to allow reusing existing PerRPCCredentials implementations (as long as they don't depend on handshake details) outside of the standard gRPC client environment (i.e. with alternative implementations of grpc.ClientConnInterface that reuse gRPC code generator machinery and APIs, but don't use the actual implementations).

Fixes #8080.

RELEASE NOTES: 

- credentials: add NewRequestInfoContext to allow testing PerRPCCredentials implementations.

### DIFF
--- a/credentials/credentials.go
+++ b/credentials/credentials.go
@@ -215,6 +215,21 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
 	return ri, ok
 }
 
+// NewRequestInfoContext creates a new context with the given RequestInfo
+// attached to it.
+//
+// This RequestInfo will be accessible via RequestInfoFromContext.
+//
+// Intended to be used from tests for PerRPCCredentials implementations (that
+// often need to check connection's SecurityLevel). Should not be used from
+// non-test code: the gRPC client already prepares a context with the correct
+// RequestInfo attached when calling PerRPCCredentials.GetRequestMetadata.
+//
+// This API is experimental.
+func NewRequestInfoContext(ctx context.Context, ri RequestInfo) context.Context {
+	return icredentials.NewRequestInfoContext(ctx, ri)
+}
+
 // ClientHandshakeInfo holds data to be passed to ClientHandshake. This makes
 // it possible to pass arbitrary data to the handshaker from gRPC, resolver,
 // balancer etc. Individual credential implementations control the actual
--- a/credentials/credentials.go
+++ b/credentials/credentials.go
@@ -207,15 +207,19 @@ type RequestInfo struct {
 	AuthInfo AuthInfo
 }
 
+// requestInfoKey is a struct to be used as the key to store RequestInfo in a
+// context.
+type requestInfoKey struct{}
+
 // RequestInfoFromContext extracts the RequestInfo from the context if it exists.
 //
 // This API is experimental.
 func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
-	ri, ok = icredentials.RequestInfoFromContext(ctx).(RequestInfo)
+	ri, ok = ctx.Value(requestInfoKey{}).(RequestInfo)
 	return ri, ok
 }
 
-// NewRequestInfoContext creates a new context with the given RequestInfo
+// NewContextWithRequestInfo creates a new context with the given RequestInfo
 // attached to it.
 //
 // This RequestInfo will be accessible via RequestInfoFromContext.
@@ -226,8 +230,8 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
 // RequestInfo attached when calling PerRPCCredentials.GetRequestMetadata.
 //
 // This API is experimental.
-func NewRequestInfoContext(ctx context.Context, ri RequestInfo) context.Context {
-	return icredentials.NewRequestInfoContext(ctx, ri)
+func NewContextWithRequestInfo(ctx context.Context, ri RequestInfo) context.Context {
+	return context.WithValue(ctx, requestInfoKey{}, ri)
 }
 
 // ClientHandshakeInfo holds data to be passed to ClientHandshake. This makes
--- a/credentials/credentials.go
+++ b/credentials/credentials.go
@@ -219,8 +219,7 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
 	return ri, ok
 }
 
-// NewContextWithRequestInfo creates a new context with the given RequestInfo
-// attached to it.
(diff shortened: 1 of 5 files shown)


credentials/credentials.go
```diff
@@ -215,6 +215,21 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
...
+// attached to it.
+//
+// This RequestInfo will be accessible via RequestInfoFromContext.
+//
+// Intended to be used from tests for PerRPCCredentials implementations (that
+// often need to check connection's SecurityLevel). Should not be used from
+// non-test code: the gRPC client already prepares a context with the correct
+// RequestInfo attached when calling PerRPCCredentials.GetRequestMetadata.
+//
+// This API is experimental.
```
eshitachandwani, line 228: // # Experimental
//
// Notice: This type is EXPERIMENTAL and may be changed or removed in a
// later release.

internal/transport/http2_client.go
```diff
@@ -545,7 +545,7 @@ func (t *http2Client) createHeaderFields(ctx context.Context, callHdr *CallHdr)
 		Method:   callHdr.Method,
 		AuthInfo: t.authInfo,
 	}
-	ctxWithRequestInfo := icredentials.NewRequestInfoContext(ctx, ri)
+	ctxWithRequestInfo := credentials.NewRequestInfoContext(ctx, ri)
```
eshitachandwani, line 548: Can this can still call the internal function, is there a reason to change this also?

vadimsht: No real reason. It just already uses credentials.RequestInfo couple of lines above, I felt like it might be more consistent to use credentials.NewRequestInfoContext as well, but it doesn't matter.

vadimsht:
Updated. PTAL. Let me know if you want to me to revert other changes to icredentials.NewRequestInfoContext

eshitachandwani (approved):

eshitachandwani:
looks good to me , adding @dfawley for another review.

vadimsht:
Looks like the presubmit check wants "RELEASE NOTES". Please review/modify the line I've added to the description. Not sure how prominently this change should be highlighted in the notes (if at all).

credentials/credentials.go
```diff
@@ -215,6 +215,21 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
...
+func NewRequestInfoContext(ctx context.Context, ri RequestInfo) context.Context {
```
dfawley, line 229: `NewContextWithRequestInfo` seems like it would be more natural.  Also matches e.g. https://pkg.go.dev/google.golang.org/grpc#NewContextWithServerTransportStream

credentials/credentials.go
```diff
@@ -215,6 +215,21 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
...
+	return icredentials.NewRequestInfoContext(ctx, ri)
```
dfawley, line 230: Can we move the internal functions here instead of passing calls through?

credentials/credentials.go
```diff
@@ -215,6 +215,21 @@ func RequestInfoFromContext(ctx context.Context) (ri RequestInfo, ok bool) {
 	return ri, ok
 }
 
+// NewRequestInfoContext creates a new context with the given RequestInfo
+// attached to it.
```
dfawley, line 219: `NewContextWithRequestInfo attaches the RequestInfo to ctx`?

dfawley (commented):
LGTM except for these comments

credentials/credentials.go
vadimsht: I chose "creates a new context with ... attached" because many other existing functions are documented that way:

```
// NewIncomingContext creates a new context with incoming md attached.
// NewOutgoingContext creates a new context with outgoing md attached.
// NewContext creates a new context with peer information attached.
// NewContextWithServerTransportStream creates a new context from ctx and attaches stream to it.
```

credentials/credentials.go
vadimsht: Done, but internal/credentials/credentials.go still remains since it still has `ClientHandshakeInfoFromContext` and `NewClientHandshakeInfoContext`. Removing it will require making `NewClientHandshakeInfoContext` public as well.

Do you want me to do this? 

Strictly speaking it may also be useful for testing PerRPCCredentials.

credentials/credentials.go
vadimsht: Done.

credentials/credentials.go
dfawley: Of these, the last one is most accurate.  I'm also fine with `creates a new context from ctx and attaches ri to it`, too.

credentials/credentials.go
dfawley: I would be fine with this, too, if you are interested in doing it.  The same caveats would apply: it should only be used for testing.  But if you don't need it, then nobody else has asked, so I'm fine with waiting until someone needs it, instead.

vadimsht: I looked into how `ClientHandshakeInfoFromContext` is used (by actual prod code) and it is only used by xDS credentials implementation. I think for xds-related tests outside of grpc package (if any) to be useful, they will need access to other xDS internal parts (internal/xds package). For that reason I think it makes little sense to make it public now (also no one really needs it public).

credentials/credentials.go
vadimsht: Changed it to `NewContextWithRequestInfo creates a new context from ctx and attaches ri to it.`

dfawley (approved):
Thank you!

PR #8199, opened 2025-03-25
delegatingresolver: Call ResolveNow in a new goroutine to avoid deadlocks

arjan-bal:
While handling updates from child resolvers, the delegating resolver locks it's mutex. If the channel returns an error on updating the state, the delegating resolver calls `ResolveNow` on the other child. This can result in a deadlock if the other child attempts to call `cc.UpdateState` synchronously while handling the `ResolveNow` call. This change makes delegatingResovler call `ResolveNow` in a new goroutine to avoid such deadlocks.

RELEASE NOTES:
* resolver/delegatingresolver: Avoid deadlocks when an http proxy is configured and a resolver calls `UpdateState` synchronously while handling a `ResolveNow` call.


arjan-bal:
Changes merged into https://github.com/grpc/grpc-go/pull/8195 since all non-test code changes were common.

Issue #8192, opened 2025-03-24
[Data Race]:a potential data race occurs in the prevGoAwayID field in http2_client.go


ririli:
Hi, I found through static analysis that the prevGoAwayID field in that code was read or written to a total of four times, three of which were protected by the mutex and one of which was unprotected. So if prevGoAwayID is accessed concurrently, then there is a data race at [L1343 ](https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_client.go#L1343)which results in unintended values being passed into connectionErrorf().

Please could you help me to confirm this issue?

purnesh42H:
@ririli i didn't find any access of `prevGoAwayID` without locking the mutex. At https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_client.go#L1343, we unlock right before throwing the error. I guess we can store the value in temp variable and then unlock but I don't that will make much difference. Are you noticing any race? Could you provide the debugging logs?

dfawley:
> I guess we can store the value in temp variable and then unlock but I don't that will make much difference.

It's either correct or it's not correct.

If it could be accessed concurrently, then it's not correct.

In this case, though, it is _not_ accessed concurrently.  This function is only called in a loop from a single goroutine.

purnesh42H:
Closing this since the field is not being accessed concurrently https://github.com/grpc/grpc-go/pull/8200#issuecomment-2752562557

PR #8200, opened 2025-03-25
internal/transport: temporarily store prevGoAwayId before releasing the lock in http2_client

purnesh42H:
Fixes: #8192

RELEASE NOTES: None


arjan-bal (approved):

dfawley:
All accesses of t.prevGoAwayId happen in this one function, which is never called concurrently.  So I don't believe any change is needed here.

Just because some field is accessed with the lock held sometimes but not others, does not mean that it always must be accessed with the lock held.  I think we can close the issue as not-a-bug.

PR #8203, opened 2025-03-27
xdsclient: add grpc.xds_client.server_failure counter mertric

purnesh42H:
Internal ticket b/401438450

RELEASE NOTES:
- xds: add `grpc.xds_client.server_failure` counter metric on xDS client to record connectivity errors.

### DIFF
--- a/xds/internal/xdsclient/clientimpl.go
+++ b/xds/internal/xdsclient/clientimpl.go
@@ -424,6 +424,8 @@ func (cs *channelState) adsStreamFailure(err error) {
 		return
 	}
 
+	xdsClientServerFailureMetric.Record(cs.parent.metricsRecorder, 1, cs.parent.target, cs.serverConfig.ServerURI())
+
 	cs.parent.channelsMu.Lock()
 	defer cs.parent.channelsMu.Unlock()
 	for authority := range cs.interestedAuthorities {
--- a/xds/internal/xdsclient/clientimpl.go
+++ b/xds/internal/xdsclient/clientimpl.go
@@ -76,6 +76,13 @@ var (
 		Labels:      []string{"grpc.target", "grpc.xds.server", "grpc.xds.resource_type"},
 		Default:     false,
 	})
+	xdsClientServerFailureMetric = estats.RegisterInt64Count(estats.MetricDescriptor{
+		Name:        "grpc.xds_client.server_failure",
+		Description: "A counter of xDS servers going from healthy to unhealthy. A server goes unhealthy when we have a connectivity failure or when the ADS stream fails without seeing a response message, as per gRFC A57.",
+		Unit:        "failure",
+		Labels:      []string{"grpc.target", "grpc.xds.server"},
+		Default:     false,
+	})
 )
 
 // clientImpl is the real implementation of the xDS client. The exported Client
--- a/xds/internal/xdsclient/clientimpl.go
+++ b/xds/internal/xdsclient/clientimpl.go
@@ -424,7 +424,9 @@ func (cs *channelState) adsStreamFailure(err error) {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


xds/internal/xdsclient/metrics_test.go
```diff
@@ -147,3 +147,169 @@ func (s) TestResourceUpdateMetrics(t *testing.T) {
...
+		Handle:    xdsClientResourceUpdatesValidMetric.Descriptor(),
+		IntIncr:   1,
+		LabelKeys: []string{"grpc.target", "grpc.xds.server", "grpc.xds.resource_type"},
+		LabelVals: []string{"Test/ServerFailureMetrics_AfterResponseRecv", mgmtServer.Address, "ListenerResource"},
+	}
+	if err := tmr.WaitForInt64Count(ctx, mdWant); err != nil {
+		t.Fatal(err.Error())
+	}
+	// Server failure should have no recording point.
+	if got, _ := tmr.Metric("grpc.xds_client.server_failure"); got != 0 {
```
dfawley, line 292: So... I see you do have a test case for this...  Why is this passing?  Is it a race?

xds/internal/xdsclient/clientimpl.go
```diff
@@ -417,6 +424,8 @@ func (cs *channelState) adsStreamFailure(err error) {
 		return
 	}
 
+	xdsClientServerFailureMetric.Record(cs.parent.metricsRecorder, 1, cs.parent.target, cs.serverConfig.ServerURI())
```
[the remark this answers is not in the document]
purnesh42H: yeah so in that case if the response is received and stream breaks after, metrics shouldn't be recorded. I modified the description. That's why i wrote that test case separately. Below condition should prevent it. @markdroth is that correct?

```
if msgReceived {
		err = xdsresource.NewError(xdsresource.ErrTypeStreamFailedAfterRecv, err.Error())
	}
```

xds/internal/xdsclient/metrics_test.go
purnesh42H: what do you mean? At this point the stream is not broken so server_failure should be 0

xds/internal/xdsclient/metrics_test.go
```diff
@@ -147,3 +147,169 @@ func (s) TestResourceUpdateMetrics(t *testing.T) {
...
+	}
+
+	mdWant = stats.MetricsData{
+		Handle:    xdsClientServerFailureMetric.Descriptor(),
+		IntIncr:   1,
+		LabelKeys: []string{"grpc.target", "grpc.xds.server"},
+		LabelVals: []string{"Test/ServerFailureMetrics_AfterResponseRecv", mgmtServer.Address},
+	}
+	// Server failure should still have no recording point.
+	if err := tmr.WaitForInt64Count(ctx, mdWant); err == nil {
```
purnesh42H, line 312: @dfawley i believe you meant to refer to this? So, because we already had a response before and stream broke after, the server_failure should still be not recorded.

xds/internal/xdsclient/clientimpl.go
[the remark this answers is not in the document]
purnesh42H: ah looks like my file with if condition for err while metric recording was not saved so the change didn't get pushed. But i figured the reason why the test case was passing without that. Watch was expiring due to short watch expiry timeout so i removed setting watch expiry while creating client. 

Another thing I did was to use the restartable listener so that we can restart the server after stopping briefly to prevent new attempt for creating ADS stream because that will call `OnStreamError` with different error if server is not restarted. This made sure that `OnStreamError` get called only once when we stop the listener.

xds/internal/xdsclient/metrics_test.go
purnesh42H: yeah i think i fixed it. I thought you are referring to this particular line.

dfawley (approved):

PR #8204, opened 2025-03-28
credentials/alts: Optimize reads

arjan-bal:
## Background

Internal issue: b/400312873

While benchmarking Google Cloud Storage read performance, the flame graphs showed significant time being spent in `google.golang.org/grpc/credentials/alts/internal/conn.(*conn).Read`.  These changes are done to reduce the time spent in this function. The following optimizations are done:

1. In the handshaker, allocate a buffer only once, outside the `for` loop.
2. Use 32KB (previously 4KB) buffers to read from the network.
3. When a complete ALTS frame doesn't fit into the buffer, directly expand the buffer to the required capacity instead of incrementing in steps of 4KB.
4. If the buffer passed to `Read()` is large enough, use it to store the decrypted message, avoiding an extra copy to `buf`. The default buffer size used by gRPC is 16KB, which also happens to be the size of ALTS records from GCS. This means we usually avoid the extra copy.

## Benchmarks

branch: master
```
go test ./credentials/alts/internal/conn -timeout=2s -bench="Bench"
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.60GHz
BenchmarkLargeMessage-48              13          90139719 ns/op
PASS
ok      google.golang.org/grpc/credentials/alts/internal/conn   2.316s
```

branch: perfromance (this PR)
```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.60GHz
BenchmarkLargeMessage-48              14          80422425 ns/op
PASS
ok      google.golang.org/grpc/credentials/alts/internal/conn   2.139s
```

RELEASE NOTES:
* credentials/alts: Improve read performance by optimizing buffer copies and allocations.

### DIFF
--- a/credentials/alts/internal/conn/record.go
+++ b/credentials/alts/internal/conn/record.go
@@ -191,7 +191,7 @@ func (p *conn) Read(b []byte) (n int, err error) {
 
 		// Decrypt directly into the buffer, avoiding a copy from p.buf if
 		// possible.
-		if cap(b) >= len(ciphertext) {
+		if len(b) >= len(ciphertext) {
 			dec, err := p.crypto.Decrypt(b[:0], ciphertext)
 			if err != nil {
 				return 0, err
--- a/credentials/alts/internal/conn/record.go
+++ b/credentials/alts/internal/conn/record.go
@@ -184,7 +184,7 @@ func (p *conn) Read(b []byte) (n int, err error) {
 					// header.
 					panic(fmt.Sprintf("protected buffer length shorter than expected: %d vs %d", len(p.protected), MsgLenFieldSize))
 				}
-				newProtectedBuf := mem.DefaultBufferPool().Get(int(length))
+				newProtectedBuf := mem.DefaultBufferPool().Get(int(length) + MsgLenFieldSize)
 				copy(*newProtectedBuf, p.protected)
 				p.protected = (*newProtectedBuf)[:len(p.protected)]
 				mem.DefaultBufferPool().Put(p.protectedPointer)
--- a/credentials/alts/internal/conn/record.go
+++ b/credentials/alts/internal/conn/record.go
@@ -23,10 +23,12 @@ package conn
 import (
 	"encoding/binary"
 	"fmt"
+	"io"
 	"math"
 	"net"
 
 	core "google.golang.org/grpc/credentials/alts/internal"
+	"google.golang.org/grpc/mem"
 )
 
 // ALTSRecordCrypto is the interface for gRPC ALTS record protocol.
@@ -63,6 +65,8 @@ const (
 	// The maximum write buffer size. This *must* be multiple of
 	// altsRecordDefaultLength.
 	altsWriteBufferMaxSize = 512 * 1024 // 512KiB
+	// The initial buffer used to read from the network.
+	altsReadBufferInitialSize = 32 * 1024 // 32KiB
 )
 
 var (
@@ -84,11 +88,19 @@ type conn struct {
 	crypto ALTSRecordCrypto
 	// buf holds data that has been read from the connection and decrypted,
 	// but has not yet been returned by Read.
-	buf                []byte
+	buf []byte
+	// bufPointer holds the entire decrypted record, even bytes that have
+	// been returned by read. It is used to restore buf to it's initial
+	// capacity after each frame is decrypted. It is also used to return the
+	// buffer to the buffer pool.
+	bufPointer         *[]byte
 	payloadLengthLimit int
 	// protected holds data read from the network but have not yet been
 	// decrypted. This data might not compose a complete frame.
 	protected []byte
+	// protectedPointer holds a pointer to the protected buffer. It is used to
+	// return the protected buffer to the buffer pool.
+	protectedPointer *[]byte
 	// writeBuf is a buffer used to contain encrypted frames before being
 	// written to the network.
 	writeBuf []byte
@@ -96,6 +108,7 @@ type conn struct {
 	nextFrame []byte
 	// overhead is the calculated overhead of each frame.
 	overhead int
+	isClosed bool
 }
 
 // NewConn creates a new secure channel instance given the other party role and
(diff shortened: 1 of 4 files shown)


credentials/alts/internal/conn/record.go
```diff
@@ -153,20 +170,30 @@ func (p *conn) Read(b []byte) (n int, err error) {
 		// Check whether the next frame to be decrypted has been
 		// completely received yet.
 		if len(framedMsg) == 0 {
-			copy(p.protected, p.nextFrame)
-			p.protected = p.protected[:len(p.nextFrame)]
+			p.protected = p.protected[:copy(p.protected, p.nextFrame)]
```
gtcooke94, line 173: I think this line is a little challenging to reason about since it's modifying `p.protectd` in several ways. Can it be broken up without performance hits, or will not immediately slicing to the right length cause an issue?

credentials/alts/internal/conn/record_test.go
```diff
@@ -188,6 +189,48 @@ func (s) TestLargeMsg(t *testing.T) {
...
+		t.Fatalf("Read() = %v, %v; want %v, <nil>", n, err, len(rcvMsg))
+	}
+	if !reflect.DeepEqual(msg, rcvMsg) {
+		t.Fatalf("Write()/Server Read() = %v, want %v", rcvMsg, msg)
+	}
+}
+
+// BenchmarkLargeMessage measures the performance of ALTS conns for sending and
+// receiving a large message.
+func BenchmarkLargeMessage(b *testing.B) {
```
rockspore, line 215: @gtcooke94 Do we have any GitHub actions that already run these benchmarks or should we add any if not?

credentials/alts/internal/conn/record.go
```diff
@@ -111,39 +124,43 @@ func NewConn(c net.Conn, side core.Side, recordProtocol string, key []byte, prot
...
-		protectedBuf = make([]byte, 0, 2*altsRecordDefaultLength-1)
-	} else {
-		protectedBuf = make([]byte, len(protected))
-		copy(protectedBuf, protected)
-	}
+	// We pre-allocate protected to be of size 32KB during initialization.
+	// We increase the size of the buffer by the required amount if can't hold a
+	// complete encrypted record.
+	protectedPointer := mem.DefaultBufferPool().Get(max(altsReadBufferInitialSize, len(protected)))
+	protectedBuf := (*protectedPointer)[:copy(*protectedPointer, protected)]
```
gtcooke94, line 131: I think this has the same readability concern where we are indexing into a buffer that we are at the same time copying into. I think it'd be easier to read as
```
protectedLen := copy(*protectedPointer, protected)
protectedBuf := (*protectedPointer([:protectedLen])
```

credentials/alts/internal/conn/record.go
```diff
@@ -153,20 +170,30 @@ func (p *conn) Read(b []byte) (n int, err error) {
...
+				// We can parse the length header to know exactly how large
+				// the buffer needs to be to hold the entire frame.
+				length, didParse := ParseMessageLength(p.protected)
+				if !didParse {
+					// The protected buffer is initialized with a capacity of
+					// larger than 4B. It should always hold the message length
+					// header.
+					panic(fmt.Sprintf("protected buffer length shorter than expected: %d vs %d", len(p.protected), MsgLenFieldSize))
+				}
+				tmp := mem.DefaultBufferPool().Get(int(length))
```
gtcooke94, line 190: I suggest naming this something other than `tmp`
If I'm understanding, I think the intention is the `tmp` actually is going to be the new buffer that `p.protected` and `p.protectedPointer` point at? In which case I'd say something more representative like `newProtectedBuf` or something?

credentials/alts/internal/conn/record.go
```diff
@@ -185,17 +212,27 @@ func (p *conn) Read(b []byte) (n int, err error) {
...
 
-		// Decrypt requires that if the dst and ciphertext alias, they
-		// must alias exactly. Code here used to use msg[:0], but msg
-		// starts MsgLenFieldSize+msgTypeFieldSize bytes earlier than
-		// ciphertext, so they alias inexactly. Using ciphertext[:0]
-		// arranges the appropriate aliasing without needing to copy
-		// ciphertext or use a separate destination buffer. For more info
-		// check: https://golang.org/pkg/crypto/cipher/#AEAD.
-		p.buf, err = p.crypto.Decrypt(ciphertext[:0], ciphertext)
+		// Decrypt directly into the buffer, avoiding a copy to p.buf if
```
gtcooke94, line 215: Is there any potential issue here where we are missing the update of `p.buf` that is guaranteed on the non-error return in the old structure? Specifically `p.buf = p.buf[n:]`

credentials/alts/internal/conn/record.go
```diff
@@ -185,17 +212,27 @@ func (p *conn) Read(b []byte) (n int, err error) {
 		}
 		ciphertext := msg[msgTypeFieldSize:]
 
-		// Decrypt requires that if the dst and ciphertext alias, they
```
gtcooke94, line 188: Can you explain why we get to ignore this now?

arjan-bal: I changed this to use a separate destination buffer for decryption, so the aliasing requirement mentioned here was no longer applicable. I've reverted this change and restored the previous behaviour of using the input slice as the destination as there's no performance benefit of using a separate destination slice.

I introduced the new destination buffer to hold multiple decrypted frames, but that didn't help so I reverted it.

credentials/alts/internal/conn/record.go
arjan-bal: We enter the outermost `for` loop when `len(p.buf) == 0`. So when we return early without updating `p.buf`, we leave `p.buf` empty. This is the same as the following:
1. Filling `p.buf` with the decrypted ciphertext.
2. Copying entire `p.buf`  to `b` (the argument to `Read`).
3. Resizing `p.buf` to 0.

credentials/alts/internal/conn/record.go
arjan-bal: Renamed as suggested.

credentials/alts/internal/conn/record_test.go
gtcooke94: I think someone else on the Go team would be the person to ask here - I don't know much about the CI setup. @arjan-bal do you know about the grpc-go github CI and benchmarking?

gtcooke94 (approved):
I think this looks good, thanks! The question remaining is how the benchmark you added interacts with the CI - I'm good with this being a follow-up if it'll take a lot of extra time.

credentials/alts/internal/conn/record_test.go
arjan-bal: We don't run benchmarks as part of CI. We have benchmarks [here](https://github.com/grpc/grpc-go/blob/2d4daf3475905868585a914ab39fdc09618869dc/benchmark/benchmain/main.go#L24-L39) that we ask authors to run when reviewing PRs that effect performance. We can have a similar benchmark for ALTS or modify the existing benchmark to support ALTS.

We have performance dashboard for all languages here, but we don't have alerts setup for regressions: https://grafana-dot-grpc-testing.appspot.com/?orgId=1

arjan-bal:
The CI has caught a data race where gRPC is calling `conn.Read` and `conn.Close` concurrently: https://github.com/grpc/grpc-go/actions/runs/14238675443/job/39903307175?pr=8204

There are a few solutions to fix the race:
1. Don't add the `protectedBuf` back to the buffer pool. It will be garbage collected like before.
2. Add synchronization in the http client to serialize calls Read and Close.
3. Add synchronization inside `alts.conn` to serialize Read, Write and Close.

I'll try to see if I can make the fix for 2. If that's not possible, 1 seems fine too.

arjan-bal:
Talked to @dfawley and decided that we don't need to use the buffer pool here since there is only one buffer that is re-used for reads. The buffer is re-allocated when a larger ALTS record needs to be processed, but it should become stable after a while.

Doug raised a concern about the buffer only growing but not shrinking, which may lead to higher memory usage if the peers sent large ALTS records earlier and there are a large number of ALTS conns. I'm checking what other languages do.

gtcooke94:
Is this ready for another review pass, and are the benchmarks in the description up to date with the change? Or are you still experimenting with doing it without the buffer pools?

arjan-bal:
> Is this ready for another review pass, and are the benchmarks in the description up to date with the change? Or are you still experimenting with doing it without the buffer pools?

Decided to remove the buffer pool after discussion with Doug. I re-ran the benchmark and the performance seems to be very slightly better after removing the buffer pool. The PR description is updated.

> Doug raised a concern about the buffer only growing but not shrinking, which may lead to higher memory usage if the peers sent large ALTS records in the past and there are a large number of ALTS conns. I'm checking what other languages do.

Decided not to solve the buffer shrinking problem since the current implementation already suffers from this and it's not something introduced in this PR.

From discussions with other language maintainers Java has a scatter gather to read multiple buffers from the socket. The list of buffers is being copied to a contiguous array for decryption, so it's not zero copy. I tried implementing a similar [scatter gather style in Go](https://github.com/grpc/grpc-go/compare/master...arjan-bal:alts-scatter-gather?expand=1) and the performance was ~10% worse than the existing implementation on the master branch.

credentials/alts/internal/conn/record.go
```diff
@@ -162,11 +156,21 @@ func (p *conn) Read(b []byte) (n int, err error) {
...
+				// the buffer needs to be to hold the entire frame.
+				length, didParse := parseMessageLength(p.protected)
+				if !didParse {
+					// The protected buffer is initialized with a capacity of
+					// larger than 4B. It should always hold the message length
+					// header.
+					panic(fmt.Sprintf("protected buffer length shorter than expected: %d vs %d", len(p.protected), MsgLenFieldSize))
+				}
+				oldProtectedBuf := p.protected
+				p.protected = make([]byte, int(length)+MsgLenFieldSize)
```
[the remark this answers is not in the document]
arjan-bal: The new buffer that we allocate must be able to hold the entire encrypted record. After reading the message header, we know the length of the record is: `length parsed from the message header` + `size of the message length header`. This is the capacity, but the length of the new buffer should be set to the number of bytes that are already read. So we set the length to the same as the length of the existing buffer and copy its contents. Let me raise a PR with some commentry.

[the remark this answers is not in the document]
arjan-bal: Created a PR to add clarity, PTAL: https://github.com/grpc/grpc-go/pull/8232

PR #8209, opened 2025-03-29
Add doc about `mem.*Unsafe` consuming the input buffer

Groxx:
I've been reading the new buffer-pool-related code due to recent OOM problems with other libraries misusing `sync.Pool`, and the varying behavior between `Buffer` and `SliceBuffer` had me confused for quite a while.

Buffer mutates itself when `read`, and implicitly calls `Free` when fully consumed:
https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/mem/buffers.go#L194-L200

While SliceBuffer returns a sub-slice but does not modify itself: https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/mem/buffers.go#L263-L267

The only way these `*Unsafe` funcs can be used correctly is by replacing the input-buffer with the result-buffer(s), as the current code does: https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/internal/transport/transport.go#L139

... which seems worth documenting since it feels like a _huge_ footgun otherwise: you only trigger the self-mutating-and-freeing behavior for relatively large data, which is less common in tests.

Granted, this isn't a technical barrier at all, but the behavior differences between implementations mislead me for a while and this "must not use input Buffer" requirement doesn't seem obviously-necessary to me when reading the code.  Might as well save future-people that effort.

---

On a related note, `Buffer` feels... a bit concerningly strange in regards to concurrency, and I didn't see it called out in the PR that introduced it so I think it might not have been noticed.

E.g. `Buffer`s are not concurrency-safe: https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/mem/buffers.go#L39-L41
But it uses an atomic refcount: https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/mem/buffers.go#L75-L80
... but it doesn't use it in an atomically-safe way, as it sets the value to nil when freeing: https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/mem/buffers.go#L160

And if you include the fact that the ref came from a pool: https://github.com/grpc/grpc-go/blob/5edab9e55414068e74320716117a2659c5d2174e/mem/buffers.go#L157
it seems like this code cannot possibly be correct:
- non-racing code is paying for atomics it does not need
- racing code can avoid some checks by using a type that *is* race-safe
  - this might not be reachable in practice though
- racing code can double-`Put` refcounts and buffers (with a racing `Ref`, `Free`, `Ref`, `Free` sequence)
- racing code can leak refcount changes _across Buffer instances_ due to pooled reuse

So this is introducing difficult-to-investigate failure modes that non-atomic or non-pooled instances would not have.  It seems like it'd be better to either just use an `int` field or not-reuse atomics (if the goal is to detect races).

Also I would be quite surprised if pooled `*int` values perform better than non-pooled, given the cheap initialization and `sync.Pool`'s overhead, but I don't have any evidence either way.

Is there some benefit to this I'm missing?  Maybe something about it can catch tricky races or misuse that would otherwise be missed?  It seems misleading and dangerous to me otherwise.

RELEASE NOTES: none

### DIFF
--- a/mem/buffers.go
+++ b/mem/buffers.go
@@ -207,6 +207,9 @@ func (b *buffer) String() string {
 
 // ReadUnsafe reads bytes from the given Buffer into the provided slice.
 // It does not perform safety checks.
+//
+// The returned Buffer MUST be used in place of the one passed in, as it may
+// have been modified or freed.
 func ReadUnsafe(dst []byte, buf Buffer) (int, Buffer) {
 	return buf.read(dst)
 }
@@ -214,6 +217,9 @@ func ReadUnsafe(dst []byte, buf Buffer) (int, Buffer) {
 // SplitUnsafe modifies the receiver to point to the first n bytes while it
 // returns a new reference to the remaining bytes. The returned Buffer
 // functions just like a normal reference acquired using Ref().
+//
+// The returned Buffer values MUST be used in place of the one passed in, as it
+// may have been modified or freed.
 func SplitUnsafe(buf Buffer, n int) (left, right Buffer) {
 	return buf.split(n)
 }


eshitachandwani:
@Groxx Can you please sign the CLA ?

eshitachandwani:
@Groxx , the PR looks good to me , but for the other part , can you open a issue and we can have a discussion there.
Also adding Purnesh for a review.

eshitachandwani (approved):

purnesh42H:
Defer to @dfawley for concurrency concern mentioned.

dfawley (approved):

dfawley:
@Groxx we will need you to complete the CLA in order to merge this, thanks!

PapaCharlie:
Regarding the atomics, each goroutine is intended to have its own `Buffer`. In other words, a single buffer is not meant to be accessed concurrently. The docs mention the following:
> // Note that a Buffer is not safe for concurrent access and instead each
> // goroutine should use its own reference to the data, which can be acquired via
> // a call to Ref().

However, you are right that it doesn't explicitly mention that the `Ref` call should still be synchronous. Namely, the code should look like this:
```go
ref := buf.Ref()
go func() {
  // use ref here
)()
buf.Free()
```
Rather than like this:
```go
go func() {
  ref := buf.Ref()
  // use ref here
}()
buf.Free()
```
Though I would argue that the second form counts as "concurrent access" :) 

The atomic counters are used to determine when the buffer can be safely freed, as the references get counted down.

PapaCharlie:
As for the proposed change in general, the doc looks good to me. The original motivation for `ReadUnsafe` and `SplitUnsafe` was that they should only be used in the `internal/transport` package, where they would be used as expected, but it definitely pays to have better documentation around it! 

> since it feels like a huge footgun otherwise

Absolutely, hence calling them `Unsafe` in the first place. However, without that specific unsafe usage being made available by the `mem` package, it wasn't really possible to introduce this feature without tanking the performance.

Groxx:
> Regarding the atomics, each goroutine is intended to have its own `Buffer`. In other words, a single buffer is not meant to be accessed concurrently. The docs mention the following:
> 
> > // Note that a Buffer is not safe for concurrent access and instead each
> > // goroutine should use its own reference to the data, which can be acquired via
> > // a call to Ref().
> 
> However, you are right that it doesn't explicitly mention that the `Ref` call should still be synchronous. Namely, the code should look like this:
> 
> ```go
> ref := buf.Ref()
> go func() {
>   // use ref here
> )()
> buf.Free()
> ```
> 
> Rather than like this:
> 
> ```go
> go func() {
>   ref := buf.Ref()
>   // use ref here
> }()
> buf.Free()
> ```
> 
> Though I would argue that the second form counts as "concurrent access" :)
> 
> The atomic counters are used to determine when the buffer can be safely freed, as the references get counted down.

So that's actually a good sample of why I think _trying_ to be race-safer here with atomics is actually _worse_ than not using atomics, even just from a safety standpoint.

```go
go func() {
  ref := buf.Ref()
  // use ref here
}()
buf.Free()
```
^ this kind of construct is a common enough mistake that https://github.com/golang/go/issues/18022 has led to adding a `go vet` check to try to prevent it.

Forgetting to wait for the goroutine is also fairly common, so that example should have been something like this:
```go
done := make(chan struct{})
go func() {
  ref := buf.Ref()
  // use ref here
  close(done)
}()
<-done
buf.Free()
```
(I fully realize your example was just sample code and this is reasonable to omit, I'm just showing it to be more concrete)

The risk that atomics bring is that in a `-race` build, _this code will be considered safe_
```go
done := make(chan struct{})
b := buffer()
go func() {
	bb := b.Ref()
	time.Sleep(time.Second)
	// and then do not do anything with the value for some reason,
	// maybe some other check no-opped it.
	// or just don't use it in a racy way, e.g. maybe only read,
	// or only racy-write to different locations
	_ = bb
	close(done)
}() //
b.Free()
<-done // note incorrect waiting, `Ref` and `Free` race
```
while without atomics it'll be a race violation:
```
WARNING: DATA RACE
Read at 0x00c000090308 by goroutine 11:
  scratch/demos.(*normalbuf).Ref()
      .../go/src/scratch/demos/grpc_test.go:33 +0x3c
  scratch/demos.TestNormal.func2.1()
      .../go/src/scratch/demos/grpc_test.go:86 +0x38

Previous write at 0x00c000090308 by goroutine 10:
  scratch/demos.(*normalbuf).Free()
      .../go/src/scratch/demos/grpc_test.go:37 +0x108
  scratch/demos.TestNormal.func2()
      .../go/src/scratch/demos/grpc_test.go:92 +0xf0
```

<details>
<summary>Full test here</summary>

```go
package demos

import (
	"sync/atomic"
	"testing"
	"time"
)

type atomicbuf struct {
	count int64 // must be used atomically
}

func (b *atomicbuf) Ref() *atomicbuf {
	atomic.AddInt64(&b.count, 1)
	return b
}
func (b *atomicbuf) Free() {
	atomic.AddInt64(&b.count, -1)
}

type normalbuf struct {
	count int64
}

func (b *normalbuf) Ref() *normalbuf {
	b.count++
	return b
}
func (b *normalbuf) Free() {
	b.count--
}

func TestAtomic(t *testing.T) {
	t.Run("safe in both constructs", func(t *testing.T) {
		done := make(chan struct{})
		b := &atomicbuf{}
		bb := b.Ref()
		go func() {
			time.Sleep(time.Second)
			// and then do not do anything with `bb` for some reason
			_ = bb
			close(done)
		}()
		b.Free()
		// ensuring the goroutine finishes, rather than ending the process early
		<-done
	})
	t.Run("unsafe but allowed with atomics", func(t *testing.T) {
		done := make(chan struct{})
		b := &atomicbuf{}
		go func() {
			bb := b.Ref() // moved inside the goroutine
			time.Sleep(time.Second)
			// and then do not do anything with the value for some reason
			_ = bb
			close(done)
		}()
		b.Free()
		// ensuring the race occurs, rather than ending the process early
		<-done
	})
}
func TestNormal(t *testing.T) {
	t.Run("safe in both constructs", func(t *testing.T) {
		done := make(chan struct{})
		b := &normalbuf{}
		bb := b.Ref()
		go func() {
			_ = bb // capture it
			time.Sleep(time.Second)
			// and then do not do anything with `bb` for some reason
			close(done)
		}()
		b.Free()
		// ensuring the goroutine finishes, rather than ending the process early
		<-done
	})
	t.Run("unsafe and blocked by race detector", func(t *testing.T) {
		done := make(chan struct{})
		b := &normalbuf{}
		go func() {
			bb := b.Ref() // moved inside the goroutine
			time.Sleep(time.Second)
			// and then do not do anything with the value for some reason
			_ = bb
			close(done)
		}()
		b.Free()
		// ensuring the race occurs, rather than ending the process early
		<-done
	})
}
```
</details>

With racy atomic code like that, it's also possible to trigger double-frees due to the non-atomic `b.refs = nil` line: that may not have propagated across threads, so old `buf` references may `Ref(); Free()` without triggering the `< 0` case with `panic("Cannot free freed buffer")`.  

So it's hiding some misuse, and not reliably preventing misuse.  I'd argue it should either _be safe_ (with added runtime cost) or _not_ (without), not something in the middle.

Groxx:
re CLA...

... tbh I don't think I can agree with a license that requires someone to publicly share their mailing address.  That seems like an absurd overreach.

dfawley:
This is the standard [Linux Foundation](https://en.wikipedia.org/wiki/Linux_Foundation) / [CNCF CLA](https://en.wikipedia.org/wiki/Cloud_Native_Computing_Foundation), and the repo requires signing the agreement in order to merge pull requests.  If you don't want to sign it, that's fine - we will just close this PR.

dfawley:
There's also some chance you may have attempted to go through the "corporate" license workflow instead of the "individual" one.  Asking around, nobody was aware that you'd need to provide a mailing address to agree to the CLA.  Unfortunately, there's no easy way for me to test this for myself.

Groxx:
> There's also some chance you may have attempted to go through the "corporate" license workflow instead of the "individual" one. Asking around, nobody was aware that you'd need to provide a mailing address to agree to the CLA. Unfortunately, there's no easy way for me to test this for myself.

No, it's the individual one.
[Individual_Contributor_License_Agreement.pdf](https://github.com/user-attachments/files/19837757/Individual_Contributor_License_Agreement.pdf)

I see that a fair number of other also have that clause, but nothing seems to describe why it's necessary... and there are also a fair number that _do not_, e.g. Google's OpenSource CLA needs just a name and username.

dfawley:
Blocked on the CLA issue; being investigated in https://github.com/communitybridge/easycla/issues/4638

dfawley:
@Groxx can you try again?  IIUC they removed the address requirement.

Issue #8211, opened 2025-04-01
Flaky Test: Test/ServerFailureMetrics_BeforeResponseRecv


purnesh42H:
https://github.com/grpc/grpc-go/actions/runs/14180815570/job/39726304420?pr=8183

Test introduced in  https://github.com/grpc/grpc-go/pull/8203

janardhanvissa:
https://github.com/grpc/grpc-go/actions/runs/14201703874/job/39789971243?pr=8056

purnesh42H:
https://github.com/grpc/grpc-go/actions/runs/14214736698/job/39828723661?pr=8183

PR #8217, opened 2025-04-02
xdsclient: fix TestServerFailureMetrics_BeforeResponseRecv test to wait for watch to start before stopping the listener

purnesh42H:
Fixes: #8211

`TestServerFailureMetrics_BeforeResponseRecv` was flaking because of race between listener watch to start and closing the listener and if watch hasn't started the stream errors won't be received and server failure metric won't be emitted. Adding the hook to wait for stream to open before stopping the listener have fixed the race.

RELEASE NOTES: None

[Successful forge run](https://fusion2.corp.google.com/invocations/ab09a308-5ffe-414c-ba71-ee7fa07b5a1b/targets/%2F%2Fthird_party%2Fgolang%2Fgrpc%2Fxds%2Finternal%2Fxdsclient:xdsclient_test/log)

### DIFF
--- a/xds/internal/xdsclient/metrics_test.go
+++ b/xds/internal/xdsclient/metrics_test.go
@@ -160,7 +160,19 @@ func (s) TestServerFailureMetrics_BeforeResponseRecv(t *testing.T) {
 	if err != nil {
 		t.Fatalf("net.Listen() failed: %v", err)
 	}
-	mgmtServer := e2e.StartManagementServer(t, e2e.ManagementServerOptions{Listener: l})
+	lis := testutils.NewRestartableListener(l)
+	streamOpened := make(chan struct{}, 1)
+	mgmtServer := e2e.StartManagementServer(t, e2e.ManagementServerOptions{
+		Listener: lis,
+		OnStreamOpen: func(context.Context, int64, string) error {
+			select {
+			case streamOpened <- struct{}{}:
+			default:
+			}
+			return nil
+		},
+	})
+
 	nodeID := uuid.New().String()
 
 	bootstrapContents, err := bootstrap.NewContentsForTesting(bootstrap.ConfigOptionsForTesting{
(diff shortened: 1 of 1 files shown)


xds/internal/xdsclient/metrics_test.go
```diff
@@ -196,15 +208,24 @@ func (s) TestServerFailureMetrics_BeforeResponseRecv(t *testing.T) {
...
 	// cause a server failure count to emit eventually.
-	l.Close()
+	lis.Stop()
 	select {
 	case <-ctx.Done():
 		t.Fatal("Timeout when waiting for ADS stream to close")
 	default:
 	}
+	// Restart to prevent the attempt to create a new ADS stream after back off.
+	lis.Restart()
```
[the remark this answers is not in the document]
purnesh42H: So, in this particular test case it won't happen because we are setting the watch expiry timeout which is way shorter than the back off. So, watch will just expire before backoff anyway

xds/internal/xdsclient/metrics_test.go
```diff
@@ -196,15 +208,24 @@ func (s) TestServerFailureMetrics_BeforeResponseRecv(t *testing.T) {
...
 
 	// Close the listener and ensure that the ADS stream breaks. This should
 	// cause a server failure count to emit eventually.
-	l.Close()
+	lis.Stop()
 	select {
 	case <-ctx.Done():
 		t.Fatal("Timeout when waiting for ADS stream to close")
 	default:
 	}
```
dfawley, line 226: This does ~nothing now - delete it?

And before, it should have been written as: `if ctx.Err() != nil {...}`

dfawley (approved):

xds/internal/xdsclient/metrics_test.go
purnesh42H: yeah not needed because both lis.Stop() and lis.Restart() are serialized by lock.

Replaced the similar select block in the other test as with `if ctx.Err() != nil {...}`

PR #8224, opened 2025-04-04
Update lrs_stream.go fix use of wrong err

alingse:
the origin `err` has been checked in line 233.

it is nil now, and after check `CheckInvalid() != nil`,

here should use the check err

RELEASE NOTES:
* xds: Fix reported error string when LRS load reporting interval is invalid.

### DIFF
--- a/xds/internal/xdsclient/transport/lrs/lrs_stream.go
+++ b/xds/internal/xdsclient/transport/lrs/lrs_stream.go
@@ -242,7 +242,7 @@ func (lrs *StreamImpl) recvFirstLoadStatsResponse(stream transport.StreamingCall
 	}
 
 	internal := resp.GetLoadReportingInterval()
-	if internal.CheckValid() != nil {
+	if err:= internal.CheckValid(); err != nil {
 		return nil, 0, fmt.Errorf("lrs: invalid load_reporting_interval: %v", err)
 	}
 	loadReportingInterval := internal.AsDuration()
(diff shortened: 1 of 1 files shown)


xds/internal/xdsclient/transport/lrs/lrs_stream.go
```diff
@@ -242,7 +242,7 @@ func (lrs *StreamImpl) recvFirstLoadStatsResponse(stream transport.StreamingCall
 	}
 
 	internal := resp.GetLoadReportingInterval()
```
arjan-bal, line 244: I believe the original author meant to name this variable `interval`. Can you please fix the name also?

xds/internal/xdsclient/transport/lrs/lrs_stream.go
```diff
@@ -242,7 +242,7 @@ func (lrs *StreamImpl) recvFirstLoadStatsResponse(stream transport.StreamingCall
...
-	if internal.CheckValid() != nil {
+	if err:= internal.CheckValid(); err != nil {
```
arjan-bal, line 245: The `vet` check runs `gofmt` and is failing because a space is needed b/w `err:=`. Can you please fix the formatting?

xds/internal/xdsclient/transport/lrs/lrs_stream.go
alingse: ok

arjan-bal (approved):
LGTM, thanks! Adding a second reviewer.

easwars (approved):

Issue #8216, opened 2025-04-02
New A72 changes for OpenTelemetry


purnesh42H:
Several minor adjustments proposed in gRFC A72 (https://github.com/grpc/proposal/pull/474) need to be implemented within the OpenTelemetry tracing code

- Add names for parent call span, attempt span and server span based on existing OpenCensus implementation.
    - https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/client_tracing.go
    - https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/server_tracing.go
- Remove superfluous and confusing sent and received words from Inbound message received and Outbound message sent. 
    - https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/trace.go

carsontham:
hi, I would like to help with this issue. 

For the `Inbound message received` and `Outbound message sent` part, the current code already adheres to gRFC A72, using messages like `Inbound compressed message` and `Outbound compressed message` in [trace.go](https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/trace.go) 

So it seems the remaining change is to update the span names in `client_tracing.go` and `server_tracing.go`. 

Am I understanding this issue correctly?

arjan-bal:
Hi @carsontham, @vinothkumarr227 already has a #8226 open to address this issue, so we'll let him handle it.

vinothkumarr227:
I'm looking into this. Please assign the issue to me.

PR #8226, opened 2025-04-04
New A72 changes for OpenTelemetry #8216

vinothkumarr227:
Fixes: #8216

RELEASE NOTES: None


stats/opentelemetry/trace.go
```diff
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+// OpenCensus's binary format for grpc-trace-bin:
```
purnesh42H, line 17: no need to add this. This was only for grfc

purnesh42H (commented):
was there no change needed for Attempt? Is it already following https://github.com/grpc/proposal/pull/474 ?

stats/opentelemetry/trace.go
vinothkumarr227: Done

vinothkumarr227:
> was there no change needed for Attempt? Is it already following [grpc/proposal#474](https://github.com/grpc/proposal/pull/474) ?

The Attempt span name already follows the grpc/proposal#474 naming convention. 
https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/client_tracing.go#L47

purnesh42H (approved):
lgtm

purnesh42H:
@dfawley for second review

yashykt (approved):

stats/opentelemetry/trace.go
```diff
@@ -58,14 +58,14 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
 			attribute.Int64("message-size", int64(rs.Length)),
 			attribute.Int64("message-size-compressed", int64(rs.CompressedLength)),
 		))
 	case *stats.OutPayload:
 		ai.countSentMsg++
-		span.AddEvent("Outbound compressed message", trace.WithAttributes(
+		span.AddEvent("Outbound message", trace.WithAttributes(
 			attribute.Int64("sequence-number", int64(ai.countSentMsg)),
 			attribute.Int64("message-size", int64(rs.Length)),
 			attribute.Int64("message-size-compressed", int64(rs.CompressedLength)),
```
yashykt, line 71: the grfc says to only add this if the message is actually compressed

stats/opentelemetry/trace.go
```diff
@@ -58,14 +58,14 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
 		// message id - "must be calculated as two different counters starting
 		// from one for sent messages and one for received messages."
 		ai.countRecvMsg++
-		span.AddEvent("Inbound compressed message", trace.WithAttributes(
+		span.AddEvent("Inbound message", trace.WithAttributes(
 			attribute.Int64("sequence-number", int64(ai.countRecvMsg)),
 			attribute.Int64("message-size", int64(rs.Length)),
 			attribute.Int64("message-size-compressed", int64(rs.CompressedLength)),
```
yashykt, line 64: the grfc says to only add this if the message is actually compressed

yashykt:
We also need tests for streaming, but that can be added in a subsequent PR

purnesh42H:
> message-size-compressed

the e2e tests contains tests for both unary and stream RPCs. Are you referring to anything else?

stats/opentelemetry/trace.go
purnesh42H: gRFC says if compression is not a separate event, just add the key `message-size-compressed` in the same event "Inbound Message" and "Outbound Message". For go implementation, if compression is not enabled `message-size-compressed` is same as actual message.  So, i guess what we have as of now is fine? or should we only add if compressed and actual are not equal?

purnesh42H:
@vinothkumarr227 please resolve the conflicts from https://github.com/grpc/grpc-go/pull/8237

vinothkumarr227:
> @vinothkumarr227 please resolve the conflicts from #8237

I have resolved the conflicts

stats/opentelemetry/trace.go
yashykt: The text says - 
```If compression needed, add key message-size-compressed with integer value of compressed message size. ```

**Comment by yashykt** [comment]
> the e2e tests contains tests for both unary and streaming RPCs. Are you referring to anything else?

In all of the tests, I see only a single message being sent/received, which means that the logic on the sequence numbers incremented is untested.

File: stats/opentelemetry/trace.go (hunk 2 above)
Comment (purnesh42H) [reply stats/opentelemetry/trace.go:70]: Thanks. @vinothkumarr227 could you do similar to what Java has https://github.com/grpc/grpc-java/blob/f79ab2f16f7e037da0e8c9985e917eca552c22cb/opentelemetry/src/main/java/io/grpc/opentelemetry/OpenTelemetryTracingModule.java#L439 which is to add message-size-compressed only if compression was enabled.

Do for both inbound and outbound. For compression enabled, check if message-size is not equal to compressed length https://github.com/grpc/grpc-go/blob/732f3f32f5365f4f5b89a3f1be19534cd28db794/stats/stats.go#L89

File: stats/opentelemetry/trace.go (hunk 3 above)
Comment (purnesh42H) [reply stats/opentelemetry/trace.go:63]: Thanks. @vinothkumarr227 could you do similar to what Java has https://github.com/grpc/grpc-java/blob/f79ab2f16f7e037da0e8c9985e917eca552c22cb/opentelemetry/src/main/java/io/grpc/opentelemetry/OpenTelemetryTracingModule.java#L439 which is to add message-size-compressed only if compression was enabled.

Do for both inbound and outbound. For compression enabled, check if message-size is not equal to compressed length https://github.com/grpc/grpc-go/blob/732f3f32f5365f4f5b89a3f1be19534cd28db794/stats/stats.go#L89

Comment (vinothkumarr227) [reply stats/opentelemetry/trace.go]: @purnesh42H  I have added both inbound and outbound. For compression to be considered enabled

File: stats/opentelemetry/trace.go [hunk 4]
```diff
@@ -57,18 +57,24 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
 		// message id - "must be calculated as two different counters starting
 		// from one for sent messages and one for received messages."
-		span.AddEvent("Inbound compressed message", trace.WithAttributes(
+		attrs := []attribute.KeyValue{
 			attribute.Int64("sequence-number", int64(ai.countRecvMsg)),
 			attribute.Int64("message-size", int64(rs.Length)),
-			attribute.Int64("message-size-compressed", int64(rs.CompressedLength)),
-		))
+		}
+		if rs.CompressedLength > 0 && rs.CompressedLength != rs.Length {
```
purnesh42H, line 64: nit: no need to check > 0?

stats/opentelemetry/trace.go
```diff
@@ -57,18 +57,24 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
 		ai.countRecvMsg++
 	case *stats.OutPayload:
-		span.AddEvent("Outbound compressed message", trace.WithAttributes(
+		attrs := []attribute.KeyValue{
 			attribute.Int64("sequence-number", int64(ai.countSentMsg)),
 			attribute.Int64("message-size", int64(rs.Length)),
-			attribute.Int64("message-size-compressed", int64(rs.CompressedLength)),
-		))
+		}
+		if rs.CompressedLength > 0 && rs.CompressedLength != rs.Length {
```
purnesh42H, line 74: nit: no need to check > 0?

vinothkumarr227: Done

dfawley (approved):

PR #8230, opened 2025-04-06
resolver/dns: add RefreshInterval parameter to enable periodic DNS record refresh

r9deyes:
This pull request introduces a new feature to the DNS resolver by adding a `RefreshInterval` variable, which defines the rate at which the resolver refreshes its DNS records. The default value is set to `0`, disabling the refresh unless configured otherwise. 

Key changes include:

- **New `RefreshInterval` Variable**: Allows for periodic DNS record refreshes, with a requirement that the value must be at least `MinResolutionInterval`.
- **Updated `watcher` Function**: Incorporates a new channel (`refreshCh`) to handle refresh signals, enabling automatic updates of DNS records alongside resolution requests.
- **Setter Function**: A new `SetRefreshInterval` function is added to allow users to configure the refresh interval while ensuring it meets the minimum requirement.

These enhancements improve the DNS resolver's responsiveness and accuracy in dynamic environments. Notably, similar functionality is available in the gRPC implementation for ASP.NET Core, which also features DNS address caching and refresh mechanisms (see [ASP.NET Core gRPC Load Balancing Documentation](https://learn.microsoft.com/en-us/aspnet/core/grpc/loadbalancing?view=aspnetcore-8.0#dns-address-caching)). This alignment with existing practices enhances consistency across platforms.

RELEASE NOTES:
- resolver/dns: add RefreshInterval parameter to enable periodic DNS record

### DIFF
--- a/internal/resolver/dns/dns_resolver.go
+++ b/internal/resolver/dns/dns_resolver.go
@@ -58,6 +58,10 @@ var (
 	// after initialization as it's not thread-safe for concurrent modification.
 	ResolvingTimeout = 30 * time.Second
 
+	// RefreshInterval is used as DNS refresh rate. Resolver refresh itself on a periodic interval.
+	// The value configured must be at least MinResolutionInterval. Default value is 0 as disabled.
+	RefreshInterval = time.Duration(0)
+
 	logger = grpclog.Component("dns")
 )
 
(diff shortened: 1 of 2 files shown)


arjan-bal:
The scenario when new k8s pods are added and the LB policy uses all the addresses (e.g round robin), refreshing the resolved addresses could be useful. I checked the DNS resolver in [gRPC Java](https://github.com/grpc/grpc-java/blob/176f3eed12922bfa460b4604ccb0b8c50f31194f/core/src/main/java/io/grpc/internal/DnsNameResolver.java), but didn't find the ability to do a periodic refresh. Checking with maintainers of other languages if we need consistent behaviour for this.

arjan-bal:
Hi @r9deyes, after discussing this with other maintainers, I found that not configuring a client side refresh interval intentional. We don't want servers to rely on hard-coded client behavior for their load balancing stability. Updating clients is very slow. There is a discussion about a similar request here: https://github.com/grpc/grpc-java/issues/2514

The general suggestion is to use MAX_CONNECTION_AGE on server-side to force the client to re-resolve. This is the approach of internet-scale load balancing, though it may be less efficient.

```go
serverOpts = append(serverOpts, grpc.KeepaliveParams(keepalive.ServerParameters{
	MaxConnectionAge: 5 * time.Minute,
}))
```

I'll be closing this PR, but will re-open if there are any questions.

Issue #6862, opened 2023-12-18
client: implement Is() and Unwrap() in status.Error to not mask context.Canceled


pracucci:
### Use case(s) - what problem will this feature solve?

The gRPC client (e.g. `grpc.ClientStream`) returns a gRPC error when the context is canceled or the deadline is exceeded. The actual standard golang error (e.g. `context.Canceled`) is masked by the gRPC error and so `errors.Is(grpcErr, context.Canceled)` returns `false` even if the error has been actually generated by context being canceled.

This behaviour makes checking for context cancellation (or deadline exceed) a bit cumbersome in our application because we have to re-map the returned gRPC error into a `context.Canceled` or `context.DeadlineExceeded`, given the rest of the application code is already instrumented to check it but not gRPC-specific error codes. In other words, what I mean is that we're used to check gRPC-specific error codes in the code blocks directly interacting with gRPC but when these errors are returned by functions, the callers are typically gRPC agnostic.

### Proposed Solution

I propose to implement `Is(err error) bool` and `Unwrap() error` on `status.Error`:

- `status.Error.Is(context.Canceled)` would return `true` if the status error is `codes.Canceled()` (and similar for `context.DeadlineExceeded`
- `status.Error.Unwrap()` would return `context.Canceled` if the status error is `codes.Canceled()` (and similar for `context.DeadlineExceeded`

_Does it make sense or are there good reasons why we shouldn't do it? If we'll get a consensus (either on this proposal or a different one after the discussion) I'm open to work on the code changes._

dfawley:
Interesting idea.  There is one hard case to consider:

* `CANCELED` and `DEADLINE_EXCEEDED` are status codes that can be returned arbitrarily from the server even if the client didn't cancel the RPC or reach its deadline.  Is it proper to return `context` errors (or something that `Is()` them) in that case?

> when these errors are returned by functions, the callers are typically gRPC agnostic.

In these cases, wouldn't the callers usually expect to be able to do a direct comparison to `context.Canceled` vs. an `errors.Is()` call?

If you're making a library that wraps gRPC in an attempt to hide gRPC from your users, then you should really consider doing this conversion (and converting all other errors as well) before returning them.  If you directly pass through our return values, our API becomes part of your API.

pracucci:
[...]

PR #8241, opened 2025-04-10
feat: export context errors

costela:
Since #6862 was closed as wontfix, this is a follow-up proposal to actuall pass along errors the caller may be interested it.

The current wrapped errors are only used for the specific cases of context cancellation and deadline. The rationale being: these are the errors that are caused via passed in context arguments from the caller.


arjan-bal:
There were 3 issues mentioned in the previously closed issue: https://github.com/grpc/grpc-go/issues/6862#issuecomment-1864991538

> 1. I think it would not be appropriate to wrap context.Canceled/DeadlineExceeded if the server returns those statuses, because the context itself may still be valid.
> 2. We could possibly (? -it might be easy, or it might be very very hard) detect whether the cancelation/deadline happened on the client-side and only conditionally wrap.
> 3. If we did 2, it's still theoretically possible there could be a race between the server deciding the client's deadline was exceeded and the client noticing the context expired, meaning we could get it wrong sometimes.

For 1 & 2, the error types modified in the PR are only used for client side context cancellation, so we should be good. For 3, the client would still see status DEADLINE_EXCEEDED with the message set here, which would probably give a nil error on unwrapping:
https://github.com/grpc/grpc-go/blob/732f3f32f5365f4f5b89a3f1be19534cd28db794/internal/transport/http2_client.go#L1238-L1245

@dfawley would like to get your thoughts on this since you handled https://github.com/grpc/grpc-go/issues/6862.

costela:
@arjan-bal thanks for the summary! Yes, the intention was to side-step the issues mentioned in the previous PR.

dfawley:
It would be better if you could follow the standard feature request issue template instead of creating a PR (https://github.com/grpc/grpc-go/issues/new?template=feature.md).  Specifically, what problem are you attempting to solve by this?

PR #8244, opened 2025-04-13
Add unit tests for registering a codec

isaiahvita:
The existing unit tests do not test the ability to test registering a codec via `encoding.RegisterCodecV2`.
This PR checks that the existing behavior thats currently implemented in the function will continue to work as it currently does.
Additionally, this brings the coverage report of encoding_v2.go from 75% to 100%.

I tested this by running the unit test and seeing it pass.

RELEASE NOTES: None

### DIFF
--- a/encoding/encoding_test.go
+++ b/encoding/encoding_test.go
@@ -363,3 +363,101 @@ func (s) TestForceCodecName(t *testing.T) {
 		t.Fatalf("ss.Client.EmptyCall(_, _) = _, %v; want _, nil", err)
 	}
 }
+
+type noopCodec struct {
+	encoding.CodecV2
+	ID        int
+	CodecName string
+}
+
+func (m *noopCodec) Name() string {
+	return m.CodecName
+}
+
+// TestRegisterCodecV2 tests the existing behavioral assumptions made by
+// RegisterCodeV2. The behavioral assumptions are:
+// 1. invalid codecs (nil, empty name) should panic
+// 2. same name codecs are valid, they just overwrite previous.
+func TestRegisterCodecV2(t *testing.T) {
+	tests := []struct {
+		name        string
+		codec1      encoding.CodecV2
+		codec2      encoding.CodecV2
+		expectPanic bool
+	}{
+		{
+			name:        "nil codes, panic",
+			expectPanic: true,
+		},
+		{
+			name: "empty codecs, panic",
+			codec1: &noopCodec{
+				CodecName: "",
+			},
+			codec2: &noopCodec{
+				CodecName: "",
+			},
+			expectPanic: true,
+		},
+		{
+			name: "one of the codec has valid name",
+			codec1: &noopCodec{
+				CodecName: "first",
+			},
+			expectPanic: false,
+		},
+		{
+			name: "new codec overrides original codec if name is same",
+			codec1: &noopCodec{
+				ID:        1,
+				CodecName: "first",
+			},
+			codec2: &noopCodec{
+				ID:        2,
+				CodecName: "first",
+			},
+			expectPanic: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
+			func() {
+				defer func() {
+					r := recover()
+					// If a panic occurred
+					if r != nil {
+						if !test.expectPanic {
+							t.Errorf("[%v] Unexpected panic: %v\n", test.name, r)
+						}
+					} else {
+						// If no panic occurred
+						if test.expectPanic {
+							t.Errorf("[%v] Expected panic but function did not panic\n", test.name)
+						}
+					}
+				}()
+
+				// test behavior on initial registration
+				encoding.RegisterCodecV2(test.codec1)
+				actualCodec := encoding.GetCodecV2(test.codec1.Name())
+				if !cmp.Equal(test.codec1, actualCodec) {
+					t.Errorf("[%v] Expected returned codec to be equal.\n", test.name)
+				}
+				if test.codec2 == nil {
+					// only test behavior on subsequent registrations
+					// if the test case prescribes it by setting codec2
+					t.Skip()
+				}
+
+				// test behavior on subsequent registrations
+				encoding.RegisterCodecV2(test.codec2)
+				actualCodec = encoding.GetCodecV2(test.codec1.Name())
+				if !cmp.Equal(actualCodec, test.codec2) {
+					t.Errorf("[%v] Expected returned codec to be equal.\n", test.name)
+				}
+			}()
+		})
+	}
+}


encoding/encoding_test.go
```diff
@@ -62,6 +63,23 @@ func (mockNamedCompressor) Name() string {
 	return "mock-compressor"
 }
 
+type mockNoopCodec struct {
```
purnesh42H, line 66: may be we can just wrap the encoding.CodecV2 and just override the name()? Similar to `renameProtoCodec`.

Also, declare the proto just above the test and may be just keep it as `noopCodec`

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
 		t.Fatalf("ss.Client.EmptyCall(_, _) = _, %v; want _, nil", err)
 	}
 }
+
+// TestRegisterCodec tests the existing behavioral assumptions made by
```
purnesh42H, line 385: s/TestRegisterCodec/TestRegisterCodecV2

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+// RegisterCodeV2. The behavioral assumptions are:
+// 1. invalid codecs (nil, empty name) should panic
+// 2. same name codecs are valid, they just overwrite previous.
+func TestRegisterCodec(t *testing.T) {
+	cases := map[string]struct {
```
purnesh42H, line 390: you can do something like this instead of having a map
```
tests := []struct {
		name          string
		origCodec  encoding.CodecV2
		newCodec  encoding.CodecV2
		wantPanic bool
	}
```

purnesh42H: example to refer https://github.com/grpc/grpc-go/blob/eb4b6877646bd942f73a2a1e8c730448f8efe7bd/clientconn_authority_test.go#L38

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+		codec1      encoding.CodecV2
+		codec2      encoding.CodecV2
+		expectPanic bool
+	}{
+		"nil code panics": {
```
purnesh42H, line 395:
```suggestion
		"nil codes, panic": {
```

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+			expectPanic: true,
+		},
+		"invalid codec name panics": {
```
purnesh42H, line 398:
```suggestion
		"empty codecs, panic": {
```

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+			codec1: &mockNoopCodec{
+				name: "",
+			},
+			codec2: &mockNoopCodec{
+				name: "",
+			},
+			expectPanic: true,
+		},
+		"valid codec": {
```
purnesh42H, line 407:
```suggestion
		"one of the codec has valid name": {
```

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+			codec1: &mockNoopCodec{
+				name: "first",
+			},
+			expectPanic: false,
+		},
+		"codec overridden": {
```
purnesh42H, line 413:
```suggestion
		"new codec overrides original codec if name is same": {
```

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+						if c.expectPanic {
+							t.Errorf("[%v] Expected panic but function did not panic\n", caseName)
+						}
+					}
+				}()
+
+				encoding.RegisterCodecV2(c.codec1)
+
+				actualCodec := encoding.GetCodecV2(c.codec1.Name())
+				if !reflect.DeepEqual(c.codec1, actualCodec) {
```
purnesh42H, line 448: can we use cmp instead of reflect?

encoding/encoding_test.go
```diff
@@ -363,3 +381,84 @@ func (s) TestForceCodecName(t *testing.T) {
...
+					t.Errorf("[%v] Expected returned codec to be equal.\n", caseName)
+				}
+				if c.codec2 == nil {
```
purnesh42H, line 451: it looks like we have too many conditions in one test. may be should separate the tests into valid and invalid codec cases? For invalid, we just need one codec then.

isaiahvita: hmm im not sure i agree with that. separating into two different tests would only allow the `defer` to be in its own test, and then well have to duplicate the equality conditions.

let me add some comments to make it more readable, and let me know what you think about that. if you still would like to have two separate tests, then i can change it.

encoding/encoding_test.go
```diff
@@ -363,3 +363,101 @@ func (s) TestForceCodecName(t *testing.T) {
...
+	encoding.CodecV2
+	ID        int
+	CodecName string
+}
+
+func (m *noopCodec) Name() string {
+	return m.CodecName
+}
+
+// TestRegisterCodecV2 tests the existing behavioral assumptions made by
```
purnesh42H, line 377:
```suggestion
// TestRegisterCodecV2 tests verify the following behaviors of RegisterCodeV2:
- invalid codecs (nil, empty name) should panic
- if new codecs has same name as existing, existing codec is overwritten
- if codec with same name does not exist, it is registered
```

encoding/encoding_test.go
```diff
@@ -363,3 +363,101 @@ func (s) TestForceCodecName(t *testing.T) {
...
+// 2. same name codecs are valid, they just overwrite previous.
+func TestRegisterCodecV2(t *testing.T) {
+	tests := []struct {
+		name        string
+		codec1      encoding.CodecV2
+		codec2      encoding.CodecV2
+		expectPanic bool
+	}{
+		{
+			name:        "nil codes, panic",
```
purnesh42H, line 389: typo: nil codecs, panics

encoding/encoding_test.go
```diff
@@ -363,3 +363,101 @@ func (s) TestForceCodecName(t *testing.T) {
...
+}
+
+// TestRegisterCodecV2 tests the existing behavioral assumptions made by
+// RegisterCodeV2. The behavioral assumptions are:
+// 1. invalid codecs (nil, empty name) should panic
+// 2. same name codecs are valid, they just overwrite previous.
+func TestRegisterCodecV2(t *testing.T) {
+	tests := []struct {
+		name        string
+		codec1      encoding.CodecV2
```
purnesh42H, line 384: might be good to name codec1 and codec2 as existingCodec and newCodec or add a comment against them

encoding/encoding_test.go
purnesh42H: I still think we should have separate test case for failures because there is lot of code which is not applicable to success cases. It is a good enough reason to make a separate test.

purnesh42H:
2 more things to keep in mind for future reference
1. Please do not force-push. Add a new commit for any change after review
2. Please do not resolve the comments. Reviewers can resolve comments once they are addressed to expectation. You can respond to the comments stating how you have addressed them.

Issue #8233, opened 2025-04-07
protoc-gen-go-grpc support split single proto file gen to diffrent package


qianlongzt:
### Use case(s) - what problem will this feature solve?

opentelemetry-go[1] supports protobuf over gRPC and over HTTP, but its [proto file][4] defines both the service and related messages in the same file.
When using protobuf over HTTP, it's necessary to remove the gRPC dependency to reduce the binary size.

### Proposed Solution

Add an option to rename grpc file.GoImportPath [5] 	
```diff
-	g := gen.NewGeneratedFile(filename, file.GoImportPath)
+	path := file.GoImportPath
+ 	if n, ok := rename[file.Desc.Path()]; ok {
+ 		path = n
+	}
+	g := gen.NewGeneratedFile(filename, path)
```

```text
-grpc_rename log.proto=github.com/aaa/bbb/xxx -grpc_rename trace.proto=github.com/aaa/bbb/yyy
```

### Alternatives Considered

1. opentelemetry-proto split proto file
2. opentelemetry-proto-go add same shim file https://github.com/open-telemetry/opentelemetry-proto-go/pull/248

### Additional Context

https://github.com/open-telemetry/opentelemetry-go/issues/2579

[1]: https://github.com/open-telemetry/opentelemetry-go
[2]: https://github.com/open-telemetry/opentelemetry-proto-go
[3]: https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/otlp/otlplog
[4]: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/logs/v1/logs_service.proto#L30
[5]: https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/grpc.go#L140
[6]: https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/main.go#L60

arjan-bal:
Hi @qianlongzt, I was going though the linked issue but I don't completely understand how the proposed change addresses the issue. I found this comment that summarises the change: https://github.com/open-telemetry/opentelemetry-go/issues/2579#issuecomment-2609059459

Can you elaborate how your proposed solution will solve the problem of splitting gRPC client/server code and protobuf request/response messages into two separate modules?

qianlongzt:
@arjan-bal https://github.com/grpc/grpc-go/pull/8245 I created a pr.

arjan-bal:
[...]

PR #8245, opened 2025-04-13
feat(cmd/gen): support mapping _grpc.pb.go files to different Go packages from .pb.go

qianlongzt:
https://github.com/grpc/grpc-go/issues/8233


arjan-bal:
Thanks for sharing the code, I'll close the PR for now as I don't want it to be reviewed yet. Once we agree on the approach in the issue, we will re-open this.

Issue #8228, opened 2025-04-04
Make least request available by default (remove GRPC_EXPERIMENTAL_ENABLE_LEAST_REQUEST)


atollena:
The gRFC https://github.com/grpc/proposal/blob/master/A48-xds-least-request-lb-policy.md states that the environment variable will be removed when the feature has proven stable. The last substantial changes to least request have been made in https://github.com/grpc/grpc-go/pull/6603 (1.58, Sep 2023).

We have successfully used least request in production for about 1 year, only discovering one minor issue last summer (fixed in https://github.com/grpc/grpc-go/pull/7467, 1.66).

We should consider removing the environment variable and always making least request available.

arjan-bal:
Based on the [comment from Eric](https://github.com/grpc/grpc-java/issues/11996#issuecomment-2787250309) assigning to @dfawley to share the result of the discussion.

davinci26:
+1 on what @atollena is saying, we are also using it in `prod` for 6m+ with no issues

dfawley:
#8253 is merged to enable it by default.  Let's leave this open to track removing the flag after the 1.73 release.

PR #8248, opened 2025-04-14
xds: remove temporary environment variable for least request

atollena:
Fixes https://github.com/grpc/grpc-go/issues/8228.

RELEASE NOTES:
- Remove the `GRPC_EXPERIMENTAL_ENABLE_LEAST_REQUEST` environment variable that allows disabling the least request balancer with xDS. Least request was made available by default with xDS in v1.72.0.


atollena:
As discussed I'll keep this PR for later and work on a change that just flips the environment variable default to `true`.

arjan-bal:
@dfawley is it sufficient for one minor version (1.72) to have the flag defaulted to `true` before removing the env var completely? If yes, we can  review and merge this.

dfawley (approved):
In this particular case, I think it's fine.  I am not really concerned that registering this LB policy by default is going to cause problems.  For other things that might be more risky, because we know our users don't always upgrade in a timely fashion, I think it's better to wait at least 2-3 releases before removing the flag.

arjan-bal:
I re-checked and found that 1.72.0 doesn't have https://github.com/grpc/grpc-go/pull/8253 that set the default to true. SO we need to wait for https://github.com/grpc/grpc-go/pull/8253 to be released with 1.73 and remove the env var in 1.74.

atollena:
> I re-checked and found that 1.72.0 doesn't have #8253 that sets the default to true. So we need to wait for #8253 to be released with 1.73 and remove the env var in 1.74.

Sounds good to me, I'll resume this when 1.73 has been released.

arjan-bal (approved):

Issue #8213, opened 2025-04-01
Consider moving ring hash out of the xds package


atollena:
The implementation of gRFC A76 (https://github.com/grpc/grpc-go/pull/8159) due in 1.72.0 makes it possible to use the ring hash balancer without xds. Most of the implementation currently lies in the `xds/internal/balancer/ringhash` package.

We should move it somewhere where it can be imported without pulling in all of xds.

atollena:
Waiting on https://github.com/grpc/grpc-go/pull/8159/files to be merged.

PR #8249, opened 2025-04-14
ringhash: move out of the xds package

atollena:
fixes https://github.com/grpc/grpc-go/issues/8213

There is a new API in that ring hash can now be used by importing `google.golang.org/grpc/balancer/ringhash`
instead of importing all of `google.golang.org/grpc/xds`.

RELEASE NOTES: N/A


internal/xds/xds.go
```diff
@@ -40,3 +42,24 @@ func GetXDSHandshakeClusterName(attr *attributes.Attributes) (string, bool) {
...
 }
+
+// xdsHashKey is the type used as the key to store request hash in the context
+// used when combining the Ring Hash load balancing policy with xDS.
+type xdsHashKey struct{}
+
+// GetXDSRequestHash returns the request hash in the context and true if it was set
+// from the xDS config selector. If the xDS config selector has not set the hash,
+// it returns 0 and false.
+func GetXDSRequestHash(ctx context.Context) (uint64, bool) {
```
atollena, line 53: For those two functions I decided to follow the convention established above with Get/SetXDSHandshakeClusterName despite the redundancy with the xds package name.

XDSHeaderName violates the rule that says that function names should not start with the package name, so I added the Get prefix, but I could also go with `xds.SetRequestHash` and `xds.RequestHash`. Let me know if you have a preference.

arjan-bal:
This test failure is in a new ringhash test, but probably not related to this PR: https://github.com/grpc/grpc-go/actions/runs/14454284376/job/40533854683?pr=8249

arjan-bal (approved):

internal/ringhash/ringhash.go
dfawley, line 1: Curious: is setting the xds request hash something that anyone else would need to be able to do?  E.g. for testing?  I'm fine with leaving this as-is for now, but I'm also OK with making that functionality public if we think it might be needed.

balancer/ringhash/ringhash.go
dfawley, line 1: It might be nice to point to the gRFC in the package comments:

https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md

internal/ringhash/ringhash.go
atollena: The main problem here is a cross-language concern raised by Mark: see https://github.com/grpc/grpc/issues/33356#issuecomment-1894196648:

> However, I am a little concerned about the complexity involved in exposing an interface to programmatically set the request's hash. When we discussed this earlier, we were thinking that this would probably just be something simple like "use a specific request header", which could just be built into the ring_hash policy. Exposing an API to allow applications to do this programmatically is a much broader API surface, and before going down that road, I'd like to be convinced that there's no simpler way to do this. Is there a way that we can handle your use-case without that complexity?

So I think that in terms of cross-language supports, it's better to limit the visibility to the xds use case.

balancer/ringhash/ringhash.go
atollena: Done.

balancer/ringhash/e2e/ringhash_balancer_test.go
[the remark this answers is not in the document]
atollena: Done.

PR #8257, opened 2025-04-16
ringhash: fix flaky e2e tests

atollena:
Fixes https://github.com/grpc/grpc-go/issues/8247.

1. Fix Test/RingHash_RequestHashKeyRandom by setting minRingSize to a higher value as we were doing in other tests. It lowers the skew that may occur from an imbalanced ring.

2. Fix Test/RingHash_RequestHashKeyConnecting by waiting for at least one connection attempt before proceeding to check for other connection attempts. That accounts for the unlikely case where the connection attempt has not been made by the time the channel transitions to CONNECTING state.

Testing done:
```
go test -run='Test/RingHash_RequestHashKey(Connecting|Random)' ./xds/internal/balancer/ringhash/e2e -count 10000
```

RELEASE NOTES: N/A

### DIFF
--- a/xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
+++ b/xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
@@ -81,6 +81,11 @@ const (
 	errorTolerance = .05 // For tests that rely on statistical significance.
 
 	virtualHostName = "test.server"
+
+	// minRingSize is the minimum ring size to use when testing randomly a
+	// backend for each request. It lowers the skew that may occur from
+	// an imbalanced ring.
+	minRingSize = 10000
 )
 
 // fastConnectParams disables connection attempts backoffs and lowers delays.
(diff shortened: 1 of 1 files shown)


xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -2847,11 +2861,16 @@ func (s) TestRingHash_RequestHashKeyConnecting(t *testing.T) {
 	}()
 	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
 
-	// Check that only one connection attempt was started.
+	// Wait for at least one connection attempt.
 	nConn := 0
-	for _, hold := range holds {
-		if hold.IsStarted() {
-			nConn++
+	for nConn == 0 {
```
arjan-bal, line 2866: nit: We should wait for a small duration like 1 millis b/w each iteration of the loop to avoid too much unnecessary work.

xds/internal/balancer/ringhash/e2e/ringhash_balancer_test.go
```diff
@@ -2847,11 +2861,16 @@ func (s) TestRingHash_RequestHashKeyConnecting(t *testing.T) {
 	}()
 	testutils.AwaitState(ctx, t, cc, connectivity.Connecting)
```
arjan-bal, line 2862: Pickfirst will report `CONNECTING` as soon as it is created. It doesn't wait for a subchannel to report `CONNECTING`. That's the reason this check wasn't sufficient. Should we remove it? The check for 1 connection attempt below seems to be sufficient.

arjan-bal (approved):
LGTM

PR #8260, opened 2025-04-16
cleanup: status formatting bug and comment grammar fix

purnesh42H:
RELEASE NOTES: None

### DIFF
--- a/xds/internal/resolver/serviceconfig.go
+++ b/xds/internal/resolver/serviceconfig.go
@@ -142,7 +142,7 @@ type erroringConfigSelector struct {
 }
 
 func newErroringConfigSelector(err error, xdsNodeID string) *erroringConfigSelector {
-	return &erroringConfigSelector{err: annotateErrorWithNodeID(status.Errorf(codes.Unavailable, err.Error()), xdsNodeID)}
+	return &erroringConfigSelector{err: annotateErrorWithNodeID(status.Errorf(codes.Unavailable, "%s", err.Error()), xdsNodeID)}
 }
 
 func (cs *erroringConfigSelector) SelectConfig(iresolver.RPCInfo) (*iresolver.RPCConfig, error) {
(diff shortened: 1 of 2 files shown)


xds/internal/resolver/serviceconfig.go
```diff
@@ -142,7 +142,7 @@ type erroringConfigSelector struct {
 }
 
 func newErroringConfigSelector(err error, xdsNodeID string) *erroringConfigSelector {
-	return &erroringConfigSelector{err: annotateErrorWithNodeID(status.Errorf(codes.Unavailable, err.Error()), xdsNodeID)}
+	return &erroringConfigSelector{err: annotateErrorWithNodeID(status.Errorf(codes.Unavailable, "%s", err.Error()), xdsNodeID)}
```
dfawley, line 145:
```suggestion
	return &erroringConfigSelector{err: annotateErrorWithNodeID(status.Error(codes.Unavailable, err.Error()), xdsNodeID)}
```

My suggestion is style vs. your fix, but the fix is a bug, not style.  If `err.Error()` outputs `%s` or something, then it would print weird error-looking stuff instead.

https://go.dev/play/p/BxccGnfgqas

```
Hello, %!s(MISSING)
```

xds/server_test.go
```diff
@@ -701,7 +701,7 @@ func (s) TestServeAndCloseDoNotRace(t *testing.T) {
 	// resource error notifications for the invalid listener resource leading
 	// to service mode change to "not serving" each time.
 	//
-	// Even if the the server is currently NOT_SERVING, in the case (where we
+	// Even if the server is currently NOT_SERVING, in the case (where we
 	// are NOT_SERVING and the new mode is also NOT_SERVING), the update is not
```
dfawley, line 705: This is hard to read anyway. Maybe while you're here:

```suggestion
	// Even if the server is currently NOT_SERVING and the new mode is also
	// NOT_SERVING, the update is not
```

dfawley (approved):

xds/internal/resolver/serviceconfig.go
purnesh42H: I see. thanks! changed.

xds/server_test.go
purnesh42H: Done

PR #8273, opened 2025-04-24
xds: Remove redundant proto checks

arjan-bal:
Fixing findings from internal analyzer: [go/gobugs-findings#protonil](http://go/gobugs-findings#protonil)
> The Go Proto API is nil safe: it is safe to call Has/Get-methods on nil messages. If a Has-method is called on a nil proto, it returns false. If a Get-method is called on a nil proto, it returns the field's default value. This nil safety makes it possible to chain Get-methods calls without intermediate nil checks.

RELEASE NOTES: N/A


purnesh42H (approved):
lgtm

PR #8278, opened 2025-04-29
grpc: Fix cardinality violations in  non-server streaming RPCs

Pranjali-2501:
Partially addresses: #7286

Client should ensure an "Internal" error is returned for client-streaming RPCs if the server doesn't send a message before returning status OK.
Currently it is returning "EOF", which should not be the case for cardinality violations.

RELEASE NOTES:
* grpc: return status code INTERNAL when servers send 0 response messages in unary and client streaming RPCs.


stream.go
```diff
@@ -542,6 +542,8 @@ type clientStream struct {
 
 	sentLast bool // sent an end stream
 
+	recvMsg bool // received msg from server
```
arjan-bal, line 545: nit: Boolean field names should be like questions with yes/no answers, example "didReceiveFirstMessage".

stream.go
```diff
@@ -1134,11 +1136,17 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
 			if statusErr := a.transportStream.Status().Err(); statusErr != nil {
 				return statusErr
 			}
+			if cs.desc.ClientStreams && !cs.desc.ServerStreams && !cs.recvMsg{
```
arjan-bal, line 1139: Can we leave out `cs.desc.ClientStreams` from this check? I would assume that the server needs to send a message for unary RPCs also. Are unary RPCs handled by this code path? If yes, we should add a test for unary RPCs also.

stream.go
```diff
@@ -1134,11 +1136,17 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
...
+				return status.Errorf(codes.Internal, "client streaming cardinality violation")
+			}
 			return io.EOF // indicates successful end of stream.
 		}
 
 		return toRPCErr(err)
 	}
+	if cs.desc.ClientStreams {
+		cs.recvMsg = true
```
arjan-bal, line 1148: Can we set this boolean without the check? Otherwise we would need to document the variable in a way to make it obvious that this is set only for client streaming RPCs. Even then users may accidentally use it for other types of RPCs.

stream.go
arjan-bal, line 1139: The vet checks is failing, probably because of the missing space before `{`. Can you please fix the vet issues?

arjan-bal:
Do we need a similar check on the server side also, i.e. change the status to internal if the server attempts to close without writing a message?

stream.go

stream.go
Pranjali-2501: Yes, Unary RPCs handled by this code path. I have removed `cs.desc.ClientStreams` from the check . Now it will check for both unary and clientStreaming RPCs.
Working on adding a test for the same.

stream.go
Pranjali-2501: Changed it to `!cs.desc.ServerStreams`, so that it will mark `cs.recvFirstMsg=true` for both unary and clientStreaming RPCs.

stream.go

stream.go
```diff
@@ -1134,11 +1136,17 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
 			if statusErr := a.transportStream.Status().Err(); statusErr != nil {
 				return statusErr
 			}
+			if !cs.desc.ServerStreams && !cs.recvFirstMsg {
+				return status.Errorf(codes.Internal, "client streaming cardinality violation")
```
dfawley, line 1140: Let's be more specific about what happened here so that a user could understand what's happening when they see the error.

`"received no response message from non-streaming RPC"`?

stream.go
```diff
@@ -1134,11 +1136,17 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
...
+			}
 			return io.EOF // indicates successful end of stream.
 		}
 
 		return toRPCErr(err)
 	}
+	if !cs.desc.ServerStreams {
+		cs.recvFirstMsg = true
```
dfawley, line 1148: Should there also be something here that fails if `recvFirstMsg` is already set?  (That would indicate we received multiple messages in a unary RPC.)

Pranjali-2501: For both unary and client-streaming RPCs, client.recvmsg() will wait for the trailers after receiving its first msg over [here](https://github.com/grpc/grpc-go/blob/54e7e26a1f323438c1dad002f2af4ed6ed4c5f11/stream.go#L1165C2-L1170C97) .

I'll change the status code [here](https://github.com/grpc/grpc-go/blob/54e7e26a1f323438c1dad002f2af4ed6ed4c5f11/stream.go#L1170) from `UNKNOWN` to `INTERNAL` in upcoming PR with tests. It will handle the case when trailers are not send from server side for unary and client-streaming RPCs.

stream.go
Pranjali-2501: Sure.

stream.go
dfawley: Yes, I think that should work and avoids adding new state to track.

test/end2end_test.go
```diff
@@ -3736,6 +3736,87 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	defer cc.Close()
+
+	go func() {
+		conn, err := lis.Accept()
+		if err != nil {
+			t.Errorf("lis.Accept() = %v", err)
+			return
+		}
+		defer conn.Close()
+		framer := http2.NewFramer(conn, conn)
```
dfawley, line 3763: Realistically this test shouldn't need low level framer access.  You should be able to implement it with a `grpc.Server` that has its `EmptyCall` handler configured as a streaming handling, and with an implementation that just returns `nil` immediately without sending a message.

I do think that would be preferable, because the framing stuff is tedious and error prone, and it makes the test much more complex.  If you're having too much trouble with that approach, let me know and I can try to help, or if it somehow turns out to be a lot more work than I was expecting, feel free to push back.

Pranjali-2501: Initially, I attempted to resolve this by returning nil from EmptyCall. However, this approach resulted in an empty, successful response rather than a failure, which wasn't the desired outcome.

The scenario that sends no response cannot be replicated within grpc-go itself, as the framework requires a response to be sent.

This test aims to verify behavior in a cross-language context, where a unary client might interact with a unary server implemented in a different language, potentially allowing the server to send no response.

Please let me know if that is not the case or I overlook something.

test/end2end_test.go
```diff
@@ -3736,6 +3736,87 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%s) = %v", lis.Addr().String(), err)
```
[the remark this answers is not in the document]
Pranjali-2501: Noted and made the changes.

test/end2end_test.go
```diff
@@ -3736,6 +3736,87 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	}
+	defer cc.Close()
+
+	go func() {
+		conn, err := lis.Accept()
+		if err != nil {
+			t.Errorf("lis.Accept() = %v", err)
```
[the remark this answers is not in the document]

test/end2end_test.go
```diff
@@ -3736,6 +3736,87 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+				if frame.Header().StreamID != 1 {
+					t.Errorf("Expected stream ID 1, got %d", frame.Header().StreamID)
+					return
+				}
+
+				var buf bytes.Buffer
+				enc := hpack.NewEncoder(&buf)
+				_ = enc.WriteField(hpack.HeaderField{Name: ":status", Value: "200"})
+				_ = enc.WriteField(hpack.HeaderField{Name: "content-type", Value: "application/grpc"})
+				_ = enc.WriteField(hpack.HeaderField{Name: "grpc-status", Value: "0"})
```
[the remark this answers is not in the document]
Pranjali-2501: Removed it.

test/end2end_test.go
arjan-bal: I think @dfawley is saying that we can register a streaming handler for a unary method on the server. The client will call `EmptyCall`, but the server will call a streaming handler.
```go
	srv := grpc.NewServer()
	serviceDesc := grpc.ServiceDesc{
		ServiceName: "grpc.testing.TestService",
		HandlerType: (*any)(nil),
		Methods:     []grpc.MethodDesc{},
		Streams: []grpc.StreamDesc{
			{
				StreamName: "EmptyCall",
				Handler: func(any, grpc.ServerStream) error {
					return nil
				},
				ClientStreams: true,
			},
		},
		Metadata: "grpc/testing/test.proto",
	}
	srv.RegisterService(&serviceDesc, &testServer{})
	go srv.Serve(lis)
	defer srv.Stop()

```

stream.go

test/end2end_test.go
Pranjali-2501: Thanks for clarifying. I had made the changes.

stream.go
```diff
@@ -1134,6 +1134,10 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
 			if statusErr := a.transportStream.Status().Err(); statusErr != nil {
 				return statusErr
 			}
+			// received no msg and status ok for non-server streaming rpcs.
+			if !cs.desc.ServerStreams {
```
arjan-bal, line 1138: There is a similar method on addrConnStream that also needs to be updated.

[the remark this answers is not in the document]
Pranjali-2501: For non-streaming RPCs:
> Server can't send more than one msg, as mentioned in this [comment](https://github.com/grpc/grpc-go/pull/8278#discussion_r2086372287).

> For successful RPC, Client can call RecvMsg() twice. Second RecvMsg() call will return with a io.EOF and close the stream.

> If will never be the case for non-streaming RPCs, that recvFirstMsg is already set.

test/end2end_test.go
dfawley: Another option (maybe easier, but no need to change now) is to use the unknown service handler, and don't register anything on the server:  https://pkg.go.dev/google.golang.org/grpc#UnknownServiceHandler

stream.go
dfawley: Ping - I think we can remove `recvFirstMsg` now?

Pranjali-2501: I had reverted that in this commit bbd83666804d26f92254d19ab4366c818e9edff2.
But than @arjan-bal  [pointed](https://github.com/grpc/grpc-go/pull/8278#discussion_r2107639397) out that in case of successful RPC, if user call RecvMsg() twice, the second RecvMsg() call should return with io.EOF instead of cardinality violation. 
That's why I reverted back to the older code to fulfill that condition.

dfawley: As I mentioned before, it's not defined what happens if `RecvMsg` is called after returning an error.

There are other scenarios where calling `recvMsg` repeatedly will not cause the previously returned error to occur.  E.g. if a received message is too large, `recv` will return an error.  But I believe a subsequent call to `RecvMsg` would actually return the next message from the stream, or `io.EOF` at the end of the stream.

test/end2end_test.go
dfawley: The idea is to make the server think it's a bidi stream, but the client thinks it's not.  The client's setting is what the code snippet you are showing is checking.  The client has no knowledge of how the server is treating the stream.

stream.go
arjan-bal: Acknowledged, if we're fine with changing the existing behaviour of returning `io.EOF` on subsequent calls after an RPC ends successfully, we can get rid of the extra variable to track state.

Pranjali-2501: I have removed the extra variable and the tests that were added to check the scenario when recvMsg() called twice.

stream.go
Pranjali-2501: Done.

arjan-bal (approved):

test/end2end_test.go
arjan-bal: @Pranjali-2501 did you try to use `UnknownServiceHandler`? I think it should reduce the amount of test code.

Pranjali-2501: Made the changes.

stream.go
```diff
@@ -1138,6 +1138,10 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
 			if statusErr := a.transportStream.Status().Err(); statusErr != nil {
 				return statusErr
 			}
+			// received no msg and status ok for non-server streaming rpcs.
```
dfawley: Please make sure comments use proper grammar (or sentence fragments if applicable without a period at the end).  So, e.g. capitalize the first letter.

(Throughout)

stream.go
```diff
@@ -1138,6 +1138,10 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
...
+			if !cs.desc.ServerStreams {
+				return status.Errorf(codes.Internal, "cardinality violation: received no response message from non-streaming RPC")
```
dfawley: Nit: use Error instead of Errorf since there is no formatting.

test/end2end_test.go
```diff
@@ -3740,6 +3737,109 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	}
+	defer cc.Close()
+
+	client := testgrpc.NewTestServiceClient(cc)
+	if _, err = client.EmptyCall(ctx, &testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests that client will receive cardinality violation in subsequent calls to RecvMsg().
```
dfawley:
```suggestion
// Tests that client will receive cardinality violations when calling RecvMsg() multiple times
// for non-streaming response streams.
```

test/end2end_test.go
```diff
@@ -3740,6 +3737,109 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	if err := stream.RecvMsg(resp); err != nil {
+		t.Fatalf("stream.RecvMsg() = %v , want <nil>", err)
+	}
+
+	if err = stream.RecvMsg(resp); status.Code(err) != codes.Internal {
+		t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests that client will receive cardinality violation in the subsequent calls to RecvMsg().
```
dfawley: Similar here

test/end2end_test.go
```diff
@@ -3740,6 +3737,109 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+func (s) TestClientStreaming_ClientCallRecvMsgTwice(t *testing.T) {
```
dfawley, line 3810: In what way is this different from the previous test case?  The server side is different, but actually does the same thing.

test/end2end_test.go
```diff
@@ -3740,6 +3737,109 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	}
 }
 
+// Tests that a client receives a cardinality violation error for unary
+// RPCs if the server doesn't send a message before returning status OK.
+func TestUnaryRPC_ServerSendsOnlyTrailersWithOK(t *testing.T) {
```
arjan-bal: Please make this a method on the `grpctest.Tester`. The `Tester` tracks leaked goroutines, buffers and fails tests when they log an error.

test/end2end_test.go
Pranjali-2501: Both tests does the same thing, only difference is `StreamDesc.ClientStreams` is set to `false` in this test(Unary RPC) and set to `true` in previous test(Client Streaming RPC)

test/end2end_test.go

test/end2end_test.go

test/end2end_test.go

stream.go
Pranjali-2501: Replaced `Errorf` with `Error`.

stream.go
Pranjali-2501: Sure.

test/end2end_test.go
arjan-bal: This PR is actually a fix for for "non-server streaming" RPCs instead of "client streaming RPCs". I've updated the PR title to reflect the same. So the value of `StreamDesc.ClientStreams` should not matter, but it's okay to have test coverage for this.

I noticed that none of the tests send 2 response messages from the server. This shows up as missing coverage reported by Codecov around line 1486. Can you please have the server send back two response messages here? You should use the `UnknownServiceHandler` for this to ensure the server doesn't skip sending the message after noticing it's a non-server streaming RPC.

arjan-bal: Discussed offline: line 1486 is in addrConnStream, not csAttempt. I had initially thought the missing coverage was in csAttempt. addrConnStream is largely similar to csAttempt but is only used for establishing health streams. There is a plan to unify it with csAttempt in the future. Adding test coverage for addrConnStream may be difficult and is not necessary at this point.

PR #8279, opened 2025-04-30
grpc: add new Dial/ServerOptions to set the size like the existing ones, but which do not also disable BDP estimation

janardhanvissa:


arjan-bal:
Closing in favour of https://github.com/grpc/grpc-go/pull/8283.

PR #8280, opened 2025-04-30
feat(cmd/gen): add flag use_separate_service_package

qianlongzt:
https://github.com/grpc/grpc-go/issues/8233


scripts/regenerate.sh
```diff
@@ -99,6 +99,14 @@ Mgrpc/testing/test.proto=google.golang.org/grpc/interop/grpc_testing,\
 Mgrpc/testing/payloads.proto=google.golang.org/grpc/interop/grpc_testing,\
 Mgrpc/testing/empty.proto=google.golang.org/grpc/interop/grpc_testing
 
+# test feature use separate service package and only test one proto file
+protoc --go-grpc_out="${OPTS},use_separate_service_package=true:${WORKDIR}/out" \
+    "examples/route_guide/routeguide/route_guide.proto"
+# move service package to right place.
+mkdir -p "${WORKDIR}/out/google.golang.org/grpc/examples/route_guide/routeguide_grpc"
+mv "${WORKDIR}"/out/google.golang.org/grpc/examples/route_guide/routeguide/*_grpc.pb.go \
+  "${WORKDIR}/out/google.golang.org/grpc/examples/route_guide/routeguide_grpc"
```
arjan-bal, line 108: By convention, the directory name and the package names should be the same, see https://go.dev/blog/package-names#package-paths. Here, the package name can't be changed, so we should place the package under a directory named `routeguide`, e.g: `${WORKDIR}/out/google.golang.org/grpc/examples/route_guide/grpc/routeguide`

scripts/regenerate.sh
```diff
@@ -99,6 +99,14 @@ Mgrpc/testing/test.proto=google.golang.org/grpc/interop/grpc_testing,\
 Mgrpc/testing/payloads.proto=google.golang.org/grpc/interop/grpc_testing,\
 Mgrpc/testing/empty.proto=google.golang.org/grpc/interop/grpc_testing
 
+# test feature use separate service package and only test one proto file
```
dfawley, line 102: This is our script to regenerate the proto packages we maintain; it is not intended for testing.  Can we find another way to test this?

cmd/protoc-gen-go-grpc/main.go
```diff
@@ -58,6 +61,7 @@ func main() {
 	var flags flag.FlagSet
 	requireUnimplemented = flags.Bool("require_unimplemented_servers", true, "set to false to match legacy behavior")
 	useGenericStreams = flags.Bool("use_generic_streams_experimental", true, "set to true to use generic types for streaming client and server objects; this flag is EXPERIMENTAL and may be changed or removed in a future release")
+	useSeparateServicePackage = flags.Bool("use_separate_service_package", false, "set to true to use separate service package, service package will import protobuf package")
```
dfawley, line 64: How about something like: name this `output_package` and make it a string?  Then the import path could be `file.GoImportPath` with everything after the last `/` stripped and replaced by what's specified here?  That would allow more flexibility for how the code is generated.

qianlongzt: The import path is useless after file generate. 

https://github.com/grpc/grpc-go/pull/8245.
https://github.com/grpc/grpc-go/issues/8233#issuecomment-2816990330

scripts/regenerate.sh
qianlongzt: I have no idea where the goldenfile should be placed.

1. How about putting it in `cmd/protoc-gen-go-grpc/testdata/routeguide_grpc/route_guide_grpc.pb.go`, adding a generation script in `cmd/protoc-gen-go-grpc/regenerate.sh`, and `creating cmd/protoc-gen-go-grpc/doc.go` with a `go:generate` directive?

2. Alternatively, remove the goldenfile and use:
```
grep 'routeguide "google.golang.org/grpc/examples/route_guide/routeguide"' \
${TEMPDIR}/examples/route_guide/routeguide/route_guide_grpc.pb.go
```

qianlongzt: I removed the goldenfile and used grep to check.

scripts/regenerate.sh
qianlongzt: I removed the goldenfile and used grep to check.

cmd/protoc-gen-go-grpc/main.go
arjan-bal: I think the most user friendly solution to split package would be to have a proto option like  `go_grpc_package` which has semantics similar to `go_package`. The probuf complier would then need to handle splitting the service and message code when `go_package` and `go_grpc_package` are different.

dfawley: Is that something we can do?  I thought the options were defined by protobuf.  We could propose a `go_services_package` if so.  The Go protobuf team seems to prefer this ideology anyway (generate services and messages into separate packages).

dfawley: This would be my preference, then.

We should think about whether renaming the package completely would be better than just adding a suffix.  I'm not sure which one would be better, but either would be better than the previous proposal.

PR #8282, opened 2025-05-01
transport: writeStatus: Skip Status.Proto() without details

evanj:
Add Status.HasDetails() to check for details. Change http2server.writeStatus to check to avoid unnecessary clones in the common case when there are no details. This seems to use about 0.3% of CPU for a high QPS gRPC service, according to the Go profiler.

While looking at profiles for a high QPS gRPC service at Datadog, I noticed that about 0.3% of CPU time is being spent calling Status.Proto(). I believe this is unnecessary: this clones the protobuf status message, only to check if there are details. If there are not (the common case, e.g. for status code OK), it throws the protobuf message away.

I tried quickly but was not able to reliably measure the impact of this change from any end-to-end benchmarks, but it does seem to make this function go away according to the profiler. This should be a small win for every gRPC server.

Screenshot 2025-05-01 at 9 49 01 AM

RELEASE NOTES:
* server: improve performance of server RPC handling by avoid a status proto copy

### DIFF
--- a/internal/status/status_test.go
+++ b/internal/status/status_test.go
@@ -16,22 +16,25 @@
  *
  */
 
-package status
+package status_test
 
 import (
 	"testing"
 
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/internal/status"
 )
 
 func TestRawStatusProto(t *testing.T) {
-	spb := RawStatusProto(nil)
+	spb := status.RawStatusProto(nil)
 	if spb != nil {
-		t.Errorf("RawStatusProto(nil) must return nil (was %v)", spb)
+		t.Errorf("RawStatusProto(nil) = %v; must return nil", spb)
 	}
-	s := New(codes.Internal, "test internal error")
-	spb = RawStatusProto(s)
-	if spb != s.s {
-		t.Errorf("RawStatusProto(s) must return s.s=%p (was %p)", s.s, spb)
+	s := status.New(codes.Internal, "test internal error")
+	spb1 := status.RawStatusProto(s)
+	spb2 := status.RawStatusProto(s)
+	// spb1 and spb2 should be the same pointer: no copies
+	if spb1 != spb2 {
+		t.Errorf("RawStatusProto(s)=%p then %p; must return the same pointer", spb1, spb2)
 	}
 }
--- a/status/status_test.go
+++ b/status/status_test.go
@@ -355,7 +355,7 @@ func (s) TestStatus_ErrorDetails(t *testing.T) {
 		}
 		details := s.Details()
 		if len(details) != len(tc.details) {
-			t.Fatalf("len(s.Details()) = %v, want = %v.", len(details), len(tc.details))
+			t.Fatalf("len(s.Details()) = %d, want = %d.", len(details), len(tc.details))
 		}
 		for i := range details {
 			if !proto.Equal(details[i].(protoreflect.ProtoMessage), tc.details[i].(protoreflect.ProtoMessage)) {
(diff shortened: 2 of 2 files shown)


dfawley:
This changes our external API since `grpc/status/Status` is a type alias of `grpc/internal/Status/Status`.  It may be preferable to implement a `grpc/internal/status/RawStatusProto(s *Status) *sbp.Proto` that just returns the internals, and then operate on them directly instead.  The reason we split status in two places is because we want to be able to do things like this that would be unsafe for users to directly access, but improves efficiency of our internal code.

evanj:
I have updated the PR to add the suggested `RawStatusProto` function instead with a test (although the test is pretty silly and could be removed).

I'm happy to make any other suggested improvements. Thanks for the suggestion and the code review!

One unimportant ignorable thought: Another option would be to make this internal function more restricted: `HasDetails(...) bool` instead. This would avoid revealing the entire internals of the `Status` type. This seems fairly unimportant. I went with RawStatusProto because I have WAY less context on this project.

dfawley (approved):
Thanks!

dfawley:
@arjan-bal can you take a second pass please?

evanj:
I have added a copyright to the new test file by copy/pasting and editing the header from status.go.

internal/status/status_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+func TestRawStatusProto(t *testing.T) {
+	spb := RawStatusProto(nil)
+	if spb != nil {
+		t.Errorf("RawStatusProto(nil) must return nil (was %v)", spb)
+	}
+	s := New(codes.Internal, "test internal error")
+	spb = RawStatusProto(s)
+	if spb != s.s {
+		t.Errorf("RawStatusProto(s) must return s.s=%p (was %p)", s.s, spb)
```
arjan-bal, line 35: To be consistent with the Google style guide, we should print the actual returned value before the expected value.

https://google.github.io/styleguide/go/decisions#got-before-want

status/status_test.go
```diff
@@ -354,6 +354,9 @@ func (s) TestStatus_ErrorDetails(t *testing.T) {
 			t.Fatalf("(%v).WithDetails(%+v) failed: %v", str(s), tc.details, err)
 		}
 		details := s.Details()
+		if len(details) != len(tc.details) {
+			t.Fatalf("len(s.Details()) = %v, want = %v.", len(details), len(tc.details))
```
arjan-bal, line 358: nit: We should use %d to format integers as this helps linters catch data type mismatches.

internal/status/status_test.go
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package status
```
arjan-bal, line 19: nit: Can we name this package `status_test`? Making the test package different from the package being tested encourages testing the exported APIs, making tests less brittle.

arjan-bal (commented):
Minor comments, otherwise LGTM.

internal/status/status_test.go
evanj: Done, thanks.

evanj: Okay this did keep me honest and I think the test is actually better: I now call it twice and ensure that it returned the same pointer value, which is a reasonable check that this is not returning a copy.

status/status_test.go
evanj: I copied this missing check from the other test, so I updated it in both places.

internal/status/status_test.go
evanj: Done, although this message and check changed to avoid testing the internals of the status message.

Issue #8281, opened 2025-04-30
streaming: should return cancel error instead of Internal error with unexpected EOF


fuweid:
### What version of gRPC are you using?

v1.72.0

### What version of Go are you using (`go version`)?

v1.24.2

### What operating system (Linux, Windows, …) and version?

Linux

### What did you do?

In the ETCD integration test, there is a case that verifies the client receives a context canceled error when reading data from a streaming RPC and the context is subsequently canceled. The steps would be like

* Create a context with a one-second timeout (ctx).
* Use that ctx to initiate the streaming RPC, and start a new goroutine to read the stream and write received data into a pipe.
* The client sleeps for two seconds and then attempts to read from the pipe.
    * NOTE: The background goroutine may read one message before timeout. https://github.com/etcd-io/etcd/blob/0a0fde11ace45eaf596862df35aae833b8e6d649/client/v3/maintenance.go#L284-L293
* The read should return a context canceled error.

```bash
# in etcd main branch
$ cd ./tests/integration/clientv3
$ go test -v ./ -count=10 --failfast -run TestMaintenanceSnapshotTimeout
```

### What did you expect to see?

I should receice context canceled error

### What did you see instead?

received `rpc error: code = Internal desc = unexpected EOF`

I can't reproduce this issue with v1.71.0 version. Right now, I found that https://github.com/grpc/grpc-go/blob/a43eba6fed49b81b84cfdba85c356aca22086d7e/rpc_util.go#L676 - the p.r.Read returns io.EOF and then convert it into unexpectedEOF. 

REF: https://github.com/etcd-io/etcd/pull/19833#issuecomment-2840562143

cc @ahrtr @dfawley

ahrtr:
To add more context to clarify this issue from users perspective.

When etcd calls the `Recv` method (generated gRPC code, see below, also see [here](https://github.com/etcd-io/etcd/blob/8da854dc69729868531d8f3e360b77c4c046eae1/api/etcdserverpb/rpc.pb.go#L7573-L7579)), when the context (the context used when creating the gRPC stream, see [here](https://github.com/etcd-io/etcd/blob/d610b0431bdde9ff12b97ee70993137c78681592/api/etcdserverpb/rpc.pb.go#L7549) ) timeout, 
- previously (on grpc-go 1.71), the `Recv` always returned a `rpc error: code = DeadlineExceeded desc = context deadline exceeded` error. 
[...]

PR #8289, opened 2025-05-06
transport: Propagate status code on receiving RST_STREAM during message read

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/8281

RCA: https://github.com/grpc/grpc-go/issues/8281#issuecomment-2848692246

This PR  uses a status error instead of an `io.UnexpectedEOF` to fail RPCs that have partially read a gRPC message.

RELEASE NOTES:
* grpc: Fix bug that causes RPCs to fail with status `INTERNAL` instead of `CANCELLED` or `DEADLINE_EXCEEDED` when the client receives an RST_STREAM frame while reading a gRPC message.

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -1242,7 +1242,8 @@ func (t *http2Client) handleRSTStream(f *http2.RSTStreamFrame) {
 			statusCode = codes.DeadlineExceeded
 		}
 	}
-	t.closeStream(s, io.EOF, false, http2.ErrCodeNo, status.Newf(statusCode, "stream terminated by RST_STREAM with error code: %v", f.ErrCode), nil, false)
+	st := status.Newf(statusCode, "stream terminated by RST_STREAM with error code: %v", f.ErrCode)
+	t.closeStream(s, st.Err(), false, http2.ErrCodeNo, st, nil, false)
 }
 
 func (t *http2Client) handleSettings(f *http2.SettingsFrame, isFirst bool) {
--- a/test/transport_test.go
+++ b/test/transport_test.go
@@ -19,16 +19,20 @@ package test
 
 import (
 	"context"
+	"encoding/binary"
 	"io"
 	"net"
 	"sync"
 	"testing"
 
+	"golang.org/x/net/http2"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/internal/grpcsync"
 	"google.golang.org/grpc/internal/stubserver"
+	"google.golang.org/grpc/internal/testutils"
 	"google.golang.org/grpc/internal/transport"
 	"google.golang.org/grpc/status"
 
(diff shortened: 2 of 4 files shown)


internal/transport/http2_client.go
```diff
@@ -1242,7 +1242,8 @@ func (t *http2Client) handleRSTStream(f *http2.RSTStreamFrame) {
 			statusCode = codes.DeadlineExceeded
 		}
 	}
-	t.closeStream(s, io.EOF, false, http2.ErrCodeNo, status.Newf(statusCode, "stream terminated by RST_STREAM with error code: %v", f.ErrCode), nil, false)
+	st := status.Newf(statusCode, "stream terminated by RST_STREAM with error code: %v", f.ErrCode)
+	t.closeStream(s, st.Err(), false, http2.ErrCodeNo, st, nil, false)
```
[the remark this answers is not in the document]
dfawley: Where is that conversion happening?

test/transport_test.go
```diff
@@ -153,3 +157,95 @@ func (s) TestClientTransportRestartsAfterStreamIDExhausted(t *testing.T) {
...
+		for {
+			frame, err := framer.ReadFrame()
+			if err != nil {
+				return
+			}
+			switch frame := frame.(type) {
+			case *http2.HeadersFrame:
+				// When the client creates a stream, write a partial gRPC
+				// message followed by an RST_STREAM.
+				go func() {
```
dfawley, line 212: Why is this in a goroutine, but the ping ack is written inline?

Also...can we just not reply to the ping and then remove the lock?

test/transport_test.go
```diff
@@ -153,3 +157,95 @@ func (s) TestClientTransportRestartsAfterStreamIDExhausted(t *testing.T) {
...
+	client := testgrpc.NewTestServiceClient(cc)
+	_, err = client.EmptyCall(ctx, &testpb.Empty{})
+
+	s, ok := status.FromError(err)
+	if !ok {
+		t.Fatalf("client.EmptyCall() returned non-status error: %v", err)
+	}
+	if s.Code() != codes.Canceled {
+		t.Fatalf("client.EmptyCall() returned status %v with code %v, want %v", s, s.Code(), codes.Canceled)
+	}
```
dfawley, line 250: This is equivalent and simpler:

```suggestion
	if _, err := client.EmptyCall(ctx, &testpb.Empty{}); status.Code(err) != codes.Canceled {
		t.Fatalf("client.EmptyCall() returned %v; want status with code %v", err, codes.Canceled)
	}
```

internal/transport/http2_client.go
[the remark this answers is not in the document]
arjan-bal: Here:
https://github.com/grpc/grpc-go/blob/a43eba6fed49b81b84cfdba85c356aca22086d7e/rpc_util.go#L674-L680

test/transport_test.go
arjan-bal: Removed the ping handling and the mutex. Removed the separate goroutine used for writing data frames. 
The removed things were unnecessary. I copied the loop that reads frames from an existing test and didn't remove the extra stuff.
https://github.com/grpc/grpc-go/blob/950a7cfdfd571b75f6b869ae0f625f9d65b19842/internal/transport/transport_test.go#L1414-L1450

test/transport_test.go
```diff
@@ -209,27 +208,14 @@ func (s) TestRSTDuringMessageRead(t *testing.T) {
...
-						mu.Unlock()
-						return
-					}
-					framer.WriteRSTStream(1, http2.ErrCodeCancel)
-					mu.Unlock()
-				}()
-			case *http2.RSTStreamFrame:
-				if frame.Header().StreamID != 1 || http2.ErrCode(frame.ErrCode) != http2.ErrCodeFlowControl {
-					t.Errorf("RST stream received with streamID: %d and code: %v, want streamID: 1 and code: http2.ErrCodeFlowControl", frame.Header().StreamID, http2.ErrCode(frame.ErrCode))
+				messageLen := 2048
```
dfawley, line 211: `const`?

dfawley (approved):

PR #8290, opened 2025-05-06
transport: Reject non-positive timeout values in server

evanj:
Negative timeouts should be forbidden. The previous implementation called ParseInt and allowed negative timeouts. Change it to call ParseUint to reject these values. This is also slightly faster, since ParseInt eventually calles ParseUint. Extend the test to cover more cases.

RELEASE NOTES:
* transport: Non-positive `grpc-timeout` header values are now rejected by servers. This is consistent with the [gRPC protocol spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests).

### DIFF
--- a/internal/transport/http_util_test.go
+++ b/internal/transport/http_util_test.go
@@ -22,13 +22,14 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"reflect"
 	"testing"
 	"time"
 )
 
-func (s) TestTimeoutDecode(t *testing.T) {
+func (s) TestDecodeTimeout(t *testing.T) {
 	for _, test := range []struct {
 		// input
 		s string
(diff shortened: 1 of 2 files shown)


internal/transport/http_util_test.go
```diff
@@ -22,28 +22,46 @@ import (
...
+		{"00000010m", time.Millisecond * 10, nil},
 		{"1234S", time.Second * 1234, nil},
+		{"00000001M", time.Minute, nil},
+		{"09999999S", time.Second * 9999999, nil},
+		{"99999999S", time.Second * 99999999, nil},
+		{"99999999M", time.Minute * 99999999, nil},
+		{"2562047H", time.Hour * 2562047, nil},
+		{"2562048H", time.Duration(math.MaxInt64), nil},
+		{"99999999H", time.Duration(math.MaxInt64), nil},
+		{"-1S", 0, fmt.Errorf("strconv.ParseUint: parsing %q: invalid syntax", "-1")},
```
arjan-bal, line 53: This error string is not part of gRPC and can cause the test to break when run with a different version of golang. You can instead use cmp.Equal() with cmpopts.EquateErrors(). You can use `cmpopts.AnyError` as the expected value.

https://google.github.io/styleguide/go/decisions#test-error-semantics

internal/transport/http_util_test.go
```diff
@@ -22,28 +22,46 @@ import (
...
-func (s) TestTimeoutDecode(t *testing.T) {
+func (s) TestDecodeTimeout(t *testing.T) {
 	for _, test := range []struct {
 		// input
 		s string
 		// output
 		d   time.Duration
 		err error
 	}{
+		{"0S", 0, nil},
```
arjan-bal, line 40: The [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests) mentions that `TimeoutValue` should be a positive integer. So this should ideally return an error on validation.

arjan-bal (commented):
Thanks for adding these test cases!

internal/transport/http_util_test.go
evanj: Good suggestion! However, since this function is returning `fmt.Errorf`, none of the expected errors can be compared, since  I changed the error check to just be for "has an error". We could change the function to return some other typed error instead, like maybe a status error? I think the test checking for existence of an error is probably okay.

internal/transport/http_util_test.go
evanj: Done.

evanj:
Thanks for the suggestion! This caused me to add a check for `t != 0` to the function, and to revise the test to avoid comparing error strings, since `fmt.Errorf` does not compare with `errors.Is`, so this matches the style guide: "If a test uses cmpopts.EquateErrors but all of its wantErr values are either nil or cmpopts.AnyError, then [...] Simplify the code by making the want field a bool."

arjan-bal (approved):

dfawley (approved):
Looks fine and correct, but can you tell us why you wanted to make this change?  Is this something you're seeing happen in practice?

evanj:
I've been investigating grpc overhead in services at Datadog that do a lot of requests/second, which means they spend a lot of time in grpc-go. I noticed this function and wondered why `decodeTimeout` shows up (47ms out of 87s CPU time or 0.05% in other words: irrelevant), but then I noticed it allows negative values, and switching to ParseUint theoretically saves some of this time so it is theoretically correct and theoretically faster. 

So ... Maybe not worth our time discussing and reviewing the change, but I thought it wasn't a total waste.

ejona86:
In b/427338711 at Google we're seeing a regression caused by this and grpc-java sending timeouts of `0n`. It is pretty trivial to change Java, as we only need to change a `0` to a `1`, but it will take a while for people to pick up the new version:
https://github.com/grpc/grpc-java/blob/919370172d65a819bb1e7ccb47f2285b0ccfd17e/core/src/main/java/io/grpc/internal/AbstractClientStream.java#L127

This makes me believe that _no_ gRPC implementation was actually requiring the timeout to be positive, only non-negative. Makes me wonder if we should change the spec.

evanj:
Wow oops thanks for the update and sorry for breaking stuff :(. A request with a 0 timeout would previously fail with deadline exceeded, and now I think this error causes the entire http2 connection to be closed IIRC, is that right? (I want to take a quick look to see if we might be running into this error at Datadog ...)

dfawley:
@evanj 

> Wow oops thanks for the update and sorry for breaking stuff :(. A request with a 0 timeout would previously fail with deadline exceeded, and now I think this error causes the entire http2 connection to be closed IIRC, is that right? (I want to take a quick look to see if we might be running into this error at Datadog ...)

This should not cause connection closure.  Just an RPC failing with a different status.  I sent #8439 to allow a 0s timeout again.

Issue #8288, opened 2025-05-05
[question] godoc of `CloseSend() error`


miparnisari:
https://github.com/grpc/grpc-go/blob/7fb5738f9989442050d9c50dd36a150787c9a1b3/stream.go#L104-L107

It says "It closes the stream when non-nil error is met". Specifically, when  _what_ error is non-nil? This function doesn't take a parameter so this is a bit confusing.

It'd also probably be good to document when or why this function returns an error.

arjan-bal:
Hi @miparnisari, the error return type and the godoc are a historical artifact. After talking to @dfawley I found that  `CloseSend` used to return the RPC status. See the following snippet from the v1.9.x branch.
https://github.com/grpc/grpc-go/blob/95869d0274dc5f61cad15f9ef42e060b9c9e0a3a/stream.go#L495-L510

Since v1.10, the method stopped returning errors.
https://github.com/grpc/grpc-go/blob/820c18582a3bce3ffc3931e4e4df5a381397bfe3/stream.go#L474-L485

This change was made to ensure users read all the pending message from the stream before they can get the RPC status.

I've raise https://github.com/grpc/grpc-go/pull/8292 to update the documentation.

PR #8292, opened 2025-05-06
grpc: Update ClientStream.CloseSend docs

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/8288

RELEASE NOTES: N/A

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -1380,7 +1380,7 @@ func (as *addrConnStream) Trailer() metadata.MD {
 
 func (as *addrConnStream) CloseSend() error {
 	if as.sentLast {
-		// TODO: return an error and finish the stream instead, due to API misuse?
+		// Return a nil error on repeated calls to this method.
 		return nil
 	}
 	as.sentLast = true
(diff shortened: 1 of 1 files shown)


stream.go
```diff
@@ -1013,7 +1013,8 @@ func (cs *clientStream) CloseSend() error {
 			binlog.Log(cs.ctx, chc)
 		}
 	}
-	// We never returned an error here for reasons.
+	// We don't return an error here as we expect users to read all messages
+	// from the stream and get the RPC status from RecvMsg().
```
dfawley, line 1017: How about being a little more explicit why it's OK to not return an error here, but we must when doing `SendMsg`:

```suggestion
	// We don't return an error here as we expect users to read all messages
	// from the stream and get the RPC status from RecvMsg().  Note that
	// SendMsg() must return an error when one occurs so the application
	// knows to stop sending messages, but that does not apply here.
```

arjan-bal: Accepted the suggestion.

dfawley (approved):

PR #8293, opened 2025-05-06
[alts] add keepalive params to the alts handshaker client dial option

rockspore:
The most important number here is the 5 seconds timeout. It means the client will close the connection if there is no activity for 5 seconds. It helps with the situation when the connection is somehow broken.

RELEASE NOTES: None

### DIFF
--- a/credentials/alts/internal/handshaker/service/service.go
+++ b/credentials/alts/internal/handshaker/service/service.go
@@ -56,8 +56,8 @@ func Dial(hsAddress string) (*grpc.ClientConn, error) {
 			grpc.WithTransportCredentials(insecure.NewCredentials()),
 			grpc.WithDisableServiceConfig(),
 			grpc.WithKeepaliveParams(keepalive.ClientParameters{
-				Timeout: 5 * time.Second,
-				Time:    time.Hour,
+				Timeout: 10 * time.Second,
+				Time:    10 * time.Minute,
 			}),
 		}
 		hsConn, err = grpc.NewClient(hsAddress, opts...)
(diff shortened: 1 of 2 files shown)


matthewstevenson88 (approved):

credentials/alts/internal/handshaker/service/service.go
```diff
@@ -50,7 +52,16 @@ func Dial(hsAddress string) (*grpc.ClientConn, error) {
 		// Disable the service config to avoid unnecessary TXT record lookups that
 		// cause timeouts with some versions of systemd-resolved.
 		var err error
-		hsConn, err = grpc.NewClient(hsAddress, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithDisableServiceConfig())
+		opts := []grpc.DialOption{
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
+			grpc.WithDisableServiceConfig(),
+			grpc.WithKeepaliveParams(keepalive.ClientParameters{
+				Timeout:             5 * time.Second,
```
apolcyn, line 59: I can't see any other defaults (internally) setting the timeout lower than 10 seconds, so might set this to 10 seconds just to match other defaults.

I'm debating about `Time` to 1 hour. 10 minutes would also be conservative but small enough to be frequently exercised in tests...

cc @ejona86 FYI

rockspore: Ack. I changed the timeout to 10 seconds, and time to 10 minutes. Will check how this changes the behavior we are interested in.

rockspore:
@matthewstevenson88 Please feel free to edit as needed and merge it while I'm gone. Thanks!

apolcyn (approved):

PR #8308, opened 2025-05-09
grpcsync/event: Simplify synchronization

arjan-bal:
This change avoids using `sync.Once` and reuses the existing atomic boolean to ensure an event is fired only once.

RELEASE NOTES: N/A

### DIFF
--- a/internal/grpcsync/event.go
+++ b/internal/grpcsync/event.go
@@ -32,7 +32,8 @@ type Event struct {
 
 // Fire causes e to complete.  It is safe to call multiple times, and
 // concurrently.  It returns true iff this call to Fire caused the signaling
-// channel returned by Done to close.
+// channel returned by Done to close. If Fire returns false, it is possible
+// the Done channel has not been closed yet.
 func (e *Event) Fire() bool {
 	if e.fired.CompareAndSwap(false, true) {
 		close(e.c)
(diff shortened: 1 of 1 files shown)


dfawley (approved):
There is a subtle behavior change in that `Fire` can now return before the channel is closed.  E.g.:

```go
go e.Fire()  // Pretend this wins, but blocks for a second before close(e.c) happens
e.Fire() // This loses and returns false
select {
	case <-e.Done(): // This channel is still open
	default: panic("used to be impossible")
}
```

But, I don't think this is a behavior that is (or should be) relied upon, so it should be fine.

internal/grpcsync/event.go
```diff
@@ -21,28 +21,24 @@
...
 type Event struct {
-	fired int32
+	fired atomic.Bool
 	c     chan struct{}
-	o     sync.Once
 }
 
 // Fire causes e to complete.  It is safe to call multiple times, and
 // concurrently.  It returns true iff this call to Fire caused the signaling
 // channel returned by Done to close.
```
dfawley, line 35: Now that I think about it, we should maybe document the new behavior as a detail:

```suggestion
// Fire causes e to complete.  It is safe to call multiple times, and
// concurrently.  It returns true iff this call to Fire caused the signaling
// channel returned by Done to close. If Fire returns false, it is possible
// the Done channel has not been closed yet.
```

PR #8309, opened 2025-05-09
leakcheck: Fix flaky test TestCheck

### LINKED ISSUE #8276: Flaky test: leakcheck.TestCheck
It seems there may have been a change in the stdlib that makes `WithTimeout` spawn a goroutine which may still be alive when we run `CheckGoroutines` in this test.

```
--- FAIL: TestCheck (2.05s)
    leakcheck_test.go:58: CheckGoroutines found 4 leaks, want 3 leaks
    leakcheck_test.go:59: leaked goroutines:
<snip out expected goroutines>
        Leaked goroutine: goroutine 8 [runnable]:
        context.WithDeadlineCause.func2()
        	/opt/hostedtoolcache/go/1.24.2/x64/src/context/context.go:643
        created by time.goFunc
        	/opt/hostedtoolcache/go/1.24.2/x64/src/time/sleep.go:215 +0x38
FAIL
FAIL	google.golang.org/grpc/internal/leakcheck	12.308s
```

We can probably just add it to https://github.com/grpc/grpc-go/blob/4cedec40eb2ccfbe3f56bb15e894903111ada2d2/internal/leakcheck/leakcheck.go#L172

We should also tidy up these error messages here while we're here:

https://github.com/grpc/grpc-go/blob/4cedec40eb2ccfbe3f56bb15e894903111ada2d2/internal/leakcheck/leakcheck_test.go#L51
https://github.com/grpc/grpc-go/blob/4cedec40eb2ccfbe3f56bb15e894903111ada2d2/internal/leakcheck/leakcheck_test.go#L78

EDIT: link to failure: https://github.com/grpc/grpc-go/actions/runs/14715401045/job/41297489567?pr=8155#step:8:116

Pranjali-2501:
Fixes #8276 

leakcheck.TestCheck and leakcheck.TestCheckRegisterIgnore is flaky due to a gouroutine leak caused by [context.WithTimeout()](https://github.com/grpc/grpc-go/blob/950a7cfdfd571b75f6b869ae0f625f9d65b19842/internal/leakcheck/leakcheck_test.go#L54C2-L55C16).
Flakiness was filed in  #8276.

RELEASE NOTES: N/A

### DIFF
--- a/internal/leakcheck/leakcheck_test.go
+++ b/internal/leakcheck/leakcheck_test.go
@@ -49,13 +49,14 @@ func TestCheck(t *testing.T) {
 		go func() { <-ch }()
 	}
 	if leaked := interestingGoroutines(); len(leaked) != leakCount {
-		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
+		t.Errorf("interestingGoroutines() = %v, want length %v", len(leaked), leakCount)
 	}
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	if CheckGoroutines(ctx, e); e.errorCount < 3 {
-		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
+	if CheckGoroutines(ctx, e); e.errorCount < leakCount {
+		t.Errorf("CheckGoroutines() = %v, want count %v", e.errorCount, leakCount)
+		t.Logf("leaked goroutines:\n%v", strings.Join(e.errors, "\n"))
 	}
 	close(ch)
 	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
@@ -75,14 +76,15 @@ func TestCheckRegisterIgnore(t *testing.T) {
 		go func() { <-ch }()
 	}
 	if leaked := interestingGoroutines(); len(leaked) != leakCount {
-		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
+		t.Errorf("interestingGoroutines() = %v, want length %v", len(leaked), leakCount)
 	}
 	go func() { ignoredTestingLeak(3 * time.Second) }()
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	if CheckGoroutines(ctx, e); e.errorCount < 3 {
-		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
+	if CheckGoroutines(ctx, e); e.errorCount < leakCount {
+		t.Errorf("CheckGoroutines() = %v, want count %v", e.errorCount, leakCount)
+		t.Logf("leaked goroutines:\n%v", strings.Join(e.errors, "\n"))
 	}
 	close(ch)
 	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
--- a/internal/leakcheck/leakcheck_test.go
+++ b/internal/leakcheck/leakcheck_test.go
@@ -44,8 +44,9 @@ func (e *testLogger) Errorf(format string, args ...any) {
 
 func TestCheck(t *testing.T) {
 	const leakCount = 3
+	ch := make(chan struct{})
 	for i := 0; i < leakCount; i++ {
-		go func() { time.Sleep(2 * time.Second) }()
+		go func() { <-ch }()
 	}
 	if leaked := interestingGoroutines(); len(leaked) != leakCount {
 		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
@@ -53,9 +54,10 @@ func TestCheck(t *testing.T) {
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	if CheckGoroutines(ctx, e); e.errorCount == 0 {
+	if CheckGoroutines(ctx, e); e.errorCount < 3 {
 		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
 	}
+	close(ch)
 	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
 	defer cancel()
 	CheckGoroutines(ctx, t)
@@ -68,8 +70,9 @@ func ignoredTestingLeak(d time.Duration) {
 func TestCheckRegisterIgnore(t *testing.T) {
 	RegisterIgnoreGoroutine("ignoredTestingLeak")
 	const leakCount = 3
+	ch := make(chan struct{})
 	for i := 0; i < leakCount; i++ {
-		go func() { time.Sleep(2 * time.Second) }()
+		go func() { <-ch }()
 	}
 	if leaked := interestingGoroutines(); len(leaked) != leakCount {
 		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
@@ -78,9 +81,10 @@ func TestCheckRegisterIgnore(t *testing.T) {
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	if CheckGoroutines(ctx, e); e.errorCount == 0 {
+	if CheckGoroutines(ctx, e); e.errorCount < 3 {
 		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
 	}
+	close(ch)
 	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
 	defer cancel()
 	CheckGoroutines(ctx, t)
--- a/internal/leakcheck/leakcheck_test.go
+++ b/internal/leakcheck/leakcheck_test.go
@@ -23,14 +23,10 @@ import (
 	"fmt"
 	"strings"
 	"sync"
-
-	// "sync"
-
 	"testing"
 	"time"
 
 	"google.golang.org/grpc/internal"
-	// "google.golang.osrg/grpc/internal"
 )
 
 type testLogger struct {
--- a/internal/leakcheck/leakcheck_test.go
+++ b/internal/leakcheck/leakcheck_test.go
@@ -23,10 +23,14 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+
+	// "sync"
+
 	"testing"
 	"time"
 
 	"google.golang.org/grpc/internal"
+	// "google.golang.osrg/grpc/internal"
 )
 
 type testLogger struct {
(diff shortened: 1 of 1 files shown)


internal/leakcheck/leakcheck_test.go
```diff
@@ -47,16 +51,14 @@ func TestCheck(t *testing.T) {
...
-		t.Error("blah")
+	if leaked := interestingGoroutines(); len(leaked) != leakCount {
+		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
 	}
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
+	if CheckGoroutines(ctx, e); e.errorCount == 0 {
```
dfawley, line 60: That works, or you can do

```suggestion
	if CheckGoroutines(ctx, e); ctx.Err() == nil {
```

Or even:

```suggestion
	if CheckGoroutines(ctx, e); e.errorCount < 3 {
```

internal/leakcheck/leakcheck_test.go
```diff
@@ -23,10 +23,14 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+
+	// "sync"
+
```
dfawley, line 28: Please revert this and below.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: This check expects that the goroutines spawned by time.Sleep() should still be alive.
I have not put `e.errorcount == 3`, because ctx can cause its goroutine live longer and it will cause a errorcount of 4 (Because of this the test was flaky)

internal/leakcheck/leakcheck_test.go
Pranjali-2501: Sure.

internal/leakcheck/leakcheck_test.go
dfawley: > I have not put e.errorcount == 3, because ctx can cause its goroutine live longer and it will cause a errorcount of 4

Right, which is why I suggested changing to failing if it's less than three.  We should have at least three.  Or we can just confirm the check failed by checking `ctx.Err()` and then not have to worry about all the complexity to track `errorCount` at all.  Either way is fine.

dfawley (commented):
Please also change the PR first comment to include "Fixes #<issue number>".  This will ensure the PR and issue are linked and close the issue when the PR is merged.

internal/leakcheck/leakcheck_test.go
```diff
@@ -47,16 +47,14 @@ func TestCheck(t *testing.T) {
 	for i := 0; i < leakCount; i++ {
 		go func() { time.Sleep(2 * time.Second) }()
```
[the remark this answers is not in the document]
Pranjali-2501: Sure.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: > Right, which is why I suggested changing to failing if it's less than three. We should have at least three. Or we can just confirm the check failed by checking `ctx.Err()` and then not have to worry about all the complexity to track `errorCount` at all. Either way is fine.

Right, make sense. Both works fine. I'm going with `e.errorcount < 3`for now.

internal/leakcheck/leakcheck_test.go
```diff
@@ -44,20 +44,20 @@ func (e *testLogger) Errorf(format string, args ...any) {
...
 	const leakCount = 3
+	ch := make(chan struct{})
 	for i := 0; i < leakCount; i++ {
-		go func() { time.Sleep(2 * time.Second) }()
+		go func() { <-ch }()
 	}
-	if ig := interestingGoroutines(); len(ig) == 0 {
-		t.Error("blah")
+	if leaked := interestingGoroutines(); len(leaked) != leakCount {
+		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
```
dfawley, line 52: Go style:

```suggestion
		t.Errorf("interestingGoroutines() = %v; want length %v", leaked, leakCount)
```

internal/leakcheck/leakcheck_test.go
```diff
@@ -44,20 +44,20 @@ func (e *testLogger) Errorf(format string, args ...any) {
...
 	}
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
+	if CheckGoroutines(ctx, e); e.errorCount < 3 {
 		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
-		t.Logf("leaked goroutines:\n%v", strings.Join(e.errors, "\n"))
```
dfawley, line 59: Why delete?  This is helpful for debugging.

internal/leakcheck/leakcheck_test.go
```diff
@@ -70,21 +70,21 @@ func ignoredTestingLeak(d time.Duration) {
...
+	ch := make(chan struct{})
 	for i := 0; i < leakCount; i++ {
-		go func() { time.Sleep(2 * time.Second) }()
+		go func() { <-ch }()
 	}
-	go func() { ignoredTestingLeak(3 * time.Second) }()
-	if ig := interestingGoroutines(); len(ig) == 0 {
-		t.Error("blah")
+	if leaked := interestingGoroutines(); len(leaked) != leakCount {
+		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
```
dfawley, line 78: As above.

internal/leakcheck/leakcheck_test.go
```diff
@@ -70,21 +70,21 @@ func ignoredTestingLeak(d time.Duration) {
...
 	}
+	go func() { ignoredTestingLeak(3 * time.Second) }()
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
+	if CheckGoroutines(ctx, e); e.errorCount < 3 {
 		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
-		t.Logf("leaked goroutines:\n%v", strings.Join(e.errors, "\n"))
```
dfawley, line 86: As above.

internal/leakcheck/leakcheck_test.go
```diff
@@ -44,20 +44,20 @@ func (e *testLogger) Errorf(format string, args ...any) {
...
-		t.Error("blah")
+	if leaked := interestingGoroutines(); len(leaked) != leakCount {
+		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
 	}
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
+	if CheckGoroutines(ctx, e); e.errorCount < 3 {
```
dfawley, line 57: You should use `leakCount` again instead of copying.

internal/leakcheck/leakcheck_test.go
```diff
@@ -70,21 +70,21 @@ func ignoredTestingLeak(d time.Duration) {
...
+	if leaked := interestingGoroutines(); len(leaked) != leakCount {
+		t.Errorf("interestingGoroutines found %v leaks, want %v leaks", len(leaked), leakCount)
 	}
+	go func() { ignoredTestingLeak(3 * time.Second) }()
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
+	if CheckGoroutines(ctx, e); e.errorCount < 3 {
```
dfawley, line 84: As above.

dfawley (approved):
Looks good after a few minor nits are fixed.  Thanks!

internal/leakcheck/leakcheck_test.go
Pranjali-2501: Done.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: Added that back.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: Done.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: Sure.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: Done.

internal/leakcheck/leakcheck_test.go
```diff
@@ -70,21 +71,22 @@ func ignoredTestingLeak(d time.Duration) {
...
+		t.Errorf("interestingGoroutines() = %v, want length %v", len(leaked), leakCount)
 	}
+	go func() { ignoredTestingLeak(3 * time.Second) }()
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
-		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
+	if CheckGoroutines(ctx, e); e.errorCount < leakCount {
```
arjan-bal, line 85: This assertion will never fail because we have asserted `len(leaked) == leakCount` above. Should we add a check for the upper bound also to ensure `CheckGoroutines` is actually ignoring the spawned goroutine?

internal/leakcheck/leakcheck_test.go
```diff
@@ -44,20 +44,21 @@ func (e *testLogger) Errorf(format string, args ...any) {
...
+		t.Errorf("interestingGoroutines() = %v, want length %v", len(leaked), leakCount)
 	}
 	e := &testLogger{}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	CheckGoroutines(ctx, e)
-	if e.errorCount != leakCount {
-		t.Errorf("CheckGoroutines found %v leaks, want %v leaks", e.errorCount, leakCount)
+	if CheckGoroutines(ctx, e); e.errorCount < leakCount {
+		t.Errorf("CheckGoroutines() = %v, want count %v", e.errorCount, leakCount)
```
arjan-bal, line 58: nit: Use `%d` to format ints.

arjan-bal, line 58: "want count < %d" to more accurately describe the assertion.

Pranjali-2501: Done.

Pranjali-2501: Checking for e.errorcount >= leakcount, not less than leakcount.

Want e.errorcount >= leakcount because ctx can cause its spawned goroutine to leak, which causes flakiness in the test.

internal/leakcheck/leakcheck_test.go
Pranjali-2501: I have changed `go func() { ignoredTestingLeak(3 * time.Second) }()` to `go ignoredTestingLeak(3 * time.Second)`,to make sure that a extra goroutine is spawned and ignored.

PR #8312, opened 2025-05-12
transport: Prevent sending negative timeouts

arjan-bal:
After https://github.com/grpc/grpc-go/pull/8290 is merged, negative timeout values fail the request with an internal status instead of a deadline exceeded. This test uses negative deadlines and can fail if the server sends an internal error before the client sees the deadline expire.

Sample failure: https://github.com/grpc/grpc-go/actions/runs/14979137997/job/42078795928?pr=8308

The client checks the deadline expiry before creating a stream.
https://github.com/grpc/grpc-go/blob/709023de87a25ae63b000139295af10589edffee/stream.go#L410-L413

However, the deadline may expire after this check, but before the `grpc-timeout` header is created.

RELEASE NOTES:
* transport: Fail RPCs on the client when using extremely short contexts that expire before the `grpc-timeout` header is created.

### DIFF
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -5342,7 +5342,7 @@ func testRPCTimeout(t *testing.T, e env) {
 		ResponseSize: respSize,
 		Payload:      payload,
 	}
-	for i := -1; i <= 10; i++ {
+	for i := 1; i <= 10; i++ {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(i)*time.Millisecond)
 		if _, err := tc.UnaryCall(ctx, req); status.Code(err) != codes.DeadlineExceeded {
 			t.Fatalf("TestService/UnaryCallv(_, _) = _, %v; want <nil>, error code: %s", err, codes.DeadlineExceeded)
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -5342,12 +5342,37 @@ func testRPCTimeout(t *testing.T, e env) {
 		ResponseSize: respSize,
 		Payload:      payload,
 	}
-	for i := 1; i <= 10; i++ {
+	for i := -1; i <= 10; i++ {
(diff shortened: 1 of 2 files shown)


test/end2end_test.go
```diff
@@ -5342,7 +5342,7 @@ func testRPCTimeout(t *testing.T, e env) {
 		ResponseSize: respSize,
 		Payload:      payload,
 	}
-	for i := -1; i <= 10; i++ {
+	for i := 1; i <= 10; i++ {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(i)*time.Millisecond)
```
dfawley, line 5346: This is even worrysome.  If it takes >1ms before the timeout is computed by grpc, then it could fail.  It seems like there's probably something in our code that we should change that prevents the client from sending a negative deadline instead.

arjan-bal: Made the fix to validate the timeout on the client side and return a `DeadlineExceeded` error.

test/end2end_test.go
```diff
@@ -5351,6 +5351,31 @@ func testRPCTimeout(t *testing.T, e env) {
...
+// Tests that the client doesn't send a negative timeout to the server. If the
+// server receives a negative timeout, it would return an internal status. The
+// client checks the context error before starting a stream, however the context
+// may expire after this check and before the timeout is calculated.
+func (s) TestNegativeRPCTimeout(t *testing.T) {
+	server := stubserver.StartTestService(t, nil)
+	defer server.Stop()
+
+	if err := server.StartClient(); err != nil {
+		t.Fatal(err)
```
dfawley, line 5363: This should say what it's doing in the error message, to make debugging a failure easier.

dfawley (approved):
Nice, thank you for fixing this!

dfawley (approved):
Nice, thank you for fixing this!

Issue #8327, opened 2025-05-15
`NO_PROXY` environment variable no longer respected on custom resolver results


nicholasngai:
### What version of gRPC are you using?

v1.71.0

### What version of Go are you using (`go version`)?

```
go version go1.24.2 linux/amd64
```

### What operating system (Linux, Windows, …) and version?

Linux

### What did you do?

```go
package main

import (
	"context"
	"fmt"

	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
	"google.golang.org/grpc/health/grpc_health_v1"
	"google.golang.org/grpc/resolver"
)

// testResolver is a dummy resolver that always resolves to a static address.
type testResolver struct{}

func (testResolver) Scheme() string {
	return "test"
}

func (r testResolver) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
	cc.UpdateState(resolver.State{
		Addresses: []resolver.Address{
			{Addr: "10.1.1.1:1234"},
		},
	})
	return r, nil
}

func (testResolver) ResolveNow(_ resolver.ResolveNowOptions) {}

func (testResolver) Close() {}

func init() {
	resolver.Register(testResolver{})
}

func main() {
	cc, err := grpc.NewClient("test://",
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
	if err != nil {
		panic(err)
	}
	defer cc.Close()

	c := grpc_health_v1.NewHealthClient(cc)

	res, err := c.Check(context.Background(), &grpc_health_v1.HealthCheckRequest{})
	if err != nil {
		panic(err)
	}

	fmt.Println(res)
}
```

Execute under the following environment:

```shell
https_proxy=http://localhost:8080 NO_PROXY=10.0.0.0/8 go run .
```

### What did you expect to see?

Behavior v1.70.0 and below was that, because the resolved address 10.1.1.1:1234 matched `NO_PROXY=10.0.0.0/8`, the proxy would not be used.

### What did you see instead?

gRPC as of v1.71.0 (specifically #7881) now attempts to use a proxy for the outbound connection to `10.1.1.1:1234`.

```
panic: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:8080: connect: connection refused"
```

[...]

PR #8329, opened 2025-05-15
delegatingresolver: avoid proxy for resolved addresses in NO_PROXY env

eshitachandwani:
Fixes: #8327

Add a check for resolved addresses to avoid proxy for addresses mentioned in NO_PROXY environment variable. We only had a check for unresolved addresses.

RELEASE NOTES: 
- client: fixed a bug that was attempting to use a proxy for addresses matching NO_PROXY environment variable

### DIFF
--- a/internal/resolver/delegatingresolver/delegatingresolver.go
+++ b/internal/resolver/delegatingresolver/delegatingresolver.go
@@ -194,15 +194,15 @@ func networkTypeFromAddr(addr resolver.Address) string {
 	return networkType
 }
 
-func isTCPAddressPresent(state *resolver.State) bool {
+func needsProxyResolver(state *resolver.State) bool {
 	for _, addr := range state.Addresses {
-		if networkType := networkTypeFromAddr(addr); networkType == "tcp" {
+		if !skipProxy(addr) {
 			return true
 		}
 	}
 	for _, endpoint := range state.Endpoints {
 		for _, addr := range endpoint.Addresses {
-			if networktype := networkTypeFromAddr(addr); networktype == "tcp" {
+			if !skipProxy(addr) {
 				return true
 			}
 		}
@@ -211,6 +211,10 @@ func isTCPAddressPresent(state *resolver.State) bool {
 }
 
 func skipProxy(address resolver.Address) bool {
+	// Avoid proxy when network is not tcp.
+	if networkType := networkTypeFromAddr(address); networkType != "tcp" {
+		return true
+	}
 	req := &http.Request{URL: &url.URL{
 		Scheme: "https",
 		Host:   address.Addr,
@@ -218,10 +222,6 @@ func skipProxy(address resolver.Address) bool {
 	// Avoid proxy when address included in `NO_PROXY` environment variable.
 	url, err := HTTPSProxyFromEnvironment(req)
 	if err != nil || url == nil {
-		return true
-	}
-	// Avoid proxy when network is not tcp.
-	if networkType := networkTypeFromAddr(address); networkType != "tcp" {
 		return true
 	}
 	return false
@@ -358,7 +358,7 @@ func (r *delegatingResolver) updateTargetResolverState(state resolver.State) err
 	r.targetResolverState = &state
 	// If no addresses returned by resolver have network type as tcp , do not
 	// wait for proxy update.
-	if !isTCPAddressPresent(r.targetResolverState) {
+	if !needsProxyResolver(r.targetResolverState) {
 		return r.cc.UpdateState(*r.targetResolverState)
 	}
 
--- a/internal/resolver/delegatingresolver/delegatingresolver.go
+++ b/internal/resolver/delegatingresolver/delegatingresolver.go
@@ -354,8 +354,9 @@ func (r *delegatingResolver) updateTargetResolverState(state resolver.State) err
 		logger.Infof("Addresses received from target resolver: %v", state.Addresses)
 	}
 	r.targetResolverState = &state
-	// If no addresses returned by resolver have network type as tcp , do not
-	// wait for proxy update.
+	// If all addresses returned by the target resolver have a non-TCP network
+	// type, or are listed in the `NO_PROXY` environment variable, do not wait
+	// for proxy update.
 	if !needsProxyResolver(r.targetResolverState) {
 		return r.cc.UpdateState(*r.targetResolverState)
 	}
--- a/internal/resolver/delegatingresolver/delegatingresolver.go
+++ b/internal/resolver/delegatingresolver/delegatingresolver.go
@@ -109,6 +109,7 @@ func New(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOpti
 
 	var err error
 	r.proxyURL, err = proxyURLForTarget(target.Endpoint())
+	logger.Info("Eshita proxy url %s", r.proxyURL)
 	if err != nil {
 		return nil, fmt.Errorf("delegating_resolver: failed to determine proxy URL for target %s: %v", target, err)
 	}
@@ -210,6 +211,18 @@ func isTCPAddressPresent(state *resolver.State) bool {
 	return false
 }
 
+func isProxyNeeded(address string) bool {
+	req := &http.Request{URL: &url.URL{
+		Scheme: "https",
+		Host:   address,
+	}}
+	url, err := HTTPSProxyFromEnvironment(req)
+	if err != nil || url == nil {
+		return false
+	}
+	return true
+}
+
 // updateClientConnStateLocked constructs a combined list of addresses by
 // pairing each proxy address with every target address of type TCP. For each
 // pair, it creates a new [resolver.Address] using the proxy address and
@@ -241,7 +254,7 @@ func (r *delegatingResolver) updateClientConnStateLocked() error {
 	var addresses []resolver.Address
 	for _, targetAddr := range (*r.targetResolverState).Addresses {
 		// Avoid proxy when network is not tcp.
-		if networkType := networkTypeFromAddr(targetAddr); networkType != "tcp" {
+		if networkType := networkTypeFromAddr(targetAddr); networkType != "tcp" || !isProxyNeeded(targetAddr.Addr) {
 			addresses = append(addresses, targetAddr)
 			continue
 		}
@@ -259,7 +272,7 @@ func (r *delegatingResolver) updateClientConnStateLocked() error {
 		var addrs []resolver.Address
 		for _, targetAddr := range endpt.Addresses {
 			// Avoid proxy when network is not tcp.
-			if networkType := networkTypeFromAddr(targetAddr); networkType != "tcp" {
+			if networkType := networkTypeFromAddr(targetAddr); networkType != "tcp" || !isProxyNeeded(targetAddr.Addr) {
 				addrs = append(addrs, targetAddr)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


purnesh42H:
I think release notes should mention the bug fix. How about something like `client: fixed a bug that was attempting to use a proxy for addresses matching NO_PROXY environment variable`?

purnesh42H:
I don't have much knowledge about the recent changes here. So I will defer to @easwars and @arjan-bal

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -241,7 +253,7 @@ func (r *delegatingResolver) updateClientConnStateLocked() error {
 	var addresses []resolver.Address
 	for _, targetAddr := range (*r.targetResolverState).Addresses {
 		// Avoid proxy when network is not tcp.
-		if networkType := networkTypeFromAddr(targetAddr); networkType != "tcp" {
+		if networkType := networkTypeFromAddr(targetAddr); networkType != "tcp" || !isProxyNeeded(targetAddr.Addr) {
```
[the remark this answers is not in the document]
eshitachandwani: Done.

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -210,6 +210,18 @@ func isTCPAddressPresent(state *resolver.State) bool {
 	return false
 }
 
+func isProxyNeeded(address string) bool {
```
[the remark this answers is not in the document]
eshitachandwani: Done

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go
```diff
@@ -968,3 +953,81 @@ func (s) TestDelegatingResolverForMixNetworkType(t *testing.T) {
...
+		// `NO_PROXY` env variable.
+		if req.URL.Host == noproxyresolvedTargetAddr1 {
+			return nil, nil
+		}
+		return &url.URL{
+			Scheme: "https",
+			Host:   envProxyAddr,
+		}, nil
+	}
+	originalhpfe := delegatingresolver.HTTPSProxyFromEnvironment
```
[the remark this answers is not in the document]
eshitachandwani: I remember having some problem with using the proxy environment variable and so we switched to  replacing the function. 
SOmething on the lines that the function used to retrieve proxy address uses Do.Once function and that gets done the first time we run the test suite and so every time we do not get the value set in proxy environments and so we decided to replace the function in tests.

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -210,6 +210,23 @@ func isTCPAddressPresent(state *resolver.State) bool {
...
+	}}
+	// Avoid proxy when address included in `NO_PROXY` environment variable.
+	url, err := HTTPSProxyFromEnvironment(req)
+	if err != nil || url == nil {
+		return true
+	}
+	// Avoid proxy when network is not tcp.
+	if networkType := networkTypeFromAddr(address); networkType != "tcp" {
+		return true
+	}
```
arjan-bal, line 226: Can you move this check to the top? I think it makes sense to only lookup the proxy config for tcp addresses.

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go
```diff
@@ -968,3 +953,81 @@ func (s) TestDelegatingResolverForMixNetworkType(t *testing.T) {
...
 }
+
+// Tests the scenario where a proxy is configured but some addresses are
+// excluded (by using the NO_PROXY environment variable). The test verifies that
+// the delegating resolver doesn't add proxyatrribute to adresses excluded , but
+// adds the proxyattribute to all other addresses.
+func (s) TestDelegatingResolverWithNoProxyEnvUsed(t *testing.T) {
+	const (
+		targetTestAddr             = "test.target"
+		noproxyresolvedTargetAddr1 = "1.1.1.1:8080"
```
arjan-bal, line 964: nit: `noProxyResolvedTargetAddr`. The suffixes 1 and 2 can be removed as the names are unique without them.

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -210,6 +210,23 @@ func isTCPAddressPresent(state *resolver.State) bool {
...
+		Scheme: "https",
+		Host:   address.Addr,
+	}}
+	// Avoid proxy when address included in `NO_PROXY` environment variable.
+	url, err := HTTPSProxyFromEnvironment(req)
+	if err != nil || url == nil {
+		return true
+	}
+	// Avoid proxy when network is not tcp.
+	if networkType := networkTypeFromAddr(address); networkType != "tcp" {
```
arjan-bal, line 224: We can make `isTCPAddressPresent` use `skipProxy` instead and  inline `networkTypeFromAddr` . `isTCPAddressPresent` would also need to be renamed to something more appropriate then. This would unify the checks in both places.

eshitachandwani: Done.

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go

internal/resolver/delegatingresolver/delegatingresolver.go
eshitachandwani: Done.

internal/resolver/delegatingresolver/delegatingresolver.go
arjan-bal: Can we inline `networkTypeFromAddr` since there is only one caller now? This would also prevent people from accidentally bypassing the "ignore proxy" check in the future.

eshitachandwani: Right! Missed that. Done!

arjan-bal (approved):
LGTM

arjan-bal:
Please open a flake for the failing advancedtls test and assign it to someone from the security team.

https://github.com/grpc/grpc-go/actions/runs/15105499577/job/42453426886?pr=8329

nicholasngai:
FWIW, I think it is worth mentioning that this behavior is probably non-idiomatic and is really only being put in for backwards compatibility. For example, it would not make sense for `dns:///localhost:1234` to match `NO_PROXY=127.0.0.1`—browsers and anyone else checking NO_PROXY traditionally only match against the target URL, not the underlying resolved addresses.

There is potential for inconsistent behavior here: For example, this would mean that `dns:///example.com:1234` with `NO_PROXY=1.2.3.4` would route to the proxy if the DNS for example.com pointed to `5.6.7.8` one day but then not route to the proxy if the DNS pointed to `1.2.3.4` the next day—applications generally would only match against `NO_PROXY=example.com` instead.

eshitachandwani:
> FWIW, I think it is worth mentioning that this behavior is probably non-idiomatic and is really only being put in for backwards compatibility. For example, it would not make sense for `dns:///localhost:1234` to match `NO_PROXY=127.0.0.1`—browsers and anyone else checking NO_PROXY traditionally only match against the target URL, not the underlying resolved addresses.
> 
> There is potential for inconsistent behavior here: For example, this would mean that `dns:///example.com:1234` with `NO_PROXY=1.2.3.4` would route to the proxy if the DNS for example.com pointed to `5.6.7.8` one day but then not route to the proxy if the DNS pointed to `1.2.3.4` the next day—applications generally would only match against `NO_PROXY=example.com` instead.

IMO including IP addresses in the NO_PROXY environment variable should not be considered inconsistent behavior. The [Go proxy configuration](https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config) explicitly supports IP addresses in `NO_PROXY,` and it is common for users to intentionally set this environment variable to exclude certain internal or reserved IP ranges from proxying. This is a practical and expected use case—particularly in environments where internal services are hosted on private network ranges. This behavior aligns with Go’s documented proxy semantics and is consistent with standard practices in private and enterprise deployments.

dfawley:
I agree we should support IP addresses, but it seems we should also support domain names here, too.  If the primary resolver is DNS and NO_PROXY includes the hostname we are looking up, we should not use a proxy.

dfawley:
I see.  So this is happening later in the process, after addresses are resolved, when we are doing local resolution.  And since our behavior for many years had been to respect `http_proxy` and `no_proxy` for locally-resolved addresses, this change is restoring that behavior.

This is basically use case 2 from [gRFC A1](https://github.com/grpc/proposal/blob/master/A1-http-connect-proxy-support.md#use-cases).  And I think as long as we are doing local name resolution, then it is pretty reasonable to check `no_proxy` for these resolved addresses.  And since it's also restoring backward compatibility, then there's no reason to not do this.  Note that Java doesn't support use case 2, and C++ doesn't use the environment variables for that use case, but custom logic instead.

internal/resolver/delegatingresolver/delegatingresolver.go
```diff
@@ -342,7 +356,7 @@ func (r *delegatingResolver) updateTargetResolverState(state resolver.State) err
 	r.targetResolverState = &state
 	// If no addresses returned by resolver have network type as tcp , do not
 	// wait for proxy update.
```
dfawley, line 358: Please update.

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go
```diff
@@ -177,14 +177,10 @@ func (s) TestDelegatingResolverwithDNSAndProxyWithTargetResolution(t *testing.T)
 		resolvedProxyTestAddr1  = "11.11.11.11:7687"
 	)
 	hpfe := func(req *http.Request) (*url.URL, error) {
```
dfawley, line 179: There seems to be a lot of repetition of this, which required you to change it in many many places.  Can you factor this out and share it between the test cases since it seems to be identical every time?

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go
```diff
@@ -968,3 +953,81 @@ func (s) TestDelegatingResolverForMixNetworkType(t *testing.T) {
 		t.Fatalf("Unexpected state from delegating resolver. Diff (-got +want):\n%v", diff)
 	}
 }
+
+// Tests the scenario where a proxy is configured but some addresses are
+// excluded (by using the NO_PROXY environment variable). The test verifies that
+// the delegating resolver doesn't add proxyatrribute to adresses excluded , but
+// adds the proxyattribute to all other addresses.
```
dfawley, line 960: Please make sure you don't have spaces before commas.  (I've noticed this elsewhere, too.)

internal/transport/proxy_ext_test.go
```diff
@@ -97,14 +97,10 @@ func (s) TestGRPCDialWithProxy(t *testing.T) {
 
 	// Overwrite the function in the test and restore them in defer.
 	hpfe := func(req *http.Request) (*url.URL, error) {
```
dfawley, line 99: Same comments for this file re: re-use.

dfawley (approved):

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go
eshitachandwani: Done.

internal/resolver/delegatingresolver/delegatingresolver_ext_test.go

internal/transport/proxy_ext_test.go
eshitachandwani: Done.

PR #8330, opened 2025-05-15
Cardinality violations in Client streaming and Unary RPC.

Pranjali-2501:
Client should ensure an `INTERNAL` error is returned for non-server streaming RPCs if the server attempts to send more than one response.
Currently it is returning `UNKNOWN`, which should not be the case for cardinality violations.

Filed in https://github.com/grpc/grpc-go/issues/7286
RELEASE NOTES: N/A

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -1167,7 +1167,7 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
 	} else if err != nil {
 		return toRPCErr(err)
 	}
-	return toRPCErr(errors.New("grpc: client streaming protocol violation: get <nil>, want <EOF>"))
+	return status.Errorf(codes.Internal, "client streaming cardinality violation: get <nil>, want <EOF>")
 }
 
 func (a *csAttempt) finish(err error) {
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -3739,6 +3739,39 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	}
 }
 
+// Tests that a client receives a cardinality violation error for client-streaming
+// RPCs if the server call SendAndClose after calling SendAndClose.
+func (s) TestClientStreaming_ServerHandlerSendAndCloseAfterSendAndClose(t *testing.T) {
+	ss := stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
+				t.Errorf("stream.SendAndClose(_) = %v, want <nil>", err)
+			}
+			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
+				t.Errorf("stream.SendAndClose(_) = %v, want <nil>", err)
+			}
+			return nil
+		},
+	}
+	if err := ss.Start(nil); err != nil {
+		t.Fatal("Error starting server:", err)
+	}
+	defer ss.Stop()
+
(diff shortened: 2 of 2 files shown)


dfawley:
@Pranjali-2501 please use the `Assignees` field to indicate who needs to act on the PR next.  I assume you have meant to assign this to @arjan-bal and me?

test/end2end_test.go
```diff
@@ -3739,6 +3739,39 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	}
 }
 
+// Tests that a client receives a cardinality violation error for client-streaming
+// RPCs if the server call SendAndClose after calling SendAndClose.
```
dfawley, line 3743:
```suggestion
// RPCs if the server calls SendAndClose multiple times.
```

test/end2end_test.go
```diff
@@ -3739,6 +3739,39 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	defer cancel()
+	stream, err := ss.Client.StreamingInputCall(ctx)
+	if err != nil {
+		t.Fatalf(".StreamingInputCall(_) = _, %v, want <nil>", err)
+	}
+	if err := stream.Send(&testpb.StreamingInputCallRequest{}); err != nil {
+		t.Fatalf("stream.Send(_) = %v, want <nil>", err)
+	}
+	if _, err := stream.CloseAndRecv(); status.Code(err) != codes.Internal {
+		t.Fatalf("stream.CloseAndRecv() = %v, want error %s", err, codes.Internal)
```
dfawley, line 3771:
```suggestion
		t.Fatalf("stream.CloseAndRecv() = %v, want error with status code %s", err, codes.Internal)
```

dfawley (approved):
Looks great, just a couple really small things.  Thanks!

test/end2end_test.go

test/end2end_test.go

stream.go
```diff
@@ -1167,7 +1167,7 @@ func (a *csAttempt) recvMsg(m any, payInfo *payloadInfo) (err error) {
 	} else if err != nil {
 		return toRPCErr(err)
 	}
-	return toRPCErr(errors.New("grpc: client streaming protocol violation: get <nil>, want <EOF>"))
+	return status.Errorf(codes.Internal, "client streaming cardinality violation: get <nil>, want <EOF>")
```
arjan-bal, line 1170: There is a similar method on `addrConnStream` that also needs to be updated.

arjan-bal (approved):
Approving.

stream.go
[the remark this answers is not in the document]

test/end2end_test.go
```diff
@@ -3739,6 +3739,39 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	}
 }
 
+// Tests that a client receives a cardinality violation error for client-streaming
+// RPCs if the server call SendAndClose multiple times.
+func (s) TestClientStreaming_ServerHandlerSendAndCloseAfterSendAndClose(t *testing.T) {
+	ss := stubserver.StubServer{
+		StreamingInputCallF: func(stream testgrpc.TestService_StreamingInputCallServer) error {
+			if err := stream.SendAndClose(&testpb.StreamingInputCallResponse{}); err != nil {
```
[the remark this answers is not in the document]
arjan-bal: Existing refferences use `SendAndClose` probably because the intention is to close the stream (even though closing doesn't happen until the handler returns). This is not done for consistency. In this test, we want to send multiple message and not close, so I would suggest just changing this test to use `Close()`. We don't need to change the remaining tests. Both `Send` and `SendAndClose` are part of the stable API.

PR #8331, opened 2025-05-16

purnesh42H:
RELEASE NOTES: None


purnesh42H:
> Vet is failing, seems to be related to this PR. Can you PTAL?

So, looks like there was version incompatibility with `tableWriter` module in the latest version of revive https://github.com/mgechev/revive/issues/1345. Anyways, they fixed it by replacing the `tableWriter` with `fmt.Sprintf()` https://github.com/mgechev/revive/pull/1351.

For our release I think we can skip upgrading the revive because otherwise we have to wait for their patch release. So, i have reverted the version bump for revive.

arjan-bal (approved):
LGTM with a comment for opening a tracking issue for bumping revive later.

purnesh42H:
> > > Vet is failing, seems to be related to this PR. Can you PTAL?
> > 
> > 
> > So, looks like there was version incompatibility with `tableWriter` module in the latest version of revive [mgechev/revive#1345](https://github.com/mgechev/revive/issues/1345). Anyways, they fixed it by replacing the `tableWriter` with `fmt.Sprintf()` [mgechev/revive#1351](https://github.com/mgechev/revive/pull/1351).
> > For our release I think we can skip upgrading the revive because otherwise we have to wait for their patch release. So, i have reverted the version bump for revive.
> 
> Can you open an issue for this and add a TODO so that this issue is not forgotten?

Done

arjan-bal:
> In general can we do the version bump after the release is published? In case of revive though its for internal use only. Would it still need to be done as part of patch release? or we just wait for next minor release? @dfawley cc

We are bumping the dep on master (not a release branch), I don't think we need to worry about patch releases here. We can bump this dependency at any point in the future.

Issue #8332, opened 2025-05-16
Least request causes panic on ReporterError starting in 1.71.0


atollena:
### What version of gRPC are you using?

1.71.1, master.

### What version of Go are you using (`go version`)?

go version go1.24.2

### What operating system (Linux, Windows, …) and version?

Linux, MacOS

### What did you do?

I'm using the least_request_experimental balancer. A resolver reports an error.

### What did you expect to see?

Maybe an error log statement reporting the resolver error.

### What did you see instead?

gRPC panics because it's embedded balancer at https://github.com/grpc/grpc-go/blob/a51009d1d7074ee1efcd323578064cbe44ef87e5/balancer/leastrequest/leastrequest.go#L100-L102 is nil.

Calls to methods that are not overridden by `leastRequestBalancer` directly should instead be explicitly forwarded to the child balancer (or `leastRequestBalancer` should use `endpointsharding` as its embedded balancer instead of having a separate child field).

PR #8333, opened 2025-05-16
balancer/least_request : Fix panic while handling resolver errors

atollena:
There was a leftover embedded balancer in least request that was never set in the code when it got refactored to pick first as leaf. As a result, method calls forwarded to the embedded balancer would panic.

Remove the embedded balance and properly forward the missing methods.

Fixes https://github.com/grpc/grpc-go/issues/8332

RELEASE NOTES:
- balancer/least_request: fix panic on resolver errors when using least_request.

### DIFF
--- a/balancer/leastrequest/leastrequest_test.go
+++ b/balancer/leastrequest/leastrequest_test.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc/status"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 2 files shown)


balancer/leastrequest/leastrequest_test.go
```diff
@@ -706,3 +707,31 @@ func (s) TestLeastRequestEndpoints_MultipleAddresses(t *testing.T) {
...
+	mr := manual.NewBuilderWithScheme("lr-e2e")
+	defer mr.Close()
+
+	cc, err := grpc.NewClient(mr.Scheme()+":///", grpc.WithResolvers(mr), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient() failed: %v", err)
+	}
+	defer cc.Close()
+
+	cc.Connect()
```
arjan-bal, line 721: There are a few changes needed to reliably trigger `ResolverError`:
1. Use a service config that specifies least request as the LB policy. You could use a dial option or the manual resolver to send the service config.
2. Send an initial update with at least one endpoint. This is needed since endpointsharding needs to have a child on which it will call ResolverError to update the picker. Otherwise, the error will be something like "No children to pick from".
3. Wait for the channel to enter TF after making the connection fail. If the channel is not in TF, the picker with the resolver error will not get used.
4. After sending the resolver error, make RPCs repeatedly till the resolver error is present as a substring of the status message. The picker update will not happen synchronously. The following PF test does something similar: https://github.com/grpc/grpc-go/blob/f2d3e11f3057cc8b4935fc342446ae8372de33d3/balancer/pickfirst/pickfirst_ext_test.go#L873

atollena: I took all your suggestions.

arjan-bal (approved):
LGTM

Issue #8222, opened 2025-04-03
Refactor `pemfile.Options` as the intermediate struct is not needed


gtcooke94:
The only reason for using the intermediate struct is to be able to specify a default value for the refresh interval, this could be done with the encoding/json package as follows: https://stackoverflow.com/questions/30445479/how-to-specify-default-values-when-parsing-json-in-go

See https://github.com/grpc/grpc-go/pull/8180/files#r2002089915

hugehoo:
Hi, do you mind if i work on this issue?

injae-kim:
Fix PR by @hugehoo : https://github.com/grpc/grpc-go/pull/8334

arjan-bal:
During the review we decided that the fixes didn't offer a significant improvement over the existing implementation, so we're not making any changes.
https://github.com/grpc/grpc-go/pull/8334#discussion_r2100931516

PR #8334, opened 2025-05-18
pem: removed unnecessary intermediate struct for `pemfile.Options`

hugehoo:
### issues
fixes: #8222 

### changed
- added json tags in Options struct for default value
- so that i can remove unnecessary intermediate field for specifying a default value .

RELEASE NOTES: N/A

### DIFF
--- a/credentials/tls/certprovider/pemfile/builder.go
+++ b/credentials/tls/certprovider/pemfile/builder.go
@@ -23,9 +23,10 @@ import (
 	"fmt"
 	"time"
 
-	"google.golang.org/grpc/credentials/tls/certprovider"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/durationpb"
+
+	"google.golang.org/grpc/credentials/tls/certprovider"
 )
 
 const (
@@ -62,32 +63,23 @@ func pluginConfigFromJSON(jd json.RawMessage) (Options, error) {
 	// The only difference between this anonymous struct and the Options struct
 	// is that the refresh_interval is represented here as a duration proto,
 	// while in the latter a time.Duration is used.
-	cfg := &struct {
-		CertificateFile          string          `json:"certificate_file,omitempty"`
-		PrivateKeyFile           string          `json:"private_key_file,omitempty"`
-		CACertificateFile        string          `json:"ca_certificate_file,omitempty"`
-		SPIFFETrustBundleMapFile string          `json:"spiffe_trust_bundle_map_file,omitempty"`
-		RefreshInterval          json.RawMessage `json:"refresh_interval,omitempty"`
-	}{}
-	if err := json.Unmarshal(jd, cfg); err != nil {
+	opts := Options{
+		RefreshDuration: defaultRefreshInterval,
+	}
+
+	if err := json.Unmarshal(jd, &opts); err != nil {
 		return Options{}, fmt.Errorf("pemfile: json.Unmarshal(%s) failed: %v", string(jd), err)
 	}
 
-	opts := Options{
-		CertFile:            cfg.CertificateFile,
-		KeyFile:             cfg.PrivateKeyFile,
-		RootFile:            cfg.CACertificateFile,
-		SPIFFEBundleMapFile: cfg.SPIFFETrustBundleMapFile,
-		// Refresh interval is the only field in the configuration for which we
-		// support a default value. We cannot possibly have valid defaults for
-		// file paths to watch. Also, it is valid to specify an empty path for
-		// some of those fields if the user does not want to watch them.
-		RefreshDuration: defaultRefreshInterval,
+	var rawMap map[string]json.RawMessage
+	if err := json.Unmarshal(jd, &rawMap); err != nil {
+		return Options{}, fmt.Errorf("pemfile: json.Unmarshal map failed: %v", err)
 	}
-	if cfg.RefreshInterval != nil {
+
+	if rawMap["refresh_interval"] != nil {
 		dur := &durationpb.Duration{}
-		if err := protojson.Unmarshal(cfg.RefreshInterval, dur); err != nil {
-			return Options{}, fmt.Errorf("pemfile: protojson.Unmarshal(%+v) failed: %v", cfg.RefreshInterval, err)
+		if err := protojson.Unmarshal(rawMap["refresh_interval"], dur); err != nil {
+			return Options{}, fmt.Errorf("pemfile: protojson.Unmarshal(%+v) failed: %v", rawMap["refresh_interval"], err)
 		}
 		opts.RefreshDuration = dur.AsDuration()
 	}
--- a/credentials/tls/certprovider/pemfile/builder.go
+++ b/credentials/tls/certprovider/pemfile/builder.go
@@ -71,7 +71,7 @@ func pluginConfigFromJSON(jd json.RawMessage) (Options, error) {
 	if opts.RefreshInterval != "" {
 		dur, err := time.ParseDuration(opts.RefreshInterval)
 		if err != nil {
-			return Options{}, fmt.Errorf("pemfile: failed to parse refresh interval %q: %v", opts, err)
+			return Options{}, fmt.Errorf("pemfile: failed to parse refresh interval: %v", err)
 		}
 		opts.RefreshDuration = dur
 	}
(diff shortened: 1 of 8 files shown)


credentials/tls/certprovider/pemfile/builder.go
```diff
@@ -62,32 +63,23 @@ func pluginConfigFromJSON(jd json.RawMessage) (Options, error) {
...
-		// file paths to watch. Also, it is valid to specify an empty path for
-		// some of those fields if the user does not want to watch them.
-		RefreshDuration: defaultRefreshInterval,
+	var rawMap map[string]json.RawMessage
+	if err := json.Unmarshal(jd, &rawMap); err != nil {
+		return Options{}, fmt.Errorf("pemfile: json.Unmarshal map(%s) failed: %v", string(jd), err)
 	}
-	if cfg.RefreshInterval != nil {
+
+	if rawMap["refresh_interval"] != nil {
```
eshitachandwani, line 79: What is the use `rawMap` becuase we convert the time to `time.Duration` after and the fault value is set by using 
`json.Unmarshal` as mentioned in the description? Or does it not do that? Doing `json.Unmarshal` twice seems redundant and defeats the purpose. Or am I missing something and there is some other purpose of this?

hugehoo: @eshitachandwani  
this is what i understand, 
`json.Unmarshal` at first can't unmarshal the key "refresh_interval" in `jd` argument, because Options struct doesn't have matching field with "refresh_interval", it has `RefreshDuration` instead.

so that's the reason why i made rawMap for unmarshalling the "refresh_interval" key in `jd`.

credentials/tls/certprovider/pemfile/builder.go
```diff
@@ -62,32 +63,23 @@ func pluginConfigFromJSON(jd json.RawMessage) (Options, error) {
...
-		CertFile:            cfg.CertificateFile,
-		KeyFile:             cfg.PrivateKeyFile,
-		RootFile:            cfg.CACertificateFile,
-		SPIFFEBundleMapFile: cfg.SPIFFETrustBundleMapFile,
-		// Refresh interval is the only field in the configuration for which we
-		// support a default value. We cannot possibly have valid defaults for
-		// file paths to watch. Also, it is valid to specify an empty path for
-		// some of those fields if the user does not want to watch them.
-		RefreshDuration: defaultRefreshInterval,
+	var rawMap map[string]json.RawMessage
```
gtcooke94, line 74: I think this isn't quite what we want to do - here we've just swapped the `Options` struct for a map without solving the default issue

I think the idea here is rather that we don't need two structs (or most of this code) at all, so swapping one for a map doesn't do it - rather, we should be able to handle a default `RefreshDuration` during the unmarshal call without having an intermediate in the first place.

gtcooke94 (commented):
Thanks for taking this up! I think this isn't quite the right direction - see my comment about the `rawMap` that was introduced

credentials/tls/certprovider/pemfile/builder.go
hugehoo: it seems like Options struct needs RefreshInterval field for unmarshalling.

eshitachandwani:
@hugehoo , is this ready for review?

hugehoo:
> @hugehoo , is this ready for review?

I'm just struggling on CI step(dependency changes). I'll comment you again when it's ready!

hugehoo:
> @hugehoo , is this ready for review?

@eshitachandwani yes it's ready for review thx.

credentials/tls/certprovider/pemfile/watcher.go
```diff
@@ -55,22 +55,26 @@ var (
...
-	SPIFFEBundleMapFile string
+	SPIFFEBundleMapFile string `json:"spiffe_trust_bundle_map_file,omitempty"`
 	// RefreshDuration is the amount of time the plugin waits before checking
 	// for updates in the specified files.
 	// Optional. If not set, a default value (1 hour) will be used.
 	RefreshDuration time.Duration
+	// RefreshInterval is the duration string parsed to set RefreshDuration.
+	// This field must be a valid Go duration string (e.g., "10s", "5m").
+	// If unset, the default value for RefreshDuration will be used.
+	RefreshInterval string `json:"refresh_interval,omitempty"`
```
gtcooke94, line 77: I don't think we want to introduce a new setting on this `Options` struct - this could be confusing for users - e.g. "Do I set RefreshDuration or RefreshInterval, and what's the difference?"

See
https://stackoverflow.com/questions/30445479/how-to-specify-default-values-when-parsing-json-in-go
and
https://stackoverflow.com/questions/48050945/how-to-unmarshal-json-into-durations

For handling default values when unmarshaling JSON

cc: @arjan-bal since he was the first to point this out

hugehoo: @gtcooke94 I understand what you meaning, thx for the references.
so i wrap the RefreshDuration field as custom Duration type and removed confusing field(RefreshInterval).

but i'm bit worried that type of RefreshDuration would may change. how do you think of it?

credentials/tls/certprovider/pemfile/watcher.go
```diff
@@ -55,22 +56,41 @@ var (
...
+	time.Duration
+}
+
+// UnmarshalJSON allows decoding a duration string like "200s" into time.Duration.
+func (d *Duration) UnmarshalJSON(b []byte) error {
+	var s string
+	if err := json.Unmarshal(b, &s); err != nil {
+		return fmt.Errorf("duration must be a string: %w", err)
+	}
+	dur, err := time.ParseDuration(s)
```
arjan-bal, line 88: The [gRFC](https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md#file_watcher-certificate-provider) mentions the following:
* `refresh_interval`: Specifies how frequently the plugin should read the files. The value must be in the JSON format described for a Duration protobuf message.

Based on this, using `time.ParseDuration` will deviate from the spec. Parsing a duration like "1m30s" would work with `time.ParseDuration` but fail when using `protojson.Unmarshal(cfg.RefreshInterval, &durationpb.Duration{})`. We could fix this by using protjosn and durationpb to parse in the here, but I'm not convinced addressing https://github.com/grpc/grpc-go/issues/8222 reduces complexity.

credentials/tls/certprovider/pemfile/builder.go
```diff
@@ -62,34 +60,12 @@ func pluginConfigFromJSON(jd json.RawMessage) (Options, error) {
 	// The only difference between this anonymous struct and the Options struct
 	// is that the refresh_interval is represented here as a duration proto,
 	// while in the latter a time.Duration is used.
-	cfg := &struct {
```
arjan-bal, line 65: @easwars IMO, the existing way of using an anonymous struct seems better. It keeps the parsing logic contained in this single function. With the change in the PR we're adding our own Duration type and a method in another file, exposing the deserialization behaviour to all users of the `Options` struct. I wanted to get your thoughts on this.

arjan-bal: One way to reduce duplication is to embed the `Options` struct here:
```go
cfg := &struct {
    Options   `json:",inline"`
    RefreshInterval string `json:"refresh_interval,omitempty"`
}{}
```

The `RefreshInterval` of the anonymous struct would get the deserialized field from the json. We can then parse it and set the field on `cfg.Duration` before returning `cfg.Duration`.

credentials/tls/certprovider/pemfile/watcher.go
hugehoo: > but fail when using protojson.Unmarshal(cfg.RefreshInterval, &durationpb.Duration{}). We could fix this by using protjosn and durationpb for parsing here,

@arjan-bal thx for comment, i have one question.
you mean it fails when using `protojson.Unmarshal(cfg.RefreshInterval, &durationpb.Duration{})`, and it could be fixed by using protojson and durationpb? i'm just bit confusing. i thought it's same meaning.

arjan-bal: `protojson.Unmarshal(cfg.RefreshInterval, &durationpb.Duration{}` will fail to parse a durations like "1m30s". This is the correct behaviour according to the gRFC. It may be that `time.ParseDuration` can parse a superset of what `durationpb` can (I haven't verified this), but we should not use `ParseDuration` to maintain cross language consistency.  If gRPC Go users start relying on the the use of `ParseDuration`, they'll see failures when the same config is used with other gRPC implementations.

PR #8341, opened 2025-05-19
endpointsharding: Implement ExitIdler

arjan-bal:
@atollena pointed out that the endpointsharding LB policy doesn't implement `ExitIdler` but it's parent policies expect it to. If a balancer that delegates to endpointsharding implements ExitIdle, it won't be able to make the children of endpointsharding exit idle without making an RPC (triggering the idle picker). We haven't seen this issue cause problems because endpointsharding [will never stay in Idle ](https://github.com/grpc/grpc-go/blob/f2d3e11f3057cc8b4935fc342446ae8372de33d3/balancer/endpointsharding/endpointsharding.go#L320-L323) unless the `DisableAutoReconnect` option is used. Only ringhash sets the `DisableAutoReconnect` option, but [it doesn't need to call ExitIdle on its child balancer](https://github.com/grpc/grpc-go/blob/edf643f50334001e8a434087c480cd5d70e9fe91/xds/internal/balancer/ringhash/ringhash.go#L304-L307).

This PR also adds error logs in other places where the code type asserts `ExitIdler` to catch such issues in tests.

RELEASE NOTES:
- balancer/endpointsharding: Implement `balancer.ExitIdler` to allow channel to exit idle without making RPCs.

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -205,6 +205,17 @@ func (es *endpointSharding) Close() {
 	}
 }
 
+func (es *endpointSharding) ExitIdle() {
+	es.childMu.Lock()
+	defer es.childMu.Unlock()
+	for _, bw := range es.children.Load().Values() {
+		ei, ok := bw.child.(balancer.ExitIdler)
+		if ok && !bw.isClosed {
+			ei.ExitIdle()
+		}
+	}
+}
+
 // updateState updates this component's state. It sends the aggregated state,
 // and a picker with round robin behavior with all the child states present if
 // needed.
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -27,6 +27,7 @@ package endpointsharding
 
 import (
 	"errors"
+	"fmt"
 	rand "math/rand/v2"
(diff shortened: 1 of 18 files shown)


arjan-bal:
> It looks like WRR locality also doesn't implement `ExitIdle`:
> 
> https://github.com/grpc/grpc-go/blob/0c24af1c707cbb7302b52113fd0f163ecbe56b42/xds/internal/balancer/wrrlocality/balancer.go#L146
> 
> We should probably also add an error message when a balancer is registered if it doesn't support `ExitIdle`. That might help us to make it required in a future release. Then we could eat all the errors when type asserting elsewhere at runtime?

Since the registry contains `Builders`, to assert the type of the produced `Balancers`, we would probably need to wrap the builders and intercept their `Balancers`. Wrapping the `Builder` will erase the `ConfigParser` interface that `Builders` can optionally implement, so we would need to choose a wrapping struct at runtime that can implement `ConfigParser` based on the delegate. This seems a little complicated. Do you have a simpler approach in mind?

dfawley:
I think we are going to skip this implementation for now, and go about requiring ExitIdler instead?

arjan-bal:
Opened https://github.com/grpc/grpc-go/issues/8345 to track the API change.

PR #8342, opened 2025-05-19
stats/opentelemetry: record retry attempts from clientStream

vinothkumarr227:
Fixes: https://github.com/grpc/grpc-go/issues/8299

RELEASE NOTES:

- stats/opentelemetry: Retry attempts (`grpc.previous-rpc-attempts`) are now recorded as span attributes for non-transparent client retries.


stream.go
```diff
@@ -430,6 +432,7 @@ func (cs *clientStream) newAttemptLocked(isTransparent bool) (*csAttempt, error)
 			IsServerStream:            cs.desc.ServerStreams,
 			IsTransparentRetryAttempt: isTransparent,
 		}
+		ctx = context.WithValue(ctx, clientStreamKey{}, cs.numRetries)
```
purnesh42H, line 435: as @dfawley suggested, we should just compute the retry attempts within otel interceptors https://github.com/grpc/grpc-go/issues/8299#issuecomment-2887537306. While incrementing in trace.go, we should just skip if transparent retry is true. Re-using `cs.numRetries` is not advisable.

vinothkumarr227: Thanks! Updated to compute retry attempts in the otel interceptor (TagRPC) and no longer using cs.numRetries

stats/opentelemetry/client_tracing.go
```diff
@@ -121,6 +122,10 @@ func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
 // TagRPC implements per RPC attempt context management for traces.
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	if ai.previousRPCAttempts > 0 {
+		atomic.AddUint32(&ai.explicitRetryCount, 1)
```
purnesh42H, line 126: Retry attempts need to be counted by storing data in the context in a per-call interceptor. If you store it in the attemptInfo, it will get reset every attempt.

stats/opentelemetry/client_tracing.go
```diff
@@ -121,6 +122,10 @@ func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
...
+	}
+	atomic.AddUint32(&ai.previousRPCAttempts, 1)
```
purnesh42H, line 128: you can maintain single attribute. Either previousRPCAttempts or explicitRetryCount. Why do we need both?

stats/opentelemetry/trace.go
```diff
@@ -41,13 +41,17 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+			atomic.AddUint32(&ai.explicitRetryCount, ^uint32(0))
+		}
 		// Note: Go always added Client and FailFast attributes even though they are not
 		// defined by the OpenCensus gRPC spec. Thus, they are unimportant for
 		// correctness.
 		span.SetAttributes(
 			attribute.Bool("Client", rs.Client),
 			attribute.Bool("FailFast", rs.FailFast),
 			attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts)),
+			attribute.Int64("retry-attempts", int64(atomic.LoadUint32(&ai.explicitRetryCount))),
```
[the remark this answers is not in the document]
purnesh42H: See https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#tracing-information. We should be confined to attributes defined here.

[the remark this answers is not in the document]
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
vinothkumarr227: Thanks. I've removed explicitRetryCount and now use only previousRPCAttempts to track retries.

stats/opentelemetry/client_tracing.go
```diff
@@ -118,9 +119,20 @@ func (h *clientTracingHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo
...
 func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
 
+type retryCountKey struct{}
+
 // TagRPC implements per RPC attempt context management for traces.
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	var counter *int32
+	if val := ctx.Value(retryCountKey{}); val != nil {
+		counter = val.(*int32)
```
purnesh42H, line 129: this again seems wrong. We should create the retry count key in the interceptor call which is per RPC. TagRPC is for per attempt. Here we should only retrieve the current retry count value (if needed)

stats/opentelemetry/trace.go
```diff
@@ -41,13 +41,20 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
 
 	switch rs := rs.(type) {
 	case *stats.Begin:
+		retryCount := ai.previousRPCAttempts
+		if !rs.IsTransparentRetryAttempt {
+			if val := ai.ctx.Value(retryCountKey{}); val != nil {
+				// Atomic increment and get new value
+				retryCount = uint32(atomic.AddInt32(val.(*int32), 1))
```
purnesh42H, line 48: this value needs to be set back to context somewhere? otherwise we have lost the count.

stats/opentelemetry/trace.go
```diff
@@ -41,13 +41,20 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+			}
+		}
 		// Note: Go always added Client and FailFast attributes even though they are not
 		// defined by the OpenCensus gRPC spec. Thus, they are unimportant for
 		// correctness.
 		span.SetAttributes(
 			attribute.Bool("Client", rs.Client),
 			attribute.Bool("FailFast", rs.FailFast),
-			attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts)),
+			attribute.Int64("previous-rpc-attempts", int64(retryCount)),
```
purnesh42H, line 57: Please note as per https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#tracing-information, previousRPCAttempts is the count of retries before current, so we should set here the existing retryCount value (before incrementing it). However, we should increment retry count after and set back to context for future.

stats/opentelemetry/e2e_test.go
```diff
@@ -928,7 +928,7 @@ func (s) TestMetricsAndTracesOptionEnabled(t *testing.T) {
 				},
 				{
 					Key:   "previous-rpc-attempts",
-					Value: attribute.IntValue(0),
+					Value: attribute.IntValue(1),
```
purnesh42H, line 931: It should be correctly 0 because it counts previous rpc attempts which means before current so for the first attempt it should be 0.

purnesh42H (commented):
I remember we had separate tests for retries. This change should only affect that test. This change shouldn't affect tests which are doing only single attempt.

stats/opentelemetry/trace.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
```diff
@@ -48,6 +48,8 @@ func (h *clientTracingHandler) initializeTraces() {
 }
 
 func (h *clientTracingHandler) unaryInterceptor(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	ci := &callInfo{numRetries: 0}
```
purnesh42H, line 51: both these operations can be done in getOrCreateCallInfo? That we set call info in ctx only once

stats/opentelemetry/client_tracing.go
```diff
@@ -58,6 +60,8 @@ func (h *clientTracingHandler) unaryInterceptor(ctx context.Context, method stri
 }
 
 func (h *clientTracingHandler) streamInterceptor(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	ci := &callInfo{numRetries: 0}
```
purnesh42H, line 63: same here about moving both to getOrCreateCallInfo

stats/opentelemetry/opentelemetry.go
```diff
@@ -179,6 +179,8 @@ type callInfo struct {
 	// nameResolutionEventAdded is set when the resolver delay trace event
 	// is added. Prevents duplicate events, since it is reported per-attempt.
 	nameResolutionEventAdded atomic.Bool
+	// numRetries holds the count of non-transparent retry attempts.
+	numRetries int32
```
purnesh42H, line 183: better to declare it as atomic and do regular operation for it otherwise you need to explicitly mention that this field need to be accessed atomically. Former is better IMO.

stats/opentelemetry/trace.go
```diff
@@ -41,24 +41,22 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+			attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts)),
 			attribute.Bool("transparent-retry", rs.IsTransparentRetryAttempt),
 		)
-		// increment previous rpc attempts applicable for next attempt
-		atomic.AddUint32(&ai.previousRPCAttempts, 1)
+		// Increment retry count for the next attempt if not a transparent
+		// retry.
+		if !rs.IsTransparentRetryAttempt {
+			if ci, ok := getRetryCount(ai.ctx); ok {
+				atomic.AddInt32(&ci.numRetries, 1)
```
purnesh42H, line 57: As called out before, we need to set this back to context. 

Also, it might be more clear if we name this field as well to previousRPCAttempts because at the time of retrieving its not accounting the current retry.

stats/opentelemetry/opentelemetry.go
```diff
@@ -213,6 +215,16 @@ func getRPCInfo(ctx context.Context) *rpcInfo {
 	return ri
 }
 
+func setRetryCount(ctx context.Context, ci *callInfo) context.Context {
+	return context.WithValue(ctx, retryCountKey{}, ci)
+}
+
+// getRetryCount retrieves the retry count tracking struct from the context.
```
purnesh42H, line 222: no need of docstring since the function is simple enough to understand.

stats/opentelemetry/opentelemetry.go
```diff
@@ -213,6 +215,16 @@ func getRPCInfo(ctx context.Context) *rpcInfo {
...
+func getRetryCount(ctx context.Context) (*callInfo, bool) {
```
purnesh42H, line 223: nit: its preferred to avoid get prefix in these situations https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#getters

stats/opentelemetry/e2e_test.go
```diff
@@ -1708,6 +1710,32 @@ func verifyTrace(t *testing.T, spans tracetest.SpanStubs, want traceSpanInfo) {
 	}
 }
 
+func verifyPreviousRPCAttempts(t *testing.T, spans tracetest.SpanStubs) {
```
purnesh42H, line 1713: why this needs to be a separate function? We are just verifying the previous-rpc-attempts which is already being compared. We shouldn't need this.

purnesh42H (commented):
@vinothkumarr227 have you tested this to ensure its working correctly? I was under impression that TestTraceSpan_WithRetriesAndNameResolutionDelay will need changes for expected values. I think its not working as expected because you are not setting the count back to ctx after incrementing.

stats/opentelemetry/client_tracing.go
vinothkumarr227: Done

stats/opentelemetry/opentelemetry.go
vinothkumarr227: Done

stats/opentelemetry/trace.go
vinothkumarr227: Thanks for the clarification. I've updated the code accordingly

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/client_metrics.go
```diff
@@ -81,6 +81,7 @@ func getOrCreateCallInfo(ctx context.Context, cc *grpc.ClientConn, method string
 		}
 		ctx = setCallInfo(ctx, ci)
 	}
+	ctx = setRetryCount(ctx, ci)
```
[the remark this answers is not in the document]
vinothkumarr227: Done

stats/opentelemetry/opentelemetry.go
```diff
@@ -179,8 +179,8 @@ type callInfo struct {
 	// nameResolutionEventAdded is set when the resolver delay trace event
 	// is added. Prevents duplicate events, since it is reported per-attempt.
 	nameResolutionEventAdded atomic.Bool
-	// numRetries holds the count of non-transparent retry attempts.
-	numRetries int32
+	// previousRPCAttempts holds the count of non-transparent retry attempts.
```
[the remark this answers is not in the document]
vinothkumarr227: Done

stats/opentelemetry/opentelemetry.go
```diff
@@ -219,8 +219,7 @@ func setRetryCount(ctx context.Context, ci *callInfo) context.Context {
 	return context.WithValue(ctx, retryCountKey{}, ci)
 }
 
-// getRetryCount retrieves the retry count tracking struct from the context.
-func getRetryCount(ctx context.Context) (*callInfo, bool) {
+func retryCount(ctx context.Context) (*callInfo, bool) {
```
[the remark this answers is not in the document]
vinothkumarr227: Done

stats/opentelemetry/trace.go
```diff
@@ -53,8 +51,9 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
 		// Increment retry count for the next attempt if not a transparent
 		// retry.
 		if !rs.IsTransparentRetryAttempt {
-			if ci, ok := getRetryCount(ai.ctx); ok {
-				atomic.AddInt32(&ci.numRetries, 1)
+			if ci, ok := retryCount(ai.ctx); ok {
+				ci.previousRPCAttempts.Add(1)
+				ai.ctx = setRetryCount(ai.ctx, ci)
```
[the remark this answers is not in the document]
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
```diff
@@ -118,9 +118,15 @@ func (h *clientTracingHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo
 // HandleConn exists to satisfy stats.Handler for tracing.
 func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
 
+type retryCountKey struct{}
```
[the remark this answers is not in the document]
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
```diff
@@ -118,9 +118,15 @@ func (h *clientTracingHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo
...
+
 // TagRPC implements per RPC attempt context management for traces.
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	if ci, ok := retryCount(ctx); ok {
+		ai.previousRPCAttempts = uint32(ci.previousRPCAttempts.Load())
+	}
+	ai.ctx = ctx
```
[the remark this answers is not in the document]
vinothkumarr227: We need to explicitly set ai.ctx = ctx here because populateSpan later uses ai.ctx to retrieve and update callInfo. If we skip this assignment, ai.ctx remains nil, leading to a nil pointer.

stats/opentelemetry/e2e_test.go
```diff
@@ -1677,31 +1678,58 @@ func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
...
 func verifyTrace(t *testing.T, spans tracetest.SpanStubs, want traceSpanInfo) {
+	t.Helper()
 	match := false
 	for _, span := range spans {
 		if span.Name == want.name && span.SpanKind.String() == want.spanKind {
 			match = true
-			if diff := cmp.Diff(want.events, span.Events, cmpopts.IgnoreFields(trace.Event{}, "Time")); diff != "" {
-				t.Errorf("Span event mismatch for %q (kind: %s) (-want +got):\n%s",
-					want.name, want.spanKind, diff)
+			if len(want.events) > 0 {
```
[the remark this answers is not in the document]
vinothkumarr227: I’ve added new test cases for TestRetrySpans_UnaryCallAttributes

stats/opentelemetry/e2e_test.go
```diff
@@ -1677,31 +1678,58 @@ func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
 				t.Fatalf("%s call failed: %v", tt.name, err)
 			}
 
-			wantSpanInfo := traceSpanInfo{
+			methodName := strings.TrimPrefix(tt.spanName, "Sent.")
```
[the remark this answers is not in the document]
vinothkumarr227: I’ve added new test cases for TestRetrySpans_UnaryCallAttributes

stats/opentelemetry/e2e_test.go
```diff
@@ -1817,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
 	}
 	validateTraces(t, spans, wantSpanInfos)
 }
+
+// TestRetrySpans_UnaryCallAttributes checks that OpenTelemetry spans
+// correctly record retry attempts during a retried unary RPC call.
```
purnesh42H: can you explain more what is the test setup, how the retries are happening and what is the expectation? Specifically the part that subsequent attempts will see higher number for "previous-rpc-attempts" than the previous attempt. Also, name this test "TestPreviousRPCAttempts"

stats/opentelemetry/client_tracing.go
```diff
@@ -121,6 +121,12 @@ func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
 // TagRPC implements per RPC attempt context management for traces.
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	if ci := getCallInfo(ctx); ci != nil {
+		ai.previousRPCAttempts = uint32(ci.previousRPCAttempts.Load())
+	}
+	if ai.ctx == nil {
```
purnesh42H: do we have to check for nil? Context won't change so we can directly assign. We should always work on latest context data.

stats/opentelemetry/e2e_test.go
```diff
@@ -1789,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
...
+func (s) TestRetrySpans_UnaryCallAttributes(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	ss := &stubserver.StubServer{
```
purnesh42H: can we write test for streaming RPC as well? Similar to how its done in NameResolutionDelay test

stats/opentelemetry/e2e_test.go
```diff
@@ -1789,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
...
+			attributes: []attribute.KeyValue{
+				attribute.Bool("Client", true),
+				attribute.Bool("FailFast", true),
+				attribute.Int("previous-rpc-attempts", 0),
+				attribute.Bool("transparent-retry", false),
+			},
+			events: nil,
+		},
+		{
+			name:     "Recv.grpc.testing.TestService.UnaryCall",
```
purnesh42H: nit: let's keep server and client metrics together. Like this can be directly under 1st wantSpan of "SpanKindServer.String()".

stats/opentelemetry/e2e_test.go
```diff
@@ -1789,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
...
+			attributes: []attribute.KeyValue{
+				attribute.Bool("Client", false),
+				attribute.Bool("FailFast", false),
+				attribute.Int("previous-rpc-attempts", 0),
+				attribute.Bool("transparent-retry", false),
+			},
+			events: nil,
+		},
+		{
+			name:     "Attempt.grpc.testing.TestService.UnaryCall",
```
purnesh42H: let's put an inline comment that this is a 1st RPC attempt

stats/opentelemetry/e2e_test.go
```diff
@@ -1789,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
...
```
purnesh42H: comment as 2nd RPC attempt (first retry)

stats/opentelemetry/e2e_test.go
```diff
@@ -1789,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
...
+	}
+
+	wantSpanInfos := []traceSpanInfo{
+		{
+			name:     "Recv.grpc.testing.TestService.UnaryCall",
+			spanKind: oteltrace.SpanKindServer.String(),
+			attributes: []attribute.KeyValue{
+				attribute.Bool("Client", false),
+				attribute.Bool("FailFast", false),
+				attribute.Int("previous-rpc-attempts", 0),
```
purnesh42H: As per A72 https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#tracing-information, previous-rpc-attempts and transparent-retry are only specified at the client side in attempt span. I think we should add them conditionally in trace.go  based on whether `rs.Client` is true or false. They will always default to false and 0 at server side because there is no notion of retry/attempt when being a receiver (server).

stats/opentelemetry/e2e_test.go
```diff
@@ -1789,3 +1789,163 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
...
+			events:     nil,
+		},
+	}
+
+	spans, err := waitForTraceSpans(ctx, exporter, wantSpanInfos)
+	if err != nil {
+		t.Fatalf("Span collection failed: %v", err)
+	}
+
+	wantSpanInfosMap := make(map[traceSpanInfoMapKey]traceSpanInfo)
```
purnesh42H: why can't we re-use validateTraces? It already does what you are doing here.

purnesh42H: Also, i can't comment above. Looks like there was another `verifyTrace` helper created which is only verifying the events. It is wrong. We should verify everything and make sure expected span is created as needed. Can you see if we can remove that and use validateTraces there as well?

stats/opentelemetry/e2e_test.go
vinothkumarr227: We no longer need a separate TestPreviousRPCAttempts. I've updated the existing TestTraceSpan_WithRetriesAndNameResolutionDelay to cover both the previous-rpc-attempts field and the name resolution delay events

stats/opentelemetry/client_tracing.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Yeah, makes sense. I've now updated the existing TestTraceSpan_WithRetriesAndNameResolutionDelay to cover both the previous-rpc-attempts field and the name resolution delay events, so  verifyTrace isn't needed anymore.

stats/opentelemetry/e2e_test.go
```diff
@@ -284,35 +285,53 @@ func validateTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []tra
...
-		if !ok {
-			t.Errorf("Unexpected span: %v", span)
+		if matchedIndex == -1 {
+			t.Errorf("Unexpected span: %q (%s)", span.Name, span.SpanKind)
 			continue
 		}
+		want := wantSpanInfos[matchedIndex]
+		// Check that the attempt span has the correct status.
+		if want.status != otelcodes.Unset {
+			got, want := span.Status.Code, want.status
```
purnesh42H: nit: got, want can be assigned in the same as the if check

stats/opentelemetry/server_tracing.go
```diff
@@ -41,6 +41,7 @@ func (h *serverTracingHandler) initializeTraces() {
 // TagRPC implements per RPC attempt context management for traces.
 func (h *serverTracingHandler) TagRPC(ctx context.Context, _ *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	ai.ctx = ctx
 	ctx, ai = h.traceTagRPC(ctx, ai)
```
purnesh42H: this is probably not the updated ctx as that's being updated in traceTagRPC. Either update the ai.ctx in traceTagRPC at the end or after it.

stats/opentelemetry/client_tracing.go
```diff
@@ -121,6 +121,10 @@ func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
 // TagRPC implements per RPC attempt context management for traces.
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	if ci := getCallInfo(ctx); ci != nil {
```
purnesh42H: can ci be ever nil at the attempt stage? I don't we need to check for nil here. If its nil, we shouldn't proceed further and return early logging an error.

stats/opentelemetry/trace.go
```diff
@@ -44,14 +42,25 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+			attrs = append(attrs,
+				attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts)),
+				attribute.Bool("transparent-retry", rs.IsTransparentRetryAttempt),
+			)
+		}
+		span.SetAttributes(attrs...)
+		// Increment retry count for the next attempt if not a transparent
+		// retry.
+		if !rs.IsTransparentRetryAttempt {
+			if ci := getCallInfo(ai.ctx); ci != nil {
```
purnesh42H: same here. I don't think ci should be nil at this stage.

stats/opentelemetry/e2e_test.go
```diff
@@ -1576,7 +1547,150 @@ func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
...
+							{
+								Name: "Outbound message",
+								Attributes: []attribute.KeyValue{
+									attribute.Int("sequence-number", 0),
+									attribute.Int("message-size", 0),
+								},
+							},
+						},
+					},
+					// RPC attempt #3
```
purnesh42H: why RPC attempt 3 doesn't have a corresponding server span?

stats/opentelemetry/e2e_test.go
```diff
@@ -1620,7 +1734,119 @@ func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
...
+					{
+						name:       "Sent.grpc.testing.TestService.FullDuplexCall",
+						spanKind:   oteltrace.SpanKindClient.String(),
+						status:     otelcodes.Ok,
+						attributes: nil,
+						events: []trace.Event{
+							{Name: delayedResolutionEventName},
+						},
+					},
+					// RPC attempt #3
```
purnesh42H: same question about corresponding server span

stats/opentelemetry/e2e_test.go
vinothkumarr227: Yeah, I missed it earlier. I've added it now.

stats/opentelemetry/trace.go
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
vinothkumarr227: Done

stats/opentelemetry/server_tracing.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
```diff
@@ -121,6 +121,13 @@ func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
...
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
+	ci := getCallInfo(ctx)
+	if ci == nil {
+		logger.Error("context passed into client side stats handler (TagRPC) has no call info")
+		return ctx
+	}
+	ai.previousRPCAttempts = uint32(ci.previousRPCAttempts.Load())
+	ai.ctx = ctx
 	ctx, ai = h.traceTagRPC(ctx, ai, info.NameResolutionDelay)
```
arjan-bal: `traceTagRPC` mutates the attempt info pointer that is passed to it, making the returned attempt info redundant. In the case of context, returning a new context makes sense since contexts are immutable. Seeing an attempt info being returned may lead readers to (incorrectly) assume the passed attemptInfo is not mutated. Sometimes arguments are mutated and returned for chaining function calls, but here that's not possible since there are two values.

Can you please remove the returned attempt info and document that the passed argument is mutated in-place?

stats/opentelemetry/opentelemetry.go
```diff
@@ -179,6 +179,9 @@ type callInfo struct {
 	// nameResolutionEventAdded is set when the resolver delay trace event
 	// is added. Prevents duplicate events, since it is reported per-attempt.
 	nameResolutionEventAdded atomic.Bool
+	// previousRPCAttempts holds the count of RPC attempts that have happened
+	// before current attempt. Transparent retries are excluded.
+	previousRPCAttempts atomic.Int32
```
arjan-bal: Can you instead make this a pointer, i.e. `*atomic.Int32`? Then we can copy the pointer into `attemptInfo`. Since both will be pointing to the same underlying Int32, you will not need to update `callInfo` at the end of each attempt.

stats/opentelemetry/client_tracing.go
vinothkumarr227: Thanks for the feedback! I’ve removed the returned attemptInfo from traceTagRPC and updated the documentation accordingly.

stats/opentelemetry/opentelemetry.go
vinothkumarr227: Done

stats/opentelemetry/client_metrics.go
```diff
@@ -79,6 +79,8 @@ func getOrCreateCallInfo(ctx context.Context, cc *grpc.ClientConn, method string
 			target: cc.CanonicalTarget(),
 			method: determineMethod(method, opts...),
 		}
+		ci.previousRPCAttempts = new(atomic.Int32)
+		ci.previousRPCAttempts.Store(0)
```
arjan-bal: No need to store 0, this is the default value.

stats/opentelemetry/client_tracing.go
```diff
@@ -121,7 +124,14 @@ func (h *clientTracingHandler) HandleConn(context.Context, stats.ConnStats) {}
 // TagRPC implements per RPC attempt context management for traces.
 func (h *clientTracingHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	ctx, ai := getOrCreateRPCAttemptInfo(ctx)
-	ctx, ai = h.traceTagRPC(ctx, ai, info.NameResolutionDelay)
+	ci := getCallInfo(ctx)
+	if ci == nil {
+		logger.Error("context passed into client side stats handler (TagRPC) has no call info")
+		return ctx
+	}
+	ai.previousRPCAttempts = uint32(ci.previousRPCAttempts.Load())
```
arjan-bal: `ai.previousRPCAttempts` should also be a `*atomic.Int32` and we should assign it `ci.previousRPCAttempts`. This way if we increment `ai.previousRPCAttempt`, `ci.previousRPCAttempts` gets incremented automatically, since it points to the same data in memory.

stats/opentelemetry/client_metrics.go
vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
vinothkumarr227: Done

stats/opentelemetry/trace.go
```diff
@@ -44,14 +42,27 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+		attrs := []attribute.KeyValue{
 			attribute.Bool("Client", rs.Client),
 			attribute.Bool("FailFast", rs.FailFast),
-			attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts)),
-			attribute.Bool("transparent-retry", rs.IsTransparentRetryAttempt),
-		)
-		// increment previous rpc attempts applicable for next attempt
-		atomic.AddUint32(&ai.previousRPCAttempts, 1)
+		}
+		if rs.Client {
```
arjan-bal: Is there a reason to stop setting the retry attributes in the server side?

stats/opentelemetry/trace.go
```diff
@@ -44,14 +42,27 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+			attrs = append(attrs,
+				attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts.Load())),
+				attribute.Bool("transparent-retry", rs.IsTransparentRetryAttempt),
+			)
+		}
+		span.SetAttributes(attrs...)
+		// Increment retry count for the next attempt if not a transparent
+		// retry.
+		if !rs.IsTransparentRetryAttempt && rs.Client {
+			ci := getCallInfo(ai.ctx)
```
arjan-bal: We don't need to get `ci` again. The reason to use the pointer is that we can increment `ai.previousRPCAttempts` and the change gets reflected in `ci.previousRPCAttempts` automatically.

stats/opentelemetry/trace.go
```diff
@@ -44,14 +42,27 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+		if rs.Client {
+			attrs = append(attrs,
+				attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts.Load())),
+				attribute.Bool("transparent-retry", rs.IsTransparentRetryAttempt),
+			)
+		}
+		span.SetAttributes(attrs...)
+		// Increment retry count for the next attempt if not a transparent
+		// retry.
+		if !rs.IsTransparentRetryAttempt && rs.Client {
```
arjan-bal: Why do we want to increment attempts for only non-transparent retries? Why did we stop incrementing the counter for transparent retries?

stats/opentelemetry/trace.go
vinothkumarr227: Thanks for the feedback! I’ve removed the redundant getCallInfo call.

stats/opentelemetry/trace.go
vinothkumarr227: Yes — per [gRPC OpenTelemetry tracing proposal (A72)](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#tracing-information.), retry information is tracked only on the client side. On the server side, there is a single span per RPC (Recv.<service>.<method>), and no per-attempt retries are recorded. Therefore, we do not set retry-related attributes like previous-rpc-attempts or transparent-retry on server spans.

stats/opentelemetry/trace.go
vinothkumarr227: We only increment the attempts count for non-transparent retries because, per the [gRPC OpenTelemetry tracing spec (A72)](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#tracing-information), the previous-rpc-attempts.

stats/opentelemetry/opentelemetry.go
```diff
@@ -241,7 +244,8 @@ type attemptInfo struct {
 	// associated call.
 	countSentMsg        uint32
 	countRecvMsg        uint32
-	previousRPCAttempts uint32
+	previousRPCAttempts *atomic.Int32
+	ctx                 context.Context
```
arjan-bal: We should not store context as a field in structs. Instead, the context should be propagated as the first param in function calls. See https://google.github.io/styleguide/go/decisions#contexts

stats/opentelemetry/trace.go
arjan-bal: Can you please raise a separate PR to remove the the client specific attributes, along with unit tests? This PR should be focussed on a single issue.

stats/opentelemetry/opentelemetry.go
arjan-bal: This new field also seems unused. It's only being set, never read.

stats/opentelemetry/trace.go
arjan-bal: Acknowledged. Can we remove the check for `rs.Client`? I think the server should not be doing any kind of retries, incrementing the counter should be a no-op.

stats/opentelemetry/trace.go
vinothkumarr227: Acknowledged. I've reverted the changes for now, and I’ll create a separate PR later to remove the client-specific attributes along with unit tests, as suggested

stats/opentelemetry/trace.go
vinothkumarr227: Done

stats/opentelemetry/opentelemetry.go
vinothkumarr227: Done

stats/opentelemetry/opentelemetry.go
```diff
@@ -241,7 +244,7 @@ type attemptInfo struct {
 	// associated call.
 	countSentMsg        uint32
 	countRecvMsg        uint32
-	previousRPCAttempts uint32
+	previousRPCAttempts *atomic.Int32
```
arjan-bal: nit: Please use `atomic.Uint32` since we're replacing a uint32 here.

stats/opentelemetry/e2e_test.go
```diff
@@ -284,35 +285,51 @@ func validateTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []tra
...
-		// span kind to compare.
-		want, ok := wantSpanInfosMap[traceSpanInfoMapKey{spanName: span.Name, spanKind: span.SpanKind.String()}]
-		if !ok {
-			t.Errorf("Unexpected span: %v", span)
+		if matchedIndex == -1 {
+			t.Errorf("Unexpected span: %q (%s)", span.Name, span.SpanKind)
 			continue
 		}
+		// Check that the attempt span has the correct status.
+		if got, want := span.Status.Code, wantSpanInfos[matchedIndex].status; want != otelcodes.Unset {
```
arjan-bal: Please set `otelcodes.Ok` in existing tests instead of converting `otelcodes.Unset` to `otelcodes.Ok` here. Test should have minimal conditional logic, they should have direct comparisons to make it easier to reason about test correctness.

stats/opentelemetry/e2e_test.go
```diff
@@ -284,35 +285,51 @@ func validateTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []tra
 		key := traceSpanInfoMapKey{spanName: info.name, spanKind: info.spanKind}
 		wantSpanInfosMap[key] = info
 	}
-
+	used := make([]bool, len(wantSpanInfos))
```
arjan-bal: Can you describe what you're trying to do here? It seems more complicated than necessary. I think there must be an easier way of doing this validation using `cmp`.

stats/opentelemetry/e2e_test.go
```diff
@@ -1753,12 +2085,6 @@ func (s) TestStreamingRPC_TraceSequenceNumbers(t *testing.T) {
 	}
 
 	wantSpanInfos := []traceSpanInfo{
-		{
```
arjan-bal, line 1756: This seems like a no-op change, please revert.

stats/opentelemetry/e2e_test.go
```diff
@@ -1550,10 +1567,10 @@ const delayedResolutionEventName = "Delayed name resolution complete"
...
 func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
 	tests := []struct {
-		name      string
-		setupStub func() *stubserver.StubServer
-		doCall    func(context.Context, testgrpc.TestServiceClient) error
-		spanName  string
+		name            string
+		setupStub       func() *stubserver.StubServer
+		doCall          func(context.Context, testgrpc.TestServiceClient) error
+		wantSpanInfosFn func() []traceSpanInfo
```
arjan-bal: Why is this a closure with 0 params instead of `[]traceSpanInfo`?

stats/opentelemetry/e2e_test.go
```diff
@@ -1677,34 +2016,27 @@ func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
...
-		if span.Name == want.name && span.SpanKind.String() == want.spanKind {
-			match = true
-			if diff := cmp.Diff(want.events, span.Events, cmpopts.IgnoreFields(trace.Event{}, "Time")); diff != "" {
-				t.Errorf("Span event mismatch for %q (kind: %s) (-want +got):\n%s",
-					want.name, want.spanKind, diff)
+			const delayedLBPickComplete = "Delayed LB pick complete"
+			// Removes events named "Delayed LB pick complete" from the slice.
+			// This is a temporary workaround to ignore this event during tests
+			// as the new Load Balancer (LB) policy and picker may cause it to
+			// appear inconsistently or unexpectedly.
```
arjan-bal: From this comment, it's not clear when the event is present and when it's absent. It's also not clear what the permanent fix is. Can you please elaborate?

vinothkumarr227: The "Delayed LB pick complete" event is triggered when stats.PickerUpdated occurs after a name resolution delay. However, this event is non-deterministic due to its dependence on timing—especially in scenarios involving retries or delayed name resolution—so it may or may not appear consistently.
This event originates from the updated Load Balancer (LB) and picker behavior introduced as part of the pick_first improvements.
In our TestTraceSpan_WithRetriesAndNameResolutionDelay test, we do not rely on this event. After discussing with purnesh, we agreed to filter it out during test validation to prevent false failures. Example Build failure (https://github.com/grpc/grpc-go/actions/runs/16137567572/job/45537142810?pr=8342)

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: We use used := make([]bool, len(wantSpanInfos)) to ensure a one-to-one mapping between actual and expected spans, even when multiple spans share the same name and kind. In tests like TestTraceSpan_WithRetriesAndNameResolutionDelay, multiple spans with identical name and kind (e.g., repeated "Attempt.grpc.testing.TestService." or "Recv.") can be generated due to retries and name resolution delays.
The used[] array helps track which expected spans have already been matched, ensuring each expected span is only matched once, regardless of order or duplication. This prevents multiple actual spans from incorrectly matching the same expected span

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Thanks,  I've updated the code to address your comments and replaced the closure with a direct []traceSpanInfo as suggested.

stats/opentelemetry/opentelemetry.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
arjan-bal: To handle ordering, you can use [SortSlices](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices).

For duplicate spans, I think the number of duplicates should be deterministic. If so, we should have the exact number of spans in the expected values.

stats/opentelemetry/e2e_test.go
arjan-bal: > this event is non-deterministic due to its dependence on timing

I don't think picker generation is non-deterministic. The picker is updates when the channel starts connecting and again when the channel becomes ready. The channel begins connecting when the first RPC is made. I would expect the `Delayed LB pick complete` complete event should always be present for the first attempt. Is this not the case?

stats/opentelemetry/e2e_test.go
vinothkumarr227: Thanks for the suggestion! I've updated the code to use SortSlices to handle ordering as recommended.

stats/opentelemetry/e2e_test.go
vinothkumarr227: You're absolutely right that picker updates themselves are deterministic — they occur when the channel starts connecting and again when it becomes ready. From the perspective of the gRPC channel state machine, we would indeed expect a "Delayed LB pick complete" event to be emitted for the first RPC attempt after a name resolution delay. However, we've observed that the emission and recording of this event in trace spans can be affected by timing at the instrumentation level. , the event may: Be emitted multiple times (e.g., due to retry logic or repeated PickerUpdated calls), Be delayed, potentially falling outside the expected trace window, or In rare cases, be missed entirely due to race conditions between channel state transitions and span collection.

arjan-bal: I had a look at the code that emits this event: https://github.com/grpc/grpc-go/blob/bed551a435ae957717c3680cc6049dd85c5c4a0f/picker_wrapper.go#L139-L151

It emits the event every time an LB policy sends queuing picker. The old pickfirst sends one queuing picker as soon as [it gets the first resolver update](https://github.com/grpc/grpc-go/blob/bed551a435ae957717c3680cc6049dd85c5c4a0f/balancer/pickfirst/pickfirst.go#L200-L203) and then a second queuing picker [when the subchannel transitions to `connecting`](https://github.com/grpc/grpc-go/blob/bed551a435ae957717c3680cc6049dd85c5c4a0f/balancer/pickfirst/pickfirst.go#L235-L243). This results in double counting the event.

The new pickfirst [de-duplicates the picker updates](https://github.com/grpc/grpc-go/blob/bed551a435ae957717c3680cc6049dd85c5c4a0f/balancer/pickfirst/pickfirstleaf/pickfirstleaf.go#L788-L799) to avoid this.

This means that the test is NOT flaky. It will always fail when you set the env var `GRPC_EXPERIMENTAL_ENABLE_NEW_PICK_FIRST=true`.

For now, please check `envconfig.NewPickFirstEnabled` and add an extra event in the expected output. Please file a bug for this issue and add a TODO in the test code referencing this issue to remove the extra event.

vinothkumarr227: @arjan-bal Thanks for the clarification. the duplicate "Delayed LB pick complete" event occurs with the old pick_first (i.e., when envconfig.NewPickFirstEnabled == false). It emits one picker on the resolver update and another on CONNECTING.

vinothkumarr227: @arjan-bal I've updated the test to add an extra Delayed LB pick complete event to the expected output. I've also filed a bug and added a TODO comment in the test referencing it. Kindly review it

vinothkumarr227: @arjan-bal I’ve merged the latest master and removed the TODO comment and the extra event-handling code from the test. Kindly review it

stats/opentelemetry/e2e_test.go
```diff
@@ -276,27 +271,35 @@ func validateTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []tra
...
+	sort.Slice(sortedSpans, func(i, j int) bool {
+		if sortedSpans[i].Name == sortedSpans[j].Name {
+			return sortedSpans[i].SpanKind.String() < sortedSpans[j].SpanKind.String()
+		}
+		return sortedSpans[i].Name < sortedSpans[j].Name
+	})
 
 	// Compare retrieved spans with expected spans.
-	for _, span := range spans {
+	for i := range sortedSpans {
```
arjan-bal: Can we run `cmp.Diff` on `wantSpanInfosMap` and `sortedSpans` directly instead?

To do this you would need to convert `sortedSpans` to `SpanStub` once. Then you can replace the calls to `sort.Slice` with `cmpopts.SortSlices` options passed to `cmp.Diff`.

stats/opentelemetry/e2e_test.go
```diff
@@ -1550,10 +1571,10 @@ const delayedResolutionEventName = "Delayed name resolution complete"
...
 func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
 	tests := []struct {
-		name      string
-		setupStub func() *stubserver.StubServer
-		doCall    func(context.Context, testgrpc.TestServiceClient) error
-		spanName  string
+		name            string
+		setupStub       func() *stubserver.StubServer
+		doCall          func(context.Context, testgrpc.TestServiceClient) error
+		wantSpanInfosFn []traceSpanInfo
```
arjan-bal: Please rename the field to remove the `Fn` suffix. It's not a function anymore.

vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

stats/opentelemetry/e2e_test.go
```diff
@@ -276,46 +270,45 @@ func validateTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []tra
...
-		if diff := cmp.Diff(want.attributes, span.Attributes, attributesSort, attributesValueComparable); diff != "" {
-			t.Errorf("Attributes mismatch for span %s (-want +got):\n%s", span.Name, diff)
-		}
-		// events
-		if diff := cmp.Diff(want.events, span.Events, attributesSort, attributesValueComparable, eventsTimeIgnore); diff != "" {
-			t.Errorf("Events mismatch for span %s (-want +got):\n%s", span.Name, diff)
+	// Converts expected span information into SpanStub format,
+	// enabling structured comparison with actual spans using cmp.Diff.
+	wantSpans := make([]tracetest.SpanStub, len(wantSpanInfos))
+	for i, info := range wantSpanInfos {
```
arjan-bal: Instead of converting both expected and got spans to `SpandStub`, can we only convert the got spans to `traceSpanInfo` for comparison? It seems to have all the 5 fields that are being set in `SpanStub` below.

stats/opentelemetry/e2e_test.go
```diff
@@ -276,46 +270,45 @@ func validateTraces(t *testing.T, spans tracetest.SpanStubs, wantSpanInfos []tra
...
-		}
-
-		// Retrieve the corresponding expected span info based on span name and
-		// span kind to compare.
-		want, ok := wantSpanInfosMap[traceSpanInfoMapKey{spanName: span.Name, spanKind: span.SpanKind.String()}]
-		if !ok {
-			t.Errorf("Unexpected span: %v", span)
-			continue
+	// Converts collected spans to SpanStub for simplified cmp.Diff comparison.
+	actualSpanStubs := make([]tracetest.SpanStub, len(spans))
```
arjan-bal: `spans` is already `[]tracetest.SpanStub`. Why is this conversion required? Is to to unset some fields?

stats/opentelemetry/e2e_test.go
vinothkumarr227: Thanks for the suggestions! Updated the test to convert only actual spans to traceSpanInfo.

stats/opentelemetry/e2e_test.go
vinothkumarr227: Removed the unnecessary conversion from []tracetest.SpanStub

stats/opentelemetry/e2e_test.go
```diff
@@ -1548,12 +1547,12 @@ const delayedResolutionEventName = "Delayed name resolution complete"
 // TestTraceSpan_WithRetriesAndNameResolutionDelay verifies that
 // "Delayed name resolution complete" event is recorded in the call trace span
 // only once if any of the retry attempt encountered a delay in name resolution
-func (s) TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
+func TestTraceSpan_WithRetriesAndNameResolutionDelay(t *testing.T) {
```
arjan-bal: Please add the `(s)` back.

arjan-bal (approved):
LGTM, adding a second reviewer. Please address the minor comment.

stats/opentelemetry/e2e_test.go
vinothkumarr227: Done

arjan-bal:
@vinothkumarr227 there is a failure in the updated test reported by `codecov`: https://github.com/grpc/grpc-go/actions/runs/16616024027/job/47008766111?pr=8342

PTAL and ensure the test is not flaky.

vinothkumarr227:
> @vinothkumarr227 there is a failure in the updated test reported by `codecov`: https://github.com/grpc/grpc-go/actions/runs/16616024027/job/47008766111?pr=8342
> 
> PTAL and ensure the test is not flaky.

Sure, I’ll check it out.

vinothkumarr227:
> @vinothkumarr227 there is a failure in the updated test reported by `codecov`: https://github.com/grpc/grpc-go/actions/runs/16616024027/job/47008766111?pr=8342
> 
> PTAL and ensure the test is not flaky.
Hi @arjan-bal , I tested the test locally and everything passed consistently. I’ve now merged the latest changes from master, and all test cases are passing.

stats/opentelemetry/trace.go
```diff
@@ -44,14 +42,28 @@ func populateSpan(rs stats.RPCStats, ai *attemptInfo) {
...
+		// prevAttempts defaults to 0 to avoid a nil pointer dereference.
+		previousRPCAttempts := int64(0)
+		if ai.previousRPCAttempts != nil {
+			previousRPCAttempts = int64(ai.previousRPCAttempts.Load())
+		}
 		span.SetAttributes(
 			attribute.Bool("Client", rs.Client),
 			attribute.Bool("FailFast", rs.FailFast),
-			attribute.Int64("previous-rpc-attempts", int64(ai.previousRPCAttempts)),
+			// TODO: Remove "previous-rpc-attempts" and "transparent-retry"
```
dfawley: Why is this a TODO?

Isn't this simply?

```go
attribs := []attribute.KeyValue{...Bool("Client", rs.Client), /* etc */}
if ai.previousRPCAttempts != nil {
	attribs = append(attribs, attribute.Int64("prev...", ..), ...)
}
```

vinothkumarr227: Thanks for the suggestion. Per @arjan-bal  recommendation, this will be handled in a separate PR, which is why it's a TODO. I've created an  issue for reference: https://github.com/grpc/grpc-go/issues/8430

dfawley: I'm not sure I understand why.  If these attributes are not supposed to be present, then wouldn't it be incorrect to include them.

vinothkumarr227: Thanks Doug. I've made changes and pushed the fix.

dfawley: Oh it was already setting these things?

Also what is this about "Go was always setting these attributes even though they aren't in the spec"?

Are we intentionally not implementing the spec?  Why are we mentioning the OC spec in the OT plugin?  I'm kind of lost here, TBH.

vinothkumarr227: Hi Doug, regarding this line "Go was always setting these attributes even though they aren't in the spec"?, I see that this one are already there, but I'm not familiar with this one. Could you please clarify that?

dfawley: Right. I don't understand this either.  Something is wrong.

Either way, this function should probably not have client/server-specific code in it, as it does not.  It should only contain things we do commonly for both.  We have separate client and server handlers for this specific reason.  If we need to delete this function or rework it so it behaves differently, then that's probably worth doing.

https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/stats/opentelemetry/client_tracing.go#L135

vinothkumarr227: Thanks for the feedback, Doug. I've refactored the code to address your concerns. The client- and server-specific logic has been moved out of populateSpan and into their dedicated handlers. The populateSpan function now only contains the common logic.

stats/opentelemetry/client_tracing.go
```diff
@@ -141,5 +142,22 @@ func (h *clientTracingHandler) HandleRPC(ctx context.Context, rs stats.RPCStats)
 		logger.Error("ctx passed into client side tracing handler trace event handling has no client attempt data present")
 		return
 	}
+
+	// Client-specific Begin attributes.
+	var previousRPCAttempts int64
+	if ri.ai.previousRPCAttempts != nil {
```
dfawley: Is it possible for this to be false?

And is there any reason this is a pointer instead of just a value?  Then we wouldn't need nil checks, or to construct it explicitly.

stats/opentelemetry/client_tracing.go
```diff
@@ -141,5 +142,22 @@ func (h *clientTracingHandler) HandleRPC(ctx context.Context, rs stats.RPCStats)
...
+		previousRPCAttempts = int64(ri.ai.previousRPCAttempts.Load())
```
dfawley: Why are we loading this here but only using it inside the `if` below?  Why not load only when needed?

stats/opentelemetry/client_tracing.go
```diff
@@ -141,5 +142,22 @@ func (h *clientTracingHandler) HandleRPC(ctx context.Context, rs stats.RPCStats)
...
+	}
+	if begin, ok := rs.(*stats.Begin); ok {
+		ri.ai.traceSpan.SetAttributes(
+			attribute.Bool("Client", begin.Client),
+			attribute.Bool("FailFast", begin.FailFast),
```
dfawley: Let's remove these now since they are not part of the spec

vinothkumarr227: Done

stats/opentelemetry/client_tracing.go
vinothkumarr227: It’s not possible for it to be nil — we always initialize it in getOrCreateCallInfo with ci.previousRPCAttempts = new(atomic.Uint32). I’ve removed the nil check as well. The pointer is used so that updates in ai are automatically reflected in ci.

stats/opentelemetry/client_tracing.go
vinothkumarr227: No need for the condition — I’ve removed the nil check as well.

stats/opentelemetry/client_tracing.go
dfawley: Oh this is in the attemptInfo.  Why are we keeping it here instead of just loading it out of the callInfo?

Also I realized that with hedging, the previous attempts accounting is racy:

- Attempt 1 starts
- Attempt 2 starts
- Attempt 3 starts
- All attempts do `Begin` simultaneously

It's possible for any of them to have any value now.  You need to instead increment and read together:

```go
previousAttempts := previousRPCAttempts.Add(1) - 1 // Add returns the new value; we need the old value
```

Then they will all see unique values.

(We don't implement hedging yet, but we need to keep in mind that we will one day.)

vinothkumarr227: previousRPCAttempts represents the number of retries before the current attempt. We should record the existing retry count in attemptInfo before incrementing it. After that, we increment the retry count.

vinothkumarr227: > Also I realized that with hedging, the previous attempts accounting is racy:
> 
> * Attempt 1 starts
> * Attempt 2 starts
> * Attempt 3 starts
> * All attempts do `Begin` simultaneously
> 
> It's possible for any of them to have any value now. You need to instead increment and read together:
> 
> ```go
> previousAttempts := previousRPCAttempts.Add(1) - 1 // Add returns the new value; we need the old value
> ```
> 
> Then they will all see unique values.
> 
> (We don't implement hedging yet, but we need to keep in mind that we will one day.)
Sure, I’ll keep that in mind; we’ll do it in a future update.

dfawley: > previousRPCAttempts represents the number of retries before the current attempt. We should record the existing retry count in attemptInfo before incrementing it. After that, we increment the retry count.

No, we need to atomically read and increment it, as I explained.

vinothkumarr227: Hi Doug, I’m a bit confused. Initially, I had the increment here — https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/client_metrics.go#L78 — but after the review comments, I’m moved to trace. Could you clarify what exactly you’d like to change here?

dfawley: If we do the load and add separately, there is a race if multiple attempts are happening at once, with hedging.  So we need to do the add and use the value it returns.  It returns the incremented value, so we need to subtract 1 from it to get the original value.  This way if multiple attempts are happening at the same time, then each one will get a unique value.

I think we should also see about removing the previousRPCAttempts pointer from the attemptInfo, and instead read it out of the callInfo, which is stored in the context too, right?

vinothkumarr227: Thanks for the suggestion! I'll update the code accordingly. I'll also double-check how it's stored in the context.

vinothkumarr227: > If we do the load and add separately, there is a race if multiple attempts are happening at once, with hedging. So we need to do the add and use the value it returns. It returns the incremented value, so we need to subtract 1 from it to get the original value. This way if multiple attempts are happening at the same time, then each one will get a unique value.
> 
> I think we should also see about removing the previousRPCAttempts pointer from the attemptInfo, and instead read it out of the callInfo, which is stored in the context too, right?

Thanks for the feedback! I've updated the code.

stats/opentelemetry/client_metrics.go
```diff
@@ -76,8 +76,9 @@ func getOrCreateCallInfo(ctx context.Context, cc *grpc.ClientConn, method string
 			logger.Info("Creating new CallInfo since its not present in context")
 		}
 		ci = &callInfo{
-			target: cc.CanonicalTarget(),
-			method: determineMethod(method, opts...),
+			target:              cc.CanonicalTarget(),
+			method:              determineMethod(method, opts...),
+			previousRPCAttempts: new(atomic.Uint32),
```
dfawley: Let's make this a non-pointer type and then we don't need the `new` here or the chance to get nil panics.

vinothkumarr227: Done

stats/opentelemetry/client_metrics.go
```diff
@@ -78,7 +78,7 @@ func getOrCreateCallInfo(ctx context.Context, cc *grpc.ClientConn, method string
 		ci = &callInfo{
 			target:              cc.CanonicalTarget(),
 			method:              determineMethod(method, opts...),
-			previousRPCAttempts: new(atomic.Uint32),
+			previousRPCAttempts: atomic.Uint32{},
```
dfawley: Please delete this line.  This is the zero value so it doesn't need explicit initialization.

vinothkumarr227: Done

PR #8347, opened 2025-05-21
balacergroup: cleanup exitIdle()

hugehoo:
### context
`exitIdle()` is called in both ExitIdle() and ExitIdleOne()

```go
func (bg *BalancerGroup) ExitIdle() {
	bg.outgoingMu.Lock()
	defer bg.outgoingMu.Unlock()
	if bg.outgoingClosed {
		return
	}
	for _, config := range bg.idToBalancerConfig {
		if !config.exitIdle() { // here
			bg.connect(config)
		}
	}
}

func (bg *BalancerGroup) ExitIdleOne(id string) {
	bg.outgoingMu.Lock()
	defer bg.outgoingMu.Unlock()
	if bg.outgoingClosed {
		return
	}
	if config := bg.idToBalancerConfig[id]; config != nil {
		if !config.exitIdle() { // here
			bg.connect(config)
		}
	}
}
```
it returns boolean value. if it returns false if-statement works.
But the current implementation of `exitIdle()` is returning only true.

### As-Is
`exitIdle` method returns boolean type but implementation returns only true. 
``` go
func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
	b := sbc.balancer
	if b == nil {
		return true
	}
	b.ExitIdle()
	return true
}
```

### To-be
so i simplified implementation as method only returns true.
And fixed comment too as `Returns a true`
``` go
// exitIdle invokes the sub-balancer's ExitIdle method. Returns a true
// indicating whether or not the operation was completed.
func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
	if b := sbc.balancer; b != nil {
		b.ExitIdle()
	}
	return true
}
```

I'm not sure this refactoring is necessary but at least the comment on exitIdle seems to be updated.

RELEASE NOTES: N/A

### DIFF
--- a/internal/balancergroup/balancergroup.go
+++ b/internal/balancergroup/balancergroup.go
@@ -106,13 +106,13 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method. Returns a true
-// indicating whether or not the operation was completed.
-func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
-	if b := sbc.balancer; b != nil {
-		b.ExitIdle()
+// exitIdle invokes the sub-balancer's ExitIdle method.
+func (sbc *subBalancerWrapper) exitIdle() {
+	b := sbc.balancer
+	if b == nil {
+		return
 	}
-	return true
+	b.ExitIdle()
 }
 
 func (sbc *subBalancerWrapper) updateClientConnState(s balancer.ClientConnState) error {
@@ -573,9 +573,7 @@ func (bg *BalancerGroup) ExitIdle() {
 		return
 	}
 	for _, config := range bg.idToBalancerConfig {
-		if !config.exitIdle() {
-			bg.connect(config)
-		}
+		config.exitIdle()
 	}
 }
 
@@ -588,9 +586,7 @@ func (bg *BalancerGroup) ExitIdleOne(id string) {
 		return
 	}
 	if config := bg.idToBalancerConfig[id]; config != nil {
-		if !config.exitIdle() {
-			bg.connect(config)
-		}
+		config.exitIdle()
 	}
 }
 
--- a/internal/balancergroup/balancergroup.go
+++ b/internal/balancergroup/balancergroup.go
@@ -106,7 +106,7 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method. Returns a boolean
+// exitIdle invokes the sub-balancer's ExitIdle method. Returns a true
 // indicating whether or not the operation was completed.
 func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
 	if b := sbc.balancer; b != nil {
--- a/internal/balancergroup/balancergroup.go
+++ b/internal/balancergroup/balancergroup.go
@@ -106,8 +106,8 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method
-// which is a graceful switch balancer.
+// exitIdle invokes the ExitIdle method on the sub-balancer, a gracefulswitch
+// balancer.
 func (sbc *subBalancerWrapper) exitIdle() {
 	b := sbc.balancer
 	if b == nil {
--- a/internal/balancergroup/balancergroup.go
+++ b/internal/balancergroup/balancergroup.go
@@ -106,7 +106,8 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method.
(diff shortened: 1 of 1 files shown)


internal/balancergroup/balancergroup.go
```diff
@@ -106,14 +106,12 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method. Returns a boolean
+// exitIdle invokes the sub-balancer's ExitIdle method. Returns a true
 // indicating whether or not the operation was completed.
```
hugehoo, line 110: nit) i suggest that `whether or not` to be `whether`.

internal/balancergroup/balancergroup.go
```diff
@@ -106,14 +106,12 @@ func (sbc *subBalancerWrapper) startBalancer() {
...
 func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
-	b := sbc.balancer
-	if b == nil {
-		return true
+	if b := sbc.balancer; b != nil {
+		b.ExitIdle()
 	}
-	b.ExitIdle()
 	return true
```
injae-kim, line 115: oh.. `return true * 2` 😅 hard to understand code.. refactor +1!

eshitachandwani:
Hey @hugehoo , thank you for the PR , I have a few suggestions since the code has changed.
Since the function only returns true , you can change it to not return anything and just call `ExitIdle` and also remove the condition in balancerGroup because it is never executed and the functionality of connecting to all subconns has been passed  down to [gracefulSwitch Balancer](https://github.com/grpc/grpc-go/blob/master/internal/balancer/gracefulswitch/gracefulswitch.go#L233) and since the balancer in `subBalancerWrapper` has been changed to gracefulSwitch balancer, it will always have `ExitIdle` function:
```
func (sbc *subBalancerWrapper) exitIdle(){
	b := sbc.balancer
	if b == nil {
		return
	}
	b.ExitIdle()
}
```

And change the these parts as follows:
```
if config := bg.idToBalancerConfig[id]; config != nil {
	config.exitIdle()
}
```

eshitachandwani (commented):
Hey @hugehoo , thank you for the PR , I have a few suggestions since the code has changed.
Since the function only returns true , you can change it to not return anything and just call `ExitIdle` and also remove the condition in balancerGroup because it is never executed and the functionality of connecting to all subconns has been passed  down to [gracefulSwitch Balancer](https://github.com/grpc/grpc-go/blob/master/internal/balancer/gracefulswitch/gracefulswitch.go#L233) and since the balancer in `subBalancerWrapper` has been changed to gracefulSwitch balancer, it will always have `ExitIdle` function:
```
func (sbc *subBalancerWrapper) exitIdle(){
	b := sbc.balancer
	if b == nil {
		return
	}
	b.ExitIdle()
}
```

And change the these parts as follows:
```
if config := bg.idToBalancerConfig[id]; config != nil {
	config.exitIdle()
}
```

internal/balancergroup/balancergroup.go
```diff
@@ -411,20 +409,6 @@ func (bg *BalancerGroup) cleanupSubConns(config *subBalancerWrapper) {
...
-	if bg.incomingClosed {
-		return
-	}
-	for sc, b := range bg.scToSubBalancer {
-		if b == sb {
-			sc.Connect()
-		}
-	}
-}
-
```
hugehoo, line 427: i removed here, it fails CI-test because it's unused.
image

eshitachandwani (commented):
> nit) i suggest that `whether or not` to be `whether`. 

Can you add a clarification in comments that the sub-balancer is a graceful switch balancer

internal/balancergroup/balancergroup.go
```diff
@@ -106,15 +106,13 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method. Returns a boolean
-// indicating whether or not the operation was completed.
-func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
+// exitIdle invokes the sub-balancer's ExitIdle method.
```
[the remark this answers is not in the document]
hugehoo: @eshitachandwani  I added it 👍🏻

internal/balancergroup/balancergroup.go
```diff
@@ -106,15 +106,14 @@ func (sbc *subBalancerWrapper) startBalancer() {
 	}
 }
 
-// exitIdle invokes the sub-balancer's ExitIdle method. Returns a boolean
-// indicating whether or not the operation was completed.
-func (sbc *subBalancerWrapper) exitIdle() (complete bool) {
+// exitIdle invokes the sub-balancer's ExitIdle method
+// which is a graceful switch balancer.
```
eshitachandwani, line 110:
```suggestion
// exitIdle invokes the ExitIdle method on the sub-balancer, a gracefulswitch
// balancer.
```

PR #8348, opened 2025-05-22
xdsclient: fix unexpectedly large  in initial load report request

purnesh42H:
Internal bug: b/416260484

RELEASE NOTES:
- xdsclient: fix unexpectedly large  in initial load report request.

### DIFF
--- a/xds/internal/clients/lrsclient/load_store.go
+++ b/xds/internal/clients/lrsclient/load_store.go
@@ -82,8 +82,9 @@ func (ls *LoadStore) ReporterForCluster(clusterName, serviceName string) *PerClu
 		return p
 	}
 	p := &PerClusterReporter{
-		cluster: clusterName,
-		service: serviceName,
+		cluster:          clusterName,
+		service:          serviceName,
+		lastLoadReportAt: time.Now(),
 	}
 	c[serviceName] = p
 	return p
--- a/xds/internal/xdsclient/load/store.go
+++ b/xds/internal/xdsclient/load/store.go
@@ -115,8 +115,9 @@ func (s *Store) PerCluster(clusterName, serviceName string) PerClusterReporter {
 		return p
 	}
 	p := &perClusterStore{
-		cluster: clusterName,
-		service: serviceName,
+		cluster:          clusterName,
+		service:          serviceName,
+		lastLoadReportAt: time.Now(),
 	}
 	c[serviceName] = p
 	return p


arjan-bal (commented):
Can you please add a test for this to prevent regressions?

purnesh42H:
> Can you please add a test for this to prevent regressions?

we can only check if LoadReportInterval is positive but not excessively large. We can't compare exactly since calculation involves time.Now(). I have added test verifications in the existing e2e tests with some tolerance to ensure reportInterval of the first report and subsequent reports is a small, positive duration but greater than configured server LoadReportingInterval.

dfawley:
> we can only check if LoadReportInterval is positive but not excessively large. We can't compare exactly since calculation involves time.Now().

Without looking at the details here... Can we hook `time.Now` and make it return predictable values for our tests?

dfawley:
> Thought that limits us to only verify with predictable report interval in the unit test of the load store and not in the e2e tests

If needed, you can hop through an internal package to be able to set from e2e tests.

xds/internal/clients/lrsclient/load_store_test.go
```diff
@@ -471,3 +472,49 @@ func TestStoreStatsEmptyDataNotReported(t *testing.T) {
...
+	// First call to stats() calculates the report interval from reporter
+	// creation time.
+	currentTime = currentTime.Add(5 * time.Second)
+	stats1 := reporter.stats()
+
+	if stats1 == nil {
+		t.Fatalf("stats1 is nil after reporting a drop, want non-nil")
+	}
+	wantInterval := 5 * time.Second
+	if stats1.reportInterval != wantInterval {
```
dfawley, line 502: Optional:

I think the best way I've been recommended of writing this kind of test comparison is:

```go
if got, want := stats1.reportInterval, 5 * time.Second; got != want {
	t.Fatalf("blah blah = %v; want %v", got, want)
}
```

xds/internal/clients/lrsclient/load_store.go
```diff
@@ -25,6 +25,9 @@ import (
 	"time"
 )
 
+// clockNow is used to get the current time. It can be overridden in tests.
+var clockNow = time.Now
```
dfawley, line 29: Sorry for the nit, but would you mind renaming this to `timeNow` to match `packagename.SymbolName` but without the dot?  That's the convention we generally use.

xds/internal/xdsclient/load/store.go
```diff
@@ -25,6 +25,9 @@ import (
 
 const negativeOneUInt64 = ^uint64(0)
 
+// clockNow is used to get the current time. It can be overridden in tests.
+var clockNow = time.Now
```
dfawley, line 29: As above.

xds/internal/xdsclient/load/store_test.go
```diff
@@ -466,3 +467,49 @@ func TestStoreStatsEmptyDataNotReported(t *testing.T) {
...
+	reporter := store.PerCluster("test-cluster", "test-service")
+	// To ensure Stats() returns non-nil data, report a dummy drop.
+	reporter.CallDropped("dummy-category")
+
+	// First call to Stats() calculates the report interval from reporter
+	// creation time.
+	currentTime = currentTime.Add(5 * time.Second)
+	stats1 := store.Stats(nil)
+
+	if stats1 == nil {
```
dfawley, line 493: It seems like you want even more from this check, since it's a slice that you dereference element 0 of immediately below.

```go
if len(stats1) == 0 { // or != 1 ?
	t.Fatalf("store.Stats(nil) = %v; want len() >= 1", stats1)  // or == 1?
}
```

dfawley (approved):
LGTM, just a couple very minor things.

purnesh42H:
> > Thought that limits us to only verify with predictable report interval in the unit test of the load store and not in the e2e tests
> 
> If needed, you can hop through an internal package to be able to set from e2e tests.

I did it for the generic lrs client tests but left it for the internal one since that will be external soon.

xds/internal/xdsclient/load/store_test.go
purnesh42H: Done

xds/internal/clients/lrsclient/load_store.go
purnesh42H: Done

PR #8385, opened 2025-06-06
grpc: Fix cardinality violations in server streaming RPC.

### LINKED ISSUE #8362: Cardinality violations in serverside streaming RPC.
In a typical server-side stream, the client sends a single request and then receives a sequence of messages from the server until an EOF or error. The official Go gRPC implementation enforces this by only allowing the [client to call Recv() and the server to call Send()](https://github.com/grpc/grpc-go/blob/e3ca7f9077fc8a64ab9de856dc6cf71f21c87e9c/stream_interfaces.go#L24C1-L55C2). This prevents clients from inadvertently calling Send() on a server-side stream, thereby avoiding a cardinality violation by design.
However, if we implement our own custom RPCs, it's theoretically possible for a client to cause cardinality violations by attempt to call Client.Send() and possible for server to cause cardinality violation by calling Server.Recv() twice.

Should I handled those cases for custom RPCs and add specific tests to address these potential cardinality violation scenarios?

Pranjali-2501:
Fixes #8362 
Partially addresses: #7286 

In non-client streaming RPCs, the client's SendMsg() method is designed to automatically close the send operation after its initial call. If someone attempts to call Client.SendMsg() twice for non-client streaming RPCs, if will return with error `Internal desc = SendMsg called after CloseSend`.
To mirror this behavior, the server-side logic has been updated so that calling RecvMsg() more than once for non-client streaming RPCs will now similarly return an `Internal` error.

RELEASE NOTES:
* grpc: return status code INTERNAL when client send more than one request in unary and server streaming RPC.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -1580,6 +1580,7 @@ type serverStream struct {
 	s     *transport.ServerStream
 	p     *parser
 	codec baseCodec
+	desc  *StreamDesc
 
 	compressorV0   Compressor
 	compressorV1   encoding.Compressor
@@ -1773,6 +1774,9 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
 				for _, binlog := range ss.binlogs {
 					binlog.Log(ss.ctx, chc)
 				}
+			}
+			if !ss.desc.ClientStreams {
+				return status.Errorf(codes.Internal, "RecvMsg is called twice")
 			}
 			return err
 		}
--- a/test/end2end_test.go
+++ b/test/end2end_test.go
@@ -3740,6 +3740,268 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	}
 }
 
+// Tests the behavior for server-side streaming when server call
+// RecvMsg twice. Second call to RecvMsg should fail with Internal
+// error.
+func (s) TestServerStreaming_ServerCallRecvMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
+		ServiceName: "grpc.testing.TestService",
+		HandlerType: (*any)(nil),
+		Methods:     []grpc.MethodDesc{},
+		Streams: []grpc.StreamDesc{
+			{
+				StreamName: "FullDuplexCall",
+				Handler: func(srv interface{}, stream grpc.ServerStream) error {
+					err := stream.RecvMsg(&testpb.Empty{})
+					if err != nil {
+						t.Errorf("stream.RecvMsg() = %v, want <nil>", err)
+					}
+
+					if err = stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+						t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+					}
+					return nil
+				},
+				ClientStreams: false,
+				ServerStreams: true,
+			},
+		},
+		Metadata: "grpc/testing/test.proto",
+	}
+	s.RegisterService(&serviceDesc, &testServer{})
+	go s.Serve(lis)
+	defer s.Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "FullDuplexCall",
+		ServerStreams: true,
+		ClientStreams: false, // This is the test case: client is *not* allowed to stream
+	}
+
+	stream, err := cc.NewStream(ctx, desc, "/grpc.testing.TestService/FullDuplexCall")
+	if err != nil {
+		t.Fatalf("cc.NewStream() failed unexpectedly: %v", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for server-side streaming when client call
+// SendMsg twice. Second call to SendMsg should fail with Internal
+// error.
+func (s) TestServerStreaming_ClientCallSendMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
+		ServiceName: "grpc.testing.TestService",
+		HandlerType: (*any)(nil),
+		Methods:     []grpc.MethodDesc{},
+		Streams: []grpc.StreamDesc{
+			{
+				StreamName: "FullDuplexCall",
+				Handler: func(srv interface{}, stream grpc.ServerStream) error {
+					err := stream.RecvMsg(&testpb.Empty{})
+					if err != nil {
+						t.Errorf("stream.RecvMsg() = %v, want <nil>", err)
+					}
+
+					return nil
+				},
+				ClientStreams: false,
+				ServerStreams: true,
+			},
+		},
+		Metadata: "grpc/testing/test.proto",
+	}
+	s.RegisterService(&serviceDesc, &testServer{})
+	go s.Serve(lis)
+	defer s.Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "FullDuplexCall",
+		ServerStreams: true,
+		ClientStreams: false,
+	}
+
+	stream, err := cc.NewStream(ctx, desc, "/grpc.testing.TestService/FullDuplexCall")
+	if err != nil {
+		t.Fatalf("cc.NewStream() failed unexpectedly: %v", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for unary RPC when server call
+// RecvMsg twice. Second call to RecvMsg should fail with Internal
+// error.
+func (s) TestUnaryRPC_ServerCallRecvMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
+		ServiceName: "grpc.testing.TestService",
+		HandlerType: (*any)(nil),
+		Methods:     []grpc.MethodDesc{},
+		Streams: []grpc.StreamDesc{
+			{
+				StreamName: "UnaryCall",
+				Handler: func(srv interface{}, stream grpc.ServerStream) error {
+					err := stream.RecvMsg(&testpb.Empty{})
+					if err != nil {
+						t.Errorf("stream.RecvMsg() = %v, want <nil>", err)
+					}
+
+					if err = stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+						t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+					}
+					return nil
+				},
+				ClientStreams: false,
+				ServerStreams: false,
+			},
+		},
+		Metadata: "grpc/testing/test.proto",
+	}
+	s.RegisterService(&serviceDesc, &testServer{})
+	go s.Serve(lis)
+	defer s.Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "UnaryCall",
+		ServerStreams: false,
+		ClientStreams: false,
+	}
+
+	stream, err := cc.NewStream(ctx, desc, "/grpc.testing.TestService/UnaryCall")
+	if err != nil {
+		t.Fatalf("cc.NewStream() failed unexpectedly: %v", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for unary RPC when client call
+// SendMsg twice. Second call to SendMsg should fail with Internal
+// error.
+func (s) TestUnaryRPC_ClientCallSendMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
+		ServiceName: "grpc.testing.TestService",
+		HandlerType: (*any)(nil),
+		Methods:     []grpc.MethodDesc{},
+		Streams: []grpc.StreamDesc{
+			{
+				StreamName: "UnaryCall",
+				Handler: func(srv interface{}, stream grpc.ServerStream) error {
+					err := stream.RecvMsg(&testpb.Empty{})
+					if err != nil {
+						t.Errorf("stream.RecvMsg() = %v, want <nil>", err)
+					}
+
+					return nil
+				},
+				ClientStreams: false,
+				ServerStreams: false,
+			},
+		},
+		Metadata: "grpc/testing/test.proto",
+	}
+	s.RegisterService(&serviceDesc, &testServer{})
+	go s.Serve(lis)
+	defer s.Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "UnaryCall",
+		ServerStreams: false,
+		ClientStreams: false,
+	}
+
+	stream, err := cc.NewStream(ctx, desc, "/grpc.testing.TestService/UnaryCall")
+	if err != nil {
+		t.Fatalf("cc.NewStream() failed unexpectedly: %v", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
 func (s) TestExceedMaxStreamsLimit(t *testing.T) {
 	for _, e := range listTestEnv() {
 		testExceedMaxStreamsLimit(t, e)
--- a/server.go
+++ b/server.go
@@ -1572,6 +1572,7 @@ func (s *Server) processStreamingRPC(ctx context.Context, stream *transport.Serv
 		s:                     stream,
 		p:                     &parser{r: stream, bufferPool: s.opts.bufferPool},
 		codec:                 s.getCodec(stream.ContentSubtype()),
+		desc:                  sd,
 		maxReceiveMessageSize: s.opts.maxReceiveMessageSize,
 		maxSendMessageSize:    s.opts.maxSendMessageSize,
 		trInfo:                trInfo,


dfawley (commented):
Is it possible to implement all of this in `server.go` to avoid adding state to the `serverStream`?

Pranjali-2501:
> Is it possible to implement all of this in `server.go` to avoid adding state to the `serverStream`?

As discussed this with @arjan-bal offline, Cardinality violations can only be detected when messages are being read from the stream. This reading process occurs specifically within the server.RecvMsg() function. Since RecvMsg() is invoked from the user-implemented handler, it's not possible to detect cardinality violations during the initial stream setup phase.

stream.go
```diff
@@ -1774,6 +1775,9 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
 					binlog.Log(ss.ctx, chc)
 				}
 			}
+			if !ss.desc.ClientStreams {
```
arjan-bal: It's not immediately clear to me how this conditions indicates RecvMsg is called twice. From my understanding, the handling of cardinality violations in the server stream should be similar to the one in client stream. The check here should be for 0 requests being received for a non-client-streaming request.
https://github.com/grpc/grpc-go/blob/e5de1e2cac527d419f53fbe4ae07628c688ff0dd/stream.go#L1142-L1144

For > 1 requests, there should be a block near the end of this method similar to the following:
https://github.com/grpc/grpc-go/blob/e5de1e2cac527d419f53fbe4ae07628c688ff0dd/stream.go#L1167-L1178

arjan-bal: Is there any issue if we try to follow the same pattern as client streams here?

Pranjali-2501: In server-streaming, client can't send more than one message. So the following code will be useless for server.recvmsg().
https://github.com/grpc/grpc-go/blob/e5de1e2cac527d419f53fbe4ae07628c688ff0dd/stream.go#L1167-L1178

>The check here should be for 0 requests being received for a non-client-streaming request. https://github.com/grpc/grpc-go/blob/e5de1e2cac527d419f53fbe4ae07628c688ff0dd/stream.go#L1142-L1144

Right, in client.recvMsg() this check is triggered in case of 0 response from server.
But in server-stream, it is not possible to have 0 request. So the following case will only be triggered when server.recvmsg() is called more than once and receives an `io.EOF`.

arjan-bal: > In server-streaming, client can't send more than one message.

What is stopping a client from doing this?

Pranjali-2501: Clientstream maintains a variable named `sentlast` which will be set to `true` when client sends a msg in case of non client-streaming RPCs.
https://github.com/grpc/grpc-go/blob/e5de1e2cac527d419f53fbe4ae07628c688ff0dd/stream.go#L905-L910

arjan-bal: As we discussed earlier, can we have the client (incorrectly) create a bidi stream by manipulating the strem desc? 

Ideally, a well behaving client should never cause cardinality violations. The reason for adding these checks is to fail the RPCs with good error codes and messages when a client misbehaves.

Pranjali-2501: Added the logic in server.recvmsg() to call recv() twice in case of non-client-stream RPCs.  Also added a test with misbehaving client.

stream.go
```diff
@@ -1774,6 +1775,9 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
...
+				return status.Errorf(codes.Internal, "RecvMsg is called twice")
```
arjan-bal: This doesn't necessarily indicate RecvMsg being called twice. If the code reaches here, it means that the server handler was trying to read a message, but the stream ended. This can happen in two situations:
1. The client sent 0 request messages, or
2. The server handler got an `io.EOF` after reading the first message, but still called `RecvMsg` again.

I belive the behaviour in case 2 is undefined and we should tailor the message to case 1. This is similar to the client side change in previous PRs. So error message should be similar to:
> cardinality violation: received no request message from non-streaming RPC

stream.go
```diff
@@ -1800,6 +1804,12 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
 			binlog.Log(ss.ctx, cm)
 		}
 	}
+	// Special handling for non-client-stream rpcs.
+	if !ss.desc.ClientStreams {
+		if err := recv(ss.p, ss.codec, ss.s, ss.decompressorV0, m, ss.maxReceiveMessageSize, payInfo, ss.decompressorV1, true); err != io.EOF {
```
arjan-bal: Similar to the client stream check, pass `nil` instead of `payInfo` and add a comment:
> This recv expects EOF or errors, so we don't collect inPayload.

stream.go
```diff
@@ -1800,6 +1804,12 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
...
+			return status.Errorf(codes.Internal, "cardinality violation: expected <EOF> for non client-streaming RPCs, but received another message")
```
arjan-bal: We're converting non-nil errors to an internal status here. This is different from the client stream implementation which only converts `nil` errors to internal status. I think we should do the same here. 

Nit: I think we should keep the same structure as the client stream code, i.e.
```go
	if cs.desc.ClientStreams {
		// Subsequent messages should be received by subsequent RecvMsg calls.
		return nil
	}
	// Special handling for non-client-stream rpcs.
	// This recv expects EOF or errors, so we don't collect inPayload.
	if err := recv(a.parser, cs.codec, a.transportStream, a.decompressorV0, m, *cs.callInfo.maxReceiveMessageSize, nil, a.decompressorV1, false); err == io.EOF {
		return nil
	} else if err != nil {
		return err
	}
	return status.Error(codes.Internal, "cardinality violation: expected <EOF> for non client-streaming RPCs, but received another message")
```

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
+		ServiceName: "grpc.testing.TestService",
+		HandlerType: (*any)(nil),
+		Methods:     []grpc.MethodDesc{},
+		Streams: []grpc.StreamDesc{
+			{
+				StreamName: "FullDuplexCall",
```
arjan-bal: nit: We should change the stream name to `ServerStreaming` or something similar since this not a bi-di stream. Please fix throughout.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+					if err = stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+						t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+					}
+					return nil
+				},
+				ClientStreams: false,
+				ServerStreams: true,
+			},
+		},
+		Metadata: "grpc/testing/test.proto",
```
arjan-bal: nit: Is this field required? If not, it can be omitted.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	}
+	s.RegisterService(&serviceDesc, &testServer{})
+	go s.Serve(lis)
+	defer s.Stop()
+
+	// s := grpc.NewServer(ss)
```
arjan-bal: This comment should be removed.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+func (s) TestUnaryRPC_ClientCallSendMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	ss := grpc.UnknownServiceHandler(func(any, grpc.ServerStream) error {
+		return nil
+	})
```
arjan-bal: The RPC is failed on the client side, right? If so, we should use a real server instead of an `UnknownServiceHandler`.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for unary RPC when client call
```
arjan-bal: nit: `a client calls`.

arjan-bal: Please wrap lines at 80 columns, not earlier. Format throughout.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for unary RPC when server call
```
arjan-bal: nit: `when a server calls`. Please use complete sentences in comments.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "FullDuplexCall",
+		ServerStreams: true,
+		ClientStreams: false, // This is the test case: client is *not* allowed to stream
```
arjan-bal: nit: Omit `This is the test case:`, it doesn't add any information.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for unary RPC when server call
+// RecvMsg twice. Second call to RecvMsg should fail with Internal
+// error.
+func (s) TestUnaryRPC_ServerCallRecvMsgTwice(t *testing.T) {
```
arjan-bal: Can this be combined with `TestServerStreaming_ServerCallRecvMsgTwice` into a single table drive test? They seem to be exactly the same except for a couple of lines. Similarly for the `ClientCallSendMsgTwice` tests.

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+func (s) TestServerStreaming_ClientBehaveAsBidiStreaming(t *testing.T) {
```
arjan-bal: I think a better suffix is `ClientSendsMultipleMessages` because we want to test the case when a client sends multiple messages. Making the client bi-di is just a way to bypass the client side checks that stop a non-streaming client from sending multiple messages. 

Can you please add a doc comment describing the test?

test/end2end_test.go
```diff
@@ -3740,6 +3740,299 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
```
arjan-bal, line 3967: Please add a comment describing the need to use a bi-di streaming client for this test.

test/end2end_test.go
arjan-bal, line 1: The PR is missing a test for the case when a client sends 0 request messages. We should add that.

Pranjali-2501: As we have discussed earlier, from the client side a request is always initiated by a sendMsg() call.

However, from the server side, it is possible to return nil without sending any messages, effectively handling a "zero message" scenario.

arjan-bal: > from the client side a request is always initiated by a sendMsg() call.

This is not always true. You can create a stream using `NewStream` and immediately call `CloseSend` on it, effectively closing the send side of the stream without sending a gRPC message.

test/end2end_test.go
Pranjali-2501: Already have a test with  suffix `ClientCallSendMsgTwice()` to test what happen when a correctly configured client for server-side streaming attempts to call SendMsg() twice.

I'll add a comment to describe what this test is doing.

test/end2end_test.go
Pranjali-2501: Yes, RPC fails on client side.
Why do we need to use the real server, the check is on the client side.

test/end2end_test.go
Pranjali-2501: Done

test/end2end_test.go

test/end2end_test.go

test/end2end_test.go

test/end2end_test.go
Pranjali-2501: Done.

test/end2end_test.go
Pranjali-2501: I have added a test for zero request messages from client.

test/end2end_test.go

test/end2end_test.go

test/end2end_test.go
arjan-bal: We should use real implementations in tests as much as possible so that the entire RPC stack is used and there are no unintended interactions. By mocking the server, there is a chance that the client side check is removed and the test continues to pass because the fake server doesn't have the required functionality to handle the call.

test/end2end_test.go
arjan-bal: `ClientSendsMultipleMessages` is not the same as `ClientCallSendMsgTwice`. The former implies that the client is actually sending two messages, whereas the latter indicates that the client is calling the SendMsg method twice.

test/end2end_test.go
arjan-bal: I don't see the comment here.

test/end2end_test.go
```diff
@@ -3740,6 +3740,352 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+				ClientStreams: false,
+				ServerStreams: true,
+			},
+		},
+	}
+	s.RegisterService(&serviceDesc, &testServer{})
+	go s.Serve(lis)
+	defer s.Stop()
+
+	// s := grpc.NewServer(ss)
```
arjan-bal: Please remove this commented code.

test/end2end_test.go
```diff
@@ -3740,6 +3740,352 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+// Tests the behavior for server-side streaming RPC when client sends zero request message.
+func (s) TestServerStreaming_ClientSendsZeroRequest(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
```
arjan-bal: I believe we don't need to change the stream descriptors on the server side for this test. If so, please use a stubserver and override one of the handlers.

test/end2end_test.go
```diff
@@ -3740,6 +3740,352 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+// Second call to RecvMsg should fail with Internal error.
+func (s) TestServerStreaming_ServerCallRecvMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
+	serviceDesc := grpc.ServiceDesc{
```
arjan-bal: Do we need to fake the stream descriptors in this test? If not, we should use the test service client and server.

test/end2end_test.go
arjan-bal: Bump on this.

test/end2end_test.go
Pranjali-2501: Make sense, made the changes.

test/end2end_test.go
Pranjali-2501: I had added a comment before test.
Now added a comment here as well.

test/end2end_test.go
Pranjali-2501: sure.

test/end2end_test.go

test/end2end_test.go
Pranjali-2501: Sure.

test/end2end_test.go
Pranjali-2501: `StreamingOutputCall` have predefined calls to sendmsg() and recvmsg().
This test is to check the case when client sends no msg.

test/end2end_test.go
Pranjali-2501: Test are different now.
PTAL.

test/end2end_test.go

test/end2end_test.go
```diff
@@ -3740,6 +3740,345 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	ss := stubserver.StubServer{
+		StreamingOutputCallF: func(_ *testpb.StreamingOutputCallRequest, stream testgrpc.TestService_StreamingOutputCallServer) error {
+			// This is second call to RecvMsg(), the initial call having been performed by the server handler.
+			if err = stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
```
Pranjali-2501: @arjan-bal , as mentioned in the comment I'm using stubserver here.
But `StreamingOutputCall` handler inherently calls RecvMsg() and second call to RecvMsg() in `StreamingOutputCallF` cause a race to happen.
https://github.com/grpc/grpc-go/actions/runs/16136673126/job/45534355675?pr=8385

Should I revert it back to previous implementation and register a new service on the server to test?

arjan-bal: The race is in the writes performed on the `err` variable declared at line 3746. We should declare a new `err` variable inside the goroutine.

stream.go
```diff
@@ -1800,7 +1804,19 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
...
+		return nil
+	}
+	// Special handling for non-client-stream rpcs.
+	// This recv expects EOF or errors, so we don't collect inPayload.
+	if err := recv(ss.p, ss.codec, ss.s, ss.decompressorV0, m, ss.maxReceiveMessageSize, nil, ss.decompressorV1, true); err == io.EOF {
+		return nil
+	} else if err != nil {
+		return err
+	}
+	return status.Error(codes.Internal, "cardinality violation: expected <EOF> for non client-streaming RPCs, but received another message")
```
dfawley: How about something more like the first message:

```suggestion
	return status.Error(codes.Internal, "cardinality violation: received multiple request messages for non-client-streaming RPC")
```

stream.go
```diff
@@ -1774,6 +1775,9 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
 					binlog.Log(ss.ctx, chc)
 				}
 			}
+			if !ss.desc.ClientStreams {
+				return status.Error(codes.Internal, "cardinality violation: received no request message from non-client-stream RPC")
```
dfawley: Nit:

```suggestion
				return status.Error(codes.Internal, "cardinality violation: received no request message from non-client-streaming RPC")
```

stream.go
dfawley: Is this a behavior change?  Users that call `RecvMsg` would previously get `io.EOF` and now they'll get a cardinality violation?

test/end2end_test.go
```diff
@@ -3740,6 +3740,345 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+		t.Fatalf("cc.NewStream() failed unexpectedly: %v", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); err != nil {
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
```
dfawley: Can we confirm that this cancels the RPC, too, as observed from the server?  If the stream ends on the client side it must perform a RST_STREAM with the server.

stream.go
Pranjali-2501: I'm trying to keep it consistent with client.RecvMsg().
https://github.com/grpc/grpc-go/blob/a5e7cd6d4c2c31b1e6649789c2ddc9a82ad6b5fa/stream.go#L1180-L1181

Let me know if `multiple request messages` looks better, I'll change it.

stream.go

stream.go
Pranjali-2501: Yes, there is a behaviour change.

For non-client-streaming RPCs, it will return `internal` error for the following 2 cases.
* When there is 0 request message from Client.
* When Server call RecvMsg() twice.

test/end2end_test.go
Pranjali-2501: No, it did not do a RST_STREAM.
Following code causes this.
https://github.com/grpc/grpc-go/blob/a5e7cd6d4c2c31b1e6649789c2ddc9a82ad6b5fa/stream.go#L911-L917

test/end2end_test.go
arjan-bal: The server behaviour shouldn't matter for this test, right? When the client sends 0 request messages, the server should send an internal error immediately. Are you seeing that this is not happening with stubserver?

test/end2end_test.go
```diff
@@ -3740,6 +3740,345 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+func (s) TestServerStreaming_ServerCallRecvMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	ss := stubserver.StubServer{
+		StreamingOutputCallF: func(_ *testpb.StreamingOutputCallRequest, stream testgrpc.TestService_StreamingOutputCallServer) error {
+			// This is second call to RecvMsg(), the initial call having been performed by the server handler.
```
arjan-bal: > the initial call having been performed by the server handler.

nit: I think a clearer way to mention the reason is that the `the initial call to receive the request is made by the generated code`.

test/end2end_test.go
```diff
@@ -3740,6 +3740,345 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", lis.Addr(), err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "UnaryCall",
+		ServerStreams: false,
+		ClientStreams: false,
+	}
```
arjan-bal: Since the test is for a misbehaving server, can we use a real test service client instead of manipulating `StreamDesc`s?

test/end2end_test.go
```diff
@@ -3740,6 +3740,345 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+// Tests the behavior for server-side streaming when client calls SendMsg twice.
+// Second call to SendMsg should fail with Internal error.
+func (s) TestServerStreaming_ClientCallSendMsgTwice(t *testing.T) {
+	lis, err := testutils.LocalTCPListener()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer lis.Close()
+
+	s := grpc.NewServer()
```
arjan-bal: Similar to previous comments, can we use a stubserver if we don't need the server to misbehave? Also applicable to `TestUnaryRPC_ClientCallSendMsgTwice`.

stream.go
dfawley: I guess the second behavior change is only possible if you are using the generic API, so wouldn't affect 99% of our users.  But I still don't think this is a change we want to make.  I would _probably_ be in favor of it if we were not 1.0, but we need to keep backward compatibility unless there's enough justification, which doesn't seem to be the case here.

test/end2end_test.go
dfawley: Did you confirm a RST_STREAM doesn't happen?

Because tracing through the code...that error results in a call to `cs.finish` which will call `csAttempt.finish` which calls `transportStream.Close`.  That should hopefully result in a RST_STREAM.

test/end2end_test.go

stream.go
Pranjali-2501: As discussed offline, I have added an additional state in `serverstream` to track the first call to server.recvMsg().

Based on that server will only return `Internal` error when there is 0 request message from Client.

test/end2end_test.go
Pranjali-2501: Test is no longer required.

test/end2end_test.go
Pranjali-2501: Test is no longer required.

test/end2end_test.go
Pranjali-2501: You’re right — I traced it again and confirmed that the RST_STREAM does happen. The flow is:
`cs.finish → csAttempt.finish → transportStream.Close → http2Client.closeStream → controlBuf.executeAndPut → cleanupStream.isTransportResponseFrame()`.

stream.go
```diff
@@ -1588,6 +1589,8 @@ type serverStream struct {
 
 	sendCompressorName string
 
+	recvFirstMsg bool // recv frist msg from client
```
dfawley: This comment doesn't really say anything.  How about:

```suggestion
	recvFirstMsg bool // set after the first message is received
```

stream.go
dfawley: These error messages go out on the wire. To clients written in any language. So they should generically explain the problem and not use Go-specific terminology, or internal implementation details like "expected EOF".

That's probably not appropriate on the client side, either, but slightly less bad.  But since you're changing things, that one should probably look like:

```go
	return status.Errorf(codes.Internal, "cardinality violation: received multiple response messages for non-server-streaming RPC")
```

test/end2end_test.go
dfawley: Great.  Can we update the test to confirm the stream is getting cancelled on the server side?  I think you could just make it do `send` operations and eventually see it error?

stream.go

test/end2end_test.go

stream.go
Pranjali-2501: Made the change.

stream.go
dfawley: I don't see it.

stream.go
dfawley: I don't see it.

test/end2end_test.go
```diff
@@ -3740,6 +3740,182 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
 	}
 }
 
+// Tests the behavior for server-side streaming when client calls SendMsg twice.
+// Second call to SendMsg should fail with Internal error and result in closing
+// the connection with a RST_STREAM.
+func (s) TestServerStreaming_ClientCallSendMsgTwice(t *testing.T) {
+	// To ensure server.recvMsg() is successfully completed.
```
dfawley: Is this actually needed?  Why?  If the client were to attempt to send its second message immediately, and that caused a RST_STREAM, that would all happen after the server processed the headers, and the handler should get invoked regardless, right?

test/end2end_test.go
```diff
@@ -3740,6 +3740,182 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+		t.Errorf("stream.SendMsg() = %v, want <nil>", err)
+	}
+
+	if err := stream.RecvMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), codes.Internal)
+	}
+}
+
+// Tests the behavior for server-side streaming RPC when client sends zero request message.
+func (s) TestServerStreaming_ClientSendsZeroRequest(t *testing.T) {
```
dfawley: Nit:

```suggestion
// Tests the behavior for server-side streaming RPC when client sends zero request messages.
func (s) TestServerStreaming_ClientSendsZeroRequests(t *testing.T) {
```

test/end2end_test.go
```diff
@@ -3740,6 +3740,182 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	cc, err := grpc.NewClient(ss.Address, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("grpc.NewClient(%q) failed unexpectedly: %v", ss.Address, err)
+	}
+	defer cc.Close()
+
+	desc := &grpc.StreamDesc{
+		StreamName:    "StreamingOutputCall",
+		ServerStreams: true,
+		ClientStreams: false,
```
dfawley, line 3970: It seems like we should have a test case that sets this `true`, also, and ensures that the server errors?  This is theoretically catching a client-side check that errors if zero requests are sent before `CloseSend` is called when the client knows it is not client-streaming.

(And if we don't have such a check we may want to add one.)

test/end2end_test.go
Pranjali-2501: The first call to server.RecvMsg is made by the generated handler before the test's server handler is invoked. If this call fails, the server handler in the test will not be executed at all.

To handle this, I synchronise the calls in the following way:
- Client.SendMsg()
- Server.RecvMsg()
- Client.SendMsg() – Expected to fail with an Internal error and close connection with RST_STREAM. 
- Server.SendMsg() – Expected to fail with a Canceled error

An alternative approach would be to use a fake server implementation, which would allow us to directly observe and assert the error returned from server.RecvMsg().

test/end2end_test.go
Pranjali-2501: Just to clarify — you're suggesting me to add a test where the client behaves as client/bidi-streaming and sends zero request, while server behave as server-streaming, and then assert that it fails on the server side due to a cardinality violation. Is that correct?

test/end2end_test.go
Pranjali-2501: done.

stream.go
Pranjali-2501: Sorry, I missed committing the changes.

test/end2end_test.go
dfawley: How does that help?  The first `Server.RecvMsg` should be hung waiting for the client to send an END_STREAM, shouldn't it?

stream.go
```diff
@@ -1820,7 +1820,7 @@ func (ss *serverStream) RecvMsg(m any) (err error) {
 	} else if err != nil {
 		return err
 	}
-	return status.Error(codes.Internal, "cardinality violation: expected <EOF> for non client-streaming RPCs, but received another message")
```
dfawley: Non-_client_-streaming was right before, wasn't it?

test/end2end_test.go
dfawley: Correct.  This exact test, pretty much, but set this field to `true`.

test/end2end_test.go
Pranjali-2501: No, it will not hung.
For non-client-streaming RPCs, initial call to `client.SendMsg` will close the client stream with EOF. 
https://github.com/grpc/grpc-go/blob/ac13172781fae5593fd97ce07c3019c4044a71cd/stream.go#L1102-L1102

dfawley: Oh yes, since the client knows it's not client-streaming, too, then it will do END_STREAM along with the first call to SendMsg.

So then the reason for the synchronization here is that, if the client does a RST_STREAM, that might propagate to the server and cause its first recv to fail -- even though the first request message was received perfectly normally -- because we cancel the stream's context upon RST_STREAM receipt?

Pranjali-2501: Yes, correct.

So, should we keep the test as it is or change it with alternative approach?
```
An alternative approach would be to use a fake server implementation, which would allow us to directly observe and assert the error returned from server.RecvMsg().
```

stream.go
Pranjali-2501: Right, didn't realise while copying it.
Made the changes.

test/end2end_test.go
Pranjali-2501: I have modified the test to a table-driven test, where server will run against multiple streamdesc including client-streaming, server-streaming and bidi-streaming.

dfawley: The key difference between the two cases that I'd like to see is whether the client knows it's required to send a message.

In the case where the client knows (`ClientStreams: false`), we should detect the error locally and send a RST_STREAM to the server, but if the client doesn't know (`ClientStreams: true`), the server should detect the error and end the stream with an INTERNAL error.  Can we confirm these things are happening?  (And AFACT the test will fail since the client doesn't check whether it has sent a message in `CloseSend`, so it's fine to make that test case `Skip` until we fix it in another PR.)

test/end2end_test.go
```diff
@@ -3740,6 +3740,238 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	<-recvDoneOnServer
+	if err := stream.SendMsg(&testpb.Empty{}); status.Code(err) != codes.Internal {
+		t.Errorf("stream.SendMsg() = %v, want error %v", err, codes.Internal)
+	}
+	<-handlerDone
+}
+
+// Tests the behavior for unary RPC when client calls SendMsg twice. Second call
+// to SendMsg should fail with Internal error.
+func (s) TestUnaryRPC_ClientCallSendMsgTwice(t *testing.T) {
```
dfawley, line 3808: This is testing client-side checks.  We should also have an equivalent test for server-side checks of the same condition (which does not exist).  We do have tests for logic on the server for server-streaming RPCs, but not for unary.  Unary handling goes through a different code path.

We can add that as a follow-on PR.

test/end2end_test.go
dfawley: No, but can you explain more like how I did, to say _why_ the synchronization is needed?  Because there would be a race between client cancellation and the server reading the first request message.

test/end2end_test.go
```diff
@@ -3740,6 +3740,238 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+	// The initial call to recvMsg made by the generated code, will return the error.
+	ss := stubserver.StubServer{}
+	if err := ss.Start(nil); err != nil {
+		t.Fatal("Error starting server:", err)
+	}
+	defer ss.Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+	cc, err := grpc.NewClient(ss.Address, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
dfawley: Can you switch all these tests to use local credentials instead of insecure?

test/end2end_test.go
```diff
@@ -3740,6 +3740,238 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
+
+		if err := stream.RecvMsg(&testpb.Empty{}); status.Code(err) != tc.wantCode {
+			t.Errorf("stream.RecvMsg() = %v, want error %v", status.Code(err), tc.wantCode)
+		}
+	}
+}
+
+// Tests the behavior of client for server-side streaming RPC when client sends zero request messages.
+func (s) TestServerStreaming_ClientSendsZeroRequests(t *testing.T) {
+	t.Skip()
```
dfawley: Please add a string here and include the issue number (#7286)

test/end2end_test.go
Pranjali-2501: Sure, added a TODO for that.

test/end2end_test.go

test/end2end_test.go

test/end2end_test.go
dfawley: Please leave the function-level comment to just a description of the test, and put the explanation of the various steps in intra-function comments.

test/end2end_test.go
dfawley: Please use the `t.Skip` string instead of adding a comment.

test/end2end_test.go

test/end2end_test.go
```diff
@@ -3741,13 +3741,9 @@ func (s) TestClientStreaming_ReturnErrorAfterSendAndClose(t *testing.T) {
...
-// second SendMsg, it triggers a RST_STREAM which cancels the stream context on the
-// server. There would be a race, the RST_STREAM could cause the server’s first RecvMsg
-// to fail, even if the request message was already delivered. By synchronizing, we
-// ensure that the server has read the first message before the client triggers RST_STREAM
-// and validating expected error codes.
+// Tests the behavior for server-side streaming when client calls SendMsg twice.
+// Second call to SendMsg should fail with Internal error and result in closing
+// the connection with a RST_STREAM.
 func (s) TestServerStreaming_ClientCallSendMsgTwice(t *testing.T) {
 	// To ensure server.recvMsg() is successfully completed.
```
dfawley: What I was hoping for was more like this:

```suggestion
	// To ensure server.recvMsg() is successfully completed.  Otherwise, if the client application
	// attempts to send a second request message, that will trigger a RST_STREAM from the
	// client due to the application violating the RPC's protocol.  The RST_STREAM will prevent
	// the method handler from being called.
```

Issue #8371, opened 2025-05-30
cleanup: Add child balancer in stub.BalancerData


arjan-bal:
Almost all references of [`BalancerData.Data`](https://github.com/grpc/grpc-go/blob/4275c5bdd8e86c450d84f66c9445560fefadc4ec/internal/balancer/stub/stub.go#L47-L54) are storing child balancers in it. This field is of type `any`, so callers have to cast it to a `balancer.Balancer` before every access. Example:
https://github.com/grpc/grpc-go/blob/4275c5bdd8e86c450d84f66c9445560fefadc4ec/internal/idle/idle_e2e_test.go#L103-L106

We should introduce a new field of type `Balancer` to make using the stub balancer more ergonomic. We should update existing usages use the new field wherever possible.

vinothkumarr227:
@arjan-bal Please assign the issue to me.

PR #8393, opened 2025-06-11
cleanup: Add child balancer in stub.BalancerData

vinothkumarr227:
Fixes https://github.com/grpc/grpc-go/issues/8371

RELEASE NOTES: N/A

### DIFF
--- a/internal/balancer/stub/stub.go
+++ b/internal/balancer/stub/stub.go
@@ -51,6 +51,9 @@ type BalancerData struct {
 	BuildOptions balancer.BuildOptions
 	// Data may be used to store arbitrary user data.
 	Data any
+	// ChildBalancer holds the child balancer directly, avoiding the need
+	// to type-assert from Data.
+	ChildBalancer balancer.Balancer
 }
 
 type bal struct {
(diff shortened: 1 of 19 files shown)


internal/balancer/stub/stub.go
```diff
@@ -51,6 +51,9 @@ type BalancerData struct {
 	BuildOptions balancer.BuildOptions
 	// Data may be used to store arbitrary user data.
 	Data any
+	// ChildBalancer holds the child balancer directly, avoiding the need
+	// to type-assert from Data.
```
eshitachandwani: I dont think the second line is needed.

arjan-bal: +1, we should just say `ChildBalancer` holds a child balancer, omitting "directly".

internal/balancer/stub/stub.go
```diff
@@ -51,6 +51,9 @@ type BalancerData struct {
 	BuildOptions balancer.BuildOptions
 	// Data may be used to store arbitrary user data.
 	Data any
```
arjan-bal: If there are no remaining references for `Data`, can you remove the variable?

vinothkumarr227: Done

internal/balancer/stub/stub.go
vinothkumarr227: Done

PR #8410, opened 2025-06-25
Update gRPC-Go's dependency versions

Pranjali-2501:
RELEASE NOTES: N/A


examples/go.mod
```diff
@@ -1,18 +1,20 @@
 module google.golang.org/grpc/examples
 
-go 1.23.0
+go 1.24
```
arjan-bal: gRPC Go supports the last two major releases of Go, which is presently go 1.24 and go 1.23. So we can't bump the major version here. You would need to exclude the dependencies that are causing the go major version bump to 1.24.

arjan-bal:
Had a meeting to debug the problems while excluding dependencies causing the Go version to be updated. Raised https://github.com/grpc/grpc-go/pull/8434.

PR #8419, opened 2025-06-30
xds: Avoid error logs when setting fallback bootstrap config

arjan-bal:
Initially, there was [no error log](https://github.com/grpc/grpc-go/blob/947e2a4be2bac21c84b99114c12c40bb74860afe/internal/xds/bootstrap/bootstrap.go#L819-L829) when setting the fallback bootstrap config. The [GetConfiguration](https://github.com/grpc/grpc-go/blob/947e2a4be2bac21c84b99114c12c40bb74860afe/internal/xds/bootstrap/bootstrap.go#L588-L618) function would first attempt to read the bootstrap configuration from environment variables and then fall back to the configuration provided by the c2p resolver.

A [subsequent change](https://github.com/grpc/grpc-go/pull/7898/files#diff-8c0a0efa1bb340ffe5846ebebb64f9182dcbb1218d0a51a6b3d6d340072ab07e), which added an xDS client pool, modified this behavior. It read the bootstrap config within an init() function in clientimpl.go, preventing the c2p resolver from setting its fallback configuration after channel creation. A fix was implemented ([#8050](https://github.com/grpc/grpc-go/pull/8050)) to allow setting a fallback bootstrap config on an xDS client pool, but this introduced an error log when setting the fallback config twice. This happens when multiple c2p resolvers are created in the same process.

After that, [#8164](https://github.com/grpc/grpc-go/pull/8164) reverted the behavior of reading the bootstrap in the init() function. This allows us to return to the original way—reading from environment variables and then falling back to the c2p resolver's config—without requiring an error log.

RELEASE NOTES:
* googledirectpath: Avoid error the error log `Attempt to set a bootstrap configuration...` when creating multiple directpath channels.

### DIFF
--- a/internal/xds/bootstrap/bootstrap_test.go
+++ b/internal/xds/bootstrap/bootstrap_test.go
@@ -389,7 +389,7 @@ func (s) TestGetConfiguration_Failure(t *testing.T) {
 			testGetConfigurationWithFileContentEnv(t, name, true, nil)
 		})
 	}
-	name := "empty"
+	const name = "empty"
 	t.Run(name, func(t *testing.T) {
 		testGetConfigurationWithFileNameEnv(t, name, true, nil)
 		// If both the env vars are empty, a nil config with a nil error must be
--- a/internal/xds/bootstrap/bootstrap_test.go
+++ b/internal/xds/bootstrap/bootstrap_test.go
@@ -383,12 +383,19 @@ func (s) TestGetConfiguration_Failure(t *testing.T) {
 	cancel := setupBootstrapOverride(bootstrapFileMap)
 	defer cancel()
 
-	for _, name := range []string{"nonExistentBootstrapFile", "empty", "badJSON", "noBalancerName", "emptyXdsServer"} {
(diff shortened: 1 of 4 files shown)


arjan-bal:
FYI @apolcyn

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -383,12 +383,19 @@ func (s) TestGetConfiguration_Failure(t *testing.T) {
...
 	defer cancel()
 
-	for _, name := range []string{"nonExistentBootstrapFile", "empty", "badJSON", "noBalancerName", "emptyXdsServer"} {
+	for _, name := range []string{"nonExistentBootstrapFile", "badJSON", "noBalancerName", "emptyXdsServer"} {
 		t.Run(name, func(t *testing.T) {
 			testGetConfigurationWithFileNameEnv(t, name, true, nil)
 			testGetConfigurationWithFileContentEnv(t, name, true, nil)
 		})
 	}
+	name := "empty"
```
dfawley: Nit: `const` please.

arjan-bal: Changed.

PR #8422, opened 2025-06-30
xdsclient: relay marshalled bytes of complete resource proto to decoders

purnesh42H:
In internal usages of xdsclient, it was found that some resource type implementations (like MintCollection) expect wrapped resource to retrieve some information like `ResourceName`. Currently, xdsclient is sending the underlying raw resource to decoders which can break `MintCollection` resource type implementation ([internal discussion](http://cl/775504257/depot/google3/net/grpc/go/directpath/mint_collection.go?left=r13&right=s5#91L)).

This PR make the change to relay complete resource (including wrapped) to decoders and let them decode as they want.


xds/internal/xdsclient/xdsresource/resource_type.go
```diff
@@ -185,13 +186,16 @@ type GenericResourceTypeDecoder struct {
 // Decode deserialize and validate resource bytes of an xDS resource received
 // from the xDS management server.
 func (gd *GenericResourceTypeDecoder) Decode(resourceBytes []byte, gOpts xdsclient.DecodeOptions) (*xdsclient.DecodeResult, error) {
-	raw := &anypb.Any{TypeUrl: gd.ResourceType.TypeURL(), Value: resourceBytes}
+	rProto := &anypb.Any{}
+	if err := proto.Unmarshal(resourceBytes, rProto); err != nil {
```
dfawley: It seems like such a waste to decode, encode, and re-decode this.

An `Any` is just a `TypeURL` and a `[]byte` with the serialized proto data in it.  Can we pass those two fields directly instead of the `resourceBytes []byte` that is a serialized proto message containing those two fields?  This could then fill those in.  Maybe just make a struct called `AnyProto` so it can extensibly mirror `anypb.Any` without putting a foreign symbol into our API?

purnesh42H: Thanks. That's a good idea. It requires to change the `Decode` interface but its fine since no one is using the new client. Changed.

dfawley (approved):
Thanks for the fix!

xds/internal/clients/xdsclient/helpers_test.go
```diff
@@ -62,17 +63,24 @@ var (
...
 )
 
-func unmarshalListenerResource(r []byte) (string, listenerUpdate, error) {
+func unmarshalListenerResource(rProto *anypb.Any) (string, listenerUpdate, error) {
+	rProto, err := xdsresource.UnwrapResource(rProto)
+	if err != nil {
+		return "", listenerUpdate{}, fmt.Errorf("failed to unwrap resource: %v", err)
+	}
+	if !xdsresource.IsListenerResource(rProto.GetTypeUrl()) {
+		return "", listenerUpdate{}, fmt.Errorf("unexpected listener resource type: %q ", rProto.GetTypeUrl())
```
arjan-bal: nit: Extra space at end of the error string.

purnesh42H: Done

PR #8428, opened 2025-07-04
xds/cdsbalancer: correctly remove the unwanted cds watchers

eshitachandwani:
RELEASE NOTES: 
- xds/cdsbalancer: Fixes panic in cdsBalancer by correctly removing cds watchers for clusters no longer in the cluster tree.

In the code for `onClusterUpdate` after the cluster tree has been updated, we want to remove the watchers for clusters that are not currently in the tree. The existing [code](https://github.com/grpc/grpc-go/blob/aa57e6af6cbc8e1285315b338b092420f014c732/xds/internal/balancer/cdsbalancer/cdsbalancer.go#L533C2-L540C3) will always panic because of the error in condition. This change fixes the removal of unwanted watchers.

### DIFF
--- a/xds/internal/balancer/cdsbalancer/cdsbalancer.go
+++ b/xds/internal/balancer/cdsbalancer/cdsbalancer.go
@@ -443,6 +443,11 @@ func (b *cdsBalancer) annotateErrorWithNodeID(err error) error {
 	return fmt.Errorf("[xDS node id: %v]: %w", nodeID, err)
 }
 
+var onwatcherUpdated = func() {
+	// This function is a no-op, but can be overridden in tests to signal that
+	// the watchers map has been updated.
+}
+
 // Handles a good Cluster update from the xDS client. Kicks off the discovery
 // mechanism generation process from the top-level cluster and if the cluster
 // graph is resolved, generates child policy config and pushes it down.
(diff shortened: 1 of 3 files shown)


xds/internal/balancer/cdsbalancer/cdsbalancer.go
```diff
@@ -443,6 +443,11 @@ func (b *cdsBalancer) annotateErrorWithNodeID(err error) error {
 	return fmt.Errorf("[xDS node id: %v]: %w", nodeID, err)
 }
 
+var onwatcherUpdated = func() {
+	// This function is a no-op, but can be overridden in tests to signal that
+	// the watchers map has been updated.
+}
```
arjan-bal: I think you can avoid adding a function specifically for tests and instead verify that the unsubscription happens by intercepting the requests in the xDS management server. When creating the xDS management server, you can provide an `OnStreamRequest` in `e2e.ManagementServerOptions`. Once the management server removes the cluster, you can wait for an update CDS discovery request from the xds client which should not have the removed cluster.

eshitachandwani: Done. Changed the test.

arjan-bal (approved):
LGTM, thanks for finding and fixing this! Assigning a second reviewer.

arjan-bal:
Please add release notes.

PR #8438, opened 2025-07-08
endpointsharding: shuffle endpoint order before updating children

dfawley:
See b/412468630, and esp. comment 46.  Essentially, if the connection order is not randomized then the first address gets hammered with all the client's initial requests when traffic is bursty.  If multiple clients are all using the same order and many start up around the same time, then the first address can see that problem magnified even further.  Randomizing at least spreads the load between the clients, even if it still results in one address getting hammered at startup.

The diffs are much smaller than it shows.  Basically, the existing `e.s._test.go` was moved to `..._ext_test.go` and the only new code is in `..._test.go`.  `_ext_test.go` is completely unchanged from the old `_test.go`.  Weightedtarget's tests needed changes because they assumed the order in which subchannels were created matched the order in which endpoints were provided.

RELEASE NOTES:
* round_robin: randomize the order in which addresses are connected to in order to spread out initial RPC load between clients.

### DIFF
--- a/balancer/endpointsharding/endpointsharding.go
+++ b/balancer/endpointsharding/endpointsharding.go
@@ -37,6 +37,8 @@ import (
 	"google.golang.org/grpc/resolver"
 )
 
+var randIntN = rand.IntN
+
 // ChildState is the balancer state of a child along with the endpoint which
 // identifies the child balancer.
 type ChildState struct {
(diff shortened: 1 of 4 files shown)


balancer/endpointsharding/endpointsharding.go
```diff
@@ -112,6 +114,21 @@ type endpointSharding struct {
 	mu sync.Mutex
 }
 
+// rotateEndpoints returns a slice of all the input endpoints rotated a random
+// amount.
+func rotateEndpoints(es []resolver.Endpoint) []resolver.Endpoint {
```
markdroth, line 119: It looks like this actually changes the order of the address list, rather than simply starting to connect from a random index.  But maybe that amounts to the same thing, since each new picker starts from a random index anyway in RR?

dfawley: This function rotates the slice a random amount -- it puts all elements after the random point at the start of a new slice, and the elements before it at the end.

balancer/weightedtarget/weightedtarget_test.go
```diff
@@ -670,6 +680,10 @@ func (s) TestWeightedTarget_TwoSubBalancers_MoreBackends(t *testing.T) {
 	}); err != nil {
 		t.Fatalf("failed to update ClientConn state: %v", err)
 	}
+
+	// The sc# corresponding to addr3 is not known due to possible randomization
+	// in connection order in RR.  Check the address instead of sc and set
```
eshitachandwani: If I understand correctly , we have changed sc3 to correspond to address 3 in the above change. I am not sure what this comment accomplishes?

dfawley: Oops, I changed the way I did this and left this comment from the old way behind.  Removed

PR #8439, opened 2025-07-08
server: allow 0s grpc-timeout header values, as java is known to be able to send them

dfawley:
https://github.com/grpc/grpc-java/pull/12201 was recently created to prevent this, but older versions will still do it.  This reverts our old behavior of allowing 0 second timeouts (pre- #8290).

I would propose that we do not revert the change in http_util and allow the RPC to instantly timeout instead of throw an internal error.

cc @ejona86 

RELEASE NOTES:
* server: allow 0s grpc-timeout header values, which older gRPC-Java versions could send

### DIFF
--- a/internal/transport/http2_server.go
+++ b/internal/transport/http2_server.go
@@ -600,10 +600,24 @@ func (t *http2Server) operateHeaders(ctx context.Context, frame *http2.MetaHeade
 			return nil
 		}
 	}
+
+	if s.ctx.Err() != nil {
+		// Early abort in case the timeout was zero or so low it already fired.
+		t.controlBuf.put(&earlyAbortStream{
+			httpStatus:     http.StatusOK,
+			streamID:       s.id,
+			contentSubtype: s.contentSubtype,
+			status:         status.New(codes.DeadlineExceeded, context.DeadlineExceeded.Error()),
+			rst:            !frame.StreamEnded(),
+		})
+		return nil
+	}
+
 	t.activeStreams[streamID] = s
 	if len(t.activeStreams) == 1 {
 		t.idle = time.Time{}
 	}
+
 	// Start a timer to close the stream on reaching the deadline.
 	if timeoutSet {
 		// We need to wait for s.cancel to be updated before calling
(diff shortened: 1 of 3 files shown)


arjan-bal (approved):

colega:
Please note that this change has changed the behavior of `ServerInHandle` (`gprc.InTapHandle` option).

Previously, when `ServerInHandle` returned no error, the request would be processed, so any resource changes (like inflight requests counter) performed in `ServerInHandle` could be later undone though a stats handler.

Now, however, there's no such guarantee anymore.

I wonder if it would be possible to move this early return before the inTapHandle call?

dfawley:
The reason I put this after the intaphandle call is because that is able to change the context to one that has already expired.

I'm also curious about your use case.  Can you not track in-flight requests entirely within the stats handler (increment on `Begin`)?  Or if you need to _restrict_ parallelism, note that you can also do that via `MaxConcurrentStreams`, which will let the client know about your limit as well.

colega:
_Sorry for the delayed response, this has been sitting on my TODO list for a while._

We do want to _restrict_ parallelism, but we want a global limit, not per-client. We want to restrict the total amount of connections that the given server may handle, and `MaxConcurrentStreams` doesn't seem to help with that.

dfawley:
Do you want to limit the total number of streams on a server across all connections?  Or just the total number of connections?  The latter can be accomplished by via https://pkg.go.dev/golang.org/x/net/netutil#LimitListener.

You could also use interceptors instead of the tap/stats combo you are using now.  That is the approach that would translate to other gRPC language implementations which don't have anything like our tap handler thing.  There's a small amount of extra CPU cycles that go into interceptors vs. the tap handler, but it shouldn't be too bad.

But also unless your clients are looking at the error and doing some kind of custom backoff logic, then I'm not sure the general approach is ideal.  Using `MaxConcurrentStreams` and a `LimitListener` instead would help you protect your server and also advertise the stream limit to clients so they would naturally get pushback when they try to create too many streams.  This approach also prevents one client from stealing all of the available streams from the other clients.

Issue #8311, opened 2025-05-12
stats: Update documentation and examples to demonstrate enabling `experimental` metrics


purnesh42H:
There has been several issues opened recently like [this](https://github.com/grpc/grpc-go/issues/8234) asking about not being able to see experimental metrics from wrr and xds. It is because they are not enabled by default and needs to be added to the DefaultMetrics in the dial and server options.

This is the list of all experimental metrics the list of all experimental metrics https://grpc.io/docs/guides/opentelemetry-metrics/#instruments. We should include this part in the example and also sample code for adding experimental metrics.

vinothkumarr227:
@purnesh42H I'm looking into this. Please assign the issue to me.

purnesh42H:
Noticed that we have go doc examples here https://github.com/grpc/grpc-go/blob/master/stats/opentelemetry/example_test.go. We should add one for appending the experimental metrics here as well. We have for removing a metric from `DefaultMetrics` but not for adding.

Also, the godoc doesn't show all the examples from `example_test.go` https://pkg.go.dev/google.golang.org/grpc/stats/opentelemetry#pkg-examples. We should update the names as per specifications here https://go.dev/blog/examples

PR #8388, opened 2025-06-09
examples/opentelemetry: demonstrate enabling experimental metrics

vinothkumarr227:
Fixes https://github.com/grpc/grpc-go/issues/8311

RELEASE NOTES:
* examples: Add examples to demonstrate enabling experimental metrics using the opentelemetry plugin.

Issue #8234, opened 2025-04-08
Expose xDS updates metrics for library clients


sevaorlov:
Hi, how can I get xDS updates metrics?

I am interested in monitoring xDS updates frequency to our service and noticed there was metrics recorder added in this PR https://github.com/grpc/grpc-go/pull/8038. 

I don't understand is how I can provide my own metrics recorder implementation and get those metrics.

What I ended up doing is fetching the resolver and wrapping it together with metrics recorder.

```
base := resolver.Get("xds")
resolver.Register(wrappedResolver{baseResolver: base, context: ctx, recordUpdate: recordUpdate})
```

```

type wrappedResolver struct {
	baseResolver resolver.Builder
	recordUpdate metricRecorderFunc
}

func (r wrappedResolver) Build(target resolver.Target, cc resolver.ClientConn,
	opts resolver.BuildOptions) (resolver.Resolver, error) {
	opts.MetricsRecorder = wrappedMetricsRecorder{
		base:         opts.MetricsRecorder,
		recordUpdate: r.recordUpdate,
	}

	return r.baseResolver.Build(target, cc, opts)
}

func (r wrappedResolver) Scheme() string {
	return r.baseResolver.Scheme()
}
```

```
type metricRecorderFunc func(ctx context.Context, incr int64, name string)

type wrappedMetricsRecorder struct {
	base stats.MetricsRecorder
	recordUpdate metricRecorderFunc
}

func (w wrappedMetricsRecorder) RecordInt64Count(handle *stats.Int64CountHandle, incr int64, labels ...string) {
	if w.base != nil {
		w.base.RecordInt64Count(handle, incr, labels...)
	}

	w.recordUpdate(incr, handle.Name)
}

func (w wrappedMetricsRecorder) RecordFloat64Count(handle *stats.Float64CountHandle, incr float64, labels ...string) {
	if w.base != nil {
		w.base.RecordFloat64Count(handle, incr, labels...)
	}
}

func (w wrappedMetricsRecorder) RecordInt64Histo(handle *stats.Int64HistoHandle, incr int64, labels ...string) {
	if w.base != nil {
		w.base.RecordInt64Histo(handle, incr, labels...)
	}
}

func (w wrappedMetricsRecorder) RecordFloat64Histo(handle *stats.Float64HistoHandle, incr float64, labels ...string) {
	if w.base != nil {
		w.base.RecordFloat64Histo(handle, incr, labels...)
	}
}

[...]

PR #8441, opened 2025-07-10
examples/opentelemetry:  enabling experimental metrics

vinothkumarr227:
Fixes https://github.com/grpc/grpc-go/issues/8311

RELEASE NOTES:

### DIFF
--- a/examples/features/opentelemetry/client/main.go
+++ b/examples/features/opentelemetry/client/main.go
@@ -70,8 +70,8 @@ func main() {
 			// up-to-date list of metrics, see:
 			// https://grpc.io/docs/guides/opentelemetry-metrics/#instruments
 			Metrics: opentelemetry.DefaultMetrics().Add(
-				"grpc.client.attempt.started",
-				"grpc.client.attempt.duration",
+				"grpc.xds_client.resource_updates_valid",
+				"grpc.xds_client.resource_updates_invalid",
 			),
 		},
 		TraceOptions: oteltracing.TraceOptions{TracerProvider: traceProvider, TextMapPropagator: textMapPropagator},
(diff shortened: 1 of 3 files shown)


eshitachandwani:
Please link the comment in response to which this PR has been raised. @vinothkumarr227

vinothkumarr227:
@arjan-bal This fix was made following the review comment from in [this discussion](https://github.com/grpc/grpc-go/pull/8388#discussion_r2160773043)

examples/features/opentelemetry/client/main.go
```diff
@@ -70,8 +70,8 @@ func main() {
 			// up-to-date list of metrics, see:
 			// https://grpc.io/docs/guides/opentelemetry-metrics/#instruments
 			Metrics: opentelemetry.DefaultMetrics().Add(
-				"grpc.client.attempt.started",
-				"grpc.client.attempt.duration",
+				"grpc.xds_client.resource_updates_valid",
```
eshitachandwani: Correct me if I am wrong but I dont think we are using xds in this example , so using xds metrics might not make sense. Can you change to use some other metrics? Example : `grpc.lb.pick_first.
connection_attempts_succeeded`

vinothkumarr227: @eshitachandwani , Thanks for the review! I don’t think we need to change this example — we’ve already mentioned that it’s meant to demonstrate experimental metrics. Also, this is aligned with the discussion here: [grpc-go#8234 (comment)](https://github.com/grpc/grpc-go/issues/8234#issuecomment-2812415022).

eshitachandwani: But the example in my opinion should be a working example , like if someone is trying to run the example , they should be able to see some metrics. And since the example is not using xds , there will be no metrics for the xds client.

vinothkumarr227: Thanks for the feedback! I’ve updated the example to use pick_first with a valid address, and it now emits metrics

eshitachandwani:
Adding @arjan-bal for a second review.

Issue #8421, opened 2025-06-30
jwt audience URI vs uir passed to PerRPCCredentials.GetRequestMetadata


ukai:
re: https://issues.chromium.org/427932818 Credential helper not called with the full endpoint

grpc-go strips method from uri when passing to PerRPCCredentials.GetRequestMetadata
https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_client.go#L645

credentials/oauth jwt code also strips service name from uri
https://github.com/grpc/grpc-go/blob/v1.73.0/credentials/oauth/oauth.go#L61

but jwt documentation says
>  Service account JWTs are limited to the specified audience. You must set the audience to the exact URL and path of the IAP-secured resource
https://cloud.google.com/iap/docs/authentication-howto#creating_the_jwt

don't we need to pass exact URI to GetRequestMetadata to create valid jwt?

matthewstevenson88:
Deferring to @gtcooke94 here.

gtcooke94:
This is a pretty old section of code which makes me hesitant to change - I worry it would break a lot of existing workflows.

I found the PR where it was added - https://github.com/grpc/grpc-go/pull/4713, which sounds like it was explicitly changed to work with a different spec. It looks like that got discussed on the Chromium issue mentioning that the JWTs are not handled consistently.

It also seems like there's not one "correct" way to handle this - would a environment flag that can optionally disable the service name stripping work?

ukai:
hmm, why google designed inconsistently?  is there any comprehensive documents?

yeah, env to disable service name/method name stripping will work.

chrisstaite-menlo:
Hi, I was just wondering if there was any traction here to resolve this?  Happy to make a commit if necessary, although I did already suggest a solution to retrieve the method from the Context: https://chromium-review.googlesource.com/c/infra/infra/+/6686992

gtcooke94:
Feel free to make a commit! It should be a simple env var boolean check to not execute the code block, with the default behavior being the current behavior

chrisstaite-menlo:
Don't suppose it's possible to get this in to the 1.74 release?

dfawley:
Sorry, but this should have been discussed cross-language (or at least more broadly within the gRPC-Go team) before being approved and merged, so we are rolling it back.

[...]

PR #4713, opened 2021-08-27
oauth: Allow access to Google API regional endpoints via Google Default Credentials

yihuazhang:
The self-signed JWT used to access Google API regional endpoints do not have a correct audience format dictated in https://google.aip.dev/auth/4111. That is, it includes a trailing RPC service name while the spec requires only the presence of hostname (or endpoint).

RELEASE NOTES:
* oauth: Allow access to Google API regional endpoints via Google Default Credentials

PR #8442, opened 2025-07-10
credentials: allow audience to be configured (#8421)

chrisstaite-menlo:
Fixes: https://github.com/grpc/grpc-go/issues/8421
There are competing specifications around whether a method should be included in a JWT audience or not.  For example #4713 specifically excluded the method referencing https://google.aip.dev/auth/4111 whereas GCE IAP requires the full URI https://cloud.google.com/iap/docs/authentication-howto.

In order to facilitate both methods, we introduce a new environment variable, namely GRPC_AUDIENCE_IS_FULL_PATH, to allow the method stripping to be disabled.  This defaults to the existing behaviour of stripping the method, but can be set to avoid this.

RELEASE NOTES:
* credentials: Add support to disable gRPC method name stripping for JWT audience calculation when the environment variable `GRPC_AUDIENCE_IS_FULL_PATH` is set to true.

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -39,6 +39,7 @@ import (
 	"google.golang.org/grpc/internal"
 	"google.golang.org/grpc/internal/channelz"
 	icredentials "google.golang.org/grpc/internal/credentials"
+	"google.golang.org/grpc/internal/envconfig"
 	"google.golang.org/grpc/internal/grpclog"
 	"google.golang.org/grpc/internal/grpcsync"
 	"google.golang.org/grpc/internal/grpcutil"
@@ -645,6 +646,10 @@ func (t *http2Client) createAudience(callHdr *CallHdr) string {
 	// Construct URI required to get auth request metadata.
 	// Omit port if it is the default one.
 	host := strings.TrimSuffix(callHdr.Host, ":443")
+	if envconfig.AudienceIsFullPath {
+		return "https://" + host + callHdr.Method
+	}
+
 	pos := strings.LastIndex(callHdr.Method, "/")
 	if pos == -1 {
 		pos = len(callHdr.Method)
--- a/test/creds_test.go
+++ b/test/creds_test.go
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/internal/envconfig"
 	"google.golang.org/grpc/internal/testutils"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/resolver"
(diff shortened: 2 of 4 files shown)


test/creds_test.go
```diff
@@ -543,3 +544,57 @@ func (s) TestServerCredsDispatch(t *testing.T) {
 		t.Errorf("Read() = %v, %v; want n>0, <nil>", n, err)
 	}
 }
+
+type audienceTestCreds struct{}
+
+func (a *audienceTestCreds) GetRequestMetadata(_ context.Context, uri ...string) (map[string]string, error) {
+	endpoint := ""
```
eshitachandwani:
```suggestion
	var endpoint string
```

test/creds_test.go
```diff
@@ -543,3 +544,57 @@ func (s) TestServerCredsDispatch(t *testing.T) {
...
+	if len(uri) > 0 {
+		endpoint = uri[0]
+	}
+	return nil, status.Error(codes.Unknown, endpoint)
+}
+
+func (a *audienceTestCreds) RequireTransportSecurity() bool { return false }
+
+func (s) TestGRPCMethodInAudienceWhenEnvironmentSet(t *testing.T) {
+	oldAudienceIsFullPath := envconfig.AudienceIsFullPath
```
eshitachandwani: This can be changed to use the testutil function to set and unset the environment config.
`testutils.SetEnvConfig(t, &envconfig.AudienceIsFullPath, true)`
This will set to true and reset to original value in the end.

test/creds_test.go
```diff
@@ -543,3 +544,57 @@ func (s) TestServerCredsDispatch(t *testing.T) {
...
+	defer cancel()
+	if _, err := tc.EmptyCall(ctx, &testpb.Empty{}); status.Convert(err).Message() != wantMethod {
+		t.Fatalf("ss.client.EmptyCall(_, _) = _, %v; want _, _.Message()=%q", err, wantMethod)
+	}
+
+	if _, err := tc.EmptyCall(ctx, &testpb.Empty{}, grpc.WaitForReady(true)); status.Convert(err).Message() != wantMethod {
+		t.Fatalf("ss.client.EmptyCall(_, _) = _, %v; want _, _.Message()=%q", err, wantMethod)
+	}
+
+	envconfig.AudienceIsFullPath = false
```
eshitachandwani: Here , we are setting the env variable but never resetting it back to original. These can be two seperate test or one test with table cases for two different cases. [Example](https://github.com/grpc/grpc-go/blob/af2600d31cafb081f5e6f6b91a597c526c530ac3/resolver/resolver_test.go#L37)

eshitachandwani:
Please add release notes.

eshitachandwani:
Adding @gtcooke94 as another reviewer.

test/creds_test.go
chrisstaite-menlo: Nice :)

test/creds_test.go
chrisstaite-menlo: It's reset in the original defer.

chrisstaite-menlo:
@eshitachandwani I'm sorry, I'm not sure what you mean by release notes, I don't see any in the repository.  I have updated the documentation around the environment variable however.

Oh, I think I see... it should be set in the PR description.

I'm confused, I edited the PR description and the validation is still failing.  Are you able to help me figure out where I've gone wrong please?

Final edit: figured it out, I had a space before the *, reading the RegExp helped.

eshitachandwani:
LGTM , adding @gtcooke94 to review.

internal/transport/http2_client.go
```diff
@@ -646,7 +647,7 @@ func (t *http2Client) createAudience(callHdr *CallHdr) string {
 	// Omit port if it is the default one.
 	host := strings.TrimSuffix(callHdr.Host, ":443")
 	pos := strings.LastIndex(callHdr.Method, "/")
-	if pos == -1 {
+	if pos == -1 || envconfig.AudienceIsFullPath {
```
gtcooke94: What do you think of making this more explicit:
```
	host := strings.TrimSuffix(callHdr.Host, ":443")
        if envconfig.AudienceIsFullPath {
         return "https://" + host + callHdr.Method       
         }

	pos := strings.LastIndex(callHdr.Method, "/")
	if pos == -1  {
		pos = len(callHdr.Method)
	}
	return "https://" + host + callHdr.Method[:pos]
```

test/creds_test.go
```diff
@@ -543,3 +544,58 @@ func (s) TestServerCredsDispatch(t *testing.T) {
...
+	te.startServer(&testServer{security: te.e.security})
+	defer te.tearDown()
+
+	cc := te.clientConn(grpc.WithPerRPCCredentials(&audienceTestCreds{}))
+	tc := testgrpc.NewTestServiceClient(cc)
+
+	tests := []struct {
+		name     string
+		endpoint string
+		fullPath bool
```
gtcooke94: Suggest renaming to `audienceIsFullPath`

internal/transport/http2_client.go
chrisstaite-menlo: Sure, I was going for minimal change, but happy to be explicit.

gtcooke94 (approved):
LGTM - by the way, the commits will automatically be squashed into 1 when the PR is merged, so you don't need to worry about squashing and force-pushing a single commit each update

Thanks for the PR!

chrisstaite-menlo:
Thanks, a lot of those were to force a re-build because the arm64 tests seem *super* flaky.

eshitachandwani:
Reverting this PR to allow for broader cross-language discussion prior to making a decision on incorporating this feature.

chrisstaite-menlo:
That's rather frustrating... this has a severe effect on our ability to use Google Cloud due to its implementation of IAP.

PR #8444, opened 2025-07-10
alts: improve alts handshaker error logs

rockspore:


credentials/alts/internal/handshaker/handshaker.go
```diff
@@ -298,11 +298,11 @@ func (h *altsHandshaker) doHandshake(req *altspb.HandshakerReq) (net.Conn, *alts
 
 func (h *altsHandshaker) accessHandshakerService(req *altspb.HandshakerReq) (*altspb.HandshakerResp, error) {
 	if err := h.stream.Send(req); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to send handshaker request: %w", err)
```
matthewstevenson88: Can we include ALTS in the error message, as handshaker might be too generic? For example, "failed to send ALTS handshaker request"? Similarly for the response.

rockspore: Good point. Done.

matthewstevenson88 (approved):

PR #8445, opened 2025-07-10
xds: give up pool lock before closing xdsclient channel

dfawley:
This prevents deadlocks with nested xds channels.

I believe this must be cherry-picked into the 1.74 release branch, too.

RELEASE NOTES: none

### DIFF
--- a/xds/internal/xdsclient/pool.go
+++ b/xds/internal/xdsclient/pool.go
@@ -241,6 +241,9 @@ func (p *Pool) clientRefCountedClose(name string) {
 	}
 	p.mu.Unlock()
 
+	// This attempts to close the transport to the management server and could
+	// theoretically call back into the xdsclient package again and deadlock.
+	// Hence, this needs to be called without holding the lock.
 	client.Close()
 
 	xdsClientImplCloseHook(name)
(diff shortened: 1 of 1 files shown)


purnesh42H (approved):
Thanks for finding and fixing it back. Perhaps we should file an issue for writing the test?

xds/internal/xdsclient/pool.go
```diff
@@ -227,7 +227,6 @@ func (p *Pool) clientRefCountedClose(name string) {
 	}
 	delete(p.clients, name)
 
-	client.Close()
 	for _, s := range client.bootstrapConfig.XDSServers() {
```
purnesh42H, line 230: nit: Even these 2 cleanups of authorities and xds servers were being done after releasing the lock before because they were within client's close method. May be we should move them down as well after closing the generic client.

arjan-bal:
> Thanks for finding and fixing it back. Perhaps we should file an issue for writing the test?

In my opinion, we should have added a test when we first fixed the issue. Now that it has occurred again, I think we should add the test as soon as possible. I can try to write it before we attempt to import the fix back into google3.

xds/internal/xdsclient/pool.go
dfawley: I'm worried about actual races here, since the `bootstrapConfig` appears to be a pointer, and could be modified later.

PR #8448, opened 2025-07-14
xds: add a test for deadlocks in nested xDS channels

arjan-bal:
This PR adds a test to catch deadlocks when using xDS to resolve the address of the xDS management server, resulting in the creation of nested xDS channels. The fix for the bug was initially submitted in https://github.com/grpc/grpc-go/pull/8011 and later in https://github.com/grpc/grpc-go/pull/8445.

## Tested
Verified that the deadlock can be reproduced in 2-3 runs when the fix in https://github.com/grpc/grpc-go/pull/8445 is not included. The deadlock is resolved after including https://github.com/grpc/grpc-go/pull/8445.

RELEASE NOTES: N/A

### DIFF
--- a/xds/internal/xdsclient/pool/pool_test.go
+++ b/xds/internal/xdsclient/pool/pool_test.go
@@ -19,17 +19,33 @@
 package pool_test
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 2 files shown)


xds/internal/xdsclient/pool/pool_test.go
```diff
@@ -19,18 +19,34 @@
 package pool_test
```
dfawley: Aside: it seems this file should have been named `pool_ext_test.go` to prepare for the day we might want `pool_test.go` that is inside `package pool`.

arjan-bal: Renamed the file and package.

PR #8456, opened 2025-07-17
Retract v1.74.0 and v1.74.1

dfawley:
RELEASE NOTES: none

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,3 +42,6 @@ require (
 	golang.org/x/text v0.26.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20250707201910-8d1bb00bc6a7 // indirect
 )
+
+// v1.74.0 was published prematurely with known issues.
+retract [v1.74.0, v1.74.1]


go.mod
```diff
@@ -42,3 +42,6 @@ require (
 	golang.org/x/text v0.25.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20250528174236-200df99c418a // indirect
 )
+
+// v1.74.0 was published prematurely with known issues.
+retract [v1.74.0, v1.74.1]
```
arjan-bal, line 47: Why are we retracting 1.74.1 (unreleased)?

dfawley:
To retract a release, you need to perform a new, latest release.  That one will retract itself.  This is following along with the guide published by the Go team here:

https://play-with-go.dev/retract-module-versions_go119_en/

> To retract v1.0.0 you will need to publish v1.0.1. But that means you will also need to retract version v1.0.1. You do so using a closed range.

Issue #8453, opened 2025-07-16
Duplicate Delayed LB pick complete event emitted by old pick_first policy.


vinothkumarr227:
The old pick_first LB sends the "Delayed LB pick complete" event twice due to redundant picker updates—once on the first resolver update and again when the subchannel connects. The new pick_first correctly de-duplicates these events.

Reference Issues: https://github.com/grpc/grpc-go/pull/8342#discussion_r2207440697

dfawley:
This is a bug with how we implemented the stats callout; it was only intended to be emitted once per attempt, but we did it every time the LB policy asked for the attempt to be queued.  This should be fixed by #8465.

PR #8465, opened 2025-07-18
stats: add DelayedPickComplete and follow correct semantics

dfawley:
Fixes: #8453

This is part of a two-step change to eliminate the potentially-multiple calls for `PickerUpdated` as we do today. Other languages only have a maximum of one event for this, when the pick finally completed, but in Go we implemented it as a call every time the picker was updated. The plan is to implement the full change as follows:

1. Add DelayedPickComplete and make PickerUpdated a type alias of it, and change the semantics to match other languages (v1.x).
2. Delete PickerUpdated (v1.x+1)

RELEASE NOTES:
* stats: introduce `DelayedPickComplete` event, a type alias of `PickerUpdated`. This (combined) event will now be emitted only once per call, when a transport is successfully selected for the attempt.  OpenTelemetry metrics will no longer have multiple "Delayed LB pick complete" events in Go, matching other gRPC languages.  A future release will delete `PickerUpdated`.

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -76,7 +76,8 @@ func (*DelayedPickComplete) isRPCStats() {}
 // PickerUpdated indicates that the RPC is unblocked following a delay in
 // selecting a connection for the call.
 //
-// Deprecated: will be removed in a future release.
+// Deprecated: will be removed in a future release; use DelayedPickComplete
+// instead.
 type PickerUpdated = DelayedPickComplete
 
 // InPayload contains stats about an incoming payload.
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -73,6 +73,12 @@ func (*DelayedPickComplete) IsClient() bool { return true }
 
 func (*DelayedPickComplete) isRPCStats() {}
 
+// PickerUpdated indicates that the RPC is unblocked following a delay in
+// selecting a connection for the call.
+//
+// Deprecated: will be removed in a future release.
+type PickerUpdated = DelayedPickComplete
+
 // InPayload contains stats about an incoming payload.
 type InPayload struct {
 	// Client is true if this InPayload is from client side.
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -64,15 +64,14 @@ func (s *Begin) IsClient() bool { return s.Client }
 
 func (s *Begin) isRPCStats() {}
 
-// PickerUpdated indicates that the LB policy provided a new picker while the
-// RPC was waiting for one.
-type PickerUpdated struct{}
+// DelayedPickComplete indicates that the RPC is unblocked following a delay in
+// selecting a connection for the call.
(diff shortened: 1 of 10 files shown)


stats/stats.go
```diff
@@ -64,15 +64,20 @@ func (s *Begin) IsClient() bool { return s.Client }
...
+// IsClient indicates DelayedPickComplete is available on the client.
+func (*DelayedPickComplete) IsClient() bool { return true }
 
-func (*PickerUpdated) isRPCStats() {}
+func (*DelayedPickComplete) isRPCStats() {}
+
+// PickerUpdated indicates that the RPC is unblocked following a delay in
+// selecting a connection for the call.
+//
+// Deprecated: will be removed in a future release.
```
arjan-bal: Should we also specify that `DelayedPickComplete` should be used instead?

stream.go
```diff
@@ -481,6 +482,11 @@ func (a *csAttempt) getTransport() error {
 	if a.trInfo != nil {
 		a.trInfo.firstLine.SetRemoteAddr(a.transport.RemoteAddr())
 	}
+	if pick.blocked {
+		for _, sh := range a.statsHandlers {
+			sh.HandleRPC(a.ctx, &stats.DelayedPickComplete{})
```
arjan-bal, line 487: Should we also call `stats.PickerUpdated` to keep the previous behaviour until `PickerUpdated` is removed?

test/end2end_test.go
```diff
@@ -6649,13 +6649,13 @@ func (s) TestRPCBlockingOnPickerStatsCall(t *testing.T) {
...
-		if _, ok := stat.(*stats.PickerUpdated); ok {
-			pickerUpdatedCount++
+		if _, ok := stat.(*stats.DelayedPickComplete); ok {
+			delayedPickCompleteCount++
 		}
 	}
-	if pickerUpdatedCount != 1 {
-		t.Fatalf("sh.pickerUpdated count: %v, want: %v", pickerUpdatedCount, 2)
+	if delayedPickCompleteCount != 1 {
+		t.Fatalf("sh.delayedPickComplete count: %v, want: %v", delayedPickCompleteCount, 2)
```
arjan-bal: We compare `delayedPickCompleteCount` to `1`, but the log mentions `2`. I think the log message needs to be fixed.

stats/stats.go
dfawley: Yes.  Related: I think technically we're not supposed to put "Deprecated" on this until after `DelayedPickComplete` exists for a release.  But since this package is experimental, I think we can skip that and call it deprecated immediately.

stream.go
dfawley: It's a type alias so the stats handler will get one event that matches both of the events.

test/end2end_test.go
dfawley: Fixed.

Issue #8430, opened 2025-07-07
Remove client-specific retry attributes from server spans (per A72 tracing proposal)


vinothkumarr227:
Per the [gRPC OpenTelemetry tracing proposal A72](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md), retry information (e.g., previous-rpc-attempts, transparent-retry) is only tracked on the client side.

Currently, server spans incorrectly include client-specific retry attributes, which is not compliant with A72. This issue tracks the work required to remove these incorrect attributes and add appropriate unit tests.

PR #8466, opened 2025-07-21
stats/opentelemetry: remove client retry attributes from server spans

vinothkumarr227:
Fixes: https://github.com/grpc/grpc-go/issues/8430

RELEASE NOTES:

stats/opentelemetry: Removed client-specific retry attributes from server spans to align with tracing proposal A72.


vinothkumarr227:
This PR should be put on hold until https://github.com/grpc/grpc-go/pull/8342 is merged. We can review this PR after that.

PR #8479, opened 2025-07-30
interop/xds: Avoid setting unit suffixes

arjan-bal:
The prometheus exporter changed it's behaviour to add not standard unit suffixes to metrics: https://github.com/open-telemetry/opentelemetry-go/pull/6839. As a result `grpc_server_call_started_total` becomes `grpc_server_call_started_call_total`.

From the [release notes](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.37.0):
> Changed handling of go.opentelemetry.io/otel/exporters/prometheus metric renaming to add unit suffixes when it doesn't match one of the pre-defined values in the unit suffix map. 

Similar fix in another project: https://github.com/pomerium/pomerium/pull/5749/files

## Tested
Verified by running the otel examples that the units are no longer part of the metric names.

RELEASE NOTES: N/A

### DIFF
--- a/examples/features/csm_observability/client/main.go
+++ b/examples/features/csm_observability/client/main.go
@@ -50,7 +50,7 @@ var (
 
 func main() {
 	flag.Parse()
-	exporter, err := prometheus.New()
+	exporter, err := prometheus.New(prometheus.WithoutUnits())
 	if err != nil {
 		log.Fatalf("Failed to start prometheus exporter: %v", err)
 	}
--- a/examples/features/csm_observability/server/main.go
+++ b/examples/features/csm_observability/server/main.go
@@ -57,7 +57,7 @@ func (s *server) SayHello(_ context.Context, in *pb.HelloRequest) (*pb.HelloRepl
 
 func main() {
 	flag.Parse()
-	exporter, err := prometheus.New()
+	exporter, err := prometheus.New(prometheus.WithoutUnits())
 	if err != nil {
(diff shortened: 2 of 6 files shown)


arjan-bal (commented):
I went through the otel PRs that caused the behaviour change again and I think that there is a bug in gRPCs otel plugin:
https://github.com/grpc/grpc-go/blob/ac13172781fae5593fd97ce07c3019c4044a71cd/stats/opentelemetry/server_metrics.go#L57

We're specifying the unit as `call` instead of `{call}`. Based on the [metric naming conventions](https://ucum.org/ucum#section-Grammar-of-Units-and-Unit-Terms:~:text=%C2%A76-,curly%20braces,-%E2%96%A01%20The%20full) anything enclosed in curly braces is an annotation and in the absence of a prefix, the unit is `1`. The gRPC docs also mention that non-standard units need to be enclosed in curly braces: https://grpc.io/docs/guides/opentelemetry-metrics/#server-instruments.

I believe the correct fix would be to follow the gRPC docs and enclose the required units in `{}`, however I'm not sure how this would impact existing users of the gRPC otel plugin.

@dfawley @easwars can you comment on this?

dfawley:
Oh good catch.  I would classify that as a bug fix.  We should include it in the release notes, but I don't think we have much other option here....

arjan-bal:
Closing this PR, will open another PR with the correct fix.

PR #8492, opened 2025-08-06
chore: setup golangci-lint

mmorel-35:
#### Description

Following https://github.com/grpc/grpc-go/pull/7994

Compared to the referenced issue, this provide golangci-lint along the existing linters. This will help answer https://github.com/grpc/grpc-go/pull/7994#issuecomment-2596160644 interrogations.

It adds golangci-lint in test/tools and configure it to cover the actual behavior of the linters.
It also fixes several issues raised by those linters.
There is one commit per rule so it can be reviewed individually and droped if you disagree.

RELEASE NOTES: None


dfawley:
What is the agenda for adding `golangci-lint` to our repo?  We already run `staticcheck` directly.  It seems this PR accomplishes two positive things:

- Running the spelling linter on sub-modules.
  - This can be done by moving the `misspell` check inside the loop across modules instead.
- Enabling additional linters in `staticcheck` and acting on them.
  - This can be done by updating the version of `staticcheck` (?) or by adding these checks to the list of things we have it include when it's run.

Bringing in this extra tool adds many tens of new dependencies for questionable benefit and increases the complexity of the system and results in inefficiencies like running staticcheck twice.  None of these are desirable.

The actual code changes here all look like reasonable things to do, but I'm not sure I agree that including `golangci-lint` itself is appropriate.

mmorel-35:
Thank you for the detailed feedback and for highlighting the project's priorities.

The intention of this pull request is to introduce golangci-lint alongside the existing linters, not to replace them. This approach enables a transition period where both tools operate in parallel, allowing for thorough evaluation and verification that golangci-lint provides equivalent or improved coverage compared to the current setup. Once alignment and confidence are established, consolidation to a single tool can be considered if it aligns with project goals and contributor consensus.

**Key benefits of this approach include:**

- **Centralized and Readable Configuration:** All linter rules and exclusions are managed in a single YAML file, reducing the complexity of bash scripts and making the linting setup more accessible to contributors.
- **Automatic Exclusion Management:** golangci-lint provides warnings when exclusions are no longer necessary, ensuring that the configuration remains current as the codebase evolves.
- **Consistent Linting for Submodules:** Running linters such as revive and misspell across all submodules is streamlined, reducing the risk of missed checks or fragmented logic.
- **Ecosystem Alignment:** golangci-lint is a standard tool in the Go community, offering straightforward integration of new linters and familiarity for many contributors.
- **Isolated Dependency Impact:** All additional dependencies are indirect and limited to the tools module, ensuring no impact on users of the main project modules.
- **Avoiding Redundant Runs:** Configuration can be adjusted to ensure staticcheck is only run once, avoiding inefficiencies.

While script modifications could achieve similar results for specific linters, this would increase long-term maintenance overhead and reduce transparency. Retaining both approaches for now allows for verification that all checks remain equivalent, with the option to further consolidate in the future.

If there are alternative transition strategies or additional concerns, further suggestions are welcome for discussion. Feedback and ideas for improvement are appreciated as the review continues.

mmorel-35:
I added new commits to be more aligned with what I could see is already defined in the existing vet.sh.

dfawley:
You still haven't answered my first question, which is: why are you doing this? What is your agenda?  Nobody asked for this change, so why are you doing it?

This tool adds a new layer of magic into our CI that includes a ton of new dependencies and a new configuration syntax.  Its usage brings in new risks, e.g. breaking config changes, or behavior changes.

You say "it's a standard tool", but I completely disagree.  FWIW I don't consider `staticcheck` a standard tool, either.  `gofmt` and `goimports` are standard tools, because they are distributed by the Go team alongside the compiler.  We do use `staticcheck`, though, because it gives us value in excess of the cost of using it.  The value provided by this doesn't seem to offset its costs to me.  It's doing nothing new for us.  Yes it makes some aspect of our maintenance slightly easier, but it also has a cost and risks, and those things don't offset the minor benefits it provides in configuration/etc.

mmorel-35:
Thank you for the direct response and for outlining the concerns in detail.

The motivation for proposing the introduction of golangci-lint is to address recurring maintenance complexity and improve consistency in linting across the project's submodules. While no explicit request for this change has been made, there have been previous discussions (such as in #7994 and related comments) regarding the challenges of managing linters and ensuring uniform coverage, especially as the number of submodules grows.

The current bash-based approach for running multiple linters is functional but has become increasingly complex and less transparent to contributors. Each additional linter or configuration tweak adds to this complexity. A tool like golangci-lint centralizes configuration, reduces script maintenance, and provides built-in features such as exclusion management and easier onboarding for those already familiar with it from other Go projects.

The choice to include it alongside, rather than as a replacement for, existing tools is specifically to allow the project to evaluate its practical benefits and ensure there is no loss in coverage or behavior change before considering any further consolidation. The intention is to ensure that any transition is both deliberate and fully justified by demonstrated value.

Concerns about increased dependencies, potential for breaking changes, and configuration syntax are valid. These risks are mitigated by scoping golangci-lint to the tools module and by retaining the existing workflows. No impact to downstream consumers is expected, and the configuration can be version-controlled and reviewed like any other part of the CI. Should future maintenance or risk outweigh the benefits, removal can be easily performed.

The project already uses some non-core Go tools where they provide clear value, such as staticcheck. The intention here is to evaluate, in practice, whether golangci-lint’s unified configuration and maintenance benefits justify continued use. If, after this evaluation, consensus is that the value does not offset the cost, reverting is straightforward.

If there are specific metrics or outcomes that would help in making a decision about the long-term inclusion of golangci-lint, further suggestions or criteria for evaluation are welcome.

dfawley:
Thank you for the contribution, but I don't believe the cost is worth the benefit with integrating golangci-lint.  While it has some nice features, it also brings about a lot of new costs.  We're happy with our script-based solution.

mmorel-35:
Thank you for reviewing and for your clear feedback.

I appreciate you considering the proposal, and I respect the decision to keep the current approach. If there are opportunities in the future to contribute smaller improvements or address specific pain points, I’m happy to help.

Thanks again for the constructive discussion and for maintaining clarity in project direction.

PR #8504, opened 2025-08-11
Failing to delete the stream from the `activeStreams` map leading to "REFUSED_STREAM" errors.

alanprot:
### Description

Hi, 

After upgrading `go-grpc`, we started see the following error:

```
stream terminated by RST_STREAM with error code: REFUSED_STREAM
```

Upon investigation, it appears that in some scenarios, we fail to properly delete streams from the [`activeStreams(https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1309)`](https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1309) map, resulting in this error being wrongly returned.

I added a test case that seems to trigger the problem:

PS: To test the behavior i added a "ActiveStreamTracker" hook -> im not sure this is the best way but i wanted just to show the problem

```
end2end_test.go:4015: leak streams:  5
```

The issue seems to have been introduced in the following commit:
[https://github.com/grpc/grpc-go/commit/0d6e39f679d651f4e7c2d56a9daf26e5ce789e15](https://github.com/grpc/grpc-go/commit/0d6e39f679d651f4e7c2d56a9daf26e5ce789e15)

Specifically, the check added here:
[https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2\_server.go#L1357-L1359](https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1357-L1359)

If this conditional is removed, all streams are correctly cleaned up from the `activeStreams` map, and the test passes without leaks.


alanprot:
cc @arjan-bal @dfawley

arjan-bal:
Hi @alanprot, thanks for sharing the test to repro. I haven't debugged the test, but wanted to ask if you know [why the check added](https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1357-L1359) causes stream deletion to be skipped? From a look quick look at the places `streamDone` is being set, I found only two:
https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1332-L1343
https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1350-L1360

In both places, `deleteStream` is called eventually. Do you know how the code ends up in a situation where the stream is marked as done, but the it's not deleted from the map?

Edit: It seems like the `onWrite` callback may not be getting executed.

alanprot:
Hi @arjan-bal 

As far i could see the problem is we add set the clean and add the hdr on the controlbuff at

https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/http2_server.go#L1338


And we get to:

https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/controlbuf.go#L692-L696

And at this point, the stream is not on the `activeStreams` list, so processData returns here:

https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/controlbuf.go#L947-L950

And i think this stream never become added back on the  `activeStreams` list, so we never clean it up.

So it seems that it only happens when the ServerSide is sending tons of data and waiting for quota?

arjan-bal:
Thank you @alanprot, I think I understand the sequence of events leading to this.
1. The server stream exhausts the stream quota, and starts buffering data frames.
2. The server attempts to finish the RPC gracefully by sending trailers by calling [finishStream](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/http2_server.go#L1326-L1347).
3. The write of the header frame [gets blocked](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/controlbuf.go#L692-L696) due to the pending data frames that are awaiting flow control quota.
4. Once the deadline expires, the [server's deadline monitoring timer](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/http2_server.go#L627-L630) tries to call closeStream, which is a no-op since the stream state is already marked `streamDone` by `finishStream` in point 1.
5. The client sends and RST stream, [the server calls closeStream](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/http2_server.go#L865-L870) which is again a no-op.

The simplest fix is to remove the guard block added in `closeStream` in https://github.com/grpc/grpc-go/pull/8071. We may end up pushing redundant `cleanupStream` events on to the controlbuf, but it doesn't cause correctness issues since [cleanupStreamHandler returns early](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/controlbuf.go#L803-L805) in such cases.

@dfawley what do you think?

dfawley:
> The simplest fix is to remove the guard block added in closeStream in #8071. We may end up pushing redundant cleanupStream events on to the controlbuf, but it doesn't cause correctness issues since cleanupStreamHandler returns early in such cases.

That makes sense to me.  I can't think of any other way to do this.  Does the transport properly handle the rest of this situation, especially: dropping the queued data frames when the RST_STREAM is sent (or shortly thereafter)?

alanprot:
Also maybe we wanna consider decrementing this metric only if the stream was present on the activeStreams map?

https://github.com/grpc/grpc-go/blob/18ee309ab1f096b30ad08ec03c09c964f41cb605/internal/transport/http2_server.go#L1307-L1323

arjan-bal:
> That makes sense to me. I can't think of any other way to do this. Does the transport properly handle the rest of this situation, especially: dropping the queued data frames when the RST_STREAM is sent (or shortly thereafter)?

All the pending Data and Header frames are dropped in [cleanupStreamHandler](https://github.com/grpc/grpc-go/blob/18ee309ab1f096b30ad08ec03c09c964f41cb605/internal/transport/controlbuf.go#L812-L819).

> When does our max incoming streams quota go back up?

The check for max concurrent streams [depends on the size of the activeStreams map](https://github.com/grpc/grpc-go/blob/18ee309ab1f096b30ad08ec03c09c964f41cb605/internal/transport/http2_server.go#L555).
We delete from activeStreams map only in [deleteStream()](https://github.com/grpc/grpc-go/blob/18ee309ab1f096b30ad08ec03c09c964f41cb605/internal/transport/http2_server.go#L1306-L1313).
So the question is: When is deleteStream() called?

Without timeout:
After the trailers are written, at the beginning of cleanupStreamHandler() due to [cleanupStream.onWrite()](https://github.com/grpc/grpc-go/blob/18ee309ab1f096b30ad08ec03c09c964f41cb605/internal/transport/http2_server.go#L1342-L1344). 

With timeout:
[Just before cleanupStream event is pushed](https://github.com/grpc/grpc-go/blob/18ee309ab1f096b30ad08ec03c09c964f41cb605/internal/transport/http2_server.go#L1360-L1367) into the `controlBuf`. So there is a small interval before the cleanupStreamHandler is called when an extra stream could be started. We can minimize this interval this by using `cleaupStream.onWrite()`, similar to the timeout case above, but this is not super urgent since it's been this way for a while.

arjan-bal:
@alanprot are you interested in raising a PR with the fix and a unit test for catching regressions?

alanprot:
@arjan-bal I can definitely do that, but I’m out until Tuesday, so I’d only be able to pick it up then. :)

arjan-bal:
I'll be closing this PR and using https://github.com/grpc/grpc-go/issues/8517 to track the fix as per our normal bug fixing process. When the PR with the fix is merged https://github.com/grpc/grpc-go/issues/8517 can be marked as completed.

PR #8509, opened 2025-08-13
deps: bump go version to 1.24

kevinGC:
It's generally useful to have new-and-improved Go. One specific useful feature is `b.Loop()`, which makes benchmarking easier.

Encountered while working on #8510, as I was using `b.Loop()` in my benchmarks.

RELEASE NOTES:
* Minimum supported Go version is now 1.24

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: '1.25'
           cache-dependency-path: "**/go.sum"
 
       # Run the vet-proto checks.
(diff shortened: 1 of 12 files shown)


arjan-bal:
Hi @kevinGC, can you also bump the Go versions used for CI tests similar to [this change](https://github.com/grpc/grpc-go/pull/8108/files#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023d)?

kevinGC:
Thanks for the review. Done.

arjan-bal (approved):

Just FYI, gRPC Go supports the two most recent major versions of Golang, which aligns with Go's support policy. I'm not sure if it's a coincidence that this PR came just 6 hours after Go 1.25 was released. :)

dfawley:
(Per: https://github.com/grpc/grpc-go?tab=readme-ov-file#prerequisites)

ldemailly:
personally I find it annoying when a library forces a bump without taking advantage of any actual feature - ie a just bumping while things compile fine with older

(ps not to say that that when building end product one shouldn't indeed use the most recent stable patch (1.24.8 as of this writing) but forcing cascading bumps into dependencies is annoying (and golang.org/x is a prime culprit there but it would be nice the rest of the ecosystem doesn't imitate bad things))

Issue #8510, opened 2025-08-13
TCP receive zerocopy support


kevinGC:
### Use case(s) - what problem will this feature solve?
[TCP receive zerocopy](https://lwn.net/Articles/752188/) can be used on Linux to increase throughput, decrease CPU usage, and decrease memory bandwidth pressure.

### Proposed Solution
Add support for zerocopy receive. This breaks into a few smaller issues:

- Add support for `SO_RCVLOWAT`, which reduces CPU usage and is important for zerocopy performance. [`SO_RCVLOWAT` is used in the C++ gRPC library](https://github.com/grpc/grpc/blob/2afc2e0b04252696f416987b7fb9378ec9ad5caa/src/core/lib/event_engine/posix_engine/posix_endpoint.cc#L491) as well.
- Implement support for zerocopy itself.
- Benchmarking to prove the effectiveness of zerocopy.

The implementation may be limited to ALTS connections at first, but I'd like to expand it beyond ALTS in the future.

### Alternatives Considered
N/A, as the goal here is specifically to get zerocopy working.

### Additional Context
I plan on implementing this myself.

DevSusu:
Thanks for raising and working on this! Very excited to see this :)

PR #8511, opened 2025-08-13
alts: move ParseFramedMsg out of common

kevinGC:
It's only called in one place, and is effectively a method on conn.

Part of #8510.

Note that this PR includes #8509. GitHub doesn't support proper commit chains / stacked PRs, so I'm doing this in several PRs with some (annoyingly) redundant commits. Let me know if this isn't a good workflow for you and I'll change things up.

### DIFF
--- a/credentials/alts/internal/conn/record.go
+++ b/credentials/alts/internal/conn/record.go
@@ -144,7 +144,7 @@ func NewConn(c net.Conn, side core.Side, recordProtocol string, key []byte, prot
 func (p *conn) Read(b []byte) (n int, err error) {
 	if len(p.buf) == 0 {
 		var framedMsg []byte
-		framedMsg, p.nextFrame, err = ParseFramedMsg(p.nextFrame, altsRecordLengthLimit)
+		framedMsg, err = p.parseFramedMsg(p.nextFrame, altsRecordLengthLimit)
 		if err != nil {
 			return n, err
 		}
@@ -184,7 +184,7 @@ func (p *conn) Read(b []byte) (n int, err error) {
 				return 0, err
 			}
 			p.protected = p.protected[:len(p.protected)+n]
-			framedMsg, p.nextFrame, err = ParseFramedMsg(p.protected, altsRecordLengthLimit)
+			framedMsg, err = p.parseFramedMsg(p.protected, altsRecordLengthLimit)
 			if err != nil {
 				return 0, err
 			}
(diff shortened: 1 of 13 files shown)


arjan-bal:
> Note that this PR includes https://github.com/grpc/grpc-go/pull/8509. GitHub doesn't support proper commit chains / stacked PRs, so I'm doing this in several PRs with some (annoyingly) redundant commits. Let me know if this isn't a good workflow for you and I'll change things up.

@kevinGC, it would be great if you could rebase this branch off of master so both PRs can be merged independently. We shouldn't need Go 1.24 for the changes in this PR.

arjan-bal (approved):
LGTM, with request to remove the Go 1.24 changes. Adding a reviewer from the security team.

gtcooke94:
I think the easiest way for reviewers to diff is to just get the prior PRs merged and then this will diff nicely against master

The other option is to make a PR on your fork against the other PR branch if you want us to see the diff right now, then eventually open a PR later

credentials/alts/internal/conn/record.go
```diff
@@ -225,6 +225,38 @@ func (p *conn) Read(b []byte) (n int, err error) {
...
+		return nil, nil
+	}
+	if length > maxLen {
+		return nil, fmt.Errorf("received the frame length %d larger than the limit %d", length, maxLen)
+	}
+	if len(b) < int(length)+4 { // account for the first 4 msg length bytes.
+		// Frame is not complete yet.
+		return nil, nil
+	}
+	p.nextFrame = b[MsgLenFieldSize+length:]
```
gtcooke94, line 245: At a glance I personally prefer the previous style of returning nextFrame and not modifying state in this method, I think it makes it easier to follow and test. Do we have a compelling reason to swap it to this?

kevinGC: I'm happy to leave as-is or revert back to the old version. It just seemed odd to have a function in `common.go` that's only called in 1 place and always called s.t. it modifies the same bit of struct state.

Happy with whatever the gRPC team wants -- this was more of an implementation detail on the way to my other changes.

kevinGC:
Rebased. Looks like there're differing opinions on whether this is change is worth making -- I'm cool with keeping or dropping it.

credentials/alts/internal/conn/record.go
arjan-bal: Deferring to the @gtcooke94 since the security team owns this package. Closing the PR.

Issue #8485, opened 2025-08-04
client transport: Handle 1xx HTTP status HEADERS frames correctly


dfawley:
Per the HTTP/2 spec, we should be ignoring leading HEADERS frames with a 1xx status code, unless END_STREAM is also set (in which case it would be an error).

arjan-bal:
This is where the client processes the headers: https://github.com/grpc/grpc-go/blob/9fa3267859958a7fa0141a8180102850f3d5842c/internal/transport/http2_client.go#L1436

hugehoo:
hi, do you mind if i try on this issue?

vinothkumarr227:
@arjan-bal kindly assign the issue to me

arjan-bal:
Hi @hugehoo, @vinothkumarr227 had already started work on this issue, but the issue wasn't assigned to him. Apologies for the confusion.

hugehoo:
@arjan-bal got it 👌

PR #8518, opened 2025-08-18
transport: handle 1xx HTTP status HEADERS at the client side

vinothkumarr227:
Fixes: https://github.com/grpc/grpc-go/issues/8485

RELEASE NOTES:
* client: Ignore http headers with status 1xx and `END_STREAM` flag unset.
* client: Fail RPCs with status `INTERNAL` instead of `UNKNOWN` on receiving http headers with status 1xx and  `END_STREAM` flag set.

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -1513,7 +1513,6 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 					t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
 					return
 				}
-				return
 			}
 			httpStatusCode = &statusCode
 			if hf.Value == "200" {
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -1509,16 +1509,14 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 			if statusCode >= 100 && statusCode < 200 {
 				if endStream {
 					se := status.New(codes.Internal, fmt.Sprintf(
-						"protocol error: received 1xx informational header with END_STREAM set: %d", statusCode))
+						"protocol error: informational header with status code %d must not have END_STREAM set", statusCode))
 					t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
 				}
 				return
 			}
 			httpStatusCode = &statusCode
-			if hf.Value == "200" {
+			if statusCode == 200 {
 				httpStatusErr = ""
-				statusCode := 200
-				httpStatusCode = &statusCode
 				break
 			}
 
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -1499,13 +1499,6 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		case "grpc-message":
 			grpcMessage = decodeGrpcMessage(hf.Value)
 		case ":status":
-			if hf.Value == "200" {
-				httpStatusErr = ""
-				statusCode := 200
-				httpStatusCode = &statusCode
-				break
-			}
-
 			c, err := strconv.ParseInt(hf.Value, 10, 32)
 			if err != nil {
 				se := status.New(codes.Internal, fmt.Sprintf("transport: malformed http-status: %v", err))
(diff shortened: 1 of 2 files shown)


internal/transport/http2_client.go
```diff
@@ -1499,21 +1499,28 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
 			if err != nil {
 				se := status.New(codes.Internal, fmt.Sprintf("transport: malformed http-status: %v", err))
 				t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
 				return
 			}
 			statusCode := int(c)
+			if statusCode >= 100 && statusCode < 200 {
+				if endStream {
+					se := status.New(codes.Internal, fmt.Sprintf(
+						"protocol error: received 1xx informational header with END_STREAM set: %d", statusCode))
```
arjan-bal: The present message just tells what happened, but now why it caused an error and how it can be fixed. A better error message would be one that tells what went wrong and how to fix things. 
```go
fmt.Sprintf( "protocol error: informational header with status code %d must not have END_STREAM set", statusCode)
```

internal/transport/http2_client.go
```diff
@@ -1499,21 +1499,28 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+					t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
+				}
+				return
+			}
 			httpStatusCode = &statusCode
+			if hf.Value == "200" {
```
arjan-bal: Since the status code is already parsed, we can use an integer comparison now.

internal/transport/http2_client.go
```diff
@@ -1499,21 +1499,28 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+				httpStatusErr = ""
+				statusCode := 200
+				httpStatusCode = &statusCode
```
arjan-bal: We should not need this since the status code is already parsed.

internal/transport/transport_test.go
arjan-bal, line 1: Please also add a test for the case when the header is ignored. The test should fail on the master branch, but pass with the changes in this PR.

internal/transport/http2_client.go
vinothkumarr227: Done

internal/transport/http2_client.go
vinothkumarr227: Done

internal/transport/http2_client.go
vinothkumarr227: Done

internal/transport/transport_test.go
vinothkumarr227: Done

internal/transport/transport_test.go
```diff
@@ -3125,3 +3125,82 @@ func (s) TestServerSendsRSTAfterDeadlineToMisbehavedClient(t *testing.T) {
...
+				0: ts,
+			},
+			controlBuf: newControlBuffer(make(<-chan struct{})),
+		}
+	}
+
+	for _, test := range []struct {
+		name            string
+		metaHeaderFrame *http2.MetaHeadersFrame
+		endStreamFlag   http2.Flags
```
arjan-bal: This field should be renamed to something more general like `httpFlags`.  `EndStream` is one value for a flag, the field here can store any http flag value, not just end stream.

internal/transport/transport_test.go
```diff
@@ -3125,3 +3125,82 @@ func (s) TestServerSendsRSTAfterDeadlineToMisbehavedClient(t *testing.T) {
...
+				},
+			}
+
+			s.operateHeaders(test.metaHeaderFrame)
+
+			got := ts.status
+			want := test.wantStatus
+
+			if got.Code() != want.Code() || got.Message() != want.Message() {
+				t.Fatalf("operateHeaders(%v); status = \ngot: %v\nwant: %v", test.metaHeaderFrame, got, want)
```
arjan-bal:
```go
t.Fatalf("operateHeaders(%v); status = %v, want %v", test.metaHeaderFrame, got, want)
```
See https://google.github.io/styleguide/go/decisions#useful-test-failures

internal/transport/transport_test.go
vinothkumarr227: Done

internal/transport/transport_test.go
vinothkumarr227: Done

Issue #8514, opened 2025-08-14
Race condition involving trailer metadata


akehlenbeck:
### What version of gRPC are you using?

v1.72.0

### What version of Go are you using (`go version`)?

go version go1.24.2 linux/amd64

### What operating system (Linux, Windows, …) and version?

$ uname -a
Linux 6.8.0-1030-gcp #32~22.04.1-Ubuntu SMP Tue Apr 29 23:17:09 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

### What did you do?

We observed the following panic in grpc code.

```
fatal error: concurrent map iteration and map write

goroutine 13860950 [running]:
internal/runtime/maps.fatal({0x28de412?, 0x45df57?})
        /usr/local/go/src/runtime/panic.go:1058 +0x18
internal/runtime/maps.(*Iter).Next(0x224e100?)
        /usr/local/go/src/internal/runtime/maps/table.go:683 +0x86
google.golang.org/grpc/metadata.MD.Copy(...)
        /go/src/github.com/lightstep/vendor/google.golang.org/grpc/metadata/metadata.go:101
google.golang.org/grpc/internal/transport.(*serverHandlerTransport).writeStatus(0xc00a544ee0, 0xc0274c3320, 0xc016ef6388)
        /go/src/github.com/lightstep/vendor/google.golang.org/grpc/internal/transport/handler_server.go:282 +0x3c5
google.golang.org/grpc/internal/transport.(*ServerStream).WriteStatus(...)
        /go/src/github.com/lightstep/vendor/google.golang.org/grpc/internal/transport/server_stream.go:76
google.golang.org/grpc.(*Server).processUnaryRPC(0xc000a63200, {0x2c782f0, 0xc02152a6f0}, 0xc0274c3320, 0xc001ac2cc0, 0x43e0550, 0x0)
        /go/src/github.com/lightstep/vendor/google.golang.org/grpc/server.go:1418 +0x1562
google.golang.org/grpc.(*Server).handleStream(0xc000a63200, {0x2c794a8, 0xc00a544ee0}, 0xc0274c3320)
        /go/src/github.com/lightstep/vendor/google.golang.org/grpc/server.go:1815 +0xb88
google.golang.org/grpc.(*Server).serveStreams.func2.1()
        /go/src/github.com/lightstep/vendor/google.golang.org/grpc/server.go:1035 +0x7f
[...]

PR #8519, opened 2025-08-18
transport: ensure header mutex is held while copying trailers in handler_server

arjan-bal:
Fixes: https://github.com/grpc/grpc-go/issues/8514

The mutex that guards the trailers should be held while copying the trailers. We do lock the mutex in [the regular gRPC server transport](https://github.com/grpc/grpc-go/blob/9ac0ec87ca2ecc66b3c0c084708aef768637aef6/internal/transport/http2_server.go#L1140-L1142), but have missed it in the std lib http/2 transport. The only place where a write happens is `writeStatus()` is when the status contains a proto. 

https://github.com/grpc/grpc-go/blob/4375c784450aa7e43ff15b8b2879c896d0917130/internal/transport/handler_server.go#L251-L252

RELEASE NOTES:
* transport: Fix a data race while copying headers for stats handlers in the std lib http2 server transport.

### DIFF
--- a/internal/transport/handler_server.go
+++ b/internal/transport/handler_server.go
@@ -277,11 +277,13 @@ func (ht *serverHandlerTransport) writeStatus(s *ServerStream, st *status.Status
 	if err == nil { // transport has not been closed
 		// Note: The trailer fields are compressed with hpack after this call returns.
 		// No WireLength field is set here.
+		s.hdrMu.Lock()
 		for _, sh := range ht.stats {
 			sh.HandleRPC(s.Context(), &stats.OutTrailer{
 				Trailer: s.trailer.Copy(),
 			})
 		}
+		s.hdrMu.Unlock()
 	}
 	ht.Close(errors.New("finished writing status"))
 	return err
--- a/internal/transport/handler_server_test.go
+++ b/internal/transport/handler_server_test.go
@@ -247,13 +247,16 @@ type handleStreamTest struct {
 	ht    *serverHandlerTransport
 }
 
-type mockStatsHandler struct{}
+type mockStatsHandler struct {
+	rpcStatsCh chan stats.RPCStats
+}
 
 func (h *mockStatsHandler) TagRPC(ctx context.Context, _ *stats.RPCTagInfo) context.Context {
 	return ctx
 }
 
-func (h *mockStatsHandler) HandleRPC(context.Context, stats.RPCStats) {
+func (h *mockStatsHandler) HandleRPC(_ context.Context, s stats.RPCStats) {
+	h.rpcStatsCh <- s
 }
 
 func (h *mockStatsHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo) context.Context {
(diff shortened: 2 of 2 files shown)


internal/transport/handler_server.go
```diff
@@ -277,11 +277,13 @@ func (ht *serverHandlerTransport) writeStatus(s *ServerStream, st *status.Status
 	if err == nil { // transport has not been closed
 		// Note: The trailer fields are compressed with hpack after this call returns.
 		// No WireLength field is set here.
+		s.hdrMu.Lock()
```
arjan-bal, line 280: We're holding two locks here, this and `ht.writeStatusMu` (acquired at line 229).  `ht.writeStatusMu` is only  referenced in this method, so there shouldn't be a chance of deadlocks.

dfawley: Note that `hdrMu` is also already taken on 249, although, I'm not sure if that closure is run in the current goroutine or another one.

internal/transport/handler_server_test.go
```diff
@@ -485,6 +507,12 @@ func (s) TestHandlerTransport_HandleStreams_ErrDetails(t *testing.T) {
 
 	hst := newHandleStreamTest(t)
 	handleStream := func(s *ServerStream) {
+		if err := s.SendHeader(metadata.New(map[string]string{})); err != nil {
+			t.Error(err)
+		}
+		if err := s.SetTrailer(metadata.Pairs("custom-trailer", "Custom trailer value")); err != nil {
+			t.Error(err)
+		}
```
dfawley, line 565: How hard would it be to test this in a new test instead of in an existing test that's intended for testing error details?

internal/transport/handler_server_test.go
```diff
@@ -260,7 +277,12 @@ func newHandleStreamTest(t *testing.T) *handleStreamTest {
 		Body: bodyr,
 	}
 	rw := newTestHandlerResponseWriter().(testHandlerResponseWriter)
-	ht, err := NewServerHandlerTransport(rw, req, nil, mem.DefaultBufferPool())
+	// Add mock stats handlers to exercise the stats handler code path.
+	statsHandlers := make([]stats.Handler, 0, 5)
+	for range 5 {
+		statsHandlers = append(statsHandlers, &mockStatsHandler{})
+	}
+	ht, err := NewServerHandlerTransport(rw, req, statsHandlers, mem.DefaultBufferPool())
```
dfawley: Can we parameterize this, or find some other way to cause this configuration, instead of doing it for all existing tests?

internal/transport/handler_server_test.go
arjan-bal: Changed to use a new test. My thought process was that modifying existing tests would give us better coverage for interactions b/w different features.

internal/transport/handler_server.go
arjan-bal: The callback is executed in an event loop in a separate goroutine:
https://github.com/grpc/grpc-go/blob/9ac0ec87ca2ecc66b3c0c084708aef768637aef6/internal/transport/handler_server.go#L465-L474

arjan-bal: I thought about calling the stats handler in the callback above, but I noticed that the http2_server transport also schedules the network write in the background and calls the stats handlers.
https://github.com/grpc/grpc-go/blob/9ac0ec87ca2ecc66b3c0c084708aef768637aef6/internal/transport/http2_server.go#L1136-L1143

internal/transport/handler_server_test.go
arjan-bal: Done.

PR #8523, opened 2025-08-19
client: Roll-forward PR #8278(with changes): Restore the existing behavior to return io.EOF on repeated RecvMsg() calls for client-streaming RPCs

Pranjali-2501:
Partially addresses: https://github.com/grpc/grpc-go/issues/7286

This reverts commit https://github.com/grpc/grpc-go/commit/20bd1e7dfad3535675e3dff3d73fa894256f1f79

Changes:
- Modifies client.RecvMsg() so that successive calls after stream ends return io.EOF.
- Adds extra state to track calls to client.recvmsg(required to return Cardinality Violation only in case zero response)

RELEASE NOTES:
* client: Return status code INTERNAL when a server sends 0 response messages for a unary or client streaming RPC.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -549,8 +549,6 @@ type clientStream struct {
 
 	sentLast bool // sent an end stream
 
-	recvFirstMsg bool // set after the first message is received
-
 	methodConfig *MethodConfig
 
 	ctx context.Context // the application's context, wrapped by stats/tracing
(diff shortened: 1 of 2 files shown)


arjan-bal:
Can you please keep the revert commit as part of the PR? This will make it easier to review only the new changes.

Also, could you please update the PR title to accurately reflect what is being changed? Since this is an impure rollforward, the current title is a bit misleading. You can mention that it's rolling forward the reverted changes in the PR description.

Finally, please add release notes for these changes.

Thanks!

test/end2end_test.go
```diff
@@ -3981,7 +4137,7 @@ func (s) TestServerStreaming_ClientSendsZeroRequests(t *testing.T) {
 }
 
 // Tests that a client receives a cardinality violation error for client-streaming
-// RPCs if the server call SendMsg multiple times.
+// RPCs if the server call SendMsg() multiple times.
```
dfawley:
```suggestion
// RPCs if the server calls SendMsg() multiple times.
```

dfawley (approved):
Minor grammar nit otherwise LGTM

test/end2end_test.go

stream.go
```diff
@@ -549,6 +549,8 @@ type clientStream struct {
 
 	sentLast bool // sent an end stream
 
+	recvFirstMsg bool // set after the first message is received
```
arjan-bal: nit: Please expand `recv` to `received`. The abbreviation is a bit ambiguous and could be interpreted as either `receive` or `received`.

arjan-bal (approved):
LGTM with a minor comment.

stream.go

Issue #8517, opened 2025-08-18
Failing to delete the stream from the activeStreams map leading to REFUSED_STREAM errors


arjan-bal:
Original PR with repro: https://github.com/grpc/grpc-go/pull/8504

After upgrading `go-grpc`, we started see the following error:

```
stream terminated by RST_STREAM with error code: REFUSED_STREAM
```

Upon investigation, it appears that in some scenarios, we fail to properly delete streams from the [`activeStreams`](https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1309) map, resulting in this error being wrongly returned.

I added a test case that seems to trigger the problem:

PS: To test the behavior i added a "ActiveStreamTracker" hook -> im not sure this is the best way but i wanted just to show the problem

```
end2end_test.go:4015: leak streams:  5
```

The issue seems to have been introduced in the following PR: https://github.com/grpc/grpc-go/pull/8071

Specifically, the check added here:
[https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2\_server.go#L1357-L1359](https://github.com/grpc/grpc-go/blob/57b69b47a2da17098c387c8c64c4aab5295cbd3e/internal/transport/http2_server.go#L1357-L1359)

If this conditional is removed, all streams are correctly cleaned up from the `activeStreams` map, and the test passes without leaks.

## Root cause
1. The server stream exhausts the stream quota, and starts buffering data frames.
2. The server attempts to finish the RPC gracefully by sending trailers by calling [finishStream](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/http2_server.go#L1326-L1347).
3. The write of the header frame [gets blocked](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/controlbuf.go#L692-L696) due to the pending data frames that are awaiting flow control quota.
4. Once the deadline expires, the [server's deadline monitoring timer](https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/internal/transport/http2_server.go#L627-L630) tries to call closeStream, which is a no-op since the stream state is already marked `streamDone` by `finishStream` in point 1.
[...]

PR #8528, opened 2025-08-20
transport: allow stream cancellation on the server when blocked on flow control

arjan-bal:
Fixes: #8517

This change allows `t.closeStream()` to be executed even if the stream state is `done`. This is required to allow streams to be cancelled to timed out. See issue for detailed root cause.

RELEASE NOTES:
* server: Fix bug preventing streams from being cancelled or timed out when blocked on flow control.


alanprot:
The fix did not make into the release?

arjan-bal:
> The fix did not make into the release?

We didn't want to delay the 1.75 release too much. We will perform patch releases once this PR is merged. Reviews have been slow due to the annual gRPConf that happened a couple of days ago.

PR #8532, opened 2025-08-21
xds/clusterresolver: [alternative] Revise configbuilder childname generator to fix reuse issues with merge and split scenarios

erenboz:
Alternative implementation for https://github.com/grpc/grpc-go/pull/8531

RELEASE NOTES:
xds: Revise name generator to fix xds priority-lb reuse issues where localities shuffle between priorities

### DIFF
--- a/internal/xds/balancer/clusterresolver/configbuilder_childname.go
+++ b/internal/xds/balancer/clusterresolver/configbuilder_childname.go
@@ -31,21 +31,28 @@ import (
 // struct keeps state between generate() calls, and a later generate() might
 // return names returned by the previous call.
 type nameGenerator struct {
-	existingNames map[clients.Locality]string
-	prefix        uint64
-	nextID        uint64
+	currentNames       map[clients.Locality]string
+	firstAssignedNames map[clients.Locality]string
+	prefix             uint64
+	nextID             uint64
 }
 
 func newNameGenerator(prefix uint64) *nameGenerator {
-	return &nameGenerator{prefix: prefix}
+	return &nameGenerator{
+		prefix:             prefix,
+		currentNames:       make(map[clients.Locality]string),
+		firstAssignedNames: make(map[clients.Locality]string),
+	}
 }
 
 // generate returns a list of names for the given list of priorities.
 //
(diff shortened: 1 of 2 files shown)


internal/xds/balancer/clusterresolver/configbuilder_childname.go
```diff
@@ -59,30 +66,41 @@ func (ng *nameGenerator) generate(priorities [][]xdsresource.Locality) []string
...
-			if name, ok := ng.existingNames[locality.ID]; ok {
+			if name, ok := ng.currentNames[locality.ID]; ok {
+				if !usedNames[name] {
+					nameFound = name
+					break
+				}
+			}
+		}
+
+		for _, locality := range priority {
```
erenboz: Ah I noticed I just ignore currentNames altogether and it just works unintuitively :) I guess it works due to locality is not reused in multiple priorities in any tests which is expected to be valid use-case for this name generator I believe.

erenboz: Fixed and added some more corner test cases like removal of a locality

internal/xds/balancer/clusterresolver/configbuilder_childname_test.go
```diff
@@ -29,80 +29,313 @@ func Test_nameGenerator_generate(t *testing.T) {
...
+				},
+				{
+					{{ID: clients.Locality{Zone: "L1"}}, {ID: clients.Locality{Zone: "L2"}}},
+				},
+				{
+					{{ID: clients.Locality{Zone: "L2"}}},
+					{{ID: clients.Locality{Zone: "L1"}}},
+				},
+			},
+			want: []string{"priority-0-0", "priority-0-2"},
```
erenboz: Leads to leaks

arjan-bal:
Thanks for raising this change. In the grpc-go repo, we don't usually keep draft PRs. Since the discussion is going on in https://github.com/grpc/grpc-go/pull/8531, I'm closing this PR for the time being. I'll re-open this if we decide to go forward with this approach.

Issue #8425, opened 2025-07-02
Transport close after keepalive failure blocks on unresponsive reader


jgold2-stripe:
### What version of gRPC are you using?
`v1.71.2`

### What version of Go are you using (`go version`)?
`go version go1.24.4 darwin/arm64`

### What operating system (Linux, Windows, …) and version?
MacOS

### What did you do?
I have a connection open to a streaming RPC using client keepalives. When the network goes down, the keepalive fires as expected, but the transport still waits much longer than is necessary to finally close.

### What did you expect to see?
I assumed that the transport would close almost immediately after the keepalive fails.

### What did you see instead?
The transport does close, but not until the underlying network connection fails its current call to `Read`.

Just as some extra color, I think the issue stems from the following:
  - We [detect the keepalive failure](https://github.com/grpc/grpc-go/blob/v1.71.2/internal/transport/http2_client.go#L1725).
  - Before returning with this error, we [call `Close`](https://github.com/grpc/grpc-go/blob/v1.71.2/internal/transport/http2_client.go#L1699).
  - The close proceeds quickly (I verified with logging) until [it blocks on `readerDone`](https://github.com/grpc/grpc-go/blob/v1.71.2/internal/transport/http2_client.go#L1016-L1018).
  - The `readerDone` channel is [closed when returning from `reader()`](https://github.com/grpc/grpc-go/blob/v1.71.2/internal/transport/http2_client.go#L1622), which itself makes blocking read calls (via the `golang.org/x/net/http2` package) on the underlying `net.Conn`.
  - While the actual network is down, TCP does not know this yet, leaving us to wait for a significant amount of time (40s in some cases) to finally close `readerDone`.
  - Once `readerDone` is closed, the transport completes its close quickly from there.

I noticed that [a write deadline is set](https://github.com/grpc/grpc-go/blob/v1.71.2/internal/transport/http2_client.go#L973) (though I was surprised to find that the error is ignored -- maybe it should be logged at least?), but not a read deadline. Testing with a local version where I also invoke `SetReadDeadline`, I can confirm that the transport does close quickly as I'd expect.

Is the absence of a call to `SetReadDeadline` intentional? I wonder if I'm missing a good reason for it. In our case, since the failure to receive a keepalive ACK necessarily indicates that the connection to the server is dead, we absolutely expect any read to time out, so would prefer not to wait unnecessarily for it. I can propose code changes here, but want to step back and see if I am misunderstanding what *should* be happening.

[...]

PR #8534, opened 2025-08-21
Keepalive

jgold2-stripe:
This PR adds a call to `net.Conn.SetWriteDeadline`, as discussed in https://github.com/grpc/grpc-go/issues/8425#issuecomment-3057938248. Additionally, it updates the previous call to `SetReadDeadline` to log any non-nil error value (this doesn't affect behavior but proved helpful in some earlier debugging).

### DIFF
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -992,6 +992,11 @@ func (t *http2Client) Close(err error) {
 	if err := t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10)); err != nil {
 		logger.Warningf("Failed to set write deadline on t.conn during Close(): %s", err)
 	}
+	// For background on the deadline value chosen here, see
+	// https://github.com/grpc/grpc-go/issues/8425#issuecomment-3057938248 .
+	if err := t.conn.SetReadDeadline(time.Now().Add(time.Second)); err != nil {
+		logger.Warningf("Failed to set read deadline on t.conn during Close(): %s", err)
+	}
 	t.mu.Lock()
 	// Make sure we only close once.
 	if t.state == closing {
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -989,7 +989,9 @@ func (t *http2Client) closeStream(s *ClientStream, err error, rst bool, rstCode 
 // only once on a transport. Once it is called, the transport should not be
 // accessed anymore.
 func (t *http2Client) Close(err error) {
-	t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10))
+	if err := t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10)); err != nil {
+		logger.Warningf("Failed to set write deadline on t.conn during Close(): %s", err)
+	}
 	t.mu.Lock()
 	// Make sure we only close once.
 	if t.state == closing {


internal/transport/http2_client.go
```diff
@@ -989,7 +989,14 @@ func (t *http2Client) closeStream(s *ClientStream, err error, rst bool, rstCode
 // only once on a transport. Once it is called, the transport should not be
 // accessed anymore.
 func (t *http2Client) Close(err error) {
-	t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10))
+	if err := t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10)); err != nil {
+		logger.Warningf("Failed to set write deadline on t.conn during Close(): %s", err)
```
arjan-bal: nit: We should not mention internal symbol names like `t.conn` in the logs. It may be confusing for users and the message may become inaccurate after refactoring. Instead, we can say something more general like `Failed to set write deadline while closing connection`.

arjan-bal: Please use `%v` to format the error.

arjan-bal (commented):
Can you please add a test to verify the fix and catch regressions? You can refer to the test for the write deadline: https://github.com/grpc/grpc-go/blob/b0bc6dc1ce38bd93dc19552b07dc97d1f59dbd5f/internal/transport/transport_test.go#L2946-L3009

jgold2-stripe:
PTAL @arjan-bal

internal/transport/transport_test.go
```diff
@@ -3008,6 +3008,50 @@ func (s) TestClientCloseReturnsEarlyWhenGoAwayWriteHangs(t *testing.T) {
 	ct.Close(errors.New("manually closed by client"))
 }
 
+// deadlineTestConn is a net.Conn wrapper used to assert that deadlines are set
+// during http2Client.Close().
+type deadlineTestConn struct {
+	net.Conn
+	observedReadDeadline  atomic.Bool
+	observedWriteDeadline atomic.Bool
```
arjan-bal, line 3026: I believe we need these atomics because `t.Close` is being called after unblocking the reader of the `readerDone` channel here:
https://github.com/grpc/grpc-go/blob/3c743c9d3f630d12ec805055c304cc30fa6ea89d/internal/transport/http2_client.go#L1637-L1642

Using an atomic will satisfy the race detector, but will make the test flake if the deadline assertion code executed before the deadline is set.

@eshitachandwani is there a reason to close `readerDone` before closing the transport? I tried moving `close(t.readerDone)` below and some tests started failing. I'm not sure if that's a real deadlock of an issue in the tests.

internal/transport/http2_client.go
```diff
@@ -989,7 +989,14 @@ func (t *http2Client) closeStream(s *ClientStream, err error, rst bool, rstCode
 // only once on a transport. Once it is called, the transport should not be
 // accessed anymore.
 func (t *http2Client) Close(err error) {
-	t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10))
+	if err := t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10)); err != nil {
+		logger.Warningf("Failed to set write deadline when closing connection: %v", err)
```
arjan-bal: These logs are too chatty. They're appearing almost every time a connection is closed, e.g:
```
tlogger.go:126: WARNING http2_client.go:993 [transport] Failed to set write deadline when closing connection: set tcp 127.0.0.1:59286: use of closed network connection  (t=+590.186033ms)
```
I don't think we should log here because we are calling `conn.Close` in [another goroutine](https://github.com/grpc/grpc-go/blob/3c743c9d3f630d12ec805055c304cc30fa6ea89d/internal/transport/http2_client.go#L467-L477) and expecting these calls to fail.

internal/transport/transport_test.go
arjan-bal: For now, to avoid flakiness, we can use channels instead of atomics. Instead of setting atomics, we can close the channels. In the main test function we can use a select statement to wait on the `observedReadDeadline` and `ctx.Done()` events. if `ctx.Done()` happens, we should fail the test, e.g:
https://github.com/grpc/grpc-go/blob/3c743c9d3f630d12ec805055c304cc30fa6ea89d/internal/transport/transport_test.go#L817-L821

Similarly we can add a select statement for `observedWriteDeadline`.

eshitachandwani: > @eshitachandwani is there a reason to close `readerDone` before closing the transport? I tried moving `close(t.readerDone)` below and some tests started failing. I'm not sure if that's a real deadlock of an issue in the tests.

We are doing `close(t.readerDone)` before `t.Close()` because `t.Close()` will wait for a signal from `readerDone` (see [here](https://github.com/grpc/grpc-go/blob/3074bcdd8f3c52e4b5cc9527592d475780b062a8/internal/transport/http2_client.go#L1037)). This change was made because we wanted the transport to wait for all the go routines(Writer, reader, keepAlive) to complete before completely closing the transport. Change [PR](https://github.com/grpc/grpc-go/pull/7666)

If we have it the other way round, `t.Close()` will block on `readerDone` channel but `readerDone` will never be closed since it would be after `t.Close()`

arjan-bal: Oh, I see. We're calling `t.Close` twice. It is guaranteed that the deadline is set after `t.Close` returns, the reader goroutine may call it again, but it will be a no-op. The use of atomics here is fine.

jgold2-stripe: Thanks for the details to both of you. I left them as `atomic.Bool` but with some explanation and a link back here to reduce the maintenance burden for others to come.

internal/transport/http2_client.go
jgold2-stripe: Dropped the logging.

internal/transport/http2_client.go
```diff
@@ -990,6 +990,9 @@ func (t *http2Client) closeStream(s *ClientStream, err error, rst bool, rstCode
 // accessed anymore.
 func (t *http2Client) Close(err error) {
 	t.conn.SetWriteDeadline(time.Now().Add(time.Second * 10))
+	// For background on the deadline value chosen here, see
+	// https://github.com/grpc/grpc-go/issues/8425#issuecomment-3057938248 .
+	t.conn.SetReadDeadline(time.Now().Add(time.Second))
```
dfawley, line 995: I think it would be better/simpler to just use `SetDeadline` with the same 1 second deadline, if possible.

internal/transport/transport_test.go
```diff
@@ -3008,6 +3008,60 @@ func (s) TestClientCloseReturnsEarlyWhenGoAwayWriteHangs(t *testing.T) {
...
+		conn, err := net.Dial("tcp", addr)
+		if err != nil {
+			return nil, err
+		}
+		return &deadlineTestConn{Conn: conn}, nil
+	}
+	co := ConnectOptions{
+		Dialer: dialer,
+	}
+	server, client, cancel := setUpWithOptions(t, 0, &ServerConfig{}, normal, co)
```
dfawley, line 3052: If we were to set read/write deadlines during handshaking (we currently only block on the context [here](https://github.com/grpc/grpc-go/blob/29ba00196bb864af89a844c3eaf097b7af48b1ad/internal/transport/http2_client.go#L248-L258), but our server does set deadlines [here](https://github.com/grpc/grpc-go/blob/29ba00196bb864af89a844c3eaf097b7af48b1ad/server.go#L970)), then this test would be invalid.  I think we should either confirm that the values are still `false` before calling `Close`, or clear them unconditionally then.

dfawley (commented):
Thank you for the change.  A couple small things to potentially change.

internal/transport/http2_client.go
jgold2-stripe: I think it's possible to do that, but am a little concerned about throwing that in here with the other changes, only because it seems like a possible load-bearing change for which I can't really vouch myself. I took a look back at https://github.com/grpc/grpc-go/pull/7371, where the write deadline was added, and couldn't seem to find any direct discussion about why the value of 10 was chosen, or whether a value of 1 might be problematic.

I don't mind just flipping this to 1, but also wonder if you prefer that we *not* do that in this PR, and then change it later? That way if everything else is good but the change from 10 to 1 introduces some separate problem, it's a simple revert of just that one-liner PR?

Not sure how your team likes to balance risk:throughput.

internal/transport/transport_test.go
jgold2-stripe: Thanks for finding that. I clear them now unconditionally.

internal/transport/http2_client.go
jgold2-stripe: @dfawley -- I didn't realize that my comment above may not have made it to you unless I explicitly request review, so I just did that.

I can make the change to replace calls to `SetReadDeadline` and `SetWriteDeadline` with a single call to `SetDeadline`, but want to check in with you about the concerns ^^^ re combining both changes in the same PR.

dfawley: I'm OK either way, here.  It's theoretically less risky to combine them because of the behavior change, but I'm not that concerned since the original number was picked pretty arbitrarily in the first place, and neither are expected to hit unless there is a problem with the connection.

dfawley (approved):

internal/transport/http2_client.go
dfawley: That should say "_more_ risky", sorry.  Let's just go with what's in this PR, then.  Thank you!

PR #8431, opened 2025-07-07
credentials: implement file-based JWT Call Credentials (part 1 for A97)

dimpavloff:
Part one for https://github.com/grpc/proposal/pull/492 (A97). 
This is done in a new  `credentials/jwt` package to provide file-based PerRPCCallCredentials. It can be used beyond XDS. The package handles token reloading, caching, and validation as per A97 . 

There will be a separate PR which uses it in `xds/bootstrap`. 

Whilst implementing the above, I considered `credentials/oauth` and `credentials/xds` packages instead of creating a new one. The former package has `NewJWTAccessFromKey` and `jwtAccess` which seem very relevant at first. However, I think the `jwtAccess` behaviour seems more tailored towards Google services. Also, the refresh, caching, and error behaviour for A97 is quite different than what's already there and therefore a separate implementation would have still made sense. 
WRT `credentials/xds`, it could have been extended to both handle transport and call credentials. However, this is a bit at odds with A97 which says that the implementation should be non-XDS specific and, from reading between the lines, usable beyond XDS. 
I think the current approach makes review easier but because of the similarities with the other two packages, it is a bit confusing to navigate. Please let me know whether the structure should change. 

Relates to https://github.com/istio/istio/issues/53532

RELEASE NOTES:
- credentials: Add `credentials/jwt` package providing file-based JWT PerRPCCredentials (A97).

PR #8535, opened 2025-08-22
xds bootstrap: enable using JWT Call Credentials (part 2 for A97)

dimpavloff:
Part two for https://github.com/grpc/proposal/pull/492 (A97). It's stacked on top of #8431 

What this PR does is:

-  update `internal/xds/bootstrap` with support for loading multiple PerRPCCallCredentials specifed in a new `call_creds` field in the boostrap file as per A97
- adjust `xds/internal/xdsclient/clientimpl.go`to use the call credentials when constructing the client
- update `xds/bootstrap` to register the `jwtcreds` call credentials and make them available if `GRPC_EXPERIMENTAL_XDS_BOOTSTRAP_CALL_CREDS` is enabled

I have added `DialOptionsWithCallCredsForTransport` because, even though current and future call credentials are likely to all expect secure transport, I thought it would be safer to check of insecure transport just in case. If you prefer, I can just update `DialOptions` to use all call credentials regardless of the transport. 

Relates to https://github.com/istio/istio/issues/53532

RELEASE NOTES:
- xds bootstrap: add support for loading a JWT token from file and use it as Call Credentials (A97). This is guarded by GRPC_EXPERIMENTAL_XDS_BOOTSTRAP_CALL_CREDS


dimpavloff:
Apologies, I messed up the PR stacking, will open a new one

PR #8536, opened 2025-08-22
xds bootstrap: enable using JWT Call Credentials (part 2 for A97)

dimpavloff:
Part two for https://github.com/grpc/proposal/pull/492 (A97), following #8431 .

What this PR does is:

-  update `internal/xds/bootstrap` with support for loading multiple PerRPCCallCredentials specifed in a new `call_creds` field in the boostrap file as per A97
- adjust `xds/internal/xdsclient/clientimpl.go`to use the call credentials when constructing the client
- update `xds/bootstrap` to register the `jwtcreds` call credentials and make them available if `GRPC_EXPERIMENTAL_XDS_BOOTSTRAP_CALL_CREDS` is enabled

Relates to https://github.com/istio/istio/issues/53532

RELEASE NOTES:
- xds: add support for loading a JWT from file and use it as Call Credentials (A97). To enable this feature, set the environment variable `GRPC_EXPERIMENTAL_XDS_BOOTSTRAP_CALL_CREDS` to `true` (case insensitive).


xds/bootstrap/bootstrap.go
```diff
@@ -58,6 +59,9 @@ func RegisterCredentials(c Credentials) {
 // GetCredentials returns the credentials associated with a given name.
 // If no credentials are registered with the name, nil will be returned.
 func GetCredentials(name string) Credentials {
+	if name == "jwt_token_file" && !envconfig.XDSBootstrapCallCredsEnabled {
```
dimpavloff: FYI, since the convention is to register the credentials via an `init()` function in `xds/bootstrap/credentials.go`, it's not possible to opt in or out of the registration there and at the same time to be able to unit test if the feature is guarded by `XDSBootstrapCallCredsEnabled`. Hence why the check has been moved here.

arjan-bal:
Hi @dimpavloff, thank you for contributing the implementation for this gRFC. In the grpc-go repo, we don't usually keep PRs open that are not actionable. I'll be closing this PR for now and we can re-open it once https://github.com/grpc/grpc-go/pull/8431 is merged.

arjan-bal:
Re-opening since part 1 was just merged. Please re-base before we can start reviews, thanks!

internal/xds/bootstrap/bootstrap.go
```diff
@@ -165,14 +181,16 @@ func (a *Authority) Equal(other *Authority) bool {
...
+	callCreds      []CallCreds
 	serverFeatures []string
 
 	// As part of unmarshalling the JSON config into this struct, we ensure that
 	// the credentials config is valid by building an instance of the specified
 	// credentials and store it here for easy access.
-	selectedCreds    ChannelCreds
-	credsDialOption  grpc.DialOption
-	extraDialOptions []grpc.DialOption
+	selectedCreds     ChannelCreds
```
dimpavloff: I am considering renaming this (and the `SelectedCreds` method) to include `Channel` in their name, as it's more symmetric and clearer with the addition of CallCreds. WDYT?

dimpavloff:
@arjan-bal Hey, the PR is ready for review, could you please take a look?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,21 @@ func (cc ChannelCreds) String() string {
 	return cc.Type + "-" + string(b)
 }
 
+// CallCreds contains the call credentials configuration for individual RPCs.
+// It implements gRFC A97 call credentials structure.
+type CallCreds struct {
+	// Type contains a unique name identifying the call credentials type.
+	// Currently only "jwt_token_file" is supported.
```
easwars: Nit: Let's get rid of this line. There is a good chance that when someone adds the next call creds, they might forget to remove this from here.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,21 @@ func (cc ChannelCreds) String() string {
...
+	Type string `json:"type,omitempty"`
+	// Config contains the JSON configuration relevant for the call credentials.
```
easwars: Nit: (the following might be shorter with the same effect)

// Config contains the JSON configuration for the call credentials.

// Config contains the JSON configuration for this call credentials.

internal/xds/bootstrap/bootstrap.go
easwars: Sounds good.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -227,6 +257,24 @@ func (sc *ServerConfig) DialOptions() []grpc.DialOption {
...
+		(transportCreds != nil && transportCreds.Info().SecurityProtocol == "insecure")
+
+	for _, callCred := range sc.selectedCallCreds {
+		if isInsecureTransport && callCred.RequireTransportSecurity() {
+			continue
+		}
+		dopts = append(dopts, grpc.WithPerRPCCredentials(callCred))
+	}
+	return dopts
+}
```
easwars: The gRFC states the following in the documentation for the new field `call_creds` in the bootstrap configuration.

```
  // NEW FIELD!
  // List of call creds.  Optional.  Client will apply all call creds
  // types that it supports but will ignore any types that it does not
  // support.
```

My reading of this is that we *should* apply all supported call creds unconditionally. And if it so happens that an insecure transport is configured, then we can let the RPC fail and that will result in a more meaningful error to the user (something like cannot send secure call creds on an insecure connection). I will check what the other language implementations plan to do and will get back to you on this.

Also, I'm not very thrilled about the way `isInsecureTransport` is determined. This seems a bit hacky to me and would require changes if another insecure channel credentials implementation is added in the future which has a name other than `insecure`.

xds/bootstrap/bootstrap.go
easwars: The environment variable should only guard whether we process the new `call_creds` field in the bootstrap configuration. It should not guard the registration of the JWT call creds implementation.

xds/bootstrap/credentials.go
```diff
@@ -24,13 +24,15 @@ import (
...
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/internal/xds/bootstrap/jwtcreds"
 	"google.golang.org/grpc/internal/xds/bootstrap/tlscreds"
 )
 
 func init() {
 	RegisterCredentials(&insecureCredsBuilder{})
 	RegisterCredentials(&googleDefaultCredsBuilder{})
 	RegisterCredentials(&tlsCredsBuilder{})
+	RegisterCredentials(&jwtCallCredsBuilder{})
```
easwars: This doesn't seem correct to me. Quoting the gRFC:
```
The supported types for call creds use a different namespace than the supported types for channel creds.
```

This means that we should be able to support a channel creds and a call creds with the same name. So, this means that we should have completely different registries for the two. I'm OK with making a breaking change to the API here, but inserting a `Channel` to the existing APIs, like `RegisterChannelCredentials`, `ChannelCredentials` interface, `GetChannelCredentials` (although, if we are following Go style guide recommendations, this should ideally not include the `Get` prefix, but we can't do that since the name will collide with the interface name). 

And similarly we will introduce new APIs for the call creds, `RegisterCallCredentials`, `CallCredentials`, `GetCallCredentials` etc.

I would be OK with making the breaking API change in a separate PR as well. Get that merged and then merge this.

xds/bootstrap/credentials.go
```diff
@@ -68,3 +70,15 @@ func (d *googleDefaultCredsBuilder) Build(json.RawMessage) (credentials.Bundle,
 func (d *googleDefaultCredsBuilder) Name() string {
 	return "google_default"
 }
+
+// jwtCallCredsBuilder implements the `Credentials` interface defined in
+// package `xds/bootstrap` and encapsulates JWT call credentials.
+type jwtCallCredsBuilder struct{}
+
+func (j *jwtCallCredsBuilder) Build(configJSON json.RawMessage) (credentials.Bundle, func(), error) {
```
easwars: This change also makes it clear that we should have a separate interface for channel credentials. There is no reason to return a bundle when building channel credentials. Please let me know if you thought there was a good reason for doing that.

internal/xds/bootstrap/jwtcreds/bundle.go
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package jwtcreds implements JWT Call Credentials for XDS, configured via xDS
```
easwars: This package's implementation should be much simpler if we have a separate interface for call credentials just like we have today for transport credentials, and not having to deal with the bundle. (this follows from my other comments in those appropriate places).

easwars (commented):
I haven't reviewed the tests yet, but I feel they could change significantly based on some of the existing comments. So, I'll look at them a little later on.

internal/xds/bootstrap/bootstrap.go
dimpavloff: > then we can let the RPC fail and that will result in a more meaningful error to the user

That's a good point; I actually realised the RPC will fail as you say it will just the other day (as part of updating this PR) and wondered whether it would be better to do it that way. So I'll change it as you suggest, I also don't at all like the `isInsecureTransport` logic

xds/bootstrap/credentials.go
dimpavloff: > Please let me know if you thought there was a good reason for doing that.

I didn't have a good reason; I was struggling to see how to fit the RFC into the existing implementation (which I was unfamiliar with) and I think I just interpreted `Credentials` more broadly. 

p.s by `channel credentials` I assume you actually mean `call credentials`.

xds/bootstrap/credentials.go
dimpavloff: I had missed the `different namespace` bullet point and what that was implying.
Also, as mentioned in the next comment, because I wasn't knowlegeable about the area at all, I wasn't sure how things were meant to fit together with the new bootstrap Call Credentials. The `Bundle` type was adding to the confusion, too, because it contains both Transport and PerRPCCredentials and I wasn't sure whether this meant I should enhance existing bundle builders or create a new one like I did.

Introducing a `Channel` substring in the existing functions and types would be nice. If it's a simple rename, I am happy to do it (in this or another PR if you prefer). But if you think it requires some additional considerations and is a bit more involved, please do make the change. 

Also, do you have any thoughts on what the `CallCredentials` interface should look like? 
```
type CallCredentials interface {
    Build(config json.RawMessage) (credentials.PerRPCCredentials, error)
    Name() string
}
```

dimpavloff (commented):
Thanks for the prompt review, @easwars !

eyewill085-code:
23,18,55

في الخميس، ١٨ سبتمبر ٢٠٢٥ ١٨:٢٠ Dimitar Pavlov ***@***.***>
كتب:

xds/bootstrap/credentials.go
easwars: We could consider having three return parameters like what the `ChannelCredentials` interface will have.
```
    Build(config json.RawMessage) (credentials.PerRPCCredentials, func(), error)
```
The second one giving a chance to the call creds to any cleanup if they like. I see that we don't have to do any cleanup in our JWT call creds. But it might help future proof the API.

easwars: > If it's a simple rename, I am happy to do it (in this or another PR if you prefer) 

It looks like a simple rename to me at this point, unless you run into some issues when you actually end up doing it.

internal/xds/bootstrap/bootstrap.go
easwars: I checked with other grpc language leads and they seem to agree that we don't want to be doing this check. We should let the check happen at RPC time. The one thing that we need to verify though is that a meaningful error is propagated to the user when a failure happens here. Thanks.

easwars:
@dimpavloff : I see a bunch of commits. Please let me know when it is ready for another review pass. Thanks.

internal/xds/bootstrap/bootstrap.go
dimpavloff: WRT the meaningful error, I _think_ what will happen is https://github.com/grpc/grpc-go/blob/master/credentials/jwt/token_file_call_creds.go#L78 will be called and you'd get `cannot send secure credentials on an insecure connection: requires SecurityLevel PrivacyAndIntegrity; connection has NoSecurity`

dimpavloff:
@easwars I think this is ready for another review :)

internal/xds/bootstrap/bootstrap.go
```diff
@@ -256,16 +287,17 @@ func (sc *ServerConfig) Equal(other *ServerConfig) bool {
 // String returns the string representation of the ServerConfig.
 func (sc *ServerConfig) String() string {
 	if len(sc.serverFeatures) == 0 {
-		return fmt.Sprintf("%s-%s", sc.serverURI, sc.selectedCreds.String())
+		return fmt.Sprintf("%s-%s", sc.serverURI, sc.selectedChannelCreds.String())
```
dimpavloff: should this method be updated to include the CallCredentials?

internal/envconfig/xds.go
```diff
@@ -74,4 +74,9 @@ var (
 	// For more details, see:
 	// https://github.com/grpc/proposal/blob/master/A86-xds-http-connect.md
 	XDSHTTPConnectEnabled = boolFromEnv("GRPC_EXPERIMENTAL_XDS_HTTP_CONNECT", false)
+
+	// XDSBootstrapCallCredsEnabled controls if JWT call credentials can be used
```
easwars: Nit: This environment variable guards the newly added bootstrap field `call_creds`. I agree that the only call credentials that we currently support via this mechanism is the JWT call creds, but that could change any day (before we get rid of this env var). So, I would prefer we more explicitly call out what this env var guards.

easwars:
Apologies for the delay in the review. I'm a bit swamped with other things. I will get to this asap.

internal/xds/bootstrap/bootstrap.go
easwars: Sure

internal/xds/bootstrap/jwtcreds/call_creds.go
```diff
@@ -0,0 +1,51 @@
...
+	"encoding/json"
+	"fmt"
+
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/jwt"
+)
+
+// NewCallCredentials returns a credentials.PerRPCCredentials The input config
+// should match the structure specified in gRFC A97 structure.
+// See gRFC A97: https://github.com/grpc/proposal/blob/master/A97-xds-jwt-call-creds.md
```
easwars: Some nits here.

- There needs to be a period after the first sentence.
- I think the next two lines can be shortened. (this is an internal API, so we can omit the link to the actual gRFC)
- There is no mention about the second return value, `func()`

So, maybe something like
```
// NewCallCredentials returns a new JWT token based call credentials. The input
// config must match the structure specified in gRFC A97. The caller is expected 
// to invoke the second return value when they are done using the returned call creds.

File: internal/xds/bootstrap/jwtcreds/call_creds_test.go [hunk 12]
```diff
@@ -0,0 +1,162 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc/credentials"
+)
+
+func TestNewCallCredentials(t *testing.T) {
```
easwars: Please use the `Tester` defined in the `grpctest` package here: https://github.com/grpc/grpc-go/blob/d0ebcdffc75dc76f18966ab9cccafe6a949d6fb5/internal/grpctest/grpctest.go#L52

It has some setup and teardown magic that we find useful for all of our tests.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			callCreds, cleanup, err := NewCallCredentials(json.RawMessage(tt.config))
+
+			if !tt.wantSuccess {
+				if err == nil {
+					t.Fatal("Expected error, got nil")
```
easwars: Please identify the function being tested in the test error message. See: https://google.github.io/styleguide/go/decisions#identify-the-function

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+				return
+			}
+			if err != nil {
+				t.Fatalf("NewCallCredentials failed: %v", err)
+			}
+			if callCreds == nil {
+				t.Fatal("Expected non-nil bundle")
+			}
+			if cleanup == nil {
+				t.Error("Expected non-nil cleanup function")
```
easwars: If we expect a non-nil cleanup function to be returned, we need to mention that in the docstring for `NewCallCredentials`

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+					t.Fatal("Expected error, got nil")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("NewCallCredentials failed: %v", err)
+			}
+			if callCreds == nil {
+				t.Fatal("Expected non-nil bundle")
+			}
```
easwars: It feels like the test logic is complicated enough that we can split it up into two different tests. One for the failure cases and one for the success case. That way, the test logic will be simple, easy to maintain and easy to read. Thanks.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+			if cleanup == nil {
+				t.Error("Expected non-nil cleanup function")
+			} else {
+				defer cleanup()
+			}
+
+			// Test that call credentials get used
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
easwars: Please define a const for the default test timeout, like we do in our other test packages. Thanks.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+			defer cancel()
+			ctx = credentials.NewContextWithRequestInfo(ctx, credentials.RequestInfo{
+				AuthInfo: &testAuthInfo{secLevel: credentials.PrivacyAndIntegrity},
+			})
+			metadata, err := callCreds.GetRequestMetadata(ctx)
+			if err != nil {
+				t.Fatalf("GetRequestMetadata failed: %v", err)
+			}
+			if len(metadata) == 0 {
+				t.Error("Expected metadata to be returned")
```
easwars: This should be a `t.Fatal` as well. See: https://google.github.io/styleguide/go/decisions#keep-going

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+			authHeader, ok := metadata["authorization"]
+			if !ok {
+				t.Error("Expected authorization header in metadata")
```
easwars: Same here.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,162 @@
...
+	_, cleanup, err := NewCallCredentials(json.RawMessage(config))
+	if err != nil {
+		t.Fatalf("NewCallCredentials failed: %v", err)
+	}
+	if cleanup == nil {
+		t.Fatal("Expected non-nil cleanup function")
+	}
+	// Cleanup should not panic
+	cleanup()
+	// Multiple cleanup calls should be safe
```
easwars: Again, if we expect the cancel func to be idempotent, we should mention that in the docstring for `NewCallCredentials`

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,20 @@ func (cc ChannelCreds) String() string {
 	return cc.Type + "-" + string(b)
 }
 
+// CallCreds contains the call credentials configuration for individual RPCs.
+// It implements gRFC A97 call credentials structure.
+type CallCreds struct {
+	// Type contains a unique name identifying the call credentials type.
```
easwars: Hmm .. The gRFC does not state any uniqueness requirements for this.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -194,6 +211,18 @@ func (sc *ServerConfig) ServerFeatures() []string {
 	return sc.serverFeatures
 }
 
+// CallCreds returns the call credentials configuration for this server.
+func (sc *ServerConfig) CallCreds() []CallCreds {
```
easwars: Should we call this `CallCredsConfig` and the next method `CallCreds`?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -294,26 +327,51 @@ func (sc *ServerConfig) UnmarshalJSON(data []byte) error {
...
 			sc.extraDialOptions = d.DialOptions()
 		}
 		sc.cleanups = append(sc.cleanups, cancel)
 		break
 	}
+
+	if envconfig.XDSBootstrapCallCredsEnabled {
+		// Process call credentials - unlike channel creds, we use ALL supported
+		// types. Also, call credentials are optional as per gRFC A97.
+		for _, callCredConfig := range server.CallCreds {
```
easwars: Nit: Can we use a shorter name for the loop variable. Maybe `cfg`?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -294,26 +327,51 @@ func (sc *ServerConfig) UnmarshalJSON(data []byte) error {
...
+			c := bootstrap.GetCallCredentials(callCredConfig.Type)
+			if c == nil {
+				// Skip unsupported call credential types (don't fail bootstrap).
+				continue
+			}
+			callCred, cancel, err := c.Build(callCredConfig.Config)
```
easwars: We refer to them as `credentials` and not `credential` even when there is a single one. This was a major source of confusion to me when I joined the team as well :)

So, s/callCred/callCreds/

internal/xds/xdsclient/clientimpl_test.go
```diff
@@ -259,3 +260,38 @@ func (s) TestBuildXDSClientConfig_Success(t *testing.T) {
 		})
 	}
 }
+
+func TestServerConfigCallCredsIntegration(t *testing.T) {
```
easwars: This needs to have the `s` receiver on it.

easwars (commented):
I wasn't able to make a complete pass. But some comments to get you going. I will try to make another pass asap.

internal/xds/bootstrap/bootstrap.go
dimpavloff: Hm, you're right, it doesn't. I was influenced by the ChannelCreds.Type but there it makes more sense given that only one gets picked up. I've updated the comment

internal/xds/bootstrap/bootstrap.go
dimpavloff: I was trying to keep symmetry with `ChannelCreds` and `SelectedChannelCreds`. However, `SelectedCallCreds` now returns  `[]credentials.PerRPCCredentials` rather than the config, still, which breaks this symmetry. 

Realistically, I'm not sure even if the public methods are needed -- the PerRPCCredentials are added to the `ServerConfig.extraDialOptions` and used via `ServerConfig.DialOptions()` 

WDYT, shall I remove the public methods? I've renamed them in the meantime. I've used `CallCredsConfigs`, plural, because I also renamed the Go type to `CallCredsConfig` and the method returns a collection of those. I've also renamed the relevant fields in `ServerConfig` (and `ServerConfigTestingOptions`) in similar way.

internal/xds/bootstrap/bootstrap.go
dimpavloff: Haha, yes, this did confuse me a bit when starting. Thanks for spotting this

internal/xds/bootstrap/jwtcreds/call_creds_test.go
dimpavloff: Yep, much better, thanks!

internal/xds/bootstrap/bootstrap.go
dimpavloff: I defined a `type CallCredsConfigs []CallCredsConfig` and implemented `Stringer` for this purpose. LMK if you prefer a different approach

dimpavloff (commented):
Thanks for the review @easwars ; I think it's ready for another pass

internal/xds/bootstrap/bootstrap.go
easwars: I filed https://github.com/grpc/grpc-go/issues/8635 to keep track of this. Thanks.

internal/xds/bootstrap/bootstrap.go
easwars: LGTM

internal/xds/xdsclient/clientimpl_test.go
```diff
@@ -259,3 +260,38 @@ func (s) TestBuildXDSClientConfig_Success(t *testing.T) {
 		})
 	}
 }
+
+func (s) TestServerConfigCallCredsIntegration(t *testing.T) {
+	originalJWTEnabled := envconfig.XDSBootstrapCallCredsEnabled
+	envconfig.XDSBootstrapCallCredsEnabled = true
+	defer func() {
+		envconfig.XDSBootstrapCallCredsEnabled = originalJWTEnabled
+	}()
```
easwars: We have a helper function for this: https://github.com/grpc/grpc-go/blob/9ff80a7f245ceb2aad36ef6987f34c9441dbd43e/internal/testutils/envconfig.go#L26

internal/xds/bootstrap/jwtcreds/call_creds.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/jwt"
+)
+
+// NewCallCredentials returns a new JWT token based call credentials. The input
+// config must match the structure specified in gRFC A97. The caller is expected
+// to invoke the second return value when they are done using the returned call creds.
+// The cancel function is idempotent.
+func NewCallCredentials(configJSON json.RawMessage) (credentials.PerRPCCredentials, func(), error) {
```
easwars: How about adding named return values to help with the documentation.

```golang
// NewCallCredentials creates and returns JWT token-based call credentials
// using the input config (which must match gRFC A97).
//
// The caller must invoke the returned cancel function (which is idempotent),
// when they are done using the returned call credentials.
func NewCallCredentials(configJSON json.RawMessage) (c credentials.PerRPCCredentials, cancel func(), err error) {
```

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/internal/grpctest"
+)
+
+const defaultCtxTimeout = 5 * time.Second
```
easwars: Nit: s/defaultCtxTimeout/defaultTestTimeout/

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+			}`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			callCreds, cleanup, err := NewCallCredentials(json.RawMessage(tt.config))
+
+			if err == nil {
+				t.Fatal("NewCallCredentials: expected error, got nil")
```
easwars: Please consider adding the config to the error message:
`t.Fatal("NewCallCredentials(%s) succeeded when expected to fail", tt.config)`

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			if callCreds != nil {
+				t.Error("NewCallCredentials: Expected nil bundle to be returned")
```
easwars: Nit: there is no bundle anymore.
`t.Error("NewCallCredentials(%s) returned non-nil credentials", tt.config)`

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			if cleanup != nil {
+				t.Error("NewCallCredentials: Expected nil cleanup function to be returned")
```
easwars: Similar improvement to the test error message here too.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+			config: `{
+				"jwt_token_file": ""
+			}`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			callCreds, cleanup, err := NewCallCredentials(json.RawMessage(tt.config))
+
```
easwars: Nit: nix newline.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			if callCreds != nil {
+				t.Error("NewCallCredentials: Expected nil bundle to be returned")
+			}
+			if cleanup != nil {
+				t.Error("NewCallCredentials: Expected nil cleanup function to be returned")
+			}
+		})
+	}
+}
```
easwars, line 75: Nit: add newline to separate the two tests.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+func (s) TestNewCallCredentialsWithValidConfig(t *testing.T) {
+	token := createTestJWT(t)
+	tokenFile := writeTempFile(t, token)
+	config := `{"jwt_token_file": "` + tokenFile + `"}`
+
+	callCreds, cleanup, err := NewCallCredentials(json.RawMessage(config))
+	if err != nil {
+		t.Fatalf("NewCallCredentials failed: %v", err)
```
easwars: Identify the input here too: https://google.github.io/styleguide/go/decisions#identify-the-input

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	if callCreds == nil {
+		t.Fatal("NewCallCredentials: Expected non-nil bundle to be returned")
+	}
+	if cleanup == nil {
+		t.Error("NewCallCredentials: Expected non-nil cleanup function to be returned")
+	} else {
+		defer cleanup()
+	}
```
easwars, line 93: It would be nicer to always return a non-nil cancel func (even in error cases), to avoid such checks at callsites.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Test that call credentials get used
+	ctx, cancel := context.WithTimeout(context.Background(), defaultCtxTimeout)
```
easwars: Do we really need a `5s` timeout here? And this const is only used here, you could remove the global and just inline the value here. Maybe a second at best should be good enough?

internal/xds/bootstrap/jwtcreds/call_creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+	claims := "eyJhdWQiOiJodHRwczovL2V4YW1wbGUuY29tIiwiZXhwIjoyMDAwMDAwMDAwfQ"
+	signature := "fake_signature_for_testing"
+
+	return header + "." + claims + "." + signature
+}
+
+func writeTempFile(t *testing.T, content string) string {
+	t.Helper()
+	tempDir := t.TempDir()
+	filePath := filepath.Join(tempDir, "tempfile")
```
easwars: Probably doesn't matter. But could be rename the file to something more appropriate. Maybe "jwt_token"?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,40 @@ func (cc ChannelCreds) String() string {
...
+// Equal reports whether cc and other are considered equal.
+func (cc CallCredsConfig) Equal(other CallCredsConfig) bool {
+	return cc.Type == other.Type && bytes.Equal(cc.Config, other.Config)
+}
+
+func (cc CallCredsConfig) String() string {
+	if cc.Config == nil {
+		return cc.Type
+	}
+	// We do not expect the Marshal call to fail since we wrote to cc.Config
```
easwars: Nit: Here and in a few other places, please terminate comment lines with periods. https://google.github.io/styleguide/go/decisions#comment-sentences

internal/xds/bootstrap/bootstrap.go
```diff
@@ -163,16 +198,18 @@ func (a *Authority) Equal(other *Authority) bool {
 
 // ServerConfig contains the configuration to connect to a server.
 type ServerConfig struct {
-	serverURI      string
-	channelCreds   []ChannelCreds
-	serverFeatures []string
+	serverURI        string
+	channelCreds     []ChannelCreds
```
easwars, line 205: While you are here, could you please add a TODO to rename `ChannelCreds` to `ChannelCredsConfig` to match what we have with the call creds? Thanks.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,40 @@ func (cc ChannelCreds) String() string {
 	return cc.Type + "-" + string(b)
 }
 
+// CallCredsConfig contains the call credentials configuration for individual RPCs.
```
easwars: I think it might be better to clarify in this docstring that these call creds are used when talking to the xDS management server. The docstring on the `ChannelCreds` type does that. So, what do you think about something like:

```golang
CallCredsConfig contains the call credentials configuration to be used on 
RPCs to the management server.
```

I think the part about A97 is not that important and could be skipped. Thanks.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -245,9 +294,11 @@ func (sc *ServerConfig) Equal(other *ServerConfig) bool {
 		return false
 	case !slices.EqualFunc(sc.channelCreds, other.channelCreds, func(a, b ChannelCreds) bool { return a.Equal(b) }):
 		return false
+	case !slices.EqualFunc(sc.callCredsConfigs, other.callCredsConfigs, func(a, b CallCredsConfig) bool { return a.Equal(b) }):
+		return false
 	case !slices.Equal(sc.serverFeatures, other.serverFeatures):
 		return false
-	case !sc.selectedCreds.Equal(other.selectedCreds):
+	case !sc.selectedChannelCreds.Equal(other.selectedChannelCreds):
```
easwars: I know this is existing code, but I'm wondering if we can remove this field from the equality check for a couple of reasons:
- we are already checking the channel credentials configuration (from which we selected this credentials) for equality
- this type is supposed to semantically represent configuration corresponding to an xDS server and not any runtime state. We are already abusing it a little (or optimizing) by storing the selected credentials in it, so that we don't have to parse the configuration again when we actually want to use the credentials.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -256,25 +307,27 @@ func (sc *ServerConfig) Equal(other *ServerConfig) bool {
 // String returns the string representation of the ServerConfig.
 func (sc *ServerConfig) String() string {
 	if len(sc.serverFeatures) == 0 {
-		return fmt.Sprintf("%s-%s", sc.serverURI, sc.selectedCreds.String())
+		return strings.Join([]string{sc.serverURI, sc.selectedChannelCreds.String(), sc.CallCredsConfigs().String()}, "-")
 	}
 	features := strings.Join(sc.serverFeatures, "-")
-	return strings.Join([]string{sc.serverURI, sc.selectedCreds.String(), features}, "-")
+	return strings.Join([]string{sc.serverURI, sc.selectedChannelCreds.String(), features, sc.CallCredsConfigs().String()}, "-")
```
easwars, line 307: I remember you asking me if we want to include the channel creds here and I said yes. But thinking about this now, I'm wondering if the returned string is going to be too verbose, especially since the `String()` method of the call credentials config returns a JSON (as does the channel credentials). Have you see what this string actually looks like in any of your testing?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -194,6 +231,18 @@ func (sc *ServerConfig) ServerFeatures() []string {
...
+func (sc *ServerConfig) CallCredsConfigs() CallCredsConfigs {
+	return sc.callCredsConfigs
+}
+
+// CallCreds returns the built call credentials that are ready to use.
+// These are the credentials that were successfully built from the call_creds
+// configuration.
+func (sc *ServerConfig) CallCreds() []credentials.PerRPCCredentials {
+	return sc.selectedCallCreds
+}
```
easwars: Do we need this method at all (and the `selectedCallCreds` field)? We are currently storing the selected call credentials directly as dial options in the `extraDialOptions` field.

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -432,6 +578,34 @@ func (s) TestGetConfiguration_Success(t *testing.T) {
...
+	}
+}
+
+// Tests Istio-style bootstrap configurations with JWT call credentials
+func (s) TestGetConfiguration_IstioStyleWithCallCreds(t *testing.T) {
+	original := envconfig.XDSBootstrapCallCredsEnabled
+	envconfig.XDSBootstrapCallCredsEnabled = true
+	defer func() {
+		envconfig.XDSBootstrapCallCredsEnabled = original
+	}()
```
easwars: You could use the helper function here as well `testutils.SetEnvConfig`.

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -1033,6 +1207,190 @@ func Test(t *testing.T) {
...
+			want: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			result := test.cc1.Equal(test.cc2)
+			if result != test.want {
+				t.Errorf("CallCreds.Equal() = %v, want %v", result, test.want)
+			}
```
easwars:
```suggestion
			if got := test.cc1.Equal(test.cc2); got != test.want {
				t.Errorf("CallCreds.Equal() = %v, want %v", got, test.want)
			}
```

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -1033,6 +1207,190 @@ func Test(t *testing.T) {
...
+		})
+	}
+}
+
+func (s) TestServerConfig_UnmarshalJSON_WithCallCreds(t *testing.T) {
+	original := envconfig.XDSBootstrapCallCredsEnabled
+	defer func() { envconfig.XDSBootstrapCallCredsEnabled = original }()
+	envconfig.XDSBootstrapCallCredsEnabled = true
```
easwars: Please use the helper function here as well to override the env var (and elsewhere in this PR). Thanks.

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -1201,3 +1559,112 @@ func (s) TestNode_ToProto(t *testing.T) {
...
+			if len(callCredsConfig) != test.wantCallCreds {
+				t.Errorf("Call creds configs count = %d, want %d", len(callCredsConfig), test.wantCallCreds)
+			}
+			if len(callCreds) != test.wantCallCreds {
+				t.Errorf("Call creds count = %d, want %d", len(callCreds), test.wantCallCreds)
+			}
+			// Verify transport credentials are properly selected.
+			if sc.SelectedChannelCreds().Type != test.wantTransportType {
+				t.Errorf("Selected transport creds type = %q, want %q",
+					sc.SelectedChannelCreds().Type, test.wantTransportType)
```
easwars: Nit: Single line please.

internal/xds/xdsclient/clientimpl_test.go
```diff
@@ -259,3 +260,38 @@ func (s) TestBuildXDSClientConfig_Success(t *testing.T) {
...
+
+	tokenFile := "/token.jwt"
```
easwars: Nit: get rid of this variable and inline the file name in the config below.

internal/xds/xdsclient/clientimpl_test.go
```diff
@@ -259,3 +260,38 @@ func (s) TestBuildXDSClientConfig_Success(t *testing.T) {
...
+		t.Fatalf("Failed to unmarshal server config: %v", err)
+	}
+	// Verify call credentials are processed.
+	callCreds := sc.CallCredsConfigs()
+	if len(callCreds) != 1 {
+		t.Errorf("Expected 1 call credential, got %d", len(callCreds))
+	}
+	selectedCallCreds := sc.CallCreds()
+	if len(selectedCallCreds) != 1 {
+		t.Errorf("Expected 1 selected call credential, got %d", len(selectedCallCreds))
```
easwars: Please specify the `got` before the `want` in test error messages.https://google.github.io/styleguide/go/decisions#got-before-want

xds/bootstrap/bootstrap.go
```diff
@@ -31,34 +31,70 @@ import (
...
+	return nil
+}
+
+// CallCredentials interface encapsulates a credentials.PerRPCCredentials
+// builder that can be used for communicating with the xDS Management server.
+type CallCredentials interface {
+	// Build returns a PerRPCCredentials created from the provided
+	// configuration, and a function to clean up any additional resources
+	// associated with them when they are no longer needed.
+	Build(config json.RawMessage) (credentials.PerRPCCredentials, func(), error)
```
easwars, line 79: I realize that I added a comment about possibly using named return values for this method on the JWT call creds implementation. But it looks like we haven't done that for the existing channel credentials interface which has a similar set of return values. So, I'm fine with leaving this as is. Thanks.

xds/bootstrap/bootstrap_test.go
```diff
@@ -62,14 +63,15 @@ func TestRegisterNew(t *testing.T) {
...
 }
 
-func TestCredsBuilders(t *testing.T) {
+func TestChannelCredsBuilders(t *testing.T) {
 	tests := []struct {
-		typename string
-		builder  Credentials
+		typename              string
+		builder               ChannelCredentials
+		minimumRequiredConfig json.RawMessage
```
easwars: Do we really need this new field when it is always set to `nil`?

xds/bootstrap/bootstrap_test.go
```diff
@@ -78,10 +80,40 @@ func TestCredsBuilders(t *testing.T) {
...
 
-			_, stop, err := test.builder.Build(nil)
+			bundle, stop, err := test.builder.Build(test.minimumRequiredConfig)
 			if err != nil {
 				t.Fatalf("%T.Build failed: %v", test.builder, err)
 			}
+			if bundle == nil {
+				t.Errorf("%T.Build returned nil bundle, expected non-nil", test.builder)
+			}
+			stop()
```
easwars, line 88: Can we `defer` this call to `stop` right after the line where we call `Build`?

xds/bootstrap/bootstrap_test.go
```diff
@@ -78,10 +80,40 @@ func TestCredsBuilders(t *testing.T) {
...
+			}
+
+			bundle, stop, err := test.builder.Build(test.minimumRequiredConfig)
+			if err != nil {
+				t.Fatalf("%T.Build failed: %v", test.builder, err)
+			}
+			if bundle == nil {
+				t.Errorf("%T.Build returned nil bundle, expected non-nil", test.builder)
+			}
 			stop()
```
easwars: Same here.

xds/bootstrap/bootstrap_test.go
```diff
@@ -100,3 +132,25 @@ func TestTlsCredsBuilder(t *testing.T) {
 		stop()
 	}
 }
+
+func TestJwtCallCredentials_DisabledIfFeatureNotEnabled(t *testing.T) {
+	builder := GetCallCredentials("jwt_call_creds")
```
easwars: Can we explicitly set the value of the env var to `false` instead of running this test under the assumption that it is false (which it currently is). But if we explicitly set it here, then we wont have to change the test when we change the default value of the env var to `true`.

easwars:
@dimpavloff This PR is mostly looking good. Most of the comments in this pass are nits. We should be able to resolve them quickly and move on. Thank you very much!

internal/xds/bootstrap/bootstrap.go
dimpavloff: No, we don't, other than for testing; I had it for symmetry with ChannelCreds. 
I've assumed that you'd prefer that we delete it so I have done so and adjusted the unit tests a bit (in `bootstrap_test.go` and `clientimpl_test.go`). A downside is that DialOpts feels a bit distant/opaque when unit testing the call credentials config functionality.

internal/xds/bootstrap/bootstrap.go
dimpavloff: Yeah, it's a bit verbose and why I was asking about it (I should have been clearer). The returned value looks like this `xds-server:443-tls-{}-jwt_token_file-{"jwt_token_file":"/token.jwt"}`

xds/bootstrap/bootstrap.go
dimpavloff: I've kept it without for consistency.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
dimpavloff: Definitely don't need anywhere near 5 seconds.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
dimpavloff: I agree, updated. Although I think it might be good not to document it as then it might sound like the caller should call the cancel function even when an error is returned which would be an unnecessary ritual. OTOH, in this case, the test will be asserting implementation detail.. 
Or I can keep it as is. What do you prefer?

xds/bootstrap/bootstrap_test.go
dimpavloff: I was thinking that eventually the variable would just be removed and the behaviour – turned on. So my intention here is asserting that the feature is off by default rather than that it can be controlled by the env variable. 
When the default changes, the test would then rightly fail this assertion and in this case it would make sense to just delete it. 

Happy to adjust it if you still prefer.

xds/bootstrap/bootstrap_test.go
dimpavloff: I think it's because I had the Channel and Call creds together initially. Removed.

internal/xds/bootstrap/bootstrap.go
dimpavloff: I agree - because we're already comparing `sc.channelCreds` and `other.channelCreds` and because the config doesn't get dynamically updated, it makes sense. 
Shall I just remove it in this PR or send a follow-up PR in case we're missing something?

dimpavloff (commented):
Thanks for the feedback and apologies for the delay, @easwars , it's been a hectic week. 
The PR is ready for a another look, PTAL.

internal/xds/bootstrap/jwtcreds/call_creds_test.go
easwars: Makes sense. Let's keep it the way it is.

internal/xds/bootstrap/bootstrap.go
easwars: I'm fine with removing the public methods if they are not required, which in this case seems to be the case. Less code the better :)

internal/xds/bootstrap/bootstrap.go
easwars: Doing it in this PR seems fine to me. Thanks.

internal/xds/bootstrap/bootstrap.go
easwars: Let's leave it as it stands now. We can clean it up later if it becomes a problem.

internal/xds/bootstrap/bootstrap.go
easwars: We *could* add a method with a `ForTesting` suffix if that makes testing much cleaner. That is a pattern we use in a lot of places in our codebase. Thanks.

easwars:
@dimpavloff : Sorry for the delay in getting back. I was away for a while. 

Thanks for taking care of all of my comments. I will make one more pass today and will assign it to someone else for a second review.

xds/bootstrap/bootstrap_test.go
```diff
@@ -78,25 +80,73 @@ func TestCredsBuilders(t *testing.T) {
...
 			}
+			if bundle == nil {
+				t.Errorf("%T.Build returned nil bundle, expected non-nil", test.builder)
+			}
 			stop()
 		})
 	}
 }
 
+func TestCallCredsBuilders(t *testing.T) {
```
easwars: Can we please get rid of the table, and have this test be specific to JWT call creds. Thanks.

xds/bootstrap/bootstrap_test.go
```diff
@@ -78,25 +80,73 @@ func TestCredsBuilders(t *testing.T) {
...
+		{"jwt_token_file", &jwtCallCredsBuilder{}, json.RawMessage(`{"jwt_token_file":"/path/to/token.jwt"}`)},
+	}
+
+	for _, test := range tests {
+		t.Run(test.typename, func(t *testing.T) {
+			if got, want := test.builder.Name(), test.typename; got != want {
+				t.Errorf("%T.Name = %v, want %v", test.builder, got, want)
+			}
+
+			bundle, stop, err := test.builder.Build(test.minimumRequiredConfig)
```
easwars: The use of the term `bundle` here is very confusing (now that we've gotten rid of it).

xds/bootstrap/bootstrap_test.go
```diff
@@ -78,25 +80,73 @@ func TestCredsBuilders(t *testing.T) {
...
+	defer stop()
 
 	if _, stop, err := tls.Build(json.RawMessage(`{"ca_certificate_file":"/ca_certificates.pem","refresh_interval": "asdf"}`)); err == nil {
+		defer stop()
 		t.Errorf("tls.Build() succeeded with an invalid refresh interval, when expected to fail")
-		stop()
+	}
+}
+
+func TestJwtCallCredentials_DisabledIfFeatureNotEnabled(t *testing.T) {
```
easwars: Does this test even make sense now. We are not guarding the registration of this call credentials based on whether or not the env var is set, right. So, I don't quite understand the reasoning behind the existence of this test. Thanks.

easwars (approved):
LGTM, modulo minor comments in tests

xds/bootstrap/bootstrap_test.go
dimpavloff: Sorry, my bad. Nice one spotting it!

xds/bootstrap/bootstrap_test.go
dimpavloff: 🤦 No, it doesn't. Sorry, I've missed this in the refactor. Also, the test was actually passing incorrectly because of using `jwt_call_creds` when it should have used `jwt_token_file` ...

dimpavloff (commented):
Thanks for the review and for the fixes @easwars ! I've addressed the latest comments.

xds/bootstrap/bootstrap_test.go
```diff
@@ -78,25 +78,43 @@ func TestCredsBuilders(t *testing.T) {
...
+			if bundle == nil {
+				t.Errorf("%T.Build returned nil bundle, expected non-nil", test.builder)
+			}
 			stop()
 		})
 	}
 }
 
+func TestJWTCallCredsBuilder(t *testing.T) {
+
```
arjan-bal: nit: Please remove the newline at the beginning of this function.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,40 @@ func (cc ChannelCreds) String() string {
 	return cc.Type + "-" + string(b)
 }
 
+// CallCredsConfig contains the call credentials configuration to be used on
+// RPCs to the management server.
+type CallCredsConfig struct {
+	// Type contains a name identifying the call credentials type.
+	Type string `json:"type,omitempty"`
+	// Config contains the JSON configuration for this call credentials.
```
arjan-bal, line 92: I think it's worth mentioning in the godoc that this field is optional. From [A97](https://github.com/grpc/proposal/blob/master/A97-xds-jwt-call-creds.md#configuring-call-credentials-in-xds-bootstrap-file):
> // The "config" field is optional; it may be missing if the
// credential type does not require config parameters.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -83,6 +84,40 @@ func (cc ChannelCreds) String() string {
...
+
+// Equal reports whether cc and other are considered equal.
+func (cc CallCredsConfig) Equal(other CallCredsConfig) bool {
+	return cc.Type == other.Type && bytes.Equal(cc.Config, other.Config)
+}
+
+func (cc CallCredsConfig) String() string {
+	if cc.Config == nil {
+		return cc.Type
+	}
```
arjan-bal: Since `json.RawMessage` is an alias of a `[]byte`, we should use a `len(cc.Config) == 0` check instead of checking for `nil`. See the [style guide](https://google.github.io/styleguide/go/decisions#nil-slices):
> When designing interfaces, avoid making a distinction between a nil slice and a non-nil, zero-length slice, as this can lead to subtle programming errors. This is typically accomplished by using len to check for emptiness, rather than == nil

internal/xds/bootstrap/bootstrap.go
```diff
@@ -294,26 +340,51 @@ func (sc *ServerConfig) UnmarshalJSON(data []byte) error {
...
+				continue
+			}
+			callCreds, cancel, err := c.Build(cfg.Config)
+			if err != nil {
+				// Call credential validation failed - this should fail bootstrap.
+				return fmt.Errorf("failed to build call credentials from bootstrap for %q: %v", cfg.Type, err)
+			}
+			if callCreds == nil {
+				continue
+			}
```
arjan-bal: Is there a benefit to allowing the returned PerRPCCredentials to be nil?

It seems that if a credential's configuration required a no-op, the implementer could simply return a dedicated no-op implementation to achieve the same result.

My concern is that allowing nil forces every consumer of the API to add nil checks or risk a panic.

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -196,6 +199,74 @@ var (
 				"server_features" : ["ignore_resource_deletion", "xds_v3"]
 			}]
 		}`,
+		// example data seeded from
+		// https://github.com/istio/istio/blob/master/pkg/istio-agent/testdata/grpc-bootstrap.json
```
arjan-bal: nit: I think we should use a commit hash instead of the `master` branch to ensure the link doesn't break.

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -213,69 +284,145 @@ var (
...
+			Metadata:             istioNodeMetadata,
+			userAgentName:        gRPCUserAgentName,
+			userAgentVersionType: userAgentVersion{UserAgentVersion: grpc.Version},
+			clientFeatures:       []string{clientFeatureNoOverprovisioning, clientFeatureResourceWrapper},
+		},
+		certProviderConfigs:                       map[string]*certprovider.BuildableConfig{},
+		clientDefaultListenerResourceNameTemplate: "%s",
+	}
+
+	configWithIstioStyleWithTLSAndJWT = &Config{
```
arjan-bal: Since this variable is only referenced in a single test function, let's move it into that function to improve readability and avoid polluting the global namespace. Similar for other variables with a single reference.

internal/xds/bootstrap/bootstrap_test.go
```diff
@@ -432,6 +579,29 @@ func (s) TestGetConfiguration_Success(t *testing.T) {
 		{"goodBootstrap", configWithGoogleDefaultCredsAndV3},
 		{"multipleXDSServers", configWithMultipleServers},
 		{"serverSupportsIgnoreResourceDeletion", configWithGoogleDefaultCredsAndIgnoreResourceDeletion},
+		{"istioStyleInsecureWithoutCallCreds", configWithIstioStyleNoCallCreds},
```
arjan-bal, line 542: If this case isn't covered already, can you please add a test case to verify that the call creds are not read when the environment variable is unset?

arjan-bal (commented):
Mostly looks good, left some minor comments.

internal/xds/bootstrap/bootstrap.go
dimpavloff: I don't think there's any benefit to allowing them to be nil, I'm just being defensive here. It's not something that can be enforced with the current interface (`Build(config json.RawMessage) (credentials.PerRPCCredentials, func(), error)` ) other than through convention.

I could document it in the interface if you want

dimpavloff (commented):
Thanks @arjan-bal . Could you please take another look?

internal/xds/bootstrap/bootstrap_test.go
dimpavloff: Good shout; I've added one at the bottom. The call creds are still going to get parsed in the current implementation (as we're just doing `json.Unmarshal(data, &server)` ) , but they won't be used as DialOptions

internal/xds/bootstrap/bootstrap.go
arjan-bal: gRPC Go utilizes global registries for builders in many places, and the library does not expect these builders to return nil. When a builder returns nil, it is considered an illegal state and can result in a panic. Some examples:

* [Channel Creds builder](https://github.com/grpc/grpc-go/blob/dc355bc08a9a8f32cc5b06b980918ba88f57a413/internal/xds/bootstrap/bootstrap.go#L305-L309)
* [Name Resolver builder](https://github.com/grpc/grpc-go/blob/dc355bc08a9a8f32cc5b06b980918ba88f57a413/internal/resolver/delegatingresolver/delegatingresolver.go#L130-L132)
* [Load Balancing policy builder](https://github.com/grpc/grpc-go/blob/dc355bc08a9a8f32cc5b06b980918ba88f57a413/internal/balancer/gracefulswitch/gracefulswitch.go#L131-L134). Here, the nil is considered illegal, but it's handled by returning an error. There are other places where nil load balancers are not handled.

In my opinion, it is okay to omit the nil check here and document in the Build method's comments that it must not return nil.  We could also return an error if we don't want gRPC to panic. If nil PerRPCCredentials  are silently ignored, developers might begin to rely on this implicit nil check, forcing every future call site of Build to also handle nil values for consistency.

Also note that checking nil values for interfaces can be error-prone. Consider the following code:
```go
package main

import "fmt"

type Element interface{}

type SomeImplementation struct{}

func f() Element {
	return nil
}

func g() Element {
	var result *SomeImplementation
	return result
}

func main() {
	fmt.Printf("f() != nil: %v, f(): %v\n", f() == nil, f()) // Prints: false, %!p(<nil>)
	fmt.Printf("g() != nil: %v, g(): %v\n", g() != nil, g()) // Prints: true, <nil>
}
```

Go playground: https://go.dev/play/p/hbc9cnbihBk
This document has an explanation: https://go.dev/doc/faq#nil_error

dimpavloff: Yep, fair enough. My view is that it should panic as it's an indication of a bug rather than valid situation. I've updated the PR.

arjan-bal (approved):
LGTM, thank you for the contribution!

dimpavloff:
Great, thanks both!

Issue #8486, opened 2025-08-04
client transport: ignore HTTP/2 status if content-type is application/grpc


dfawley:
The gRPC spec does not indicate that we should be validating the HTTP/2 status if the content-type is set correctly as a gRPC status.  This is to allow us to translate gRPC errors to HTTP statuses on the server side.  Go should follow this behavior.

hugehoo:
@dfawley hi, can i try on this issue?

arjan-bal:
Hi @hugehoo, thanks for offering to help! I've assigned you to the issue.

dfawley:
Sorry this was supposed to be worked on by @mbissa

mbissa:
I have started looking at this issue.

PR #8548, opened 2025-09-01
client: ignore http status header for gRPC streams

mbissa:
Fixes https://github.com/grpc/grpc-go/issues/8486

When a gRPC response is received with content type application/grpc, we then do not expect any information in the http status and the status information needs to be conveyed by gRPC status only.

In case of missing gRPC status, we will throw an Internal error instead of Unknown in accordance with https://grpc.io/docs/guides/status-codes/

	Changes :
	- Ignore http status in case of content type application/grpc
	- Change the default rawStatusCode to return Internal for missing grpc status

RELEASE NOTES:
* client : Ignore the HTTP header status for gRPC streams and return Internal error for missing gRPC status.


arjan-bal:
In the grpc-go repo, we use the assignee to indicate who the review is pending on (author or reviewer). We ignore the reviewer status (the yellow pending re-review dot). Assigning to myself.

internal/transport/http2_client.go
```diff
@@ -1534,7 +1533,8 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		}
 	}
 
-	if !isGRPC || httpStatusErr != "" {
+	//if response is not grpc or endstream doesn't receive a grpc status, fall back to http error code
```
arjan-bal: At line#1500 above, the code is attempting to parse the status code as an integer and returning early if it fails.
```go
         case ":status":
			if hf.Value == "200" {
				httpStatusErr = ""
				statusCode := 200
				httpStatusCode = &statusCode
				break
			}
			c, err := strconv.ParseInt(hf.Value, 10, 32)
			if err != nil {
				se := status.New(codes.Internal, fmt.Sprintf("transport: malformed http-status: %v", err))
				t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
				return
			}
```

Based on the issue description, I think we should completely ignore the `:status` pseudo header if the content type is gRPC. 

I think the performant way to do this would be delaying the `t.closeStream` call till we've seen all the header fields. 
The alternate would be to do 2 passes over the header list, one to check if the content type is gRPC, this may have a noticeable impact on unary RPC performance since it is executed for every RPC.  We can run a benchmark to confirm if we go with this approach.
```
go run benchmark/benchmain/main.go -benchtime=60s -workloads=unary \
   -compression=off -maxConcurrentCalls=120 -trace=off \
   -reqSizeBytes=1024 -respSizeBytes=1024 -networkMode=Local -resultFile="${RUN_NAME}"   -recvBufferPool=simple
```

arjan-bal:
For the PR title, we add the name of the Go package (e.g. `transport`), with the prefix if it's ambiguous (e.g. `xds/bootstrap`), followed by the change description as a complete sentence, written as though it were an order (an imperative sentence), see [go/cl-descriptions#first-line](http://go/cl-descriptions#first-line).

internal/transport/transport_test.go
```diff
@@ -2755,6 +2777,9 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
 
 			got := ts.status
 			want := test.wantStatus
+			if test.wantStatusEndStream != nil {
+				want = test.wantStatusEndStream
+			}
```
arjan-bal: I think we should explicitly `wantStatusEndStream` for all the tests which have non `nil` values. In table-driven tests, it's fine to duplicate parameters across test cases, even if most share the same value. This practice keeps the tests declarative and simplifies the logic, often to just an equality or diff comparison.

internal/transport/transport_test.go
```diff
@@ -2695,34 +2697,55 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
...
+				},
+			},
+			wantStatusEndStream: status.New(codes.DeadlineExceeded, "Request timed out: Internal error"),
 		},
 	} {
 
 		t.Run(test.name, func(t *testing.T) {
+			if test.wantStatus == nil {
+				t.Skip()
+			}
```
arjan-bal: Instead of skipping the test, can we assert the expected behaviour of parsing such header frames?

Ideally, we should not be running two different tests on the same set of test cases and instead have two test functions.

mbissa: Done, added assertions.

internal/transport/transport_test.go
mbissa: Done, added duplicate parameters

internal/transport/http2_client.go
mbissa: Done, moved closed stream to later.

internal/transport/http2_client.go
```diff
@@ -1539,25 +1512,53 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		}
 	}
 
-	if !isGRPC || httpStatusErr != "" {
-		var code = codes.Internal // when header does not include HTTP status, return INTERNAL
+	//if not a gRPC response - evaluate entire http status and process close stream / response
+	//for 200 -> Unknown, else decode the error
```
arjan-bal: style: Please use complete sentences for comments. Also leave a after `//`. See https://google.github.io/styleguide/go/decisions#comment-sentences

internal/transport/http2_client.go
```diff
@@ -1465,17 +1465,13 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
 		grpcMessage    string
 		recvCompress   string
-		httpStatusCode *int
 		httpStatusErr  string
-		rawStatusCode  = codes.Unknown
+		rawStatusCode  = codes.Internal
 		// headerError is set if an error is encountered while parsing the headers
-		headerError string
+		headerError        string
+		receivedHTTPStatus string
```
arjan-bal: nit: We can omit the `received` prefix int the variable name, shortening it to `httpStatus` since it doesn't effect readability.

internal/transport/http2_client.go
```diff
@@ -1539,25 +1512,53 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+			c, err := strconv.ParseInt(receivedHTTPStatus, 10, 32)
+			if err != nil {
+				se := status.New(grpcErrorCode, fmt.Sprintf("transport: malformed http-status: %v", err))
+				t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
+				return
+			}
+			statusCode := int(c)
+			if statusCode >= 100 && statusCode < 200 {
+				//In case of informational headers return
+				//For trailers, since we are already in gRPC mode, we will ignore all http statuses and not enter this block
```
arjan-bal: It looks like the check for the `endStream` flag was removed.
https://github.com/grpc/grpc-go/blob/8420f3ff9ce4617369e054cedb51fda6d45c3340/internal/transport/http2_client.go#L1521-L1528

internal/transport/http2_client.go
```diff
@@ -1539,25 +1512,53 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+	if !isGRPC {
+		var grpcErrorCode = codes.Internal // when header does not include HTTP status, return INTERNAL
+		var errs []string
 
-		if httpStatusCode != nil {
+		switch receivedHTTPStatus {
+		case "":
+			httpStatusErr = "malformed header: missing HTTP status"
+		case "200":
+			grpcErrorCode = codes.Unknown
```
arjan-bal: Is this explicit check required? Since 200 isn't present in the `HTTPStatusConvTab` map, I think it may be omitted.

internal/transport/http2_client.go
```diff
@@ -1539,25 +1512,53 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
-		var errs []string
+
 		if httpStatusErr != "" {
 			errs = append(errs, httpStatusErr)
 		}
+
 		if contentTypeErr != "" {
 			errs = append(errs, contentTypeErr)
 		}
 		// Verify the HTTP response is a 200.
```
arjan-bal: This comment doesn't seem to be accurate. Should we remove it now?

internal/transport/transport_test.go
arjan-bal: Thanks, we should be able to remove this `if` block now, right?

internal/transport/transport_test.go
```diff
@@ -3159,17 +3217,14 @@ func (s) TestClientTransport_Handle1xxHeaders(t *testing.T) {
...
 	}{
 		{
-			name: "1xx with END_STREAM is error",
+			name: "1xx with END_STREAM will be ignored",
 			metaHeaderFrame: &http2.MetaHeadersFrame{
 				Fields: []hpack.HeaderField{
 					{Name: ":status", Value: "100"},
 				},
 			},
-			httpFlags: http2.FlagHeadersEndStream,
```
arjan-bal, line 3226: I think the test was correct before these changes. This test was recently added to fix a bug: https://github.com/grpc/grpc-go/issues/8485

test/http_header_end2end_test.go
```diff
@@ -103,10 +103,9 @@ func (s) TestHTTPHeaderFrameErrorHandlingInitialHeader(t *testing.T) {
 		errCode codes.Code
 	}{
 		{
-			name: "missing gRPC status",
+			name: "missing gRPC content type",
```
arjan-bal, line 106: In my opinion, a better fix for this test would be to change the assertion to `Ok` instead, keeping the test params the same. `codes.Ok` has the zero value for the integer type, so it can be omitted, see https://google.github.io/styleguide/go/decisions#zero-value-fields

test/http_header_end2end_test.go
```diff
@@ -183,20 +182,19 @@ func (s) TestHTTPHeaderFrameErrorHandlingNormalTrailer(t *testing.T) {
 				"grpc-status", "0",
 				"grpc-status-details-bin", "????",
 			},
-			errCode: codes.Unimplemented,
+			errCode: codes.Internal,
 		},
 		{
-			name: "malformed grpc-status-details-bin field with status 200",
+			name: "malformed grpc-status-details-bin field with status 404 and no content type",
```
arjan-bal, line 188: I think we should keep this test for `malformed grpc-status-details-bin` and add a new test for the `404 and no content type` case if necessary. This would ensure we have coverage for the existing scenario.

internal/transport/http2_client.go
mbissa: The idea is when no error information can be derived from http status since it is 200, we can simply return Unknown code.

internal/transport/http2_client.go
mbissa: I was confused if endstream can come when it is not a gRPC response - @dfawley confirmed that sometimes only one header comes in the response which is set with end stream and it is possible that it may not be gRPC in which case we still need to throw an error if it is informational header. I have added this code back.

internal/transport/http2_client.go
mbissa: removed it. Not sure why it was present.

internal/transport/transport_test.go
mbissa: Agreed. Reverted it back.

test/http_header_end2end_test.go
mbissa: there is a test for missing gRPC status already. The expected outcome will not be Ok, it would be an Internal error since gRPC content type is mandated to have the status.

test/http_header_end2end_test.go
mbissa: The status 200 or 404 is immaterial in case of content type being application/grpc, the previous test already caters to that and we are finally checking the malformed grpc-status-details-bin field which we were not doing earlier. Hence, expected outcome code in the previous test is no longer Unimplemented (due to 404), the malformed grpc-status-details-bin now results in an Internal error code.

internal/transport/http2_client.go
arjan-bal: The `default` case also adds a error message which seems useful. I think we should remove the special case unless it requires different handling in code. We don't need to worry too much about optimizing the error case, since they're expected to be infrequent.

internal/transport/http2_client.go
arjan-bal: Ping on this.

internal/transport/http2_client.go
```diff
@@ -1512,52 +1538,31 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 						"protocol error: informational header with status code %d must not have END_STREAM set", statusCode))
 					t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
 				}
+				//In case of informational headers return
```
arjan-bal: nit: Need a space at the beginning of the comment here.

arjan-bal (approved):
Mostly LGTM, left a few minor comments.

internal/transport/http2_client.go
```diff
@@ -1465,17 +1465,13 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		contentTypeErr = "malformed header: missing HTTP content-type"
 		grpcMessage    string
 		recvCompress   string
-		httpStatusCode *int
 		httpStatusErr  string
-		rawStatusCode  = codes.Unknown
+		rawStatusCode  = codes.Internal
```
dfawley: This name isn't a good one, I think.  Maybe call this `grpcStatusCode` and add a comment?

```suggestion
		grpcStatusCode  = codes.Internal  // the code from the grpc-status header, if present
```

internal/transport/http2_client.go
dfawley: +1

internal/transport/http2_client.go
dfawley: +1; please remove special cases when possible most of the time.  Less code = easier to understand code.

internal/transport/http2_client.go
```diff
@@ -1499,9 +1495,39 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		case "grpc-message":
 			grpcMessage = decodeGrpcMessage(hf.Value)
 		case ":status":
-			c, err := strconv.ParseInt(hf.Value, 10, 32)
+			if !isGRPC {
```
dfawley: This is making the assumption that `:status` appears after `content-type`, which AFAIK is not required.  That's why the code before does header processing by setting a bunch of locals and then looking at them later.

In this case it's not going to cause any problems, since we don't look at this variable if `isGRPC` is `true`, but I think it's better to remove it.  And it's not even necessarily a performance win, since conditional branches are potentially worse for performance than setting a Go string (which is not a full copy of the string itself).

internal/transport/http2_client.go
```diff
@@ -1499,9 +1495,39 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+				headerError = fmt.Sprintf("transport: malformed %s: %v", hf.Name, err)
+				logger.Warningf("Failed to decode metadata header (%q, %q): %v", hf.Name, hf.Value, err)
+				break
+			}
+			mdata[hf.Name] = append(mdata[hf.Name], v)
+		}
+	}
+
+	// If a non gRPC response is received, then evaluate entire http status and process close stream / response.
+	// In case http status doesn't provide any error information (status : 200), evalute response code to be Unknown.
```
dfawley: Please wrap comments to 80 columns.

internal/transport/http2_client.go
```diff
@@ -1499,9 +1495,39 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+	if !isGRPC {
+		var grpcErrorCode = codes.Internal // when header does not include HTTP status, return INTERNAL
+		var errs []string
```
dfawley: Please move this back to where it was defined before; i.e. as late as possible.

internal/transport/http2_client.go
dfawley: +1

internal/transport/http2_client.go
mbissa: done.

internal/transport/http2_client.go
mbissa: done.

internal/transport/http2_client.go
mbissa: thanks for the explanation, thats a great catch! - done.

internal/transport/http2_client.go
mbissa: done.

internal/transport/http2_client.go
mbissa: done.

internal/transport/http2_client.go
mbissa: done.

mbissa: done.

internal/transport/http2_client.go
mbissa: done.

internal/transport/http2_client.go
```diff
@@ -1512,18 +1511,18 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 		}
 	}
 
-	// If a non gRPC response is received, then evaluate entire http status and process close stream / response.
-	// In case http status doesn't provide any error information (status : 200), evalute response code to be Unknown.
+	// If a non gRPC response is received, then evaluate entire http status and
+	//  process close stream / response.
```
dfawley: Please fix the formatting here.

internal/transport/http2_client.go
```diff
@@ -1495,13 +1492,39 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+	// In case http status doesn't provide any error information (status : 200),
+	//  evalute response code to be Unknown.
+	if !isGRPC {
+		var grpcErrorCode = codes.Internal
+		switch httpStatus {
+		case "":
+			httpStatusErr = "malformed header: missing HTTP status"
+		default:
```
dfawley: Nit: I'm not sure I like the use of a `switch` instead of an `if`/`else`.  It seems this would be simpler / easier to read as `if httpStatus == "" { ... } else { ... }`

internal/transport/http2_client.go
```diff
@@ -1495,13 +1492,39 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+			// Any other status code (e.g., "404", "503"). We must parse it.
```
dfawley: Please use complete sentences.  It can still be short:

```suggestion
			// Parse the status code (e.g. "200", 404").
```

internal/transport/http2_client.go
```diff
@@ -1512,52 +1535,30 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
 						"protocol error: informational header with status code %d must not have END_STREAM set", statusCode))
 					t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
 				}
+				// In case of informational headers return
```
dfawley: Lots of options, here, e.g.:

```suggestion
				// In case of informational headers, return.
```

internal/transport/http2_client.go
```diff
@@ -1495,13 +1492,39 @@ func (t *http2Client) operateHeaders(frame *http2.MetaHeadersFrame) {
...
+		case "":
+			httpStatusErr = "malformed header: missing HTTP status"
+		default:
+			// Any other status code (e.g., "404", "503"). We must parse it.
+			c, err := strconv.ParseInt(httpStatus, 10, 32)
+			if err != nil {
+				se := status.New(grpcErrorCode, fmt.Sprintf("transport: malformed http-status: %v", err))
 				t.closeStream(s, se.Err(), true, http2.ErrCodeProtocol, se, nil, endStream)
 				return
 			}
```
dfawley, line 1544: Why do we cast to an `int` here?  It seems unnecessary.

internal/transport/transport_test.go
```diff
@@ -2618,6 +2618,10 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
 		metaHeaderFrame *http2.MetaHeadersFrame
 		// output
 		wantStatus *status.Status
+		// end stream output
```
dfawley: I'm not sure what this means.

Let's make sure all our comments are useful and easy to understand, or just remove the comment if it's mostly just redundant.

internal/transport/transport_test.go
```diff
@@ -2618,6 +2618,10 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
...
+		wantStatusEndStream *status.Status
+		// head channel closed
```
dfawley: Similar.  What's a "head channel"?  The variable name is telling me more than this comment.

internal/transport/transport_test.go
```diff
@@ -2682,6 +2723,13 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
 				codes.Internal,
 				"transport: malformed http-status: strconv.ParseInt: parsing \"xxxx\": invalid syntax",
 			),
+			// for end stream, when we are already talking gRPC, we expect
+			// grpc - status and fail for it, we will ignore http status.
```
dfawley: I'm not following this comment.

internal/transport/transport_test.go
```diff
@@ -2752,9 +2840,8 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
 			}
 
 			s.operateHeaders(test.metaHeaderFrame)
-
 			got := ts.status
-			want := test.wantStatus
+			want := test.wantStatusEndStream
```
dfawley: This test is feeling pretty complicated to me now.  Would it be better to just split this test into two separate test functions since it's running two different tests and has multiple different assertions?

internal/transport/http2_client.go
mbissa: Done.

internal/transport/http2_client.go
mbissa: Changed.

internal/transport/http2_client.go
mbissa: Done.

internal/transport/http2_client.go
mbissa: changed

internal/transport/http2_client.go
mbissa: done.

internal/transport/transport_test.go
mbissa: removed unnecessary comments.

internal/transport/transport_test.go
mbissa: removed unnecessary comments.

internal/transport/transport_test.go
mbissa: removed unnecessary comments. Essentially indicates if it is a non header-only response, then when a trailer frame is being processed, we are already sure that the header was grpc, else it should have failed during header itself.

internal/transport/transport_test.go
mbissa: Done, split the test into 2 - one for header, one for trailer

internal/transport/transport_test.go
```diff
@@ -2693,68 +2701,152 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
...
 		})
-		t.Run(fmt.Sprintf("%s-end_stream", test.name), func(t *testing.T) {
-			ts := testStream()
-			s := testClient(ts)
+	}
+}
 
-			test.metaHeaderFrame.HeadersFrame = &http2.HeadersFrame{
+// TestClientDecodeTrailer validates the handling of trailer frames, which may
+// or may not also be the initial header frame (header-only response).
```
dfawley, line 2746:
```suggestion
// or may not also be the initial header frame (trailers-only response).
```

Terminology from https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses

internal/transport/transport_test.go
```diff
@@ -2693,68 +2701,152 @@ func (s) TestClientDecodeHeaderStatusErr(t *testing.T) {
...
+			wantEndStreamStatus: status.New(codes.DeadlineExceeded, "Request timed out: Internal error"),
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cs := newTestClientStream()
+			// Mark headerChanClosed to indicate trailer frames.
+			cs.headerChanClosed = 1
+			// Simulate the state where the initial headers have already been processed.
```
dfawley, line 2835: Both of these comments seem like they're related to the same thing, which is setting `headerChanClosed`.

test/http_header_end2end_test.go
dfawley, line 106: It seems like you have changed this test instead of adding a new one?  Shouldn't we still have a test that sets these exact header fields?

test/http_header_end2end_test.go
dfawley, line 188: Similar - don't we still want to have a test with an illegal grpc-status-details-bin and status 200?  We can also have this new test case, but it seems like having more test cases is always better.

test/http_header_end2end_test.go
mbissa: missing grpc status is no longer relevant in the the initial headers - it is tested in the trailers test where grpc status is expected/mandatory.

test/http_header_end2end_test.go
mbissa: If the content type is application/grpc (as was in previous version of the test) the http status is immaterial (200/404 etc won't make a difference) and it is tested in the test case above this with 404. This was more important variation of data where we check without correct content type.

PR #8549, opened 2025-09-01
xds/clusterimpl: Convert existing unit tests to e2e style (1/N)

Pranjali-2501:
Partially addresses: #6113 

Added `TestCircuitBreaking` and `TestDropByCategory` end2end Tests in xds/clusterimpl/tests Mirroring Existing Unit Test and delete them.

RELEASE NOTES: None


internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +439,50 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+		},
+	}
+	mgmtServer.LRSServer.LRSResponseChan <- &resp
+
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
+			if cs.TotalDroppedRequests != droppedRPCCount || cs.UpstreamLocalityStats[0].TotalIssuedRequests != maxRequests {
+				t.Errorf("Failed with unexpected outputs")
```
easwars: This error message is not very useful to a developer when they see it. You could actually break the `if` check into two separate checks and have an error message which is meaningful. One that contains information about what the test actually got and what it wanted.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +439,50 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
-	for i := 0; i < 100; i++ {
+	const droppedRPCCount = 100
+	for i := 0; i < droppedRPCCount; i++ {
 		stream, err := client.FullDuplexCall(ctx)
-		if status.Code(err) == codes.Unavailable {
-			return
-		}
-		if err == nil {
+		switch {
+		case err == nil:
```
easwars: I don't know how frequently this case happens, i.e., the race specified in the comment mentioned at the top of the for loop. But when it does happen, it would mean that the new check that you've added (that compares the number of dropped requests and issued requests with the expected counts for them respectively) would fail the test.

Can you try running this on forge a million times to see if we ever hit this race. If it is found that we never hit condition, we can add a t.Fatal here instead.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -396,12 +393,13 @@ func (s) TestCircuitBreaking(t *testing.T) {
 
 	// Configure the xDS management server with default resources.
 	const serviceName = "my-test-xds-service"
-	const maxRequests = 3
+	const maxRequests = 10
```
easwars: Curious. Why did you end up changing this value?

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: I have tested it running million times and it never hit that case. Changed it to `t.Fatal` instead.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: This was not done deliberately. Changed it to 3.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -359,9 +360,9 @@ func waitForSuccessfulLoadReport(ctx context.Context, lrsServer *fakeserver.Serv
 }
 
 // Tests that circuit breaking limits RPCs E2E.
-func (s) TestCircuitBreaking(t *testing.T) {
+func TestCircuitBreaking(t *testing.T) {
```
easwars: Why are you removing the `s` receiver here?

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -402,6 +400,7 @@ func (s) TestCircuitBreaking(t *testing.T) {
 		NodeID:     nodeID,
 		Host:       "localhost",
 		Port:       testutils.ParsePort(t, server.Address),
+		SecLevel:   e2e.SecurityLevelNone,
```
easwars: Why is this line required? The default value (or zero value) for the `e2e.SecurityLevel` enum is `e2e.SecurityLevelNone` and if fields in a struct are not set, they automatically get their zero values.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+		_, err := client.FullDuplexCall(ctx)
+		switch {
+		case err == nil:
+			t.Fatalf("client.FullDuplexCall(_) failed with unexpected error got: <nil>, want: %v", err)
+		case status.Code(err) == codes.Unavailable:
+			continue
+		default:
+			t.Errorf("client.FullDuplexCall(_) failed with unexpected error = %v", err)
+		}
+		time.Sleep(10 * time.Millisecond)
```
easwars: Why do we need this sleep here?

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
 
 	// Since we are at the max, new streams should fail.  It's possible some are
 	// allowed due to inherent raciness in the tracking, however.
-	for i := 0; i < 100; i++ {
+	const droppedRPCCount = 100
+	for i := 0; i < droppedRPCCount; i++ {
+		_, err := client.FullDuplexCall(ctx)
```
easwars: Can you please check for context errors at the top of the `for` loop before doing anything here.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+		t.Fatalf("Failure waiting for initial LRS request: %v", err)
+	}
+
+	resp := fakeserver.Response{
+		Resp: &v3lrspb.LoadStatsResponse{
+			SendAllClusters:       true,
+			LoadReportingInterval: durationpb.New(10 * time.Millisecond),
+		},
+	}
+	mgmtServer.LRSServer.LRSResponseChan <- &resp
```
easwars: Nit: You could inline the response here and eliminate the `resp` local variable.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+		for _, cs := range loadStats.ClusterStats {
+			if cs.TotalDroppedRequests != droppedRPCCount {
+				t.Errorf("Failed with unexpected diff got: %v, want: %v", cs.TotalDroppedRequests, droppedRPCCount)
+			}
+			if cs.UpstreamLocalityStats[0].TotalIssuedRequests != maxRequests {
+				t.Errorf("Failed with unexpected diff got: %v, want: %v", cs.UpstreamLocalityStats[0].TotalIssuedRequests, maxRequests)
+			}
+		}
+	case <-ctx.Done():
+		t.Fatalf("Timeout while waiting for LRS stream: %v", ctx.Err())
```
easwars: This is a timeout when waiting for a message with the load report on the stream, right? Not the stream itself.

This applies to other places where the same error string is used as well.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
+			if cs.TotalDroppedRequests != droppedRPCCount {
+				t.Errorf("Failed with unexpected diff got: %v, want: %v", cs.TotalDroppedRequests, droppedRPCCount)
+			}
+			if cs.UpstreamLocalityStats[0].TotalIssuedRequests != maxRequests {
+				t.Errorf("Failed with unexpected diff got: %v, want: %v", cs.UpstreamLocalityStats[0].TotalIssuedRequests, maxRequests)
+			}
```
easwars: Can we compare full structs instead of individual fields. See: go/go-style/decisions#compare-full-structures

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+}
+
+// ComputeIdealNumRpcs calculates the ideal number of RPCs to send to test
+// probabilistic behaviors.
+//
+// It's based on the idea of a binomial distribution approximated by a normal
+// distribution. The function aims to find a number of RPCs such that
+// the observed probability is within a certain error_tolerance of the expected
+// probability (p).
+func computeIdealNumRpcs(p float64, errorTolerance float64) uint64 {
```
easwars: Nit: This line could be shortened as follows:

```golang
func computeIdealNumRpcs(p, errorTolerance float64) uint64 {
```

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	numRpcs :=
+		uint64(math.Ceil(p * (1 - p) * 5.00 * 5.00 / errorTolerance / errorTolerance))
```
easwars: Please don't break up lines this way. See: go/go-style/guide#line-length

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	return numRpcs
+}
+
+// TestDropByCategory verifies that the balancer correctly drops the picks, and
+// that the drops are reported.
+func TestDropByCategory(t *testing.T) {
```
easwars: Same here. Why is this test missing the `s` receiver. The `s` struct embeds a `grpctest.Tester` which has a setup and teardown method that gets run for every test. Please see https://github.com/grpc/grpc-go/blob/master/internal/grpctest/grpctest.go#L52

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -359,9 +360,9 @@ func waitForSuccessfulLoadReport(ctx context.Context, lrsServer *fakeserver.Serv
 }
 
 // Tests that circuit breaking limits RPCs E2E.
-func (s) TestCircuitBreaking(t *testing.T) {
+func TestCircuitBreaking(t *testing.T) {
 	// Create an xDS management server that serves ADS requests.
```
easwars: The fact that this management server serves ADS requests is less important than the fact that is also implements the load reporting service.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+func computeIdealNumRpcs(p float64, errorTolerance float64) uint64 {
+	numRpcs :=
+		uint64(math.Ceil(p * (1 - p) * 5.00 * 5.00 / errorTolerance / errorTolerance))
+	return numRpcs
+}
+
+// TestDropByCategory verifies that the balancer correctly drops the picks, and
+// that the drops are reported.
+func TestDropByCategory(t *testing.T) {
+	// Create an xDS management server that serves ADS requests.
```
easwars: Same comment about this comment, as in the previous test.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+
+	// Create an xDS resolver with the above bootstrap configuration.
+	var resolverBuilder resolver.Builder
+	var err error
+	if newResolver := internal.NewXDSResolverWithConfigForTesting; newResolver != nil {
+		resolverBuilder, err = newResolver.(func([]byte) (resolver.Builder, error))(bc)
+		if err != nil {
+			t.Fatalf("Failed to create xDS resolver for testing: %v", err)
+		}
+	}
```
easwars: I think it would make sense to rewrite this as follows:

```suggestion
	if internal.NewXDSResolverWithConfigForTesting == nil {
		t.Fatalf("internal.NewXDSResolverWithConfigForTesting is nil")
	}

	resolverBuilder, err := internal.NewXDSResolverWithConfigForTesting.(func([]byte) (resolver.Builder, error))(bc)
	if err != nil {
		t.Fatalf("Failed to create xDS resolver for testing: %v", err)
	}
```

Also, please consider making this change in the previous test as well. Thanks.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+				if _, err := stream.Recv(); err != nil {
+					return err
+				}
+			}
+		},
+	}
+	server := stubserver.StartTestService(t, f)
+	defer server.Stop()
+
+	// Configure the xDS management server with default resources.
```
easwars: I think it would make sense to be more specific about what is being configured here. Something like:

// Configure xDS resources on the management server with drops configuration
// that drops one RPC for every 50 RPCs made.

Also, please consider making a change on the same lines in the previous test as well.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+			LoadReportingInterval: durationpb.New(10 * time.Millisecond),
+		},
+	}
+	mgmtServer.LRSServer.LRSResponseChan <- &resp
+
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
+			dropRate := float64(cs.TotalDroppedRequests / rpcCount)
```
easwars: Given that the test configures a drop reason on the endpoints resource, should we be checking explicitly for RPC dropped for that specific reason instead of checking `TotalDroppedRequests` which includes all drops.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+
+	// Create a ClientConn and make a successful RPC.
+	cc, err := grpc.NewClient(fmt.Sprintf("xds:///%s", serviceName), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithResolvers(resolverBuilder))
+	if err != nil {
+		t.Fatalf("failed to dial local test server: %v", err)
+	}
+	cc.Connect()
+	defer cc.Close()
+
+	client := testgrpc.NewTestServiceClient(cc)
```
easwars, line 646: I think it would make sense to move the call to `cc.Connect` to be right before this statement and add a comment saying something to the effect that we want to wait for the gRPC channel to be READY before starting RPCs to ensure that we get an accurate drop count.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+			SendAllClusters:       true,
+			LoadReportingInterval: durationpb.New(10 * time.Millisecond),
+		},
+	}
+	mgmtServer.LRSServer.LRSResponseChan <- &resp
+
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
```
easwars: Maybe, we should also check only for cluster stats for the cluster that we care about in this test. Ideally, we should not see loads for any other cluster. So, maybe you can fail the test if you see loads for any cluster other than the one that we sending RPCs to here in this test.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+			dropRate := float64(cs.TotalDroppedRequests / rpcCount)
+			if math.Abs(dropRate-rpcDropRate) > errortolerance {
+				t.Errorf("Drop rate goes out of errortolerance got %v, want %v", math.Abs(dropRate-rpcDropRate), errortolerance)
+			}
+		}
+	case <-ctx.Done():
+		t.Fatalf("Timeout while waiting for LRS stream: %v", ctx.Err())
+	}
+
+	const (
```
easwars, line 667: Nit: Please add a comment that you are changing the drop configuration and expect the drop rate to match the new configuration.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+
+	// Configure the xDS management server with default resources.
+	const (
+		dropReason      = "test-dropping-category"
+		dropNumerator   = 1
+		dropDenominator = 50
+		serviceName     = "my-test-xds-service"
+		errortolerance  = 0.05
+	)
+	rpcDropRate := float64(dropNumerator) / float64(dropDenominator)
```
easwars: Nit: Consider renaming this to something like `wantDropRate` or `wantRPCDropRate`.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
 		}
-		time.Sleep(10 * time.Millisecond)
 	}
 
-	t.Fatalf("RPCs unexpectedly allowed beyond circuit breaking maximum")
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
+			dropRate := float64(cs.TotalDroppedRequests / rpcCount)
```
easwars: Checking for drops with the correct reason makes more sense here as well, since we have changed the drop reason as well when changing the configuration.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	}
+
+	if _, err = mgmtServer.LRSServer.LRSRequestChan.Receive(ctx); err != nil {
+		t.Fatalf("Failure waiting for initial LRS request: %v", err)
+	}
+
+	resp := fakeserver.Response{
+		Resp: &v3lrspb.LoadStatsResponse{
+			SendAllClusters:       true,
+			LoadReportingInterval: durationpb.New(10 * time.Millisecond),
```
easwars: You are setting a load reporting interval of `10ms`. Are you sure the test is able to able to start all its RPCs by this duration? Otherwise you might see flakes because the drop rate might not match your expected rate.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	// Configure the xDS management server with default resources.
+	const (
+		dropReason      = "test-dropping-category"
+		dropNumerator   = 1
+		dropDenominator = 50
+		serviceName     = "my-test-xds-service"
+		errortolerance  = 0.05
+	)
+	rpcDropRate := float64(dropNumerator) / float64(dropDenominator)
+	rpcCount := computeIdealNumRpcs(rpcDropRate, errortolerance)
```
easwars: Here and elsewhere where you use this function to compute the number of RPCs, it would make sense to log the `rpcCount` using `t.Logf` so that it is available if and when the test fails.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,255 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	}
+	mgmtServer.LRSServer.LRSResponseChan <- &resp
+
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
+			dropRate := float64(cs.TotalDroppedRequests / rpcCount)
+			if math.Abs(dropRate-rpcDropRate) > errortolerance {
+				t.Errorf("Drop rate goes out of errortolerance got %v, want %v", math.Abs(dropRate-rpcDropRate), errortolerance)
```
easwars: Here and elsewhere, consider also printing the actual number of drops seen and the actual number of issued requests seen (and their expected values), because that will be more helpful for debugging compared to the drop rate.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Removed.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Removed.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Removed unintentionally.
Added `s`.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Changed for all tests.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Raised the interval to 1 second.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Using `DroppedCount` for specific category increase the flakiness of the test.
Moved back to use `cs.TotalDroppedRequests`.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
easwars: This could be a single line:
```
	return uint64(math.Ceil(p * (1 - p) * 5.00 * 5.00 / errorTolerance / errorTolerance))
```

internal/xds/balancer/clusterimpl/tests/balancer_test.go
easwars: If you are removing these receivers to be able to run a single test from your IDE, you can do it from the command line instead. 

For example, the command to run this test would be:
```
go test -race google.golang.org/grpc/internal/xds/balancer/clusterimpl/tests/... -run Test/DropByCategory
```

If you want to enable complete logs, you would do prefix the above command with setting the following environment variables likes:
```
GRPC_GO_LOG_VERBOSITY_LEVEL=99 GRPC_GO_LOG_SEVERITY_LEVEL=info <above command>
```

internal/xds/balancer/clusterimpl/tests/balancer_test.go
easwars: Why does it make the test flaky?

internal/xds/balancer/clusterimpl/tests/balancer_test.go
easwars: Increasing this to `1s` would cause the tests to run for abnormally long. We want our tests to run as fast as they possibly can.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,270 @@ func (s) TestCircuitBreaking(t *testing.T) {
 
 	// Since we are at the max, new streams should fail.  It's possible some are
 	// allowed due to inherent raciness in the tracking, however.
-	for i := 0; i < 100; i++ {
+	const droppedRPCCount = 100
+	for i := 0; i < droppedRPCCount; i++ {
+		if ctx.Err() != nil {
+			t.Fatalf("Context error: %v", ctx.Err())
+		}
+		_, err := client.FullDuplexCall(ctx)
```
easwars: I think the following would work:

```golang
if _, err := client.FullDuplexCall(ctx); status.Code(err) != codes.Unavailable {
  t.Fatalf("client.FullDuplexCall() returned status %q, want %q", status.Code(err), codes.Unavailable)
}
```

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,270 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	const (
+		dropReason      = "test-dropping-category"
+		dropNumerator   = 1
+		dropDenominator = 50
+		serviceName     = "my-test-xds-service"
+		errorTolerance  = 0.05
+	)
+	wantRPCDropRate := float64(dropNumerator) / float64(dropDenominator)
+	rpcCount := computeIdealNumRpcs(wantRPCDropRate, errorTolerance)
+	t.Logf("Computed rpcCount: %d", rpcCount)
```
easwars: Please make this log statement more useful to a developer when they get here because the test failed. Something like

"About to send %d RPCs to test drop rate of %v", rpcCount, wantRPCDropRate

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,270 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	}
+
+	// Create a ClientConn and make a successful RPC.
+	cc, err := grpc.NewClient(fmt.Sprintf("xds:///%s", serviceName), grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithResolvers(resolverBuilder))
+	if err != nil {
+		t.Fatalf("failed to dial local test server: %v", err)
+	}
+	defer cc.Close()
+
+	// Ensure the gRPC channel is READY before issuing RPCs to get accurate drop count.
```
easwars: Here and probably elsewhere, please wrap comment lines at 80-cols. See: go/go-style/decisions#comment-line-length

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,270 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	testutils.AwaitState(ctx, t, cc, connectivity.Ready)
+
+	for i := 0; i < int(rpcCount); i++ {
+		if ctx.Err() != nil {
+			t.Fatalf("Context error: %v", ctx.Err())
+		}
 		stream, err := client.FullDuplexCall(ctx)
-		if status.Code(err) == codes.Unavailable {
-			return
+		switch {
```
easwars: Same here as the other comment. Might be possible to simplify this and get rid of the `switch`.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,270 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+	}
+	mgmtServer.LRSServer.LRSResponseChan <- &resp
+
+	select {
+	case req := <-mgmtServer.LRSServer.LRSRequestChan.C:
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest)
+		for _, cs := range loadStats.ClusterStats {
+			gotRPCDropRate := float64(cs.TotalDroppedRequests) / float64(rpcCount)
+			if math.Abs(gotRPCDropRate-wantRPCDropRate) > errorTolerance {
+				t.Errorf("Drop rate goes out of errortolerance got: %v, want: %v, totalDroppedRequest: %v, totalIssuesRequest: %v", math.Abs(gotRPCDropRate-wantRPCDropRate), errorTolerance, cs.TotalDroppedRequests, cs.UpstreamLocalityStats[0].TotalIssuedRequests)
```
easwars: Does it make sense to continue the test at this point? See: go/go-style/decisions#keep-going

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,270 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+			},
+		})}
+
+	if err := mgmtServer.Update(ctx, resources); err != nil {
+		t.Fatal(err)
+	}
+
+	wantRPCDropRate = float64(dropNumerator2) / float64(dropDenominator2)
+	rpcCount = computeIdealNumRpcs(wantRPCDropRate, errorTolerance)
+	t.Logf("Computed rpcCount: %d", rpcCount)
```
easwars: Same here.

Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Yes I think. Second check is not dependent on this, it could pass or fail based on the drop count of second drop category.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: I don't think it can be simplified further. I can use `if-else` instead of `switch ` if you want.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Changed it to `50ms`.
Does that work?

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Modify the test to check `DroppedCount` instead of `TotalDroppedRequests` and remove the flakiness also.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Ack

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
easwars: Sure.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -425,6 +428,7 @@ func (s) TestCircuitBreaking(t *testing.T) {
 	}
 	defer cc.Close()
 
+	cc.Connect()
```
easwars: Do we need this call here? If so, could you please add a comment saying why. Thanks.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,290 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+		if err != nil {
+			continue
+		}
+		loadStats := req.(*fakeserver.Request).Req.(*v3lrspb.LoadStatsRequest).ClusterStats
+		if len(loadStats) != 1 {
+			continue
+		}
+		for _, cs := range loadStats {
+			if cs.ClusterName != resources.Clusters[0].Name {
+				t.Fatalf("Received stats for unexpected cluster got: %v, want: %v", cs.ClusterName, resources.Clusters[0].Name)
```
easwars: Nit: Consider using the `%q` formatting directive to add quotes around things like names. So, this could be written as:

t.Fatalf("Received stats for unexpected cluster %q, want %q", cs.ClusterName, resources.Clusters[0].Name)

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,290 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+			}
+			gotRPCDropRate := float64(cs.TotalDroppedRequests) / float64(rpcCount)
+			if (math.Trunc(math.Abs(gotRPCDropRate-wantRPCDropRate)*100) / 100) > errorTolerance {
+				t.Errorf("Drop rate goes out of errortolerance got: %v, want: %v, totalDroppedRequest: %v, totalIssuesRequest: %v", (math.Trunc(math.Abs(gotRPCDropRate-wantRPCDropRate)*100) / 100), errorTolerance, cs.TotalDroppedRequests, cs.UpstreamLocalityStats[0].TotalIssuedRequests)
```
easwars: This might be easier to read (both in code and the output) as two separate error statements:

```golang
t.Errorf("Drop rate goes out of errortolerance got: %v, want: %v", (math.Trunc(math.Abs(gotRPCDropRate-wantRPCDropRate)*100) / 100), errorTolerance)

t.Errorf("totalDroppedRequest: %v, totalIssuesRequest: %v", cs.TotalDroppedRequests, cs.UpstreamLocalityStats[0].TotalIssuedRequests)
```

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,290 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+				{
+					Backends: []e2e.BackendOptions{{Ports: []uint32{testutils.ParsePort(t, server.Address)}}},
+					Weight:   1,
+				},
+			},
+			DropPercents: map[string]int{
+				dropReason2: dropNumerator2 * 100 / dropDenominator2, // drops one RPC for every 40 RPCs made
+			},
+		})}
+
```
easwars, line 688: Nit: nix this newline

internal/xds/balancer/clusterimpl/tests/balancer_test.go
```diff
@@ -436,21 +440,290 @@ func (s) TestCircuitBreaking(t *testing.T) {
...
+						t.Errorf("Drop rate goes out of errortolerance got: %v, want: %v, totalDroppedRequest: %v, totalIssuesRequest: %v", (math.Trunc(math.Abs(gotRPCDropRate-wantRPCDropRate)*100) / 100), errorTolerance, cs.TotalDroppedRequests, cs.UpstreamLocalityStats[0].TotalIssuedRequests)
+					}
+				}
+			}
+			if !found {
+				continue
+			}
+		}
+		break
+	}
```
easwars, line 699: This block of code is very similar to the one above (except that you check for the drop category here). Would it be possible to make a helper function that accepts whatever is required `(ctx, client, rpcCount, lrsServer, wantCluster, wantDropCategory, wantDropRate)`, and returns an `error`, that when non-nil, the test goroutine would fail the test.

easwars (commented):
Also, same comment about the PR title as made in the other PR.

And, we should ideally be getting rid of the existing test as part of it moving it an e2e style, unless we have a good reason not to, in which case, it might make sense to include it in the PR description.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: We don't need it here.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: I have moved this logic to a helper function and need to return as soon as it fails.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

internal/xds/balancer/clusterimpl/tests/balancer_test.go
Pranjali-2501: Done.

PR #8550, opened 2025-09-02
GoogleC2P: remove dependency on metadata server for IPv6 node metadata

apolcyn:
Remove reliance on metadata server since it's result is no longer needed, hardcode IPv6 support in node metadata instead.

Related c++ change: https://github.com/grpc/grpc/pull/40571

Note we preserve prior behavior in case experiment `NewPickFirstEnabled` is disabled, because our testing/qualification has not covered that being disabled.

Related: internal issue b/407587619

RELEASE NOTES: n/a

### DIFF
--- a/xds/googledirectpath/googlec2p.go
+++ b/xds/googledirectpath/googlec2p.go
@@ -184,7 +184,7 @@ func newNodeConfig(zone string, ipv6Capable bool) map[string]any {
 	}
 	if envconfig.NewPickFirstEnabled {
 		// Enable dualstack endpoints in TD.
-		// TODO(apolcyn): remove IPv6 metadata server queries entirely after old pick first is removed.
+		// TODO(https://github.com/grpc/grpc-go/issues/8561): remove IPv6 metadata server queries entirely after old pick first is removed.
 		ipv6Capable = true
 	} else {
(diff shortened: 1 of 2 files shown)


apolcyn:
@dfawley I'm having some trouble getting the PR validation to pass `This PR must have a label starting with 'Type:'. I don't have permissions to set labels.

Besides that, sending for review

xds/googledirectpath/googlec2p.go
```diff
@@ -182,6 +182,13 @@ func newNodeConfig(zone string, ipv6Capable bool) map[string]any {
 		"id":       fmt.Sprintf("C2P-%d", randInt()),
 		"locality": map[string]any{"zone": zone},
 	}
+	if envconfig.NewPickFirstEnabled {
+		// Enable dualstack endpoints in TD.
+		// TODO(apolcyn): remove IPv6 metadata server queries entirely after old pick first is removed.
```
dfawley: If there isn't already one, could you please file an issue (internal or on github) to track this work?  Thanks!

dfawley (approved):
Thanks for the PR!

xds/googledirectpath/googlec2p.go
apolcyn: Thanks for the review!

Done, filed https://github.com/grpc/grpc-go/issues/8561

PR #8556, opened 2025-09-04
github,test: fix internal CI build

arjan-bal:
Fix the following errors:
* only use one space to separate words
* use service codedgen import name for stream type

RELEASE NOTES: N/A


dfawley:
> only use one space to separate words

This isn't "fixing an error", FWIW.  Two spaces after a period is pretty common to encounter in the real world.  Any checker that pedantic should be disabled, IMO, as it does more harm than good to police something like this.  It is not rendered any differently by HTML anyway.

